PR #5, opened 2023-09-05
Restructure code base and ensure tests still pass

RAYemelyanova:


tests/epics/areadetector/test_hdf_streamer_det.py
```diff
@@ -103,9 +51,10 @@ async def hdf_streamer_dets():
     yield deta, detb
 
 
-async def test_hdf_streamer_dets_step(hdf_streamer_dets: List[HDFStreamerDet], RE):
-    d = DocHolder()
-    RE(bp.count(hdf_streamer_dets), d.append)
+async def test_hdf_streamer_dets_step(
+    hdf_streamer_dets: List[HDFStreamerDet], RE, doc_holder
```
coretl, line 55: Not a big fan of pytest fixtures if they don't need teardown support, as they complicate typing. I guess you could:
```suggestion
    hdf_streamer_dets: List[HDFStreamerDet], RE, doc_holder: DocHolder
```
But then you would need:
```python
from .conftest import DocHolder
``
so you might as well just create a `DocHolder` manually like the original code does...

File: docs/developer/explanations/decisions/0004-repository-structure.rst [hunk 2]
```diff
@@ -55,24 +55,40 @@ During this process, the folder structure should incrementally be changed to
...
-    │       │   │   ├── _aioca.py
-    │       │   │   └── _p4p.py
+    │       │   │   ├── signal_backend.py
+    │       │   │   └── sim.py
     │       │   ├── devices
-    │       │   ├── signals
+    │       │   │   ├── __init__.py
+    │       │   │   ├── device_collector.py
+    │       │   │   └── ...
     │       │   ├── epicsdemo
```
coretl, line 64: Please can we move `ophyd_async.core.epicsdemo` to `ophyd_async.epics.demo`

docs/developer/explanations/decisions/0004-repository-structure.rst
```diff
@@ -55,24 +55,40 @@ During this process, the folder structure should incrementally be changed to
...
     │       │   ├── backends
     │       │   │   ├── __init__.py
-    │       │   │   ├── _aioca.py
-    │       │   │   └── _p4p.py
+    │       │   │   ├── signal_backend.py
+    │       │   │   └── sim.py
     │       │   ├── devices
-    │       │   ├── signals
+    │       │   │   ├── __init__.py
+    │       │   │   ├── device_collector.py
```
coretl, line 62: Please can we make everything under `core` start with an underscore (e.g. `backends` -> `_backends`), so the top level import namespace becomes `from ophyd_async.core import SignalRW, DeviceCollector, ...`

src/ophyd_async/epics/motion/__init__.py
coretl, line 1: Please could we import from `motor` like in `areadetector/__init__.py`

coretl (changes requested):
Pretty much there, I'm more on the "multiple classes per file" side of the fence, but a quick poll of the office suggests I'm in the minority, so I think you are right with your file splitting decisions.

I would like to reduce the import namespace however, so the top level namespace should be:
- `core`
- `epics.signal`
- `epics.areadetector`
- `panda`
and those namespaces should be populated with `__init__.py`s that only import from modules that start with `_`

tests/epics/areadetector/test_hdf_streamer_det.py
RAYemelyanova: I just liked the idea of splitting up the tests. Currently, two test files need the DocHolder which is why I put them as a fixture. You are right; it does complicate the typing... perhaps just for the tests I'll do a bit of code duplication.

docs/developer/explanations/decisions/0004-repository-structure.rst
RAYemelyanova: yep, on it

docs/developer/explanations/decisions/0004-repository-structure.rst
RAYemelyanova: I'm not sure I like this, though; [PEP8](https://peps.python.org/pep-0008/#public-and-internal-interfaces) suggests the following:

> To better support introspection, modules should explicitly declare the names in their public API using the __all__ attribute. Setting __all__ to an empty list indicates that the module has no public API.

> Even with __all__ set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

My takeaway from this is; in the __init__.py for `core` we should define the public API. Any objects/functions/modules not exposed in the `__all__` should be prefixed with `_` to mark them as internal. But what you seem to be suggesting here is, we define our public API in `__all__` but then every object that is being exposed publically actually originates from a module file with an underscore. So we are tricking the reader of our code to think that all the `core` code is internal. I think this is a trick; at least as it currently works, the `DeviceCollector` for example will never be an internal tool, and will have to be used for Devices or classes that inherit from Device for certain functionality.

src/ophyd_async/epics/motion/__init__.py
RAYemelyanova: Yep, on it

tests/epics/areadetector/test_hdf_streamer_det.py
coretl: Or stick it in a `test_utils.py` file and import it in the 2 test files?

docs/developer/explanations/decisions/0004-repository-structure.rst
coretl: It doesn't specifically mention this, but it does imply it

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn’t known in advance).

> When republishing names this way, the guidelines below regarding public and internal interfaces still apply.

I would argue that we are "republish an internal interface as part of a public API", because we want the retain the ability to move functions around private modules, while keeping the same `from ophyd_async.core import DeviceCollector` API.

tests/epics/areadetector/test_hdf_streamer_det.py
RAYemelyanova: For the moment I just did what you initially suggested, and we have two DocHolders. If this becomes more widely used we can move it to a utility file in the tests.

coretl (approved):

Issue #9, opened 2023-09-07
Devices and exposing read/describe methods


RAYemelyanova:
This came out of a discussion with @coretl and @callumforrester:

Do we want devices to look like:

```python
class Sensor(EpicsDevice): #i.e. import from some base class that doesn't currently exist yet but looks something like Device
    """A demo sensor that produces a scalar value based on X and Y Movers"""

    value: A[SignalR[float], READ, pv_suffix("Value")]
    mode: A[SignalRW[EnergyMode], CONFIG, pv_suffix("Mode")]

    # Here is how you define the size of a vector
    value: A[SignalR[float], READ] = epics_signal_r("Value")
    mode: A[SignalRW[EnergyMode], CONFIG] = epics_signal_r("Mode")
    some_vector: A[DeviceVector[SignalRW[float]], CONFIG] = DeviceVector(
        {i: epics_signal_r(...) for i in range(10)}
    )

    describe, read = from_signals_tagged(READ)
    describe_configuration, read_configuration = from_signals_tagged(CONFIG)

    # We think this should probably go in the base class pending discussion with actual users
    # might chat with Dom
    async def read(self) -> Reading:
        reading = {}
        for key, value in self.__annotations__:
            if value is A and READ in value:
                reading[key] = getattr(self, key).read()

    def __init__(self, prefix: str, name="") -> None:
        # Define some signals
        self.value = epics_signal_r(float, prefix + "Value")
        self.mode = epics_signal_rw(EnergyMode, prefix + "Mode")
        # Set name and signals for read() and read_configuration()
        self.set_readable_signals(
            read=[self.value],
            config=[self.mode],
        )
        super().__init__(name=name)
```

i.e. do we make ophyd-async have a similar approach to ophyd in terms of how things like read/describe work? Or do we not define read/describe in the parent class (in this case, `EpicsDevice`) and expect users to overwrite them, either explicitly or with helper methods, because actually most use cases will be unique?

We need to actually ask users and the ophyd people what their thoughts are. Personally, I think it's best for things like read to exist in the parent class, and pull out signals that are annotated with 'READ', and just expose those. Then any new user can just tag a signal with 'READ' and it'll be read. I don't think the default behaviour for the read method should be too controversial because at the end of the day it should just spit out a dictionary with signals that it's read. The controversial bit is, which signals need to be read? In ophyd currently, all signals are read. And that can probably get a bit spammy. 

Open to more discussions on this.

PR #12, opened 2023-09-11
Add ADR for procedural device definition

coretl:
@DominicOram @callumforrester @RAYemelyanova @danielballan @mrakitin @prjemian @tacaswell I have discussed various parts of this with you at various times, and I decided to write down what we have discussed. The final iteration is that the we couldn't think of a non-magic declarative approach to nested Devices. If anyone has a good suggestion then we could revisit this ADR, otherwise I propose we merge this and stick with procedural Devices.


coretl:
Fixes #9

callumforrester (approved):
Build needs fixing but other than that

DominicOram (approved):
Sounds good to me

RAYemelyanova:
going to rebase this on top of the recent changes, then happy to approve

Issue #16, opened 2023-09-14
Add save/load


olliesilvester:
As a follow up to https://github.com/bluesky/ophyd-epics-devices/issues/28, a method should be added to save all PV's of a device to a yaml file, and a load method should be added to set the device to a state specified in the yaml file.

callumforrester:
Co-opting this into a design issue. Writeup of a discussion with @coretl, @tpoliaw and @abbiemery:

Malcolm's current setup is a large bank of settings saved to a yaml file in order and restored in reverse order. Then on top of that, particular per-device logic may be invoked that sets specific signals in a specific order. In other words a broad-stoke followed by a fine-grained approach where needed. This is useful for approximately capturing the desired state without too much code overhead, but it must always be best-effort as the starting state is unknown. It also presents difficulties if the user wishes to leave a particular setting untouched, perhaps because they are tweaking it manually and experimenting. The question is how close ophyd-async's load/save is to malcolm's

There was some discussion about an inverse system: rather than capture all possible state only add explicit logic to set the state we _know_ we care about i.e. state that makes a difference to the scan we want to run. Over time as edge-cases are discovered that add more constraints on the state, we add these to the plan/prepare method/yaml file/whatever is responsible for defining the state. The principal disadvantage of this is it is extra code overhead that must be manually maintained, but hopefully that maintenance is spread out over time.

There was also discussion about decoupling load/save from flyers, so we provide a default implementation that they _can_ use but don't have to, leaving open the possibility of others writing their own version that looks different or not using load/save at all.

coretl:
We talked a lot about requirements that we can't currently check, like "Eiger trigger connected to PandA1 TTLOUT1", that would enable checks like "Sequencer has some connection to TTLOUT1". This is a good idea, but can be put in as a future enhancement.

The decision we need to make now is how to get from "I've poked my EPICS system into a working state and it runs a scan" to "my scan runs when the beamline has been doing something else". There are some options for this:

1. Change settings in Python, either in the plan or in the Device
2. Have an intelligent save, that structures the YAML file into phases, and excludes things based on the plan that will be running
3. Have an intelligent load, that takes a blob of settings as requirements, and knows how to apply them to the device

[...]

PR #18, opened 2023-09-20
Add functions for saving device

olliesilvester:
For the first part of https://github.com/bluesky/ophyd-async/issues/16. Adds a set of functions used to build save plans of a device.


src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,122 @@
...
+            phase_dicts.append(signalRWs)
+
+        # Locate all signals in parallel
+        signals_to_locate: List[SignalRW] = []
+        for phase in phase_dicts:
+            signals_to_locate.extend(phase.values())
+        signal_values = yield Msg("locate", *signals_to_locate)
+        signal_values = [value["setpoint"] for value in signal_values]
+
+        # The table PVs are dictionaries of np arrays. Need to convert these to
```
olliesilvester, line 96: There's probably a much neater way of doing this bit

src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,122 @@
...
+        dot = ""
+        # Place a dot inbetween the upper and lower class.
+        # Don't do this for highest level class.
+        if path_prefix:
+            dot = "."
+        dot_path = f"{path_prefix}{dot}{attr_name}"
+        if type(attr) is SignalRW:
+            signalRWs[dot_path] = attr
+        get_signal_RWs_from_device(attr, path_prefix=dot_path)
+    return signalRWs
```
coretl, line 47: Using a mutable default arg will cause problems:
https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm

I suggest you return a dict and update with the child devices.

```suggestion
    signals: Dict[str, SignalRW] = {}
    for attr_name, attr in device.children():
        dot_path = f"{path_prefix}{attr_name}"
        if isinstance(attr, SignalRW):
            signals[dot_path] = attr
        attr_signals = get_signal_RWs_from_device(attr, path_prefix = dot_path + ".")
        signals.update(attr_signals)
    return signals
```

src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,122 @@
...
+from typing import Any, Dict, List
+
+import yaml
+from bluesky import Msg
+from numpy import ndarray
+
+from ophyd_async.core import Device, SignalRW
+
+
+def get_signal_RWs_from_device(
```
coretl, line 11: I think PEP8 doesn't like uppercase in functions names. How about `walk_rw_signals(device, path_prefix)` instead?

src/ophyd_async/core/_device/device_save_loader.py
coretl: I wouldn't do this here, but in the yaml serialization code

src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,122 @@
...
+        # The table PVs are dictionaries of np arrays. Need to convert these to
+        # lists for easy saving
+        for index, value in enumerate(signal_values):
+            if isinstance(value, dict):
+                for inner_key, inner_value in value.items():
+                    if isinstance(inner_value, ndarray):
+                        value[inner_key] = inner_value.tolist()
+            # Convert enums to their values
+            elif isinstance(signal_values[index], Enum):
+                signal_values[index] = value.value
```
coretl, line 105: Here we should assert `isinstance(str)` as all Enums should be instances of strings

src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,122 @@
...
+        if path_prefix:
+            dot = "."
+        dot_path = f"{path_prefix}{dot}{attr_name}"
+        if type(attr) is SignalRW:
+            signalRWs[dot_path] = attr
+        get_signal_RWs_from_device(attr, path_prefix=dot_path)
+    return signalRWs
+
+
+def save_device(device: Device, savename: str, ignore: List[str] = []):
```
coretl, line 50: I would put this phase sorting and the ignore list in a device specific plan, and also move the yaml saving to another function:
```suggestion
def save_signals(signals: List[Dict[str, SignalRW]], ignore: Sequence[str] = ()) -> List[Dict[str, Any]]:
```

olliesilvester:
Now you make a plan using a series of utility functions in `device_save_loader.py`. I've put some examples in `device_save_loader_test.py`
In general a save plan should:

- Get the RW signals with walk_rw_signals()
- Locate the values of each of these with get_signal_values(), with an optional parameter to ignore specific signals
- Optionally sort the values into separate phases with a device specific function
- Save these values to a yaml file with `save_to_yaml()`

src/ophyd_async/core/_device/device_save_loader.py
```diff
@@ -0,0 +1,132 @@
...
+            # Convert enums to their value
+            elif isinstance(value, Enum):
+                assert isinstance(
+                    value.value, str
+                ), "Enum value did not evaluate to string"
+                phase[key] = value.value
+        phase_outputs.append(phase)
+
+    with open(save_path, "w") as file:
+        yaml.dump(phase_outputs, file)
```
coretl, line 132: I think you can do the numpy formatting here with a custom yaml dumper, at least you can in ruamel.yaml, not sure about pyyaml. That might be cleaner than preprocessing the structures by hand...

coretl: Maybe? Although if we make sure we define `MyEnum(str, Enum)` then it should just serialize as it is?

olliesilvester: If we do this, it will automatically serialize it as, eg
```
 parent_sig3: !!python/object/apply:test_device_save_loader.EnumTest
  - val1
```
and can then rereconstruct the Enum and its value through yaml.load. Otherwise I can keep the logic which converts the enum object to a simple string when serialising.

coretl: Ok, we want to make sure the yaml contains a plain string, so I guess we do need to do that pre-processing

olliesilvester: OK, this is back now

tests/core/_device/test_device_save_loader.py
```diff
@@ -0,0 +1,144 @@
...
+            "child2.sig1": table_pv,
+            "parent_sig3": "val1",
+            "parent_sig1": None,
+        }
+
+        save_to_yaml(values, path.join(tmp_path, "test_file.yaml"))
+
+    RE(save_my_device())
+
+    yaml.add_constructor("tag:!numpy.ndarray", ndarray_constructor)
```
coretl, line 103: Does the YAML file have these tags in it?

coretl: If it does, then we should use a dumper that allows transformation of the data by type (which may fix the enums too):
https://stackoverflow.com/questions/65866546/how-to-make-ruamel-yaml-dump-formatted-numpy-arrays-as-simple-list

jsouter:
I have made a PR from my fork. Still using PyYAML - ruamel.yaml was more difficult to get pretty formatting out of.
https://github.com/bluesky/ophyd-async/pull/27
Now using a custom Dumper subclass that checks if the data is an Enum and if so represents its value member. A custom representer was also added for np.ndarrays to cast it to a list when representing; if needed this logic could be moved into the OphydDumper class's represent_data overload as well.

PR #24, opened 2023-10-06
Move logic pertaining to writing/controlling detectors from i22 branch

rosesyrett:
@coretl has a branch, i22, which contains logic that we need merged into main in order to make a directory provider singleton that can be created in dodal and accessed in blueapi.


tests/epics/areadetector/test_hdf_streamer_det.py
```diff
@@ -109,18 +122,18 @@ async def test_hdf_streamer_dets_step(
 # TODO: write test where they are in the same stream after
 #   https://github.com/bluesky/bluesky/issues/1558
 async def test_hdf_streamer_dets_fly_different_streams(
-    hdf_streamer_dets: List[HDFStreamerDet], RE, doc_holder: DocHolder
+    hdf_streamer_dets: List[StandardDetector[StandardController, HDFWriter]], RE, doc_holder: DocHolder
```
coretl, line 125: So if we lose genericness we would pass the objects directly in here:
```suggestion
    hdf_streamer_dets: List[Tuple[StandardDetector, ADDriver, NDFileHDF]], RE, doc_holder: DocHolder
```

coretl (commented):
I think we have to discuss structuring again, but I think this is a larger discussion than just you and me, so let's get this ready to merge before we do. Thanks for doing this, look good.

src/ophyd_async/core/_detector/detector_writer.py
```diff
@@ -0,0 +1,45 @@
...
+    async def wait_for_index(self, index: int) -> None:
+        """Wait until a specific index is ready to be collected"""
+
+    @abstractmethod
+    async def get_indices_written(self) -> int:
+        """Get the number of indices written"""
+
+    @abstractmethod
+    async def reset_index(self) -> None:
+        """Reset the index count."""
```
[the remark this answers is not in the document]
rosesyrett: Sure; for now I will also delete the place it is used, which is currently in StandardDetector.resume

src/ophyd_async/core/_detector/standard_detector.py
```diff
@@ -0,0 +1,116 @@
...
+        """Stop data writing."""
+        await self.data.close()
+
+    async def pause(self) -> None:
+        """Pause the detector."""
+        await self.control.disarm()
+
+    async def resume(self) -> None:
+        """Resume the detector."""
+        await self.data.reset_index()
```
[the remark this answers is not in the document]
rosesyrett: ah yes, I've just commented on this above. Will do.

src/ophyd_async/epics/areadetector/pilatus.py
```diff
@@ -0,0 +1,26 @@
...
+from ophyd_async.epics.areadetector.writers.nd_file_hdf import NDFileHDF
+from ophyd_async.epics.areadetector.writers.nd_plugin import NDPluginStats
+
+from .controllers.pilatus_controller import PilatusController
+from .drivers.pilatus_driver import PilatusDriver
+
+dp = StaticDirectoryProvider("/dls/p45/data/cmxxx/i22-yyy", "i22-yyy-")
+
+
+class Pilatus(StandardDetector):
```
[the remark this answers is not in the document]
rosesyrett: Here, a class make much more sense to me. I think it's because I can see a factory function like hdf_stats_pilatus and see that it returns something that is a StandardDetector... but I won't know (unless I look into the factory function itself) that this StandardDetector will have a .hdf and a .drv component, for example. In fact, in the original code you have for this, this will be opaque for me anyways because the syntax...

```
def hdf_stats_pilatus(...):
    hdf = ...
    drv = ...
    det = StandardDetector( ..., hdf=hdf, drv=drv)
    return det
```

is kind of opaque. In the background, StandardDetector sets `self.hdf` but just to someone who hasn't looked at the code this is so unreadable. 

I can understand your point of, maybe I shouldn't need to know it has a .drv and a .hdf, but just that it has all the same methods as a StandardDetector, and those are the ones I should use. But, the argument at the end of the day is; who are our users? I know from your perspective it's scientists at beamlines. And from my perspective it also includes people writing and expanding on these devices for example for dodal, who will need to interact and expand on this code beyond the command line.

tests/epics/areadetector/test_hdf_streamer_det.py
```diff
@@ -35,47 +43,52 @@ def doc_holder():
...
-            hdf=NDFileHDF("PREFIX1:HDF"),
-            dp=dp,
-        )
-        detb = HDFStreamerDet(
-            drv=ADDriver(prefix="PREFIX1:DET"),
-            hdf=NDFileHDF("PREFIX1:HDF"),
-            dp=dp,
-        )
+        drv = ADDriver(prefix="PREFIX1:DET")
+        writer = HDFWriter(NDFileHDF("PREFIX1:HDF"), dp, lambda: "test", ADDriverShapeProvider(drv))
```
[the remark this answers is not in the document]
rosesyrett: ah ok, was the drv for `detb` going to be pointing at, e.g. "PREFIX2:DET"?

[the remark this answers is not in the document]
coretl: yes

src/ophyd_async/core/_detector/standard_detector.py
```diff
@@ -0,0 +1,116 @@
...
+from .._device.device import Device
+from .._signal.signal import SignalR
+from ..async_status import AsyncStatus
+from ..utils import merge_gathered_dicts
+from .detector_control import C, DetectorTrigger
+from .detector_writer import D
+
+
+class StandardDetector(
+    Generic[C, D],
```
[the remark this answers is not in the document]
rosesyrett: Anyone who creates a `StandardDetector`, via a factory method or a class method, will need to know the control and data layers (i.e. how the detector is controlled via a driver, and how data is written). i.e. they're going to have to do something like drv= ADDriver(...) and writer = HDFWriter(...) so they will know about it anyways.

Adding the generic here just means, if they want to write plans using this detector, they can have proper typing on it.

I don't see a negative here. Let's say we just have a factory function as you suggest, and this factory function creates a StandardDetector with the pilatus specific control and data components. The user, in a command line, won't care that their "det" detector object from their factory function now has a typed det.control.some_specific_method_to_pilatus. They'll just want to use (probably) the explicit methods of StandardDetector, probably by using it in a plan.

However, a developer making another kind of detector, and maybe making some intricate plan for it, probably will want mypy to pick up that det.control.some_specific_method exists.

src/ophyd_async/epics/areadetector/pilatus.py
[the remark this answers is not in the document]
coretl: There is a key question here, is CLI access part of our type-safe interface? If so, then we need to use your approach. If not, and it's just for scientists, then we don't. The fundamental part of this is, if you add a new plugin to the chain of an areaDetector, does it become a different sort of object? I would argue no, because the logic is no different, but I can see the argument that it is different, in which case a new class could make sense. 

If we go down the class route then we should be explicit about setting things on self, and also be clear that there are going to be many sorts of Pilatus:

```python
class HDFStatsSumPilatus(StandardDetector):
    def __init__(self, prefix: str, name: str = ""):
        self.drv = PilatusDriver(prefix + "DRV:")
        self.stats = NDPluginStats(prefix + "STATS:"),
        self.hdf = NDFileHDF(prefix + "HDF:")
        super().__init__(
            PilatusController(self.drv),
            ADHDFWriter(
                self.hdf, dp, lambda: self.name, ADDriverShapeProvider(drv), sum="NDStatsSum"
            ),
            config_sigs=[drv.acquire_time, drv.acquire],
            name=name
        )
```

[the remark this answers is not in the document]
rosesyrett: After discussion with @coretl and @callumforrester, have agreed to go ahead with classes for now (but name it something more specific like the above)

src/ophyd_async/core/_detector/standard_detector.py
```diff
@@ -0,0 +1,108 @@
...
+            instance of class which inherits from :class:`DetectorControl`
+        data:
+            instance of class which inherits from :class:`DetectorData`
+        config_sigs:
+            configuration signals to be used for describing/reading the detector.
+        name:
+            detector name
+        """
+        self.control = control
+        self.data = data
```
rosesyrett, line 67: Make these private

```python
@property
def control:
    return self._control
```

src/ophyd_async/core/_detector/standard_detector.py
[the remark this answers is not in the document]
rosesyrett: After discussion with @coretl and @callumforrester , have decided to not use Generics in this case. Generics are useful if other bits of code (not tests) need to know about specific implementations, which is not the case here.

src/ophyd_async/core/detector.py
```diff
@@ -0,0 +1,176 @@
...
+    NOTE: only for step-scans.
+    """
+
+    def __init__(
+        self,
+        control: DetectorControl,
+        data: DetectorWriter,
+        config_sigs: Sequence[SignalR],
+        name: str = "",
+        **plugins: Device,
```
coretl, line 118: I think we decided to ditch `**plugins` and have people inherit from `StandardDetector` instead

src/ophyd_async/core/detector.py
```diff
@@ -0,0 +1,176 @@
...
+    Configurable,
+    Readable,
+    Triggerable,
+    WritesExternalAssets,
+):
+    """Detector with useful default behaviour.
+
+    Can be supplied extra devices such that they can also be accessed on the detector.
+    This allows for similar on-the-fly device generation as what is currently done for
+    the ophyd_async.panda.PandA object.
```
coretl, line 102: Not sure what this refers to, `config_sigs` or `plugins`?

src/ophyd_async/core/sim_signal_backend.py
```diff
@@ -152,14 +152,18 @@ def _set_value(self, value: T):
         if self.callback:
             self.callback(reading, self._value)
 
+    def _get_value(self):
+        """Method to bypass asynchronous logic, designed to only be used in tests."""
+        return self.converter.value(self._value)
+
```
coretl, line 158: Is this used anywhere?

src/ophyd_async/epics/areadetector/controllers/pilatus_controller.py
```diff
@@ -0,0 +1,42 @@
...
+    DetectorTrigger.variable_gate: TriggerMode.ext_enable,
+}
+
+
+class PilatusController(DetectorControl):
+    def __init__(self, drv: PilatusDriver) -> None:
+        self.driver = drv
+
+    async def get_deadtime(self, exposure: float) -> float:
+        return 0.002
```
coretl, line 26: I got it wrong, should be
```suggestion
        return 0.001
```

src/ophyd_async/epics/areadetector/controllers/standard_controller.py
```diff
@@ -0,0 +1,42 @@
...
+    DetectorControl,
+    DetectorTrigger,
+    set_and_wait_for_value,
+    wait_for_value,
+)
+
+from ..drivers.ad_driver import ADDriver, ImageMode
+
+
+class StandardController(DetectorControl):
```
coretl, line 16: Not sure what this is for. It seems areaDetector specific so shouldn't be named "Standard". If we want to share between the different areaDetector drivers then I suggest we abstract the contents of `arm` and `disarm` into `arm_ad_driver_for_frames` and `disarm_ad_driver` utility functions but still keep a `DetectorControl` subclass per sort of detector

tests/core/test_flyer.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # now check that you can do the first half of a plan, pause it, and it unstages and prepares again.
+
+
+""" For the moment, this test doesnt work.
+
+This is because you can't add a scanspec Path and an int.
+Flyers set a _current_frame to 0 by default, but then expect to be able to add
+any generic T value to it from _prepare...
+"""
```
coretl, line 224: Feel free to exclude the scanspec bit from this PR, it's not required for the first i22 experiment, but I thought it was when I started...

coretl (commented):
Looks good thanks, just a few points

tests/core/test_flyer.py
rosesyrett: Well that'll help the test coverage, so will do for now :+1:

src/ophyd_async/epics/areadetector/controllers/standard_controller.py
rosesyrett: will rename to ADController for now

src/ophyd_async/core/sim_signal_backend.py
rosesyrett: nevermind; turns it out is used!

rosesyrett:
getting the following linting errors:

```
src/ophyd_async/core/detector.py:166: error: Return type "AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]" of "collect_asset_docs" incompatible with return type "Iterator[Iterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]] | AsyncIterator[Iterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" in supertype "WritesExternalAssets"  [override]
src/ophyd_async/core/detector.py:169: error: "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/ophyd_async/core/detector.py:169: note: Maybe you forgot to use "await"?
src/ophyd_async/core/flyer.py:109: error: Argument 1 to "gather_list" has incompatible type "Awaitable[list[AsyncStatus]]"; expected "Iterable[Awaitable[Any]]"  [arg-type]
src/ophyd_async/core/flyer.py:109: note: Maybe you forgot to use "await"?
src/ophyd_async/core/flyer.py:113: error: Return type "AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]" of "collect_asset_docs" incompatible with return type "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" in supertype "DetectorGroupLogic"  [override]
src/ophyd_async/core/flyer.py:124: error: "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/ophyd_async/core/flyer.py:124: note: Maybe you forgot to use "await"?
src/ophyd_async/core/flyer.py:229: error: Return type "AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]" of "collect_asset_docs" incompatible with return type "Iterator[Iterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]] | AsyncIterator[Iterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" in supertype "WritesExternalAssets"  [override]
src/ophyd_async/core/flyer.py:231: error: "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/ophyd_async/core/flyer.py:231: note: Maybe you forgot to use "await"?
tests/core/test_flyer.py:94: error: Return type "AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]" of "collect_stream_docs" incompatible with return type "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" in supertype "DetectorWriter"  [override]
tests/core/test_async_status.py:68: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/ophyd_async/epics/areadetector/writers/hdf_writer.py:89: error: Return type "AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]" of "collect_stream_docs" incompatible with return type "Coroutine[Any, Any, AsyncIterator[tuple[Literal['resource'], Any] | tuple[Literal['datum'], Any] | tuple[Literal['stream_resource'], Any] | tuple[Literal['stream_datum'], Any]]]" in supertype "DetectorWriter"  [override]
```

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
```diff
@@ -0,0 +1,42 @@
...
+
+from ..drivers.ad_driver import ADDriver, ImageMode
+
+
+class ADController(DetectorControl):
+    def __init__(self, drv: ADDriver) -> None:
+        self.driver = drv
+
+    def get_deadtime(self, exposure: float) -> float:
+        return 0.002
```
rosesyrett, line 21: should this return Optional[float]?

tests/epics/areadetector/test_area_detector.py
```diff
@@ -0,0 +1,227 @@
...
+async def make_detector(prefix="", name="test"):
+    dp = StaticDirectoryProvider(CURRENT_DIRECTORY, f"test-{new_uid()}")
+
+    async with DeviceCollector(sim=True):
+        drv = ADDriver(prefix=f"{prefix}:DET:")
+        hdf = NDFileHDF(f"{prefix}:HDF:")
+        writer = HDFWriter(hdf, dp, lambda: name, ADDriverShapeProvider(drv))
+        det = StandardDetector(
+            ADController(drv), writer, config_sigs=[drv.acquire_time]
+        )
```
rosesyrett, line 36: this should belong in epics.areadetector as its own area_detector.py

docs/examples/ad_demo.py
```diff
@@ -25,7 +25,7 @@
 
 # Create v2 devices
 with DeviceCollector():
-    det3 = areadetector.HDFStreamerDet(pv_prefix)
+    det3 = areadetector.Pilatus(pv_prefix)
 
 
 # And a plan
```
rosesyrett, line 31: This is probably wrong; we shouldn't be using StandardDetectors in flyscans

src/ophyd_async/core/detector.py
```diff
@@ -0,0 +1,178 @@
...
+class DetectorControl(ABC):
+    @abstractmethod
+    def get_deadtime(self, exposure: float) -> float:
+        """For a given exposure, how long should the time between exposures be"""
+
+    @abstractmethod
+    async def arm(
+        self,
+        trigger: DetectorTrigger = DetectorTrigger.internal,
+        num: int = 0,
```
rosesyrett, line 44: change to optional int, to keep with convention of timeout being able to be None.

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
rosesyrett: resolution; This is only used for fly scanning, at which point a fly scannable detector must have some deadtime, and it cannot be 0.

rosesyrett:
once tom's async-status branch is rebased on top of bluesky, lint will pass

docs/examples/ad_demo.py
coretl: We should probably create a `class DemoSimDetector(StandardDetector)` here to demonstrate how to do areaDetector, then use it in step scans. For the fly scan this plan is now wrong, and should be replaced by something that constructs a `HardwareTriggeredFlyable` then calls `fly_and_collect` on it. This is probably something for another PR so feel free to delete this file in this one

src/ophyd_async/core/sim_signal_backend.py
coretl: Where is it used apart from `self.get_value()`?

src/ophyd_async/epics/areadetector/areadetector.py
coretl, line 1: I think this belongs in the demo module as a `DemoSimDetector`

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
coretl, line 1: I think this is `SimDetectorControl`

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
```diff
@@ -0,0 +1,42 @@
...
+    async def arm(
+        self,
+        trigger: DetectorTrigger = DetectorTrigger.internal,
+        num: int = 0,
+        exposure: Optional[float] = None,
+    ) -> AsyncStatus:
+        frame_timeout = DEFAULT_TIMEOUT + await self.driver.acquire_time.get_value()
+        await asyncio.gather(
+            self.driver.num_images.set(num),
+            self.driver.image_mode.set(ImageMode.single),
```
coretl, line 32:
```suggestion
            self.driver.image_mode.set(ImageMode.multiple),
```

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+        return await set_and_wait_for_value(
+            self.driver.acquire, True, timeout=frame_timeout
+        )
+
+    async def disarm(self):
+        # wait=False means don't caput callback. We can't use caput callback as we
+        # already used it in arm() and we can't have 2 or they will deadlock
+        await self.driver.acquire.set(False, wait=False)
+        await wait_for_value(self.driver.acquire, False, timeout=1)
```
coretl, line 42: These 2 lines are likely to be used in many areaDetector `ADControl` subclasses, should they go in a utility function with a suitable name?

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
```diff
@@ -0,0 +1,42 @@
...
+class ADController(DetectorControl):
+    def __init__(self, driver: ADDriver) -> None:
+        self.driver = driver
+
+    def get_deadtime(self, exposure: float) -> float:
+        return 0.002
+
+    async def arm(
+        self,
+        trigger: DetectorTrigger = DetectorTrigger.internal,
```
coretl, line 25: Should assert `trigger == DetectorTrigger.internal` as we don't handle anything else

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
coretl: We could raise an exception to make it clear that a `SimDetector` cannot be hardware triggered

src/ophyd_async/epics/areadetector/areadetector.py
```diff
@@ -0,0 +1,45 @@
...
+    async def connect(self, sim: bool = False):
+        await super().connect(sim=sim)
+        driver = self._controller.driver
+        self._config_sigs = [
+            *self._config_sigs,
+            *[
+                (getattr(signal, "name"), cast(SignalR, signal))
+                for signal in [driver.acquire_time, driver.acquire]
+            ],
+        ]
```
coretl, line 45: What does this do?

src/ophyd_async/core/sim_signal_backend.py
rosesyrett: aha. I've now removed it... must have had a brain fart moment, as I remember removing it and seeing tests fail, and not investigating to discover that self.get_value is the only place this is called. Whoops. Oh well, it's removed now.

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
rosesyrett: Could I modify `set_and_wait_for_value` to take an optional `wait` kwarg, so that we can replace these two with a `set_and_wait_for_value call`?

src/ophyd_async/epics/areadetector/areadetector.py
coretl: I guess this is something to do with `config_sigs` becoming `Sequence[Tuple[name, SignalR]]`. Why did it change from `Sequence[SignalR]`? I was envisioning that `config_sigs` would be named and connected as per its position in the tree, and so `StandardDetector` didn't have to do anything apart from read them.

E,g.
```python
class MyDet(StandardDetector):
    def __init__(...):
        self.drv = MyDrv(...)
        ...
        super().__init__(
            config_sigs = [self.drv.acquire_time, self.drv.gain]
        )
```
Now if we name our `MyDet` instance `"my_det"` then we would get drv and all its children connected in `connect()`, named in `set_name()`, and `my_det-drv-acquire_time` and `my_det-drv-gain` would appear as names in `read()`.

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
coretl: That would make sense from a code point of view, but I worry about writing `await set_and_wait_for_value(self.drv.acquire, False, wait=False)`. 3 waits in the same line makes my brain hurt! I'd prefer a separate function so we could write `await stop_busy_record(self.drv.acquire)` which makes more sense for an EPICS person

src/ophyd_async/epics/areadetector/areadetector.py
coretl: Different sort of SimDetector... This is a proper EPICS areaDetector that makes stripy images. We often use it for demos

coretl: So the name `SimDetector` mirrors the EPICS module. Is `ADSimDetector` better?

coretl: Actually, that's the name of the module: https://github.com/areaDetector/ADSimDetector

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
rosesyrett: Sure, I can understand this. How about,

`await set_and_wait_for_value(self.drv.acquire, False, with_callback=False)`?

src/ophyd_async/epics/areadetector/areadetector.py
rosesyrett: Yep, that name is a bit better. I'll perhaps add a comment at the top of the file specifying 'ADSimDetector != Simulated ADDetector' or something

rosesyrett: to add to this... should I change ADDriver to ADSimDriver? or leave as is?

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
rosesyrett: can we compromise at `ADSimController`? Will go with SimDetectorControl if you feel it's best, I'm only apprehensive because it uses an ADDriver which seems area detector specific.

src/ophyd_async/epics/areadetector/areadetector.py
rosesyrett: config_sigs could be any signals; they could be signals associated with the detector, or not. For this reason, when doing something like:

```
with DeviceCollector():
    my_detector = Detector(..., config_sigs=[signal_a, signal_b])
```

I realised that if signal_a and b hadn't been connected and named properly, there would be problems with the bluesky bundler when doing things like opening a run. So, I added a way to ensure they were connected and named properly; by passing these config signals as a tuple of the name you'd like to appear on a descriptor doc, and the signal, the StandardDetector class ensures they are named and connected properly.

Perhaps it's easier to discuss this one offline; but the way I see it we have two approaches to this problem:
1. constrain config_sigs to just be stuff that the detector knows about.
2. raise an error somewhere (in bluesky, maybe?) if we try to use config sigs that aren't named, and another error if we try to use config_sigs that haven't been connected.
3. do what I've done above; allow any config sigs as long as I also get a tuple with a name, so I know what to name them. Because the DeviceCollector gets that from a context.

Thinking about it, number 2 might be best. Effectively the error I get from the bundler boils down to the fact that, if the signals aren't named, we end up with a dictionary for config sigs with just one key ("") and something really doesn't like it being empty. Or, I think I want to make a test case for this, fix it, make a PR with bluesky but still keep what I've done above. Would like to know your thoughts on it.

src/ophyd_async/epics/areadetector/controllers/ad_controller.py
rosesyrett: resolution; I've made a new method in area detector utils, stop_busy_record, which does essentially this.

src/ophyd_async/epics/areadetector/areadetector.py
rosesyrett: For the moment, I've added a function that, each time you stage, checks that the config signals are connected and named.

But doing what you suggests sound like a good idea also.

src/ophyd_async/epics/areadetector/areadetector.py
rosesyrett: Names are now changed

src/ophyd_async/core/detector.py
rosesyrett: Not bothering at the moment as this is an internal mechanism

PR #25, opened 2023-10-11
Add FlyingPanda implementation files

keithralphs:


rosesyrett (approved):
looks fine, when tests pass just merge (even if codecov complains)

rosesyrett:
although the tests shouldn't be failing... I think this may be happening because event model has had a new release? But not sure. Will investigate later (please feel free to do that instead)

rosesyrett:
closing as this isn't the correct repository for a flying panda; this has been replaced with the I22 panda implementation now in dodal.

PR #29, opened 2023-10-13
Add attribute xml

coretl:


coretl:
This is for writing the NDAttributesXML file for the SAXS detector so it stamps temperature on every frame:
```python
def make_saxs_linkam_stamping_xml():
    xml = NDAttributesXML()
    xml.add_epics_pv(
        "Temperature",
        "BL22I-EA-TEMPC-01:TEMP",
        description="Current linkam temperature",
    )
    with open("/dls_sw/p38/software/bluesky/saxs_linkam_stamping.xml", "w") as f:
        f.write(str(xml))
```

coretl:
The we will `saxs.drv.ndattributesxml.set("/dls_sw/p38/software/bluesky/saxs_linkam_stamping.xml")` in the plan, either directly or as part of the config loading

PR #31, opened 2023-10-16
Rebase add_save_load on add-detector-logic

jsouter:
Rebased on add-detector-logic, tests are passing on my end but let me know if things still need moving around


coretl:
Please can you fix linting. @rosesyrett when that is fixed shall we merge?

jsouter:
> Please can you fix linting. @rosesyrett when that is fixed shall we merge?

Linting is fixed now

src/ophyd_async/core/signal_backend.py
```diff
@@ -35,6 +35,10 @@ async def get_reading(self) -> Reading:
     async def get_value(self) -> T:
         """The current value"""
 
+    @abstractmethod
+    async def get_setpoint(self) -> T:
+        """The current setpoint"""
```
rosesyrett, line 40: without going into bluesky.protocols and looking at the definition of a `Location` protocol and its comments for `setpoint` and `readback`, this docstring is a little confusing. Could we perhaps replace it with:

```suggestion
    async def get_setpoint(self) -> T:
        """The point that a signal was requested to move to."""
```

src/ophyd_async/core/signal_backend.py
```diff
@@ -35,6 +35,10 @@ async def get_reading(self) -> Reading:
     async def get_value(self) -> T:
         """The current value"""
```
rosesyrett, line 36: In line with making setpoint and readback slightly less confusing, could we replace this with:

```suggestion
    async def get_value(self) -> T:
        """The readback value of the signal."""
```

src/ophyd_async/core/sim_signal_backend.py
```diff
@@ -165,6 +165,9 @@ async def get_reading(self) -> Reading:
     async def get_value(self) -> T:
         return self._get_value()
 
+    async def get_setpoint(self) -> T:
+        return self.converter.value(self._value)
```
rosesyrett, line 169: is it worth making this just call get_value, and adding a comment that for a simsignalbackend these will always be the same?

i.e.:

```suggestion
    async def get_setpoint(self) -> T:
        """For a simulated backend, the setpoint and readback values are the same."""
        return await self.get_value()
```

tests/core/test_device_save_loader.py
```diff
@@ -0,0 +1,197 @@
...
+        phase_1 = {"child1.sig1": values["child1.sig1"]}
+        phase_2 = {"child2.sig1": values["child2.sig1"]}
+        return [phase_1, phase_2]
+
+    setattr(device, "sort_signal_by_phase", sort_signal_by_phase)
+    return device
+
+
+async def test_enum_yaml_formatting(tmp_path):
+    enums = [EnumTest(EnumTest.VAL1), EnumTest(EnumTest.VAL2)]
```
rosesyrett, line 66: I think it's easier to read this as:
```suggestion
    enums = [EnumTest.VAL1, EnumTest.VAL2]
```

tests/core/test_device_save_loader.py
```diff
@@ -0,0 +1,197 @@
...
+    assert isinstance(enums[0], EnumTest)
+    save_to_yaml(enums, path.join(tmp_path, "test_file.yaml"))
+    with open(path.join(tmp_path, "test_file.yaml"), "r") as file:
+        yaml_content = yaml.load(file, yaml.Loader)
+        assert isinstance(yaml_content, list)
+        assert yaml_content[0] == EnumTest.VAL1
+        assert yaml_content[1] == EnumTest.VAL2
```
rosesyrett, line 73: these assertions all make sense, but I'd shorten them for readability:

```suggestion
    save_to_yaml(enums, path.join(tmp_path, "test_file.yaml"))
    with open(path.join(tmp_path, "test_file.yaml"), "r") as file:
        saved_enums = yaml.load(file, yaml.Loader)
    assert saved_enums == enums
```

This effectively combines the assert statements you currently have on lines 71-73, and removes the assert statement on 67 since that will never raise an AssertionError unless there is something very wrong with python or someone's redefined EnumTest... Up to you if you want to rename yaml_content or not

tests/core/test_device_save_loader.py
```diff
@@ -0,0 +1,197 @@
...
+        phases = device_with_phases.sort_signal_by_phase(device_with_phases, values)
+        save_to_yaml(phases, path.join(tmp_path, "test_file.yaml"))
+
+    RE(save_my_device())
+
+    with open(path.join(tmp_path, "test_file.yaml"), "r") as file:
+        yaml_content = yaml.load(file, yaml.Loader)
+        assert type(yaml_content[0]["child1.sig1"]) is str
+        assert type(yaml_content[1]["child2.sig1"]["VAL1"]) is list
+        assert type(yaml_content[1]["child2.sig1"]["VAL2"]) is list
```
rosesyrett, line 197: not sure about the language here r.e. phases. It seems like these three tests share a lot in common, and can probably be combined into one test and still cover the same lines.

to me, these tests only differ from the previous few because they test if save_to_yaml can handle a List[Dict[str, Any]] as well as a Dict[str, Any]. Which is fine, but why the word 'phase'? Would changing 'phase' to 'list' be better in this instance?

rosesyrett (changes requested):
Looks good, I like how only SignalRW's are saved, that's very neat. I have a few comments about docstrings and a few about the tests, as I use tests to work out how to use a new bit of code so to me it's important they're readable and clear.
Very nearly there.

tests/core/test_device_save_loader.py
jsouter: Ah you're right, was mistaken about how Python Enums work

tests/core/test_device_save_loader.py
jsouter: Yep, I can reduce these tests down. I believe the idea with "phases" is that when loading values from a yaml to a Device (which is yet to be implemented) we want to load the groups of signals in a specified order, so all the phase 1 signals first, then phase 2 signals etc. There's possibly a clearer way to express this.

tests/core/test_device_save_loader.py
jsouter:
```
async def test_enum_yaml_formatting(tmp_path):
    enums = [EnumTest.VAL1, EnumTest.VAL2]
    save_to_yaml(enums, path.join(tmp_path, "test_file.yaml"))
    with open(path.join(tmp_path, "test_file.yaml"), "r") as file:
        saved_enums = yaml.load(file, yaml.Loader)
    # check that save/load reduces from enum to str
    assert all(isinstance(value, str) for value in saved_enums)
    # check values of enums same
    assert saved_enums == enums
```
How about this? I'd like to have something that shows that reduction to a string when saving to yaml

jsouter:
I've made those changes and added in Ollie's load_device plan as load_from_yaml, with a new test.

src/ophyd_async/core/device_save_loader.py
```diff
@@ -0,0 +1,175 @@
...
+class OphydDumper(yaml.Dumper):
+    def represent_data(self, data: Any) -> Any:
+        if isinstance(data, Enum):
+            return self.represent_data(data.value)
+        return super(OphydDumper, self).represent_data(data)
+
+
+def get_signal_values(
+    signals: Dict[str, SignalRW[Any]], ignore: Optional[List[str]] = None
+) -> Union[Generator[Dict[str, Any], None, None], Dict[str, Any]]:
```
coretl, line 27: I think this is always a generator. @callumforrester is this right for a plan?

```suggestion
) -> Generator[Msg, Sequence[Location], Dict[str, Any]]:
```

src/ophyd_async/core/device_save_loader.py
```diff
@@ -0,0 +1,175 @@
...
+    for attr_name, attr in device.children():
+        dot_path = f"{path_prefix}{attr_name}"
+        if type(attr) is SignalRW:
+            signals[dot_path] = attr
+        attr_signals = walk_rw_signals(attr, path_prefix=dot_path + ".")
+        signals.update(attr_signals)
+    return signals
+
+
+def save_to_yaml(phases: Union[Dict[str, Any], List[Dict[str, Any]]], save_path: str):
```
coretl, line 108: Not sure we ever actually use a plain dict here? Most devices are likely to need phases, so I suggest we make it `Sequence[Dict[str, Any]]` anytime we see `phases`

src/ophyd_async/core/device_save_loader.py
```diff
@@ -0,0 +1,175 @@
...
+
+    save_path : str
+        Path of the yaml file to load from
+
+    See Also
+    --------
+    :func:`ophyd_async.core.save_to_yaml`
+    """
+    with open(save_path, "r") as file:
+        data_by_phase: Sequence[Dict[str, Any]] = yaml.full_load(file)
```
coretl, line 153: Can we split this function here please. `load_from_yaml(save_path: str) -> Sequence[Dict[str, Any]]:`, then a different plan `set_signal_values(signals:  Dict[str, SignalRW], values: Sequence[Dict[str, Any]]) -> Generator[Msg, None, None]:`  that takes `signals` (as provided by `walk_rw_signals`), and `values` (as provided by `load_from_yaml`) and does the `abs_set` logic you do below

jsouter: Okay, have done this rebased onto the last passing commit for add-detector-logic

PR #32, opened 2023-10-16
Move missing types from i22 branch for Pandas

DiamondJoseph:
Required for i22 PandARepeatedTriggerLogic


src/ophyd_async/panda/panda.py
```diff
@@ -86,6 +87,27 @@ class SeqBlock(Device):
     table: SignalRW[SeqTable]
 
 
+@dataclass
+class SeqTableRow:
```
coretl, line 91: This is inspired by @tpoliaw's tables.py, but I chose dataclasses instead of namedtuples for printability and typing

src/ophyd_async/panda/panda.py
```diff
@@ -116,6 +138,75 @@ def _block_name_number(block_name: str) -> Tuple[str, Optional[int]]:
...
+        outa2=np.ndarray([row.outa2 for row in rows], dtype=np.bool_),
+        outb2=np.ndarray([row.outb2 for row in rows], dtype=np.bool_),
+        outc2=np.ndarray([row.outc2 for row in rows], dtype=np.bool_),
+        outd2=np.ndarray([row.outd2 for row in rows], dtype=np.bool_),
+        oute2=np.ndarray([row.oute2 for row in rows], dtype=np.bool_),
+        outf2=np.ndarray([row.outf2 for row in rows], dtype=np.bool_),
+    )
+
+
+def seq_table_from_arrays(
```
coretl, line 163: We will use this method later when we need more efficient batch production of tables for the motion stuff...

src/ophyd_async/panda/panda.py
```diff
@@ -116,6 +138,75 @@ def _block_name_number(block_name: str) -> Tuple[str, Optional[int]]:
     return block_name, None
 
 
+def seq_table_from_rows(*rows: SeqTableRow):
```
coretl, line 141: This is the method we will use for now for i22, as per https://github.com/DiamondLightSource/i22-bluesky/blob/b56865cb26a9cedbe7f4d2c688d269a758421fa6/src/i22_bluesky/panda/fly_scanning.py#L33-L54

coretl (commented):
Should we have some tests too? I know I didn't write any in the first place...

src/ophyd_async/panda/panda.py
DiamondJoseph: Should we use TypedDict for consistency with SeqTable?

DiamondJoseph: Should the table functions/types be moved to panda/tables.py, to keep it a little easier to read?

coretl: > Should we use TypedDict for consistency with SeqTable?

The only issue with this is the lack of defaults. `SeqTable` is a fully populated output to go to the PandA. It can be constructed with `seq_table_from_arrays` to handle consistent lengths. If we made the `SeqTableRow` a `TypedDict` then we'd have to add the defaults in `seq_table_from_rows`. Not sure which I prefer. Thoughts?

DiamondJoseph: Have moved and exported from package but left as is in the interest of time.

DiamondJoseph:
All these changes have made it into main.

PR #38, opened 2023-10-25
Add navigation_with_keys key to html_theme_options

rosesyrett:


docs/conf.py
```diff
@@ -187,6 +187,7 @@
             url="https://blueskyproject.io",
         )
     ],
+    navigation_with_keys=True
```
coretl, line 190: I suggest this should be False as per https://github.com/pydata/pydata-sphinx-theme/issues/1492.

Also, we should put this fix into skeleton too

rosesyrett: done, will make an issue in skeleton(s)

PR #43, opened 2023-10-31
Made panda use pva signal backend instead of pvi_get

rosesyrett:
Make use of PVASignalBackend standard Backend


DiamondJoseph:
Still need, but do not need to get the initial state, just get the pvi.

DiamondJoseph:
@coretl: just checking that I am going in the right direction with this change? I've gotten to the point that the new code looks correct as far as I understand, but the pva fixture isn't playing nicely (timing out tests, not letting the test suite complete). It looks like `INCOMPLETE_BLOCK_RECORD, INCOMPLETE_RECORD ,EXTRA_BLOCKS_RECORD` don't actually exist in `tests/panda/db`?

DiamondJoseph:
Also: is there a standard for "FOO" + ":BAR" vs. "FOO:" + "BAR:" ? The latter seems to be more common in dodal, so I've been using it, assuming the trailing colon won't cause issue.

coretl:
> Also: is there a standard for "FOO" + ":BAR" vs. "FOO:" + "BAR:" ? The latter seems to be more common in dodal, so I've been using it, assuming the trailing colon won't cause issue.
> 
> I see in Rose's documentation change she also enforced that prefix should end with : if it isn't already https://github.com/bluesky/ophyd-async/pull/107/files#diff-fbf444aec05ace8f7baf32ccb7fe37cbb7bb988607c665bb235f89d06dfa1c49

DLS convention used to be no colon in the prefix, then colon added in the database. This only work because DLS PVs look like `BLxxI-MY-DEVICE-01:RECORD`, it doesn't work for all facilities that might use a different separator. Much better is to include the colon in the prefix which is what Rose's change did

src/ophyd_async/panda/panda.py
```diff
@@ -242,6 +228,26 @@ def set_attribute(self, name: str, num: Optional[int], block: Device):
...
+        await backend.connect()
+        pv_info: Dict[str, Dict[str, str]] = await backend.get_value()
+
+        result = {}
+
+        for attr_name, attr_info in pv_info.items():
+            result[attr_name] = PVIEntry(**attr_info)  # type: ignore
+
+        _remove_inconsistent_blocks(result)
+        return result
```
coretl, line 249: Please can you put all the `pvi` related bits into their own file in `ophyd_async.epics.pvi` please? These will be used on things that are not PandA in the future

DiamondJoseph: Assuming that remove_inconsistent_blocks is Panda specific enough to stay, I've moved pvi_get, make_signal, PVIEntry, and the _pvi_mapping.

tests/panda/test_panda.py
```diff
@@ -135,26 +136,9 @@ async def test_panda_block_missing_signals(pva):
...
-    lines = [
-        "await panda.connect(timeout=0.01)",
-        'await pvi(self._init_prefix + ":PVI", self.ctxt, timeout=timeout)',
-        "raise NotConnected(pv) from exc",
-    ]
-
-    for idx, each_frame in enumerate(traceback.extract_tb(exc.tb)):
-        assert each_frame.filename.endswith(files[idx])
-        assert each_frame.name == funcs[idx]
-        assert each_frame.line == lines[idx]
```
DiamondJoseph, line 160: Was this level of where the exception is being thrown actually useful? Since this has been refactored from line 2 downwards, I'm only putting it back if there's serious outcry.

coretl: Can be removed

src/ophyd_async/epics/_backend/_p4p.py
```diff
@@ -141,6 +141,16 @@ def descriptor(self, source: str, value) -> Descriptor:
...
 
 
+class PvaStructureConverter(PvaConverter):
+
+    def value(self, value):
+        return value.get("pvi")
+
+    def descriptor(self, source: str, value) -> Descriptor:
+        # This is wrong, but defer until we know how to actually describe a structure
+        return dict(source=source, dtype="object", shape=[])  # type: ignore
```
coretl, line 151: Please can we turn this into a generic `DictConverter`, then not do the `pvi` extraction here, but do it in the class?

We could also error when returning the descriptor to say it can't be described in scans at the moment

src/ophyd_async/panda/panda.py
```diff
@@ -80,66 +50,31 @@ def _block_name_number(block_name: str) -> Tuple[str, Optional[int]]:
...
-
 class PandA(Device):
-    _ctxt: Optional[Context] = None
-
     pulse: DeviceVector[PulseBlock]
     seq: DeviceVector[SeqBlock]
     pcap: PcapBlock
 
-    def __init__(self, prefix: str, name: str = "") -> None:
+    def __init__(self, prefix: str, name: str) -> None:
```
coretl, line 75: Still want to default to blank name

DiamondJoseph: From BlueAPI I think we'd prefer not to allow defaulting to blank name- especially if there's multiple pandas per beamline.

coretl: But from creation via DeviceCollector then we would like this, so to ease this I suggest we put the name creation and checking in the dodal/blueapi layers, and have an option name in the ophyd layer

DiamondJoseph: ceded

tests/panda/test_panda_utils.py
```diff
@@ -13,7 +13,7 @@
 @pytest.fixture
 async def sim_panda():
     async with DeviceCollector(sim=True):
-        sim_panda = PandA("PANDA")
+        sim_panda = PandA("PANDA", "sim_panda")
```
coretl, line 16:
```suggestion
        sim_panda = PandA("PANDA")
```

tests/epics/test_signals.py
```diff
@@ -332,6 +342,24 @@ async def test_pva_table(ioc: IOC) -> None:
...
+async def test_pvi_structure(ioc: IOC) -> None:
+    if ioc.protocol == "ca":
+        # CA can't do structure
+        return
+    # Make and connect the backend
+    backend = await ioc.make_backend(Dict[str, Any], "pvi")
+    # Make a monitor queue that will monitor for updates
+    q = MonitorQueue(backend)
+    try:
+        # Check initial value
```
coretl, line 354: Where's the check of the actual value?

DiamondJoseph:
Test for getting the PVI info as a PVADict currently times out. I don't know enough about Epics DB Records to diagnose

coretl:
@evalott100 please could you take a look at the test failure?

tests/epics/test_signals.py
```diff
@@ -97,9 +100,17 @@ async def assert_updates(self, expected_value):
...
+        assert (
+            value
+            == expected_value
+            == asyncio.wait_for(self.backend.get_value(), timeout=5)
+        )
+        assert (
+            reading
+            == expected_reading
+            == asyncio.wait_for(self.backend.get_reading(), timeout=5)
+        )
```
coretl, line 113:
```suggestion
        reading, value = await asyncio.wait_for(self.updates.get(), timeout=5)
        assert (
            value
            == expected_value
            == await asyncio.wait_for(self.backend.get_value(), timeout=5)
        )
        assert (
            reading
            == expected_reading
            == await asyncio.wait_for(self.backend.get_reading(), timeout=5)
        )
```

src/ophyd_async/epics/_backend/_p4p.py
```diff
@@ -141,6 +140,20 @@ def descriptor(self, source: str, value) -> Descriptor:
         return dict(source=source, dtype="object", shape=[])  # type: ignore
 
 
+class PvaDictConverter(PvaConverter):
+    def reading(self, value):
+        value = value.todict()
+        ts = time.time()
```
DiamondJoseph, line 146: So this isn't coming from the signal?

tests/epics/test_signals.py
```diff
@@ -360,8 +358,15 @@ async def test_pvi_structure(ioc: IOC) -> None:
...
                 "rw": f"{PV_PREFIX}:{ioc.protocol}:height",
             },
-        }
+        },
+        "record": {
+            "_options": {
+                "atomic": ANY,  # the field can be true or false
+                "queueSize": 0,
+            },
+        },
```
DiamondJoseph, line 367: Is the atomic not deterministic from the dbrecord?

I guess it's only the pvi entry that is used, I'm just trying to understand

DiamondJoseph (commented):
Tom should Approve

src/ophyd_async/epics/_backend/_p4p.py
evalott100: Nah, the pva table doesn't have a timestamp (or an error state). We just add one ourselves from when we read it.

tests/epics/test_signals.py
evalott100: It's appended when you actually read the pva table:
```
pvinfo PREFIX:SEQ1:PVI
PREFIX:SEQ1:PVI
Server: 172.23.244.139:5075
Type:
    structure
        structure record
            structure _options
                int queueSize
                boolean atomic
        structure pvi
            structure active
                string r
            structure bita
                string rw
            structure bita_delay
```

`_atomic` differs between using the `backend.get_reading()` and `backend.update()`:
```
backend reading True
update value False
update reading False
backend value True
```

tests/epics/test_signals.py
```diff
@@ -332,6 +340,44 @@ async def test_pva_table(ioc: IOC) -> None:
...
+            "height": {
+                "rw": f"{PV_PREFIX}:{ioc.protocol}:height",
+            },
+        },
+        "record": {
+            "_options": {
+                "atomic": ANY,  # the field can be true or false
+                "queueSize": 0,
+            },
+        },
```
coretl, line 367: I think we don't care about any of the record specific bits

```suggestion
        "record": ANY
```

abbiemery (approved):

PR #46, opened 2023-11-01
Fix flyer

rosesyrett:


rosesyrett:
I know I've requested reviews, but I'm actually just going to merge this and make a release because this makes the linkam plan work with a tetram on i22. I will want to, in future, add some documentation about this and why the change was needed.

rosesyrett:
docs is failing but I can't recreate it locally. I'm going to merge this and fix the docs build later as it is out of date anyway and the error seems completely unrelated to my (tiny) change. For some reason this has been a common annoyance with docs recently.

PR #50, opened 2023-11-01
Add timeout to HDF5 plugin in AreaDetector and check detector final state after acquisition

callumforrester:
Changes:
- Check areadetector state after acquisition,
- Raise if the state is not valid
- Add timeout to waiting on HDF plugin after acqusition
- Raise if timeout exceeded

### DIFF
--- a/src/ophyd_async/epics/areadetector/drivers/ad_base.py
+++ b/src/ophyd_async/epics/areadetector/drivers/ad_base.py
@@ -15,6 +15,11 @@ from ..writers.nd_plugin import NDArrayBase
 
 
 class DetectorState(str, Enum):
+    """
+    Default set of states of an AreaDetector driver.
+    See definition in ADApp/ADSrc/ADDriver.h in https://github.com/areaDetector/ADCore
+    """
+
     Idle = "Idle"
     Acquire = "Acquire"
     Readout = "Readout"
@@ -28,6 +33,8 @@ class DetectorState(str, Enum):
     Aborted = "Aborted"
 
 
+#: Default set of states that we should consider "good" i.e. the acquisition
+#  is complete and went well
 DEFAULT_GOOD_STATES: FrozenSet[DetectorState] = frozenset([DetectorState.Idle])
 
 
@@ -52,7 +59,8 @@ async def start_acquiring_driver_and_ensure_status(
     good_states: Set[DetectorState] = set(DEFAULT_GOOD_STATES),
     timeout: float = DEFAULT_TIMEOUT,
 ) -> AsyncStatus:
-    """Start aquiring driver, raising ValueError if the detector is in a bad state.
+    """
+    Start acquiring driver, raising ValueError if the detector is in a bad state.
 
     This sets driver.acquire to True, and waits for it to be True up to a timeout.
     Then, it checks that the DetectorState PV is in DEFAULT_GOOD_STATES, and otherwise
@@ -61,7 +69,7 @@ async def start_acquiring_driver_and_ensure_status(
     Parameters
     ----------
     driver:
-        The driver to start aquiring. Must subclass ADBase.
+        The driver to start acquiring. Must subclass ADBase.
     good_states:
         set of states defined in DetectorState enum which are considered good states.
     timeout:
--- a/src/ophyd_async/epics/areadetector/drivers/ad_base.py
+++ b/src/ophyd_async/epics/areadetector/drivers/ad_base.py
@@ -77,7 +77,7 @@ async def start_acquiring_driver_and_ensure_status(
     status = await set_and_wait_for_value(driver.acquire, True, timeout=timeout)
 
     async def completion_task() -> None:
-        """NOTE: possible race condition here between the callback from 
+        """NOTE: possible race condition here between the callback from
         set_and_wait_for_value and the detector state updating."""
         await status
         state = await driver.detector_state.get_value()
--- a/src/ophyd_async/epics/areadetector/drivers/ad_base.py
+++ b/src/ophyd_async/epics/areadetector/drivers/ad_base.py
@@ -1,11 +1,30 @@
 import asyncio
-from typing import Sequence
+from enum import Enum
+from typing import Sequence, Set
 
-from ophyd_async.core import ShapeProvider
+from ophyd_async.core import AsyncStatus, ShapeProvider
+from ophyd_async.core.signal import set_and_wait_for_value
 
 from ...signal.signal import epics_signal_rw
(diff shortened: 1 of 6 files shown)


src/ophyd_async/core/detector.py
```diff
@@ -171,7 +171,10 @@ async def trigger(self) -> None:
         indices_written = await self.writer.get_indices_written()
         written_status = await self.controller.arm(DetectorTrigger.internal, num=1)
         await written_status
-        await self.writer.wait_for_index(indices_written + 1)
+        await asyncio.wait(
+            [self.writer.wait_for_index(indices_written + 1)],
+            timeout=DEFAULT_TIMEOUT,
+        )
```
rosesyrett, line 177: Should this be `wait` or `wait_for`?

Looking at the api, the only differences are:
1. wait accepts an iterable of futures/tasks, but wait_for accepts a single awaitable.
2. wait doesnt cancel the futures when a timeout occurs, but wait_for does.

I think it's neater to use wait_for personally.
4.

src/ophyd_async/core/detector.py
```diff
@@ -171,7 +171,10 @@ async def trigger(self) -> None:
         indices_written = await self.writer.get_indices_written()
         written_status = await self.controller.arm(DetectorTrigger.internal, num=1)
         await written_status
-        await self.writer.wait_for_index(indices_written + 1)
+        await asyncio.wait(
+            [self.writer.wait_for_index(indices_written + 1)],
```
coretl, line 175: Should we add `timeout` to `DetectorWriter.wait_for_index` so this line becomes `await self.writer.wait_for_index(indices_written + 1, timeout=DEFAULT_TIMEOUT)`? We already added a `timeout` to `wait_for_value` so this is just a case of passing that value down...

src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -18,11 +37,29 @@ def __init__(self, prefix: str, name: str = "") -> None:
         self.array_counter = ad_rw(int, prefix + "ArrayCounter")
         self.array_size_x = ad_r(int, prefix + "ArraySizeX")
         self.array_size_y = ad_r(int, prefix + "ArraySizeY")
+        self.detector_state = ad_r(DetectorState, prefix + "DetectorState")
```
rosesyrett, line 40: I think Tom wants the ADBase here to replicate that of Malcolm? (@coretl )

I also know that this will be useful for the tetramm to have, but currently although the PilatusDriver and AdAravisMakoDriver in dodal (needed for d11 for example, for p38) subclass ADBase, the TetrammDriver does not. So, this functionality wouldn't exist for the tetramm.

I'm not sure where the best place for this would be. I agree it's here. I think there's a larger discussion that needs to happen with @coretl to decide where it ought to live.

src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -18,11 +37,29 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
 
+async def arm_and_trigger_detector_and_check_status_pv(
+    driver: ADBase,
+    good_states: Set[DetectorState] = DEFAULT_GOOD_STATES,
+) -> AsyncStatus:
+    status = await set_and_wait_for_value(driver.acquire, True)
+
+    async def completion_task() -> None:
+        await status
+        state = await driver.detector_state.get_value()
```
coretl, line 54: This might work for now, but there is some code in Malcolm to cope with the race condition between the callback completing and the state updating:
https://github.com/dls-controls/pymalcolm/blob/d46319e9ea6b95c0856bfc2fe930b500b2f7949c/malcolm/modules/ca/parts/caactionpart.py#L98-L114
Maybe put this in as a comment that we could add this if it is a problem?

src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -1,13 +1,32 @@
...
+    Acquire = "Acquire"
+    Readout = "Readout"
+    Correct = "Correct"
+    Saving = "Saving"
+    Aborting = "Aborting"
+    Error = "Error"
+    Waiting = "Waiting"
+    Initializing = "Initializing"
+    Disconnected = "Disconnected"
+    Aborted = "Aborted"
```
rosesyrett, line 24: we've been bitten before by making enums for things and then realising that epics changes enum members at runtime; how comprehensively have these enums been checked? Otherwise on line 40 where you define the detector_state property on ADBase I can imagine some devices will just fail to instantiate because, e.g. the epics screen actually is missing some of these states (or some IOCs have custom ones...)

src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -18,11 +37,29 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
 
 
+async def arm_and_trigger_detector_and_check_status_pv(
+    driver: ADBase,
+    good_states: Set[DetectorState] = DEFAULT_GOOD_STATES,
+) -> AsyncStatus:
+    status = await set_and_wait_for_value(driver.acquire, True)
+
+    async def completion_task() -> None:
+        await status
```
rosesyrett, line 53: will this line not hang, meaning the rest of the logic never gets processed? 

I just remember that originally in the hdf writer wait_for_index it seems like we hung at await set_and_wait_for_value(...)

rosesyrett (changes requested):
We need to talk to @coretl about this. I agree we should change it, I'm not sure how best to.

src/ophyd_async/epics/areadetector/drivers/ad_base.py
rosesyrett: After speaking with @coretl seems like this is the right place after all :)

src/ophyd_async/epics/areadetector/drivers/ad_base.py
rosesyrett: Again after speaking with @coretl seems like these enums are pretty standard. If we come across issues we can always re-define the enum and the DetectorState definition by subclassing ADBase.

src/ophyd_async/core/detector.py
rosesyrett: solved in https://github.com/bluesky/ophyd-async/pull/55

src/ophyd_async/core/detector.py
rosesyrett: Implemented in https://github.com/bluesky/ophyd-async/pull/55

rosesyrett (approved):

PR #51, opened 2023-11-01
Change adding callback to sim signals to be a context (#49)

### LINKED ISSUE #49: Add context for callback to be removed from sim signal
When writing tests I want to simulate some basic signal behaviour e.g. writing to one signal will put to another. However, I would like to make sure this behaviour is properly cleaned up between tests. I can simulate with `set_sim_callback` and `set_sim_value` but there is no way to remove this callback.

## Acceptance Criteria
* There is a context that removes callbacks from simulated signals on cleanup

DominicOram:
Fixes #49

I like forcing it to be a context so that you will always clear it up but happy to hear others thoughts if they want a separate add and clear?

### DIFF
--- a/tests/epics/demo/test_demo.py
+++ b/tests/epics/demo/test_demo.py
@@ -99,11 +99,11 @@ async def test_mover_moving_well(sim_mover: demo.Mover) -> None:
 
 async def test_mover_stopped(sim_mover: demo.Mover):
     callbacks = []
-    set_sim_callback(sim_mover.stop_, lambda r, v: callbacks.append(v))
 
-    assert callbacks == [None]
-    await sim_mover.stop()
-    assert callbacks == [None, None]
+    with set_sim_callback(sim_mover.stop_, lambda r, v: callbacks.append(v)):
+        assert callbacks == [None]
+        await sim_mover.stop()
+        assert callbacks == [None, None]
 
 
 async def test_read_mover(sim_mover: demo.Mover):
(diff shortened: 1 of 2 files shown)


tests/epics/demo/test_demo.py
```diff
@@ -99,11 +99,11 @@ async def test_mover_moving_well(sim_mover: demo.Mover) -> None:
...
     callbacks = []
-    set_sim_callback(sim_mover.stop_, lambda r, v: callbacks.append(v))
 
-    assert callbacks == [None]
-    await sim_mover.stop()
-    assert callbacks == [None, None]
+    with set_sim_callback(sim_mover.stop_, lambda r, v: callbacks.append(v)):
+        assert callbacks == [None]
+        await sim_mover.stop()
+        assert callbacks == [None, None]
```
coretl, line 106: I wonder if `set_sim_callbacks` is what we actually want here? We add a callback to a signal, effectively monitoring it, but we use it to track when it's put to. Maybe we instead want:

```python
with track_sim_puts(sim_mover.stop_) as puts:
    assert puts == []
    await sim_mover.stop()
    assert puts == [None]
```

@rosesyrett what do you reckon?

DominicOram: I just changed it to work, have no opinion on whether the original was correct. Though I think `set_sim_callbacks` is still useful and if we are going to change this test we should add another test for `set_sim_callbacks`

DominicOram:
Fixed by https://github.com/bluesky/ophyd-async/issues/94

PR #53, opened 2023-11-01
Add more motor fields 

### LINKED ISSUE #42: Change motor internals to better match ophyd
To ease in migration it would be good if the underlying signals in `Motor` had the same name as `EpicsMotor`. For example, `user_readback` in `EpicsMotor` is `readback` in `Motor`. We should go through and try and keep consistency unless there is a good reason to change the name.

Additionally, there are a number of parameters that are not exposed:
* Limits (see https://github.com/bluesky/ophyd-async/issues/41)
* Acceleration

DominicOram:
Fixes #42


DominicOram:
Closing in favour of the more up-to-date https://github.com/bluesky/ophyd-async/pull/174

PR #55, opened 2023-11-02
Add a timeout to DetectorWriters when they wait for index to increment

rosesyrett:
In a flyscan, a common problem I have found is if the hardware is incorrectly configured, any writer (ie. the hdf writer) will hang because it's expecting to have read data (After it's recieved a trigger, say from a panda). Of course if it's not wired properly, it will just hang.

This prevents it from hanging indefinitely, and allows the timeout to be configurable on the flyer class.

### DIFF
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -68,7 +68,9 @@ class DetectorWriter(ABC):
         """
 
     @abstractmethod
-    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
+    async def wait_for_index(
+        self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT
+    ) -> None:
         """Wait until a specific index is ready to be collected"""
 
     @abstractmethod
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -110,6 +110,7 @@ class StandardDetector(
         writer: DetectorWriter,
         config_sigs: Sequence[SignalR] = (),
         name: str = "",
+        writer_timeout: float = DEFAULT_TIMEOUT,
     ) -> None:
         """
         Parameters
@@ -125,6 +126,7 @@ class StandardDetector(
         self._writer = writer
         self._describe: Dict[str, Descriptor] = {}
         self._config_sigs = list(config_sigs)
+        self._writer_timeout = writer_timeout
         super().__init__(name)
 
     @property
@@ -173,10 +175,12 @@ class StandardDetector(
         indices_written = await self.writer.get_indices_written()
         written_status = await self.controller.arm(DetectorTrigger.internal, num=1)
         await written_status
-        await self.writer.wait_for_index(indices_written + 1, timeout=DEFAULT_TIMEOUT)
+        await self.writer.wait_for_index(
+            indices_written + 1, timeout=self._writer_timeout
+        )
 
     async def read(self) -> Dict[str, Reading]:
-        """Unused method: will be deprecated."""
+        """Read the detector"""
         # All data is in StreamResources, not Events, so nothing to output here
         return {}
 
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -126,7 +126,7 @@ class StandardDetector(
         self._writer = writer
         self._describe: Dict[str, Descriptor] = {}
         self._config_sigs = list(config_sigs)
-        self._writer_timeout = writer_timeout
+        self._frame_writing_timeout = writer_timeout
         super().__init__(name)
 
     @property
@@ -176,7 +176,7 @@ class StandardDetector(
         written_status = await self.controller.arm(DetectorTrigger.internal, num=1)
         await written_status
         await self.writer.wait_for_index(
-            indices_written + 1, timeout=self._writer_timeout
+            indices_written + 1, timeout=self._frame_writing_timeout
         )
 
     async def read(self) -> Dict[str, Reading]:
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -18,7 +18,7 @@ from bluesky.protocols import (
 from .async_status import AsyncStatus
 from .device import Device
 from .signal import SignalR
-from .utils import merge_gathered_dicts
+from .utils import DEFAULT_TIMEOUT, merge_gathered_dicts
 
 
 class DetectorTrigger(str, Enum):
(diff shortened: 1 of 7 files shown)


src/ophyd_async/core/detector.py
```diff
@@ -68,7 +68,7 @@ async def open(self, multiplier: int = 1) -> Dict[str, Descriptor]:
         """
 
     @abstractmethod
-    async def wait_for_index(self, index: int) -> None:
+    async def wait_for_index(self, index: int, timeout: Optional[float] = None) -> None:
```
coretl, line 71: I suggest we make this `DEFAULT_TIMEOUT` so people have to opt into a "wait forever" behaviour
```suggestion
    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
```

src/ophyd_async/core/flyer.py
```diff
@@ -242,7 +246,9 @@ async def kickoff(self) -> None:
     async def _fly(self) -> None:
         await self._trigger_logic.start()
         # Wait for all detectors to have written up to a particular frame
-        await self._detector_group_logic.wait_for_index(self._last_frame - self._offset)
+        await self._detector_group_logic.wait_for_index(
+            self._last_frame - self._offset, timeout=self._timeout
+        )
```
coretl, line 251: By the time this is called, the triggers have been sent to the detector, so I think `DEFAULT_TIMEOUT` is a reasonable time to wait for all the frames to be written
```suggestion
        await self._detector_group_logic.wait_for_index(
            self._last_frame - self._offset, timeout=DEFAULT_TIMEOUT
        )
```

src/ophyd_async/core/flyer.py
```diff
@@ -179,6 +181,7 @@ def __init__(
         detector_group_logic: DetectorGroupLogic,
         trigger_logic: TriggerLogic[T],
         configuration_signals: Sequence[SignalR],
+        timeout: Optional[float] = None,
```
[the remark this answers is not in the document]
rosesyrett: I'll make the change to its name, and have it as configurable; just in case someone might want to change this timeout to be longer/shorter...

src/ophyd_async/core/flyer.py
```diff
@@ -63,7 +63,7 @@ def collect_asset_docs(self) -> AsyncIterator[Asset]:
         """Collect asset docs from all writers"""
 
     @abstractmethod
-    async def wait_for_index(self, index: int):
+    async def wait_for_index(self, index: int, timeout: Optional[float] = None):
```
coretl, line 66:
```suggestion
    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT):
```

src/ophyd_async/core/flyer.py
```diff
@@ -126,8 +126,10 @@ async def collect_asset_docs(self) -> AsyncIterator[Asset]:
             async for doc in writer.collect_stream_docs(indices_written):
                 yield doc
 
-    async def wait_for_index(self, index: int):
-        await gather_list(writer.wait_for_index(index) for writer in self._writers)
+    async def wait_for_index(self, index: int, timeout: Optional[float] = None):
```
coretl, line 129:
```suggestion
    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT):
```

src/ophyd_async/epics/areadetector/writers/hdf_writer.py
```diff
@@ -85,12 +85,12 @@ async def open(self, multiplier: int = 1) -> Dict[str, Descriptor]:
         }
         return describe
 
-    async def wait_for_index(self, index: int):
+    async def wait_for_index(self, index: int, timeout: Optional[float] = None):
```
coretl, line 88:
```suggestion
    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT):
```

tests/core/test_flyer.py
```diff
@@ -84,7 +64,7 @@ async def open(self, multiplier: int = 1) -> Dict[str, Descriptor]:
             )
         }
 
-    async def wait_for_index(self, index: int) -> None:
+    async def wait_for_index(self, index: int, timeout: Optional[float] = None) -> None:
```
coretl, line 67:
```suggestion
    async def wait_for_index(self, index: int, timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
```

src/ophyd_async/core/signal.py
```diff
@@ -208,8 +208,7 @@ def set(self, value: T, wait=True, timeout=USE_DEFAULT_TIMEOUT) -> AsyncStatus:
         """Set the value and return a status saying when it's done"""
         if timeout is USE_DEFAULT_TIMEOUT:
             timeout = self._timeout
-        coro = self._backend.put(value, wait=wait, timeout=timeout)
-        return AsyncStatus(coro)
+        return AsyncStatus(self._backend.put(value, wait=wait, timeout=timeout))
```
callumforrester, line 211: I think this change should be reverted as it was only for debugging purposes

src/ophyd_async/core/detector.py
```diff
@@ -171,7 +173,7 @@ async def trigger(self) -> None:
         indices_written = await self.writer.get_indices_written()
         written_status = await self.controller.arm(DetectorTrigger.internal, num=1)
         await written_status
-        await self.writer.wait_for_index(indices_written + 1)
+        await self.writer.wait_for_index(indices_written + 1, timeout=DEFAULT_TIMEOUT)
```
callumforrester, line 176: It may be better to pass the timeout into this object rather than always use the default, because that would be consistent with the other changes in this PR

callumforrester (changes requested):
See comments and needs a rebase

src/ophyd_async/core/detector.py
rosesyrett: So the writer.wait_for_timeout method gets called twice in the code base, once here and once for the flyer. In the flyer, I have a _trigger_to_frame_timeout on the Flyer that can be set, which gets propagated down to this method call. But here, I don't. I think I agree with you; I'll make the detector have a _detector_index_timeout or something similar to indicate this.

callumforrester: I would call it `_frame_wrote_timeout`

rosesyrett: `_frame_writing_timeout` ?

rosesyrett:
My finger slipped and I accidentally merged main instead of rebasing. Oops.

I don't think it particularly matters since I'll do a squash and merge anyways.

src/ophyd_async/core/detector.py
callumforrester: Sure

callumforrester (approved):

PR #56, opened 2023-11-02
 Change SignalX.execute to trigger #54 

### LINKED ISSUE #54: Change `SignalX.execute` to `trigger`
The existing code is:
https://github.com/bluesky/ophyd-async/blob/f1b4aba227ee6b568718c170ced6378a744abf99/src/ophyd_async/core/signal.py#L226-L231

This does not have a natural verb to use in plans, but `trigger` does, and the signature matches. I propose we change it to:

```python
class SignalX(Signal):
    """Signal that puts the default value"""

    async def trigger(self, wait=True, timeout=USE_DEFAULT_TIMEOUT) -> AsyncStatus:
        """Trigger the action and return a status saying when it's done"""
        if timeout is USE_DEFAULT_TIMEOUT:
            timeout = self._timeout
        coro = self._backend.put(None, wait=wait, timeout=timeout)
        return AsyncStatus(coro)
```

We can then use `bps.trigger` for this (along with its exception handling) rather than `bps.wait_for`

Bilchreis:
fixes #54

### DIFF
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -226,9 +226,12 @@ class SignalRW(SignalR[T], SignalW[T], Locatable):
 class SignalX(Signal):
     """Signal that puts the default value"""
 
-    async def execute(self, wait=True, timeout=None):
-        """Execute the action and return a status saying when it's done"""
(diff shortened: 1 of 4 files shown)


coretl (approved):
Thanks for this! The docs failure is another issue, so will fix in another PR.

PR #60, opened 2023-11-03
Set flow style to False by default, and True for lists/arrays

jsouter:
Small fix for YAML formatting, fields saved without flow style formatting unless they are lists (prevents lists from printing each element on a new line)


olliesilvester:
I tested this with saving a PandA and it makes the formatting look much nicer

PR #61, opened 2023-11-06
Fix docs errors

coretl:
Remove a couple of unnecessary imports while at it


rosesyrett (approved):
Thanks, very happy you fixed this because it would have taken me ages :+1:

coretl:
> Thanks, very happy you fixed this because it would have taken me ages 👍

No idea why the ipython directive has started requiring the `pickshare` library, but :shrug:

PR #36, opened 2023-10-23
Add save load

coretl:
As per #27 #31 #18, but now targeted on main

PR #63, opened 2023-11-07
Add panda-specific save load plans

olliesilvester:
Adds plans to save and load a panda device, including loading the phases in the correct order. This uses the functions added in https://github.com/bluesky/ophyd-async/pull/36

This also fixes an issue with hanging subprocceses in unit tests, and fixes a bug where devices wouldn't load if one of the saved PV's were ignored (set to None in the yaml file)


src/ophyd_async/panda/panda_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    signalRW_and_value = yield from get_signal_values(walk_rw_signals(panda), ignore)
+    phase_1 = {}
+    phase_2 = {}
+    for signal_name in signalRW_and_value.keys():
+        if signal_name[-5:] == "units":
+            phase_1[signal_name] = signalRW_and_value[signal_name]
+        else:
+            phase_2[signal_name] = signalRW_and_value[signal_name]
+
+    return [phase_1, phase_2]
```
coretl, line 24: I already wrote this yesterday for i22, so I'll paste in what I had in case you want to use any of it:
```python
def save_the_panda():
    signals = walk_rw_signals(panda)
    values = yield from get_signal_values(signals, ignore=["seq1.table"])
    units = {n: values.pop(n) for n in list(values) if n.endswith("units")}
    save_to_yaml([units, values], "/tmp/thing.yaml")
```

src/ophyd_async/panda/panda_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    save_to_yaml(phases, path)
+
+
+def load_panda(panda: PandA, path: str):
+    """
+    Sets all PV's to a PandA using a previously saved configuration
+    """
+    values = load_from_yaml(path)
+    signals_to_set = walk_rw_signals(panda)
+    yield from set_signal_values(signals_to_set, values)
```
coretl, line 41: Thinking about this, the `load` is completely generic, it is not specifically for PandA, so I think we should remove it from here...

src/ophyd_async/core/device_save_loader.py
```diff
@@ -177,8 +177,22 @@ def set_signal_values(
     for phase_number, phase in enumerate(values):
         # Key is signal name
         for key, value in phase.items():
+            # Skip ignored values
+            if value is None:
```
olliesilvester, line 181: Loading from a yaml file where we had a list of ignored PVs wouldn't work without this

rosesyrett: We have a 'Compression' enum now, in `NDFileHDF`: https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/areadetector/writers/nd_file_hdf.py

Any devices using this, get serialised (at the moment) so that if the compression is set to "None", it appears in the yaml as:

```
some-signal: None
```
obviously this is bad yaml; it should be 'null' or something similar. However, for the moment, this change would ignore that pv completely, which seems against what we want to happen.

But this should probably be a broader, post i22 discussion.

src/ophyd_async/core/device_save_loader.py
```diff
@@ -177,8 +177,22 @@ def set_signal_values(
...
         yield from wait(f"load-phase{phase_number}")
+
+
+def load_device(device: Device, path: str):
+    """
+    Sets all PV's to a device using a previously saved configuration
+    """
+    values = load_from_yaml(path)
+    signals_to_set = walk_rw_signals(device)
+    yield from set_signal_values(signals_to_set, values)
```
rosesyrett, line 198: If we're going to do this, we should add a `save_device` function that does the opposite. I've noticed for some reason this function currently lives in the tests for device_save_loader...

src/ophyd_async/panda/panda_utils.py
```diff
@@ -0,0 +1,22 @@
...
+from ophyd_async.panda import PandA
+
+
+def _get_panda_phases(panda: PandA, ignore: Optional[List[str]] = None):
+    # Panda has two load phases. If the signal name ends in the string "UNITS",
+    # it needs to be loaded first so put in first phase
+    signals = walk_rw_signals(panda)
+    phase_2 = yield from get_signal_values(signals, ignore=ignore)
+    phase_1 = {n: phase_2.pop(n) for n in list(phase_2) if n.endswith("units")}
+    return [phase_1, phase_2]
```
rosesyrett, line 13: Do we anticipate there will be other devices, for which we'd probably want to split units out like this?

rosesyrett (commented):
I'm not sure I like the fact we have plans in ophyd-async. I think this needs a broader discussion...

In general I think this looks good if we are heading in that direction, but personally I'd prefer if we could move this into dls-bluesky-core and keep ophyd-async bluesky agnostic where possible...

src/ophyd_async/core/device_save_loader.py
olliesilvester: The initial idea was to have generic functions which you could construct save/load plans from, functions like `walk_rw_signals` and `load_from_yaml` and `set_signal_values`, then save/load plans could be constructed from these for a specific use-case. We later realised the the `load_device` plan would be generic for all devices, so we might as well include it here. 
The `save_device` plan is not as generic as it will depend on the device-specific phases, which is why I didn't include it in this PR. I guess if most devices don't have a phase order then it makes sense to include `save_device` here too

src/ophyd_async/panda/panda_utils.py
olliesilvester: As far as I'm aware this logic was specific to the Panda but @coretl might be aware of future devices which will do the same thing

src/ophyd_async/core/device_save_loader.py
olliesilvester: Don't we want to completely ignore these values?

src/ophyd_async/panda/panda_utils.py
coretl: There are likely to be more in the future as we add more PVs (e.g. remember the trouble we had with setting TriggerMode at the same time as TriggerSource in the GigE cameras)

src/ophyd_async/core/device_save_loader.py
coretl: I guess we could have generic load and save plans, but pass an optional `sorter` function to it that takes the saved values and sorts them into phases?

src/ophyd_async/core/device_save_loader.py
coretl: No, this is not actually bad yaml, just confusing yaml. `some-signal: None` translates to python `{"some-signal": "None"}`, which is what we actually need, while `some-signal: null` translates to python `{"some-signal": None}` which would the the output if we wanted to ignore that field.

Either way I think what you have here is correct, but we need to document this confusing corner case

rosesyrett:
Gonna improve the test coverage and fix the conflict with main

rosesyrett:
@coretl and @olliesilvester I've just been adding a test case for this:

```python
def set_signal_values(
    signals: Dict[str, SignalRW[Any]], values: Sequence[Dict[str, Any]]
) -> Generator[Msg, None, None]:
    """Maps signals from a yaml file into device signals.

    ``values`` contains signal values in phases, which are loaded in sequentially
    into the provided signals, to ensure signals are set in the correct order.

    Parameters
    ----------
    signals : Dict[str, SignalRW[Any]]
        Dictionary of named signals to be updated if value found in values argument.
        Can be the output of :func:`walk_rw_signals()` for a device.

    values : Sequence[Dict[str, Any]]
        List of dictionaries of signal name and value pairs, if a signal matches
        the name of one in the signals argument, sets the signal to that value.
        The groups of signals are loaded in their list order.
        Can be the output of :func:`load_from_yaml()` for a yaml file.

    See Also
    --------
    :func:`ophyd_async.core.load_from_yaml`
    :func:`ophyd_async.core.walk_rw_signals`
    """
    # For each phase, set all the signals,
    # load them to the correct value and wait for the load to complete
    for phase_number, phase in enumerate(values):
        # Key is signal name
        for key, value in phase.items():
            # Skip ignored values
            if value is None:
                continue

            if key in signals:
                yield from abs_set(
                    signals[key], value, group=f"load-phase{phase_number}"
                )

        yield from wait(f"load-phase{phase_number}")
```

Specifically, for the 'continue' line. I've just noticed; isn't the purpose of this, so that yaml files with 'null' entries are ignored? If so, why not just remove that line from the yaml file, especially considering the confusion this can cause with enums that have 'None' values?

coretl:
> Specifically, for the 'continue' line. I've just noticed; isn't the purpose of this, so that yaml files with 'null' entries are ignored? If so, why not just remove that line from the yaml file, especially considering the confusion this can cause with enums that have 'None' values? Edit: on the other side of this, we can ensure that when we are saving to yaml, we don't save any pv's that have 'None' values outside of enums.

If the value is None, this means the signal was present on the Device, but ignored. This means you can tell the different between a new signal being added to a Device (which should be saved if you re-save), and an ignored Signal

The continue looks correct to me, if you see a null in the file (which will be translated to a `None` in the python dict), then don't set that value

rosesyrett:
Just rebased

coretl (approved):

PR #66, opened 2023-11-10
Add representers for all channel access types

rosesyrett:
we have save load functionality, but unfortunately it only works for PV access currently. So I've added some lines which let it work for channel access, by adding representers. I've also expanded on existing tests to ensure the serialisation/deserialisation works.

### DIFF
--- a/src/ophyd_async/core/device_save_loader.py
+++ b/src/ophyd_async/core/device_save_loader.py
@@ -1,5 +1,6 @@
 from enum import Enum
-from typing import Any, Dict, Generator, List, Optional, Sequence
+from functools import partial
+from typing import Any, Callable, Dict, Generator, List, Optional, Sequence, Union
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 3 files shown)


src/ophyd_async/core/device_save_loader.py
```diff
@@ -59,6 +76,10 @@ def get_signal_values(
         key: signal for key, signal in signals.items() if key not in ignore
     }
     selected_values = yield Msg("locate", *selected_signals.values())
+
+    if isinstance(selected_values, dict):
+        selected_values = [selected_values]
+
```
rosesyrett, line 82: I want to know people's opinion on this; while writing tests I found that, if get_signal_values is called using a dictionary containing just one device, the result of yield Msg("locate", ...) will be a dictionary, not a sequence. The type hinting for selected_values suggests it will always be a sequence. But it isn't... Personally I'd prefer if locate always returned a list of dicts, instead of sometimes returning a dict and sometimes returning a list of dicts.

Thoughts?

src/ophyd_async/core/device_save_loader.py
```diff
@@ -1,23 +1,40 @@
...
+from functools import partial
+from typing import Any, Callable, Dict, Generator, List, Optional, Sequence, Union
 
 import numpy as np
 import numpy.typing as npt
 import yaml
 from bluesky.plan_stubs import abs_set, wait
 from bluesky.protocols import Location
 from bluesky.utils import Msg
+from epicscorelibs.ca.dbr import ca_array, ca_float, ca_int, ca_str
```
coretl, line 11: This is the right solution, but in the wrong place. We need to make sure `epicscorelibs` imports are only within the `ophyd_async.epics` packages. 

How about a global (yuck) directory of representers that are added to on inport of packages?

rosesyrett: I think this needs a broader discussion on where stuff belongs. I don't like the fact that we have plans in ophyd_async.core, but definitely agree epics things shouldn't belong there in the long run. For the meantime, while ophyd_async is still in pre-release, I'd be keen to just merge this in anyway and come up with a better long term solution later...

callumforrester: @coretl I suggest merging, making an issue and fixing later because of time constraints

src/ophyd_async/core/device_save_loader.py
abbiemery: @rosesyrett Since we are merging this for need then coming back to it, can I just put a type ignore on this to get mypy to behave? It's the only thing flagging now

rosesyrett: Yes please. The type hints are wrong; this is probably a bluesky bug.

PR #71, opened 2023-11-13
Fix HardwareTriggeredFlyable collect asset order

coretl:
Change so all stream_resource documents are emitted first then all stream_datums. This means the NeXus writer can produce datasets first on SR, before emitting
updates on SD


rosesyrett (approved):
This looks reasonable and like it'll fix the problem we are having. I can't think of a document generation order for which this wouldn't work.

Issue #76, opened 2023-11-13
Deal with stalled timeout


rosesyrett:
We need a check in place to ensure that hdf writers can write data from triggers faster than triggers can be produced. If the hdf writer takes too long before the next trigger, it should timeout.

callumforrester:
This is no longer an issue

PR #78, opened 2023-11-13
Implement stall timeout

rosesyrett:
wip: start removing stall timeout from _hdffile and put it higher up the stack

fixes https://github.com/bluesky/ophyd-async/issues/76


DiamondJoseph:
Dropped re: TC

PR #86, opened 2023-11-20
Fix root/path kwargs for Stream Resource documents generated by HDFWriter

rosesyrett:
Previously, root was hardcoded to '/'. This removes that, and instead ensures the root and resource_path are accurate.

### DIFF
--- a/src/ophyd_async/core/_providers.py
+++ b/src/ophyd_async/core/_providers.py
@@ -37,9 +37,9 @@ class StaticDirectoryProvider(DirectoryProvider):
     def __init__(
         self,
         directory_path: Union[str, Path],
-        resource_dir: Path = Path("."),
         filename_prefix: str = "",
         filename_suffix: str = "",
+        resource_dir: Path = Path("."),
     ) -> None:
         if isinstance(directory_path, str):
             directory_path = Path(directory_path)
--- a/src/ophyd_async/core/_providers.py
+++ b/src/ophyd_async/core/_providers.py
@@ -21,10 +21,6 @@ class DirectoryInfo:
     prefix: Optional[str] = ""
     suffix: Optional[str] = ""
 
-    @property
-    def directory(self):
-        return self.root / self.cwd
-
 
 class DirectoryProvider(Protocol):
     @abstractmethod
@@ -43,7 +39,7 @@ class StaticDirectoryProvider(DirectoryProvider):
             directory_path = Path(directory_path)
         self._directory_info = DirectoryInfo(
             root=directory_path,
-            cwd=directory_path,
+            cwd=Path("."),
             prefix=filename_prefix,
             suffix=filename_suffix,
         )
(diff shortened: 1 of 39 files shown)


coretl:
Tests failing?

DiamondJoseph:
This change is good, I will finish it

evalott100 (approved):

src/ophyd_async/core/_providers.py
```diff
@@ -1,12 +1,25 @@
...
+from pathlib import Path
+from typing import Optional, Protocol, Sequence, Union
 
 
 @dataclass
 class DirectoryInfo:
-    directory_path: str
-    filename_prefix: str
+    """
+    Information about where and how to write a file.
```
callumforrester, line 10: If you can come up with a sufficiently brief explanation of why both `root` and `resource_dir` are needed, I think it would be useful to put in this docstring. Maybe something like

> The bluesky event model splits the URI for a resource into two segments to aid in different applications mounting filesystems at different mount points.

src/ophyd_async/epics/areadetector/writers/hdf_writer.py
```diff
@@ -45,15 +45,15 @@ async def open(self, multiplier: int = 1) -> Dict[str, Descriptor]:
             self.hdf.num_extra_dims.set(0),
             self.hdf.lazy_open.set(True),
             self.hdf.swmr_mode.set(True),
-            self.hdf.file_path.set(info.directory_path),
-            self.hdf.file_name.set(f"{info.filename_prefix}{self.hdf.name}"),
+            self.hdf.file_path.set(str(info.root / info.resource_dir)),
```
callumforrester, line 48: Relevant: https://github.com/bluesky/ophyd-async/issues/122

callumforrester (approved):

src/ophyd_async/epics/areadetector/writers/_hdffile.py
```diff
@@ -2,18 +2,25 @@
...
+    ) -> None:
         self._last_emitted = 0
         self._bundles = [
             compose_stream_resource(
                 spec="AD_HDF5_SWMR_SLICE",
-                root="/",
+                root=str(directory_info.root),
                 data_key=ds.name,
-                resource_path=full_file_name,
+                resource_path=str(directory_info.resource_dir / full_file_name),
```
coretl, line 23: `full_file_name` is the complete path, if you want to separate out the root then you need `full_file_name` relative to `directory_info.root / directory_info.resource_dir`...

coretl: Are we missing a test that would catch this too?

DiamondJoseph: Added a test as part of the change that should catch if the behaviour changes

coretl (approved):

PR #95, opened 2023-12-13
Add generic save_device function

rosesyrett:
Change panda_utils -> utils.py only, as it already resides in a panda submodule, and have this define the sorting function needed to save a generic device.


rosesyrett:
Going to reword some of the docstrings, but this effectively solves the use cases both for i22 (from i22-bluesky) and fits nicely into the existing PR on save loading.

src/ophyd_async/panda/utils.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any, Dict
+
+
+def phase_sorter(panda_signal_values: Dict[str, Any]):
```
coretl, line 4:
```suggestion
def phase_sorter(panda_signal_values: Dict[str, Any]) -> Sequence[Dict[str, Any]]:
```

src/ophyd_async/core/device_save_loader.py
```diff
@@ -190,9 +193,61 @@ def set_signal_values(
...
     signals_to_set = walk_rw_signals(device)
     yield from set_signal_values(signals_to_set, values)
+
+
+def save_device(
+    device: Device,
+    path: str,
+    sorter: Callable[[Dict[str, Any]], Sequence[Dict[str, Any]]] = lambda x: [x],
+    ignore: Optional[List[str]] = None,
+):
```
coretl, line 219: It would be nice to promote the lambda into a named function so we can reference it in the docs

```suggestion
def all_at_once(values: Dict[str, Any]) -> Sequence[Dict[str, Any]]:
    """Sort all the values into a single phase so they are set all at once"""
    return [values]


def save_device(
    device: Device,
    path: str,
    sorter: Callable[[Dict[str, Any]], Sequence[Dict[str, Any]]] = all_at_once,
    ignore: Optional[List[str]] = None,
):
```

coretl (commented):
Couple of minor points, otherwise looks good

olliesilvester (approved):
I like these changes, thanks!

PR #96, opened 2023-12-14
Ensure detectors receive trigger num from plans

### LINKED ISSUE #93: No longer explicitly support n=0 frames to go forever. Plan should pass this to the detectors.
Currently, the `HardwareTriggeredFlyable` has many detectors, and calls `.arm` for all of them. By default, the signature of the `DetectorControl.arm` method has a keyword argument, `num=0` which is set to 0 by default to indicate that this controller should take infinite pictures.

This should be changed, so that actually the number of images is fixed. The easiest way of doing this is,
1. Remove the default value; this has to be passed from the Flyer.
2. Make the Flyer get this value from it's constructor
3. Test it to check it still works, e.g. on p38

This can be done in isolation to the other changes to Flyers, and is a good first issue.

abbiemery:
Fixes #93 .

@rosesyrett please may you take a look at this? I think this provides the behaviour you are after, but I am unsure on how to test it how the num value needs to be supplied from elsewhere.

### DIFF
--- a/tests/epics/areadetector/test_controllers.py
+++ b/tests/epics/areadetector/test_controllers.py
@@ -35,10 +35,10 @@ async def ad(RE) -> ADSimController:
 
 async def test_ad_controller(RE, ad: ADSimController):
     with patch("ophyd_async.core.signal.wait_for_value", return_value=None):
-        await ad.arm(num=0)
+        await ad.arm(num=1)
 
     driver = ad.driver
-    assert await driver.num_images.get_value() == 0
+    assert await driver.num_images.get_value() == 1
     assert await driver.image_mode.get_value() == ImageMode.multiple
     assert await driver.acquire.get_value() is True
 
@@ -52,10 +52,10 @@ async def test_ad_controller(RE, ad: ADSimController):
 
 async def test_pilatus_controller(RE, pilatus: PilatusController):
     with patch("ophyd_async.core.signal.wait_for_value", return_value=None):
-        await pilatus.arm(num=0, trigger=DetectorTrigger.constant_gate)
+        await pilatus.arm(num=1, trigger=DetectorTrigger.constant_gate)
 
     driver = pilatus.driver
-    assert await driver.num_images.get_value() == 999_999
+    assert await driver.num_images.get_value() == 1
     assert await driver.image_mode.get_value() == ImageMode.multiple
     assert await driver.trigger_mode.get_value() == PilatusTrigger.ext_enable
     assert await driver.acquire.get_value() is True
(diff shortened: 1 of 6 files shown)


tests/epics/areadetector/test_controllers.py
```diff
@@ -35,7 +35,7 @@ async def ad(RE) -> ADSimController:
 
 async def test_ad_controller(RE, ad: ADSimController):
     with patch("ophyd_async.core.signal.wait_for_value", return_value=None):
-        await ad.arm()
+        await ad.arm(num=0)
```
rosesyrett, line 38: I think we should make it clear that `num=0` should be reserved for the fly scanning case. Maybe we should set `num=1` here and test it actually tries to take 1 image? So there's an assert statement later that checks it's set to 0, which should just change to being set to 1

tests/epics/areadetector/test_controllers.py
```diff
@@ -52,7 +52,7 @@ async def test_ad_controller(RE, ad: ADSimController):
 
 async def test_pilatus_controller(RE, pilatus: PilatusController):
     with patch("ophyd_async.core.signal.wait_for_value", return_value=None):
-        await pilatus.arm(trigger=DetectorTrigger.constant_gate)
+        await pilatus.arm(num=0, trigger=DetectorTrigger.constant_gate)
```
rosesyrett, line 55: Same as above; we should reserve `num=0` for flyscanning and use `num=1` here instead.

rosesyrett (changes requested):
This looks good. Just a couple of minor changes that I'd like to see in the tests.

rosesyrett:
As a quick comment, how the num value is supplied differs based on if we are doing a step or a fly scan. In both cases, this change happens in a `Controller`.

For step scanning, at the moment `StandardDetector` has both a `Controller` and a `Writer`, and as you've noticed whenever you `trigger` the detector you set it up to take exactly 1 picture - which makes sense, if you're step scanning you want to just take an image each time you trigger.

For fly scanning, a `Flyer` essentially has a bunch of `Controllers` and `Writers` that it controls. In the `Flyer` logic, **at the moment** it calls `controller.arm()` on all of them with num=0. You've correctly changed this to num=TriggerInfo.num, so that now at the plan level this can be passed down, or generated from the TriggerLogic. This bit is subject to change in future anyways, as we move stuff out of the Flyer.

In short, I wouldn't worry too much about if this is going to break things; there are so many breaking changes coming up that we should not be too rigorous at this stage anyways. This looks logical; if things break when we come to testing them on p38 then we will address them, but I highly doubt this will be the reason why they fail :)

tests/epics/areadetector/test_controllers.py
abbiemery: In this pilatus case we are asserting that n=0 results in `999_999` as the value, are we not wanting to retain this test case?

tests/epics/areadetector/test_controllers.py
abbiemery: Done

tests/epics/areadetector/test_controllers.py
abbiemery: Done

rosesyrett (approved):
lgtm :tada:

Issue #48, opened 2023-11-01
Improve `connect` protocol to take timeout


coretl:
At the moment we have:
https://github.com/bluesky/ophyd-async/blob/f1b4aba227ee6b568718c170ced6378a744abf99/src/ophyd_async/core/device.py#L53-L65

This doesn't take timeout, so to add timeout on top we wrap with `asyncio.wait` or similar (like other primitives in the asyncio library). Unfortunately this makes logging the error difficult as `connect` calls recursively connect child devices. The way we report reasonable errors at the moment is to catch the `CancelledError` that is injected when the task times out, and raise `NotConnectedError` with the name of the signal in question, accumulating them in the parent until we produce a single top level `NotConnectedError` with all the failing signals:

https://github.com/bluesky/ophyd-async/blob/f1b4aba227ee6b568718c170ced6378a744abf99/src/ophyd_async/core/utils.py#L26-L53

This is horrible. It also bites people who `await device.connect()` rather than using `asyncio.wait` like in https://github.com/DiamondLightSource/dodal/issues/223.

@callumforrester suggested a better way, pass the timeout down, then let each child produce a `class ConnectTimeoutError(TimeoutError)` when it times out, then assemble it at the top level with an `asyncio.gather(*coros, return_exceptions=True)`, squashing `ConnectTimeoutError`s into a single one.

This would change the signature to:

```python
async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):  
```

with each concrete class raising `ConnectTimeoutError` if it times out, then `wait_for_connection` doing the squashing of `ConnectTimeoutErrors`.

olliesilvester:
An `ophyd_async` device timed out on Hyperion yesterday, very difficult to diagnose without knowing what device / signal timed out, so this change will be very useful

PR #97, opened 2024-01-04
Modify error handling for timed out signals when connecting.

rosesyrett:
This commit puts responsibility on anything with a .connect method to raise its own ConnectionTimeoutError, to be then picked up by the DeviceCollector, or not. This simplifies the complex error handling requiring timeouts to be handled at the asyncio level and passed up the called via (now redefined) NotConnected exceptions.

fixes https://github.com/bluesky/ophyd-async/issues/48

### DIFF
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -2,20 +2,23 @@
 
 from __future__ import annotations
 
-import asyncio
-import logging
 import sys
-from typing import Any, Dict, Generator, Iterator, Optional, Set, Tuple, TypeVar
+from typing import (
+    Any,
+    Coroutine,
+    Dict,
+    Generator,
+    Iterator,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+)
 
 from bluesky.protocols import HasName
 from bluesky.run_engine import call_in_bluesky_event_loop
 
-from .utils import (
-    DEFAULT_TIMEOUT,
-    ConnectionTimeoutError,
-    NotConnected,
-    wait_for_connection,
-)
+from .utils import DEFAULT_TIMEOUT, wait_for_connection
 
 
 class Device(HasName):
@@ -150,53 +153,19 @@ class DeviceCollector:
 
     async def _on_exit(self) -> None:
         # Name and kick off connect for devices
-        tasks: Dict[str, asyncio.Task] = {}
+        connect_coroutines: Dict[str, Coroutine] = {}
         for name, obj in self._objects_on_exit.items():
             if name not in self._names_on_enter and isinstance(obj, Device):
                 if self._set_name and not obj.name:
                     obj.set_name(name)
                 if self._connect:
-                    tasks[name] = asyncio.create_task(
-                        obj.connect(self._sim, timeout=self._timeout)
+                    connect_coroutines[name] = obj.connect(
+                        self._sim, timeout=self._timeout
                     )
-        # Wait for all the signals to have finished
-        if tasks:
-            await self._wait_for_tasks(tasks)
 
-    async def _wait_for_tasks(self, tasks: Dict[str, asyncio.Task]):
-        names = tasks.keys()
-        results = await asyncio.gather(*tasks.values(), return_exceptions=True)
-
-        disconnected = 0
-        failed: Dict[str, Exception] = {}
-
-        msg = ""
-
-        for name, result in zip(names, results):
-            if result and isinstance(result, ConnectionTimeoutError):
-                msg += f"\n  {name}: {type(result).__name__}"
-                lines = str(result).splitlines()
-                if len(lines) <= 1:
-                    msg += f": {result}"
-                else:
-                    msg += "".join(f"\n    {line}" for line in lines)
-
-                disconnected += 1
-
-            elif result and isinstance(result, Exception):
-                failed[name] = result
-
-        if disconnected:
-            msg = f"{disconnected} Devices did not connect:" + msg
-            logging.error(msg)
-
-        if failed:
-            logging.error(f"{len(failed)} Devices raised an error:")
-            for name, exception in failed.items():
-                logging.exception(f"  {name}:", exc_info=exception)
-
-        if disconnected or failed:
-            raise NotConnected("Not all Devices connected")
+        # Connect to all the devices
+        if connect_coroutines:
+            await wait_for_connection(**connect_coroutines)
 
     async def __aexit__(self, type, value, traceback):
         self._objects_on_exit = self._caller_locals()
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -5,13 +5,17 @@ from __future__ import annotations
 import asyncio
 import logging
 import sys
-from contextlib import suppress
 from typing import Any, Dict, Generator, Iterator, Optional, Set, Tuple, TypeVar
 
 from bluesky.protocols import HasName
 from bluesky.run_engine import call_in_bluesky_event_loop
 
-from .utils import NotConnected, wait_for_connection
+from .utils import (
+    DEFAULT_TIMEOUT,
+    ConnectionTimeoutError,
+    NotConnected,
+    wait_for_connection,
+)
 
 
 class Device(HasName):
@@ -51,16 +55,21 @@ class Device(HasName):
             child.set_name(child_name)
             child.parent = self
 
-    async def connect(self, sim: bool = False):
+    async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
         """Connect self and all child Devices.
+
+        Contains a timeout that gets propagated to child.connect methods.
 
         Parameters
         ----------
         sim:
             If True then connect in simulation mode.
+        timeout:
+            Time to wait before failing with a TimeoutError.
         """
         coros = {
-            name: child_device.connect(sim) for name, child_device in self.children()
+            name: child_device.connect(sim, timeout=timeout)
+            for name, child_device in self.children()
         }
         if coros:
             await wait_for_connection(**coros)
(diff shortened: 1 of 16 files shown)


src/ophyd_async/core/device.py
```diff
@@ -140,40 +149,52 @@ async def __aenter__(self) -> "DeviceCollector":
...
-            for t in raised:
-                logging.exception(f"  {tasks[t]}:", exc_info=t.exception())
-        if pending or raised:
+
+        if failed:
+            logging.error(f"{len(failed)} Devices raised an error:")
+            for name, exception in failed.items():
+                logging.exception(f"  {name}:", exc_info=exception)
+
+        if disconnected or failed:
```
callumforrester, line 197: Had a go at a slightly more concise version of this, which I like because it handles timeouts and errors in the same way, what do you think?

```python
    async def _wait_for_tasks(self, tasks: Dict[str, asyncio.Task]):
        names = tasks.keys()
        results = await asyncio.gather(*tasks.values(), return_exceptions=True)

        timed_out: Dict[str, ConnectionTimeoutError] = {}
        failed: Dict[str, Exception] = {}

        for name, result in zip(names, results):
            if result and isinstance(result, ConnectionTimeoutError):
                timed_out[name] = result
            elif result and isinstance(result, Exception):
                failed[name] = result

        if len(timed_out) > 0:
            logging.error(f"{len(timed_out)} devices timed out during connection:")
            for name, exception in timed_out.items():
                logging.exception(f"  {name}:", exc_info=exception)
        if len(failed) > 0:
            logging.error(f"{len(failed)} devices raised an error:")
            for name, exception in failed.items():
                logging.exception(f"  {name}:", exc_info=exception)

        if len(timed_out) > 0 or len(failed) > 0:
            raise NotConnected("Not all Devices connected")

```

At this point the only purpose of having separate handling code for `ConnectTimeoutError` and other kinds of exceptions is to sort them, i.e. so we can have

```shell
ERROR: 2 Devices timed out
device_a: <taceback>...ConnectionTimeOutError
device_b: <taceback>...ConnectionTimeOutError

ERROR: 2 Devices errored
device_c: <taceback>...IOError
device_d: <taceback>...PVaccessError
```

So as a separate question: Do we care any more, they are all just tracebacks at the end of the day, does it matter if it says something like this?

```shell
ERROR: 4 Devices failed to
device_a: <taceback>...ConnectionTimeOutError
device_c: <taceback>...IOError
device_b: <taceback>...ConnectionTimeOutError
device_d: <taceback>...PVaccessError
```

...If not then we could simply further and just say "if exception: log it"

src/ophyd_async/core/utils.py
```diff
@@ -23,34 +23,35 @@ def __str__(self) -> str:
...
+    lines: List[str] = []
+    for name, result in zip(names, results):
+        if isinstance(result, ConnectionTimeoutError):
+            if len(result.lines) == 1:
+                lines.append(f"{name}: {result.lines[0]}")
+            else:
+                lines.append(f"{name}:")
+                lines += [f"  {line}" for line in result.lines]
+        elif isinstance(result, Exception):
+            raise result
```
callumforrester, line 51: Ditto for comment on `_wait_for_tasks` above

callumforrester (dismissed):
Definitely the right idea and I'm onboard, see individual comments and also the following about approach:

The approach in most default devices seems to be:
```python
try:
    some_device_specific_connection()
except TimeoutError:  # or CANothing or similar
    raise ConnectionTimeoutError(some_device_specific_source)
```

Does this lose us information from the original error? Would we not be better off with something like this?

```python
try:
    some_device_specific_connection()
except TimeoutError as ex:  # or CANothing or similar
    raise ConnectionTimeoutError(some_device_specific_source) from ex
```

src/ophyd_async/core/device.py
coretl: I don't think reporting hundreds of the same tracebacks when the IOC is down is very helpful, we should think of a way to squash it. My suggestion:
- If a `Signal.connect` sees the PV is disconnected, it raises `NotConnected([self.source])` from the original error
- `Device.connect` aggregates all errors into a single `NotConnected(indented(errors))`, and `log.exception`s any non-NotConnected errors
- `DeviceCollector` does the same error handling as `Device.connect`

coretl: Ok, how about this, log `NotConnected` at debug level, log all other exceptions at error level

callumforrester: I'm happy with that as long as it is clearly documented!

evalott100:
New plan:

image

src/ophyd_async/core/utils.py
```diff
@@ -16,41 +19,84 @@
...
+            self._errors[name] = (
+                f"unexpected exception {type(sub_not_connected_error).__name__}"
+            )
+        else:
+            raise RuntimeError(
+                "`NotConnected.add_sub_NotConnected` expects `sub_not_connected_error` "
+                f"to be an Exception, received {type(sub_not_connected_error)}"
+            )
+
+    def format_error_string(self, errors: Dict | str, indent=0):
```
coretl, line 56:
```suggestion
    def format_error_string(self, errors: Dict | str, indent=""):
```

src/ophyd_async/core/device.py
```diff
@@ -141,41 +145,36 @@ async def __aenter__(self) -> "DeviceCollector":
 
     async def _on_exit(self) -> None:
         # Name and kick off connect for devices
-        tasks: Dict[asyncio.Task, str] = {}
+        tasks: Dict[str, asyncio.Task] = {}
```
coretl, line 148: Suggest that rather than storing these as tasks, create a dict of name -> coros in the for loop below, then call `wait_for_connection` on it at the end. Then we can ditch _wait_for_tasks

src/ophyd_async/epics/_backend/_p4p.py
```diff
@@ -233,22 +234,35 @@ def _del_ctxt():
...
+            for task in pending:
+                task.cancel()
+
+                with suppress(asyncio.CancelledError):
+                    await task
+            message = f"signal pva://{pv} timed out"
+            logging.debug(message)
+            raise NotConnected(message)
+
+        self.initial_values[pv] = [d.result() for d in done][0]
```
coretl, line 254:
```suggestion
        try:
            self._initial_values[pv] = await asyncio.wait_for(
                self.ctxt.get(pv), timeout=timeout
            )
        except asyncio.TimeoutError as exc:
            logging.debug(f"PVA connect {pv} failed", exc_info=True)
            raise NotConnected(f"pva://{pv}") from exc
```

tests/core/test_utils.py
```diff
@@ -11,3 +23,227 @@ def test_static_directory_provider():
...
+    for protocol in ("pva", "ca"):
+        assert f"signal {protocol}://A_NON_EXISTENT_SIGNAL timed out" in messages
+
+
+def test_not_connected_error_output():
+    not_connected = NotConnected(TWO_WORKING_TWO_TIMEOUT_TWO_VALUE_ERROR_OUTPUT)
+
+    assert str(not_connected) == (
+        "timeout_child_device_ca:\n"
+        "    timeout_signal: signal ca://A_NON_EXISTENT_SIGNAL timed out\n"
```
coretl, line 204: Please can we include the error type here
```suggestion
        "    timeout_signal: NotConnected: ca://A_NON_EXISTENT_SIGNAL\n"
```

tests/core/test_utils.py
```diff
@@ -11,3 +23,227 @@ def test_static_directory_provider():
...
+        "timeout_child_device_pva:\n"
+        "    timeout_signal: signal pva://A_NON_EXISTENT_SIGNAL timed out\n"
+        "value_error_child_device1:\n"
+        "    value_error_signal: unexpected exception ValueError\n"
+        "value_error_child_device2:\n"
+        "    value_error_signal: unexpected exception ValueError\n"
```
coretl, line 210: And here:
```suggestion
        "    value_error_signal: ValueError: Some ValueError text\n"
```

src/ophyd_async/core/utils.py
```diff
@@ -11,46 +24,75 @@
...
+            else:
+                exceptions[name] = f"unexpected exception {type(result).__name__}"
+                logging.exception(
+                    f"device `{name}` raised unexpected "
+                    f"exception {type(result).__name__}",
+                    exc_info=result,
+                )
+
+    if exceptions:
+        raise NotConnected(exceptions)
```
coretl, line 95: I wanted to write this, but that would push the string formatting back down and you would lose the dictionary. It would handle multiline exception output though, so maybe this would be better?

```suggestion
    exceptions: Dict[str, Exception] = {}

    for name, result in zip(coros, results):
        if isinstance(result, Exception):
            exceptions[name] = f"{type(result).__name__}: {result}"
            if not isinstance(result, NotConnected):
                logging.exception(f"{self.name}: device `{name}` raised unexpected exception")

    if exceptions:
        raise NotConnected(exceptions)
```

coretl (commented):
Let's discuss the string formatting offline

src/ophyd_async/core/utils.py
```diff
@@ -11,46 +24,82 @@
...
+                if isinstance(value._errors, str):
+                    string += self._format_exception_string(value)
+                else:
+                    string += "\n" + value._format_error_string(
+                        value._errors, indent=(indent + self._indent_width)
+                    )
+            elif isinstance(value, Exception):
+                string += self._format_exception_string(value)
+            else:
+                raise RuntimeError(f"`{type(value)}` not a string or exception")
```
coretl, line 75: If we allow formatting like:
```
device: NotConnected:
    sigA: NotConnected: ca://...
    sigB: ValueError: 32
```

Then we can simplify to

```suggestion
        for name, value in errors.items():
            if isinstance(value, NotConnected):
                error_txt = value.format_error_string(indent + self._indent_width)
            else:
                error_txt = str(error)
            string += f"\n{indent}{name}: {type(error).__name__}"
            if error_txt:
                string += f": {error_txt}"
```

tests/panda/test_panda.py
```diff
@@ -130,3 +132,29 @@ async def test_panda_block_missing_signals(pva):
...
         )
+
+
+async def test_panda_unable_to_connect_to_pvi():
+    panda = PandA("NON-EXISTENT")
+
+    with pytest.raises(NotConnected) as exc:
+        await panda.connect(timeout=0.01)
+
+    assert exc.value._errors == "NON-EXISTENT:PVI"
```
coretl, line 143:
```suggestion
    assert exc.value._errors == "pva://NON-EXISTENT:PVI"
```

coretl (approved):

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -184,23 +185,26 @@ def __init__(self, datatype: Optional[Type[T]], read_pv: str, write_pv: str):
...
         try:
-            self.initial_values[pv] = await caget(pv, format=FORMAT_CTRL, timeout=None)
-        except CancelledError:
-            raise NotConnected(self.source)
+            self.initial_values[pv] = await caget(
+                pv, format=FORMAT_CTRL, timeout=timeout
+            )
+        except CANothing:
+            logging.debug(f"signal ca://{pv} timed out")
+            raise NotConnected(f"ca://{pv}")
```
coretl, line 195:
```suggestion
        except CANothing as exc:
            logging.debug(f"signal ca://{pv} timed out")
            raise NotConnected(f"ca://{pv}") from exc
```

PR #102, opened 2024-01-15
Fix panda name typo and add test

olliesilvester:
Had a silly typo in https://github.com/bluesky/ophyd-async/pull/101. Added a test to catch this


rosesyrett (approved):
Looks good :+1: funny that I didn't spot this but good job fixing it quickly

PR #107, opened 2024-01-26
Added more documentation

rosesyrett:


rosesyrett:
as there is currently no support for tango, I think I need to mention this somewhere at a top level. Otherwise it gets annoying having to write everywhere that you *could* do this with a tango layer instead of epics.

I'll do this next week.

DiamondJoseph:
Move code samples into epics.demo or similar. Make sure the words make sense. TODO by me.

coretl:
@callumforrester is any of this worth adding to the current docs after your additions?

callumforrester:
@coretl Will review and get back to you

coretl:
Most of this has already been merged

Issue #99, opened 2024-01-05
Move writer collect_asset_docs logic into detector


rosesyrett:
Currently, the Flyer has a `collect_asset_docs` method which calls (in parallel) the `collect_stream_docs(idx)` methods of each writer underneath it. This is essentially a way to concurrently collect the asset docs of all writers underneath the flyer.

A new issue has been added, https://github.com/bluesky/bluesky/issues/1649, which will allow this logic to be moved into the Detector. So the plan is to:

1. Remove collect_asset_docs entirely from the Flyer; the Flyer should be transparent to any document generation. Remove it from the `DetectorGroupLogic` protocol and implementation(s) also for the same reason.
2. Make the StandardDetector obey the new protocol in the above issue; that is, it should have a `collect_asset_docs(idx)` method as well as something equivalent to `get_index`, which should mean it calls it's underlying writer to figure out what index of data it has written.

In order to test this, we should probably do https://github.com/bluesky/ophyd-async/issues/91 alongside this.

Issue #91, opened 2023-12-08
Add kickoff, complete and describe_collect on StandardDetector


rosesyrett:
This requires moving the existing logic of kickoff/complete and describe_collect from the flyer to the StandardDetector.

This is part of a general work to disassemble the flyer object to purely hold logic triggers for underlying detectors, such that the stream resource/datum documents produced from a fly scan do not eject data about the flyer itself. See [the project page readme](https://github.com/orgs/bluesky/projects/5/) for more details.

abbiemery:
Closed with #116

PR #110, opened 2024-02-05
Move collect asset docs

abbiemery:
This is nominally the PR for #99 . However for this we also need to address parts of #91 as these methods are coupled in varying degrees. 

This also requires the change from Bluesky 1649. This is to allow the collect method in the bundler to accept multiple devices.


abbiemery:
It was discussed in this to remove the enforced event ordering and instead rely on the file writing services to address the ordering from a run. I agree with this but I want to check if people rely on this behaviour and warn them before-hand. @DominicOram Do you rely on this?

DominicOram:
> It was discussed in this to remove the enforced event ordering and instead rely on the file writing services to address the ordering from a run. I agree with this but I want to check if people rely on this behaviour and warn them before-hand. @DominicOram Do you rely on this?

No, I don't think we use any of this

abbiemery:
Closed as this is covered in #91

PR #113, opened 2024-02-13
Add design goals and simple explanation of flyscanning

callumforrester:
Create design goals page in docs, outlining the broad vision for ophyd-async. To explain some of the context, create a simple "what even is flyscanning?"-style page to explain the high level concepts to new people.


docs/user/explanations/design-goals.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Ophyd-async should provide the same building blocks for defining flyscans scans as malcolm_. It should support PandA and Zebra as timing masters by default, but also provide easy helpers for developers to write support for their own devices.
+
+It should enable `motor trajectory scanning <motortraj_>` and `multiple triggering rates<detectorsync_>` based around a base rate, and pausing/resuming scans. Scans should be modelled using scanspec_, which serves as a universal language for defining trajectory and time-resolved scans, and converted to the underlying format of the given motion controller. It should also be possible to define an `outer scan <outerscan_>`.
+
+
+Improved Trajectory Calculation
+-------------------------------
+
+Ophyd-async will provide and improve upon the algorithms that malcolm uses to calculate trajectories for supported hardware.
```
callumforrester, line 36: @coretl This is quite vague but it's the limit of my understanding, could you provide more detail?

docs/user/explanations/design-goals.rst
```diff
@@ -0,0 +1,48 @@
...
+- Conformity to the bluesky protocols
+- Epics (ChannelAccess) as a backend
+
+Ophyd-async should provide built-in support logic for controlling `the same set of devices as ophyd <https://blueskyproject.io/ophyd/user/reference/builtin-devices.html>`_. 
+
+
+Clean Device Definition
+-----------------------
+
+It should be easy to define devices with signals that talk to multiple backends and to cleanly organize device logic via composition.
```
callumforrester, line 20: @coretl I think it's important to expand on this, any ideas?

coretl: We need to be able to:
- Separate the Device interface from the multiple pieces of logic that might use that Device in a particular way
- Define that Signals of a particular type exist without creating them so backends like Tango or EPICS + PVI can fill them in

docs/user/explanations/flyscanning.rst
```diff
@@ -0,0 +1,63 @@
+Flyscanning
```
callumforrester, line 1: From discussion: this should go in the cookbook rather than ophyd-async

olliesilvester:
Elsewhere I've seen the term 'flyscanning' being used to mean 'constant motion scanning'. it might be worth clarifying that there's a distinction here.

docs/user/explanations/design-goals.rst
coretl: The EPICS [pmac module](https://github.com/dls-controls/pmac) supports trajectory scanning, specifying a growing array of positions, velocities and time for axes to move through to perform a scan. Ophyd-async will provide mechanisms for specifying these scans via a [ScanSpec](https://github.com/dls-controls/scanspec), calculating run-ups and turnarounds based on motor parameters, keeping the trajectory scan arrays filled based on the ScanSpec, and allowing this scan to be paused and resumed.

callumforrester:
@olliesilvester yeah this is a problem! Different facilities have different, sometimes conflicting names for the same thing, as do different departments within the same facility. There was some discussion last week about a glossary of terms going into the bluesky cookbook, so we can at least point at a document.

docs/user/explanations/design-goals.rst
abbiemery: I have put this in.

docs/user/explanations/flyscanning.rst
abbiemery: I think for the sake of progress this goes into ophyd-async now, then I pull put on the cook-book issue to move it when it becomes relevant. It's more useful existing in the wrong place than not existing in either place.

abbiemery:
> @olliesilvester yeah this is a problem! Different facilities have different, sometimes conflicting names for the same thing, as do different departments within the same facility. There was some discussion last week about a glossary of terms going into the bluesky cookbook, so we can at least point at a document.

I have added the flyscanning docs in here anyway for now because I figured it is more useful for it to exist than not. At least it explains it from the perspective of ophyd-async. Once we have these above discussions and decide on a glossary we can move it.

PR #121, opened 2024-02-22
Added a new test connecting to devices before initializing RE

evalott100:
Quick one


tests/core/test_device_collector.py
```diff
@@ -56,19 +61,69 @@ def test_device_connector_sync_run_engine_created_connects(RE):
...
+
+
 """
-# TODO: Once passing a loop into the run-engine selector works, this should pass
-async def test_device_connector_async_run_engine_same_event_loop():
-    async with DeviceCollector(sim=True):
-        sim_motor = motor.Motor("BLxxI-MO-TABLE-01:X")
+# SimSignalBackend currently allows a different event loop to set the value
+# so this test is not working
+def test_device_connector_async_run_engine_different_event_loop():
```
callumforrester, line 105: Would prefer `@pytest.skip(reason="<link to github issue>")` rather than commenting the whole test out.

tests/core/test_device_collector.py
```diff
@@ -56,19 +61,69 @@ def test_device_connector_sync_run_engine_created_connects(RE):
...
+        )
+
+    finally:
+        if RE.state not in ("idle", "panicked"):
+            try:
+                RE.halt()
+            except TransitionError:
+                pass
+        loop.call_soon_threadsafe(loop.stop)
+        checking_loop.call_soon_threadsafe(checking_loop.stop)
```
callumforrester, line 96: Could extract both loop and RE out into fixtures and handle the teardown logic there. Not essential but would make the test clearer.

tests/core/test_device_collector.py
```diff
@@ -56,19 +61,69 @@ def test_device_connector_sync_run_engine_created_connects(RE):
     assert working_device.connected
 
 
+def test_device_connector_async_run_engine_same_event_loop():
```
callumforrester, line 64: Not entirely sure from the name what this test is supposed to do, is it asserting that the device will use `loop` even if you call `checking_loop.run_until_complete`?

callumforrester (changes requested):

tests/core/test_device_collector.py
evalott100: I thought about it, but to represent the RE-independent connection of the `DeviceConnector` we'd either have to:

* Initialise the `RunEngine` with the `new_event_loop` in the fixture, and the `DeviceConnector` in the test with the same loop. I don't like this much because we want to go through the device connection prior to any exisiting run-engine - we want to check for future additions of `call_in_bluesky_event_loop` in the codebase, but also because that will be how it'll be done in real devices.
* Initialise the `DeviceCollector` in the fixture before run-engine initialisation with the same event-loop - which would lead to a fixture as complicated as the test, and only being used for one test.

tests/core/test_device_collector.py
evalott100: I should change to `device_connector_async -> async_device_connector`... I'll clean it up.

It's checking that devices can be connected to in the `DeviceConnector` prior to any run-engine initialisation, and then can be put through the run-engine later so long as you explicitly pass in the same event-loop.

The test that's now skipping checks that the above fails if you don't pass in the same event-loop to the run-engine and let it create a new one. It doesn't pass currently, I think because `SimSignalBackends` will allow connection with one event-loop and `put` with another, which real devices will not.

callumforrester: Why is the second checking loop required in that case?

evalott100: The `loop` can't be used because it's still active in the `RunEngine`. `An asyncio.run()` to get the motor setpoint works, but for some reason causes
`tests/core/test_device_save_loader.py::test_load_from_yaml` to fail. Happy to investigate why and address it in this PR.

I've noticed `tests/epics/test_signals.py::test_backend_get_put_monitor` is also timing out sometimes so maybe it would be best to make a new issue to fix these flakey tests.

callumforrester: I'd just stick it in an issue for now

Issue #130, opened 2024-02-29
Add Multipliers to prepare


abbiemery:
We want to be able to run one detector twice as fast as another but we still need to correlate the multiple fast detector frames to the single slow detector frame (In the case of twice as fast, 2:1). We do this my making `TriggerInfo.num` twice as big in the faster detector. This trigger info is passed to the detector in its prepare method. If we also set `TriggerInfo.multipler` then this can make it into the stream resource so downstream processes can zip the files together. 

Depends on #129

jwlodek:
Should `TriggerInfo.num` remain the number of triggers for the detector instance itself, or should it now be the number of triggers for the slowest detector in the scan, since we can presumably get the number of triggers for each detector based on the multiplier?

coretl:
> Should `TriggerInfo.num` remain the number of triggers for the detector instance itself, or should it now be the number of triggers for the slowest detector in the scan, since we can presumably get the number of triggers for each detector based on the multiplier?
> As best as I can tell, it would be easier to use the baseline `num` for triggers, since the original trigger_info is coming from the higher level HardwareTriggeredFlyable class, with the same one being passed to each detector in `prepare`. I guess the question becomes should each `StandardDetector` have it's `TriggerInfo` updated to the actual `num`, or should we leave it as the baseline frequency and recalculate actual num triggers whenever needed with the multiplier.


I think it should be the number of triggers sent to the detector. We will use a PandA Pulse block to take the input pulse train and multiply it up, changing the livetime and deadtime too, so it makes sense for it to calculate a new `TriggerInfo` with this information and send it to `detector.prepare()`. The only reason the detector needs the multiplier is so that it can put the right information in the descriptor and StreamResource documents.

coretl:
Depends on #127

coretl:
Do this as part of #314

coretl:
Do this as part of #314

Issue #129, opened 2024-02-29
Make StandardDetector call writer.open in prepare


abbiemery:
The writers are currently opened in stage. This needs to be move to prepare to we can use multipliers. We also need to move multiplers into prepare.

abbiemery:
Closed in favour of #314 .

Issue #123, opened 2024-02-28
Add step scan support into the prepare for the standard detector


abbiemery:
This is a follow on from #116 . The prepare method deals with the fly scanning behaviour of the detector, but not the step scanning. This is not major issue as the prepare is not currently widely used i step scanning, but it could be. As prepare is meant to be able to deal with both a reasonable behaviour for `prepare` in the standard detector would be to have prepare set trigger values for both options.  It then will go on to perform its other tasks for fly scans. The step scan `trigger` method can check that the cached trigger info from prepare is for a software trigger (so not a flyscan) if it is to use it. If there is no trigger info then prepare has not been called and we should use the existing defaults ( as they are present now), if it does exist but its fly scanning trigger info, fail with an error.

jwlodek:
One thought: if #129 is true, then wouldn't `prepare` become required to be run after `stage` and before `trigger` even in a step scan, because otherwise the `writer` will never be opened. In that case, should `trigger` just fail outright if `self._trigger_info` is `None` (indicating `prepare` was not run)?

coretl:
I think that depends on whether we want to support doing step scans of `Preparable` detectors using existing plans. If yes, then we need to lazily call `prepare()` if not already done in `trigger()`. In not, then we can do as you say and just fail if it hasn't been run

coretl:
@danielballan @tacaswell thoughts on whether we want people to be able to upgrade to ophyd-async without changing their plans at all?

coretl:
I think we need a lazy prepare in `trigger()`

coretl:
Close in favour of #314

PR #132, opened 2024-03-01
Add multiplier field to TriggerInfo dataclass. Update two tests to reflect change

jwlodek:
I will be working on #129 and #130 in this PR.

This PR depends on #116, so it is initially configured to merge into the branch for that PR.


jwlodek:
The most recent commit begins to address #123

Issue #126, opened 2024-02-29
Move PandA block making functionality into a standalone pvi function


abbiemery:
We need a method that recuses down and makes instances of the blocks based on PVI data structures. We have a method already but we need to turn it into a stand alone function.

Depends on #125 .

PR #133, opened 2024-03-05
PandA type annotations parent class

### LINKED ISSUE #125: Put the PandA type annotations in a CommonPandABlocks parent class
This Parent class will only contain type annotations for each of the blocks in a PandA and will define the common interface that is expected by all logic that uses the PandA.

This will require changing the logic on how you find type annotation in order to make the blocks.

evalott100:
Closes #125 

Working off of #43, should be merged after.


evalott100:
We'll include this change in the PR for https://github.com/bluesky/ophyd-async/issues/126

PR #137, opened 2024-03-07
Use metadata from DB records to populate DataKeys

DiamondJoseph:


src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +35,58 @@
...
+    """
+
+    # Always present keys
+    if not value.ok:
+        raise Exception(f"{source} error: {value.errorcode}: {str(value)}")
+
+    scalar = value.element_count == 1
+
+    d = DataKey(
+        source=source,
```
DiamondJoseph, line 63: source that is passed in is value.name with the protocol. Can we get the protocol from elsewhere to not need to pass the source in?

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +35,58 @@
...
+    Args:
+        value (AugmentedValue): Description of the the return type of a DB record
+
+    Returns:
+        DataKey: A rich DataKey describing the DB record
+    """
+
+    # Always present keys
+    if not value.ok:
+        raise Exception(f"{source} error: {value.errorcode}: {str(value)}")
```
DiamondJoseph, line 58: What's the correct Exception to throw here?/ at all? Previously were not checking this.
Docs of AugmentedValue says that str(value) contains the error message- does this mean after calling signal.get()?

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +35,58 @@
...
+
+    scalar = value.element_count == 1
+
+    d = DataKey(
+        source=source,
+        dtype=dbr_to_dtype[value.datatype] if scalar else "array",
+        shape=[] if scalar else [len(value)],
```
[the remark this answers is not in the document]
coretl: `element_count` is canonical, but len is derived from it so they should always be the same

src/ophyd_async/epics/_backend/_aioca.py
coretl: protocol is `ca://` for the whole of this file, so fine to put that as a global

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +35,58 @@
...
+                "lower_warning_limit": value.lower_warning_limit,
+                "upper_warning_limit": value.upper_warning_limit,
+            }
+        )
+    if value.datatype in _floating_dbr:
+        d["precision"] = value.precision
+    if value.datatype is dbr.DBR_ENUM:
+        d["choices"] = value.enums
+    if value.datatype is dbr.DBR_STRING:
+        d["timestamp"] = value.timestamp
```
coretl, line 87: Can we do `hasattr` here? There are some tricky upstream combinations of datatypes, but the value if present should always be put in the output

src/ophyd_async/epics/_backend/_aioca.py
coretl: We should never get here, as if `value.ok == False` then the `caget` should have raised. 

```suggestion
    assert value.ok, f"Error reading {source}: {value}"
```

tests/epics/test_signals.py
```diff
@@ -156,24 +149,49 @@ class MyEnum(str, Enum):
...
+    "upper_ctrl_limit": ANY,
+    "lower_ctrl_limit": ANY,
+}
 
+_metadata: Dict[str, Dict[str, Any]] = {
+    "string": {"timestamp": ANY},
+    "integer": _common_metadata,
+    "number": {**_common_metadata, "precision": ANY},
+    "enum": {},
+}
```
DiamondJoseph, line 165: I've tried a few arrangements of the metadata, but because it's partially inconsistent I wasn't confident with any of them. I'd like to be able to get it back to descriptor(value) == backend.get_descriptor() rather than descriptor(values).items() <= backend.get_descriptor().items()

src/ophyd_async/epics/_backend/_aioca.py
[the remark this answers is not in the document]
DiamondJoseph: Just noting for reference if someone comes back to this, element_count >= len(value), so stuck with element_count

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +36,59 @@
...
+    "lower_alarm_limit",
+    "upper_alarm_limit",
+    "lower_ctrl_limit",
+    "upper_ctrl_limit",
+    "lower_disp_limit",
+    "upper_disp_limit",
+    "lower_warning_limit",
+    "upper_warning_limit",
+    "precision",
+    "timestamp",
```
coretl, line 52: Not timestamp, that goes in the value

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +36,59 @@
...
+    "units",
+    "lower_alarm_limit",
+    "upper_alarm_limit",
+    "lower_ctrl_limit",
+    "upper_ctrl_limit",
+    "lower_disp_limit",
+    "upper_disp_limit",
+    "lower_warning_limit",
+    "upper_warning_limit",
+    "precision",
```
coretl, line 51: @danielballan do we want all these to appear in the descriptor document?

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -35,7 +36,59 @@
...
+        source=source,
+        dtype=dtype if scalar else "array",
+        # strictly value.element_count >= len(value)
+        shape=[] if scalar else [len(value)],
+    )
+    for key in _common_meta:
+        if hasattr(value, key):
+            attr = getattr(value, key)
+            if isinstance(attr, str) or not isnan(attr):
+                d[key] = attr
```
coretl, line 85: Nit
```suggestion
        attr = getattr(value, key, nan)
        if isinstance(attr, str) or not isnan(attr):
            d[key] = attr
```

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -88,11 +140,9 @@ def write_value(self, value: Union[Enum, str]):
...
         return self.enum_class(value)
 
-    def descriptor(self, source: str, value: AugmentedValue) -> Descriptor:
-        choices = [e.value for e in self.enum_class]
-        return dict(
-            source=source, dtype="string", shape=[], choices=choices
-        )  # type: ignore
+    def descriptor(self, value: AugmentedValue) -> DataKey:
+        # Sometimes DBR_TYPE returns as String, must pass choices still
+        return _data_key_from_augmented_value(value, choices=self.choices)
```
coretl, line 145: Don't you calculate these above anyway?

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -15,7 +16,7 @@
     caput,
 )
 from aioca.types import AugmentedValue, Dbr, Format
-from bluesky.protocols import Descriptor, Dtype, Reading
+from bluesky.protocols import DataKey, Dtype, Reading
```
coretl, line 19: Probably want to make this change in other places, and change the `SignalBackend.descriptor` method to `.datakeys` as well. That should be another ticket though, captured in #145

coretl (commented):
A couple of minor things, please could you make a ticket for doing the same for the PVA side so we get this in PandA too?

src/ophyd_async/epics/_backend/_aioca.py
DiamondJoseph: The test Enum signals are returning DBR_TYPE of String and don't have a choices field :shrug: I don't know if they are misconfigured? I did try getting rid of this type same as CaArrayConverter, but had to restore this method.

src/ophyd_async/epics/_backend/_aioca.py
DiamondJoseph: I'm happy to turn down to just units, precision. Currently for my downstream purposes I turn lower_ctrl_limit into soft_limit_min for NeXus, so if we can keep just one that's the one I want.

danielballan: Only units and precision are blessed in the schema, but IIRC the schema is open-ended (additional properties allowed) so you can add any that you need.

DiamondJoseph:
> [...] make a ticket for doing the same for the PVA side [...]

Captured as #146

src/ophyd_async/epics/_backend/_aioca.py
DiamondJoseph: Added some docstring to the EnumConverter to explain what's going on

src/ophyd_async/epics/_backend/_aioca.py
coretl: PVA also defines the same set of limits:
- [`display.limitLow` and `display.limitHight`](https://docs.epics-controls.org/en/latest/pv-access/Normative-Types-Specification.html#display-t)
- [`control.limitLow` and `control.limitHight`](https://docs.epics-controls.org/en/latest/pv-access/Normative-Types-Specification.html#control-t)
- [`valueAlarm.lowWarningLimit`, `valueAlarm.lowAlarmLimit`, `valueAlarm.highWarningLimit`, `valueAlarm.highAlarmLimit`](https://docs.epics-controls.org/en/latest/pv-access/Normative-Types-Specification.html#alarmlimit-t)

The question here is what is useful downstream.

- If I am capturing a diode PV, do I want the display limits that would be used as axis ranges for plotting?
- If I am capturing a furnace setpoint PV, do I want the control limits that show I was close to the maximum temperature the device accepts?
- If I am capturing a thermocouple PV, do I want the warning or alarm limits to show that the sample was melting?

For that last point, we currently capture severity with each value, but we questioned the usefulness of that as maybe the scan should abort if the sample got too hot?

DiamondJoseph: | CA | PVA |
| -- | -- |
| lower_alarm_limit | valueAlarm.lowWarningLimit |
| upper_alarm_limit | valueAlarm.highWarningLimit |
| lower_ctrl_limit | control.limitLow |
| upper_ctrl_limit | control.limitHigh |
| lower_disp_limit | display.limitLow |
| upper_disp_limit | display.limitHigh |
| lower_warning_limit | valueAlarm.lowWarningLimit |
| upper_warning_limit | valueAlarm.highWarningLimit |

Presumably the correct equivalents?
Do we want to do some mapping to either/both to get something consistent?

coretl: Yes, I think we should map both to the same, then add them to event model as optional.
However I only think we should do this for the useful fields...
@danielballan which fields do you think are useful?

coretl: How about we have nested objects, so:
- `limits.control.low` and `limits.control.high`
- `limits.display.low` and `limits.display.high`
- `limits.warning.low` and `limits.warning.high`
- `limits.alarm.low` and `limits.alarm.high`

DiamondJoseph:
Replaced by #367

PR #141, opened 2024-03-08
Add PandA hdf writer

olliesilvester:
Adds an HDF writer for the `PandA`.

- Upon calling `open`, which should be triggered by arming the `PCAP` block, the writer reads from all the capture signals to find which of these were actually marked for capture. It creates datasets based off these


olliesilvester:
@jwlodek This isn't ready to try out on a real panda just yet, but should be in the next few days

src/ophyd_async/panda/writers/hdf_writer.py
```diff
@@ -0,0 +1,234 @@
...
+from .panda_hdf import _HDFDataset, _HDFFile
+
+
+@dataclass()
+class HdfSignals:
+    file_path: SignalRW  # This is the directory rather than path. Path is read only
+    file_name: SignalRW
+    num_capture: SignalRW
+    num_captured: SignalR
+    capture: SignalRW
```
coretl, line 31: That's a different approach to the one I was expecting. I was thinking that you would have something like this in `panda.py`:
```
class DataBlock(Device):
    capture: SignalRW[bool]
    num_capture: SignalRW[int]
    hdf_file_path: SignalRW[str]
    ...
```
This would then be filled in in the `PandA` device, then passed to the `PandaHDFWriter.__init__(self, data: DataBlock, ...)`

(At the moment the pvi names don't have underscores where they should after https://github.com/PandABlocks/PandABlocks-ioc/issues/101)

src/ophyd_async/panda/writers/hdf_writer.py
```diff
@@ -0,0 +1,234 @@
...
+    signal_values = await asyncio.gather(*do_read)
+
+    assert len(signal_values) == len(
+        capture_signals.keys()
+    ), "Length of read signals are different to length of signals"
+    signals_to_capture: Dict[str, Dict[str, SignalR | Capture]] = {}
+    for signal_path, signal_object, signal_value in zip(
+        capture_signals.keys(), capture_signals.values(), signal_values
+    ):
+        if (signal_value.value in iter(Capture)) and (signal_value.value != Capture.No):
```
coretl, line 82: I'm wondering why this works, as the `Capture` enum defined above is not the same as the generated Enum subclass that the "fill in signals" code uses. Maybe because they both subclass str?

src/ophyd_async/panda/writers/hdf_writer.py
```diff
@@ -0,0 +1,234 @@
...
+                actual_block = attribute_path.split(".")[-3]
+                block_name = f"{actual_block}.{block_name}"
+
+            if value["capture_type"] == Capture.MinMaxMean:
+                shape = [3]
+            elif value["capture_type"] == Capture.MinMax:
+                shape = [2]
+            else:
+                shape = [1]
+                # TODO check if this is correct format
```
coretl, line 178: Actually `Min Max Mean` will generate 3 datasets, so you need something like:
```python
for suffix in value["capture_type"].value.split(" "):
    self._datasets.append(
        _HDFDataset(...., path=f"{actual_block}{block_name}.{signal_name}.{suffix}", ...)
```

coretl: Actually it strikes me that all this information is available in the capture table:
https://github.com/PandABlocks/PandABlocks-ioc/blob/acfefb3da61db47a32d6fff71f26236c35909812/src/pandablocks_ioc/_pvi.py#L183-L227

@evalott100 @olliesilvester I think we need a discussion about the PVI structures of PandA and how it will fit together

coretl (commented):
I'm inclined to say don't act on the more invasive of my comments now, just get this working with the minimum changes, then make a ticket to revisit the pvi type changes here

src/ophyd_async/panda/writers/hdf_writer.py
olliesilvester: Yeah, I was planning on putting this into `panda.py` after Eva's change which changes that file. It might not conflict though, I can add it here

coretl: The key thing is, this is a request for the pvi code to fill it in, rather than a dataclass where the already filled in signals are passed to it

tests/panda/test_writer.py
```diff
@@ -0,0 +1,181 @@
...
+        assert "source" in entry
+        assert entry.get("external") == "STREAM:"
+    expected_datakeys = [
+        "test-panda.block1.test.Min",
+        "test-panda.block1.test.Max",
+        "test-panda.block1.test.Mean",
+        "test-panda.block2.test.Value",
+    ]
+    for key in expected_datakeys:
+        assert "test-panda.block1.test.Min" in description
```
stan-dot, line 124: shouldn't this be 
`assert key in description`?

olliesilvester: yeah, thanks!

olliesilvester:
I will add some more tests, but other than that I think this is ready. I can make another issue for future improvements:

- There is a lot of code which is common between `AreaDetector`'s `hdf_writer` and this new `hdf_writer`. Would be nice to put it somewhere common
- Use information from the capture table when writing the descriptor (https://github.com/bluesky/ophyd-async/pull/141#discussion_r1521775460)
- The `HdfSignal` class can be removed after https://github.com/PandABlocks/PandABlocks-ioc/issues/101 is fixed
- `DataBlock` in `panda.py` can be changed to handle Enums properly once https://github.com/bluesky/ophyd-async/pull/152 is merged

src/ophyd_async/panda/writers/panda_hdf.py
```diff
@@ -0,0 +1,50 @@
...
+        # Indices are relative to resource
+        if indices_written > self._last_emitted:
+            indices = dict(
+                start=self._last_emitted,
+                stop=indices_written,
+            )
+            self._last_emitted = indices_written
+            for bundle in self._bundles:
+                yield bundle.compose_stream_datum(indices)
+        return None
```
DiamondJoseph, line 50: To prevent duplication of the areadetector hdf_writer, a common root HDFFile could be defined (in ophyd_async/core?) that handles common behaviour. `T` is the dataclass _HDFDataset

```
class HDFFile(Generic[T]):
    """
    :param directory_info: Contains information about how to construct a StreamResource
    :param full_file_name: Absolute path to the file to be written
    :param datasets: Datasets to write into the file
    """

    def __init__(
        self,
        directory_info: DirectoryInfo,
        full_file_name: Path,
        datasets: List[T],
    ) -> None:
        self._last_emitted = 0
        self._bundles = [
            compose_stream_resource(
                spec="AD_HDF5_SWMR_SLICE",
                root=str(directory_info.root),
                data_key=ds.name,
                resource_path=str(full_file_name.relative_to(directory_info.root)),
                resource_kwargs=asdict(ds),
            )
            for ds in datasets
        ]

    def stream_resources(self) -> Iterator[StreamResource]:
        for bundle in self._bundles:
            yield bundle.stream_resource_doc

    def stream_data(self, indices_written: int) -> Iterator[StreamDatum]:
        # Indices are relative to resource
        if indices_written > self._last_emitted:
            indices = dict(
                start=self._last_emitted,
                stop=indices_written,
            )
            self._last_emitted = indices_written
            for bundle in self._bundles:
                yield bundle.compose_stream_datum(indices)
        return None
```

Then this is just 

```suggestion
class _PandaHDFFile(HDFFile[_HDFDataset]): ...
```

Do you want the name of the dataset in both the data_key and the resource_kwargs?

coretl: This is being handled in #173

src/ophyd_async/panda/writers/hdf_writer.py
```diff
@@ -0,0 +1,220 @@
...
+            panda_device.data.hdfdirectory,
+            panda_device.data.hdffilename,
+            panda_device.data.numcapture,
+            panda_device.data.numcaptured,
+            panda_device.data.capture,
+        )
+
+        # Get capture PVs by looking at panda. Gives mapping of dotted attribute path
+        # to Signal object
+        self.capture_signals = get_capture_signals(self.panda_device)
```
coretl, line 122: Unfortunately this is too early, those signals won't exist until after `PandA.connect` so this sound be done in `open` instead

src/ophyd_async/panda/panda.py
```diff
@@ -72,9 +85,12 @@ class PandA(Device):
     pulse: DeviceVector[PulseBlock]
     seq: DeviceVector[SeqBlock]
     pcap: PcapBlock
+    data: DataBlock
 
     def __init__(self, prefix: str, name: str = "") -> None:
         super().__init__(name)
+        if not prefix.endswith(":"):
+            prefix = prefix + ":"
```
coretl, line 93: Not a fan of adding things to the prefix silently. I'd be happy with:
```suggestion
        assert prefix.endswith(":"), f"PandA prefix '{prefix}' must end in ':'"
```
Or even ditch it altogether, as site naming conventions differ a lot and we might change the PandA IOC to have a configurable PV separator when it goes to FastCS...

src/ophyd_async/panda/utils.py
```diff
@@ -1,4 +1,12 @@
-from typing import Any, Dict, Sequence
+from typing import Any, Dict, Sequence, TypedDict
+
+
+class PVIEntry(TypedDict, total=False):
+    d: str
+    r: str
+    rw: str
+    w: str
+    x: str
```
coretl, line 9: This should disappear now #142 is merged

src/ophyd_async/panda/panda.py
olliesilvester: I will do the assert for now. I seem to remember issues on the IOC side if there's no colon at the end

coretl: Yeah, at the moment give the IOC the prefix of `MY-PANDA-PREFIX` and it makes `MY-PANDA-PREFIX:SEQ1:ENABLE`, so the assert is helpful, but other sites had a different convention so might like `id3{xfg-sd}/seq1/enable` so we want to support that on the IOC side in future and will have to remove the assert when we do that

evalott100:
I'm currently blocked by this, I'm going to take the liberty of squashing it so I can easily merge with my own branch.

evalott100:
Replced with https://github.com/bluesky/ophyd-async/pull/198

PR #147, opened 2024-03-13
Enum extension

DiamondJoseph:
Allows passing an Enum datatype to a Signal, when the Signal may have additional possible values.


coretl (approved):
Excellent! @olliesilvester FYI

PR #138, opened 2024-03-07
Add detector control for PandA Pcap

tomtrafford:
Implementation for DetectorControl for PandA pcap.This resolves #124

PR #150, opened 2024-03-14
NSLS-II test with PandA

mrakitin:
Based on #141 plus #138 plus `main` minus a merge commit from #147 plus some missing pieces.

Works in tandem with https://github.com/NSLS-II-TST/profile_collection/pull/4/files.


mrakitin:
We don't need this PR/branch as we succeeded using #141 directly.

PR #153, opened 2024-03-14
Use DataKey instead of Descriptor

### LINKED ISSUE #145: Produce valid datakeys for a Table
```[tasklist]
### Tasks
- [ ] Change `SignalBackend.get_datakey(self, source: str) -> DataKey` to `SignalBackend.get_datakeys(self, name: str, source: str) -> Dict[str, DataKey]` and make `Signal.describe` safely merge the dictionaries of all the read signals
- [ ] Change `SignalBackend.get_reading(self) -> Reading` to `SignalBackend.get_readings(self, name: str) -> Dict[str, Reading]` and make `Signal.read` safely merge the dictionaries of all the read signals
- [ ] Change PVA and CA converters to match
- [ ] Make `PvaTableConverter.get_datakeys` emit a dict with an entry per column of the correct type
- [ ] Make `PvaTableConverter.get_readings` emit a dict with an entry per column of the correct type
```

DiamondJoseph:
Closes #145 

Descriptor is overloaded with a DescriptorDocument, which contains many DataKeys. A DataKey is the key to a data field, describing how it should be read.

### DIFF
--- a/src/ophyd_async/core/signal_backend.py
+++ b/src/ophyd_async/core/signal_backend.py
@@ -1,7 +1,7 @@
 from abc import abstractmethod
 from typing import Generic, Optional, Type
 
-from bluesky.protocols import Descriptor, Reading
+from bluesky.protocols import DataKey, Reading
 
 from .utils import DEFAULT_TIMEOUT, ReadingValueCallback, T
 
@@ -27,7 +27,7 @@ class SignalBackend(Generic[T]):
         """Put a value to the PV, if wait then wait for completion for up to timeout"""
 
     @abstractmethod
-    async def get_descriptor(self, source: str) -> Descriptor:
+    async def get_datakey(self, source: str) -> DataKey:
         """Metadata like source, dtype, shape, precision, units"""
 
     @abstractmethod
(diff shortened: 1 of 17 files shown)


coretl:
Is this still in progress?

DiamondJoseph:
@coretl I think it was draft as had conflicts with the bool signal, I've rebased and will see if the CI is happy

src/ophyd_async/core/signal_backend.py
```diff
@@ -27,7 +27,7 @@ async def put(self, value: Optional[T], wait=True, timeout=None):
         """Put a value to the PV, if wait then wait for completion for up to timeout"""
 
     @abstractmethod
-    async def get_descriptor(self, source: str) -> Descriptor:
+    async def get_datakey(self, source: str) -> DataKey:
```
coretl, line 30: There is a longstanding issue that tables are not representable under the current scheme as we can only put a single entry in the descriptor, where we would like to put one entry for each column in the table. To fix this we would eventually like to make this:

```suggestion
    async def get_datakey(self, name: str, source: str) -> Dict[str, DataKey]:
```

...

coretl (approved):
I started writing about what we would need to support tables, but stopped because I think it should now be a separate ticket. This looks fine as it is.

DiamondJoseph:
> I started writing about what we would need to support tables, but stopped because I think it should now be a separate ticket. 

Has that ticket been created?

Issue #122, opened 2024-02-23
Consider Path Signals


callumforrester:
Some string PVs are intended only to store file paths, e.g. the HDF5 plugin write path. Perhaps there is value in representing these with a `Signal[Path]` rather than a `Signal[str]` and performing some validation on input values.

See [ophyd_async.epics.areadetector.writers.nd_file_hdf.NDFileHDF](https://github.com/bluesky/ophyd-async/blob/641894d6404072f4e3413a82383ef257186d0735/src/ophyd_async/epics/areadetector/writers/nd_file_hdf.py#L19) for an example.

@coretl any thoughts?

DominicOram:
:+1: from me

coretl:
Could be done, but I'd say this is low priority compared with the other things we're doing

callumforrester:
Will leave it as a good first issue then

coretl:
Closing as `PathProvider` makes the above usecase less compelling. We can revisit after 1.0 if this is still needed

PR #154, opened 2024-03-15
Update hdfwriter to include full filepath in root key in resource doc, add uuid DirectoryProvider for dynamic filenames

jwlodek:
This PR does the following:

* Includes full filepath from the PandA hdf writer into the stream resource document. Before this the `root` key was hard coded to `/`
* Removes hard-coded period and device name in PandA writer filenames. I think if you want to include that in the name adding it to the passed in `DirectoryProvider` is a better solution.
* Adds a `UUIDDirectoryProvider` that generates dynamic filenames based on `uuid4` - this is how we typically run things at NSLS-II.

### DIFF
--- a/src/ophyd_async/panda/writers/hdf_writer.py
+++ b/src/ophyd_async/panda/writers/hdf_writer.py
@@ -208,7 +208,7 @@ class PandaHDFWriter(DetectorWriter):
         if indices_written:
             if not self._file:
                 self._file = _HDFFile(
-                    await self.hdf.file_name.get_value(), self._datasets
+                    await self.hdf.file_path.get_value(), await self.hdf.file_name.get_value(), self._datasets
                 )
                 for doc in self._file.stream_resources():
                     yield "stream_resource", doc
--- a/src/ophyd_async/panda/writers/hdf_writer.py
+++ b/src/ophyd_async/panda/writers/hdf_writer.py
@@ -79,6 +79,8 @@ async def get_signals_marked_for_capture(
     for signal_path, signal_object, signal_value in zip(
         capture_signals.keys(), capture_signals.values(), signal_values
     ):
+        # use .val instead of .val_capture
+        signal_path = signal_path.replace("_capture", "")
         if (signal_value.value in iter(Capture)) and (signal_value.value != Capture.No):
             signals_to_capture[signal_path] = {
                 "signal": signal_object,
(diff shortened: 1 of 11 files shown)


src/ophyd_async/panda/writers/hdf_writer.py
```diff
@@ -133,9 +133,7 @@ async def open(self, multiplier: int = 1) -> Dict[str, Descriptor]:
         info = self._directory_provider()
         await asyncio.gather(
             self.hdf.file_path.set(str(info.root / info.resource_dir)),
-            self.hdf.file_name.set(
-                f"{info.prefix}.{self.panda_device.name}{info.suffix}.h5"
-            ),
+            self.hdf.file_name.set(f"{info.prefix}{info.suffix}.h5"),
```
mrakitin, line 136: A note to reviewers: this change was to exclude the mandatory PandA device name from the file name. If needed, this can be added via `info.prefix` at the device instantiation step.

mrakitin (approved):
We tested this addition at our lab with real PandA. All worked great!

mrakitin:
The tests/linter issues need to be resolved, of course.

DiamondJoseph:
This is handled slightly differently with the [exisitng HDFFileWriter (passing the DirectoryInfo in whole)](https://github.com/bluesky/ophyd-async/pull/86/files#diff-ec10599188f51f48bf97ff110c773eae5c22856d23bb01d2659112d4a852e335), maybe allowing us to refactor the writer to use Path signals if #122 gets merged. It'd be good to settle on a common handling (not opposed to changing what's already in) or merging the implementations of HDFWriter, _HDFFile

coretl:
@mrakitin why did you need to remove the PandA name from the HDF file path? We use the same DirectoryProvider in multiple detectors, so they need some level of uniqueness, hence the device name in the path.

An alternative would be to change the signature of `DirectoryProvider` so you would `__call__` it with `Device.name` and a file extension, and it would make the whole `Path` for you...

coretl:
> I was going to include a `DirectoryProvider` that would append `Device.name` to the filename prefix with `__call__` to this PR. Should I add that back in? My only hangup was what it should be called - I had originally used `StaticDirectoryProviderWithDeviceName`, but that seems quite a bit too long of a name. 

Could pass an enum to the `__init__` to specify the file naming behaviour? `StaticDirectoryProvider("/tmp/files", FileNaming.UID)`

coretl:
@callumforrester @DiamondJoseph could we rename to `FilePathProvider`?

Issue #117, opened 2024-02-21
Add support for watchers into AsyncStatus.wrap 


abbiemery:
Pass is as an extra argument into the wrapped function.

coretl:
An idea for the signature:
```
T = TypeVar('T')
P = ParamSpec('P')
Watchers = Sequence[Callable]

class AsyncStatus(Status):
...
    @classmethod
    def wrap(cls, f: Callable[Concatenate[T, Watchers, P], Coroutine]) -> Callable[Concatenate[T, P], "AsyncStatus"]:
        @functools.wraps(f)
        def wrap_f(self, *args, **kwargs) -> AsyncStatus:
            watchers = []
            return AsyncStatus(f(self, watchers, *args, **kwargs), watchers)

        return wrap_f
```

Then we can use it to wrap `prepare` and `set` as well as `kickoff` and `complete`

coretl:
Thinking more about this, I wonder if we should support signatures with and without watchers, i.e.
```python
T = TypeVar('T')
P = ParamSpec('P')
Watchers = Sequence[Callable]

class AsyncStatus(Status):
...
    @classmethod
    def wrap(cls, f: Callable[P, Awaitable]) -> Callable[P, "AsyncStatus"]:
        @functools.wraps(f)
        def wrap_f(*args, **kwargs) -> AsyncStatus:
            return AsyncStatus(f(*args, **kwargs))

        return wrap_f

    @classmethod
    def wrap_passing_watchers(cls, f: Callable[Concatenate[T, Watchers, P], Awaitable]) -> Callable[Concatenate[T, P], "AsyncStatus"]:
        @functools.wraps(f)
        def wrap_f(self: T, *args, **kwargs) -> AsyncStatus:
            watchers = []
            return AsyncStatus(f(self, watchers, *args, **kwargs), watchers)

        return wrap_f
```

coretl:
This would mean we could do:
```python
class MyDevice(Stageable, Movable):
    @AsyncStatus.wrap
    def stage(self):
        ...
    @AsyncStatus.wrap_passing_watchers
    def set(self, watchers: Watchers, value: float):
        ...
```

Rather than passing a superfluous `watchers` to `stage()`. There is `inspect` magic we could do to put this in the same `wrap` decorator but I'm not sure it's worth it...

coretl:
@DominicOram @dperl-dls bike shedding welcome...

dperl-dls:
[...]

PR #155, opened 2024-03-20
[Waiting for Apr 5 python 3.9 deprecation] Extend AsyncStatus wrapper to allow methods and functions with args

### LINKED ISSUE #45: limited functionality of AsyncStatus wrap
At the moment, AsyncStatus.wrap does not allow wrapping:
1. methods of classes which have args or kwargs,
2. functions outside of classes

```python
    @classmethod
    def wrap(cls, f: Callable[[T], Coroutine]) -> Callable[[T], "AsyncStatus"]:
        @functools.wraps(f)
        def wrap_f(self) -> AsyncStatus:
            return AsyncStatus(f(self))

        return wrap_f
```

The above means if a function is wrapped with this class method, `my_func = AsyncStatus.wrap(my_func)`, if the function signature originally looked something like `my_func(self)` it behaves as expected, but if it has a signature like `my_func(self, a, b, c, extra=None)` the wrapping means you can no longer pass extra arguments to it.

dperl-dls:
Fixes #45

### DIFF
--- a/src/ophyd_async/core/async_status.py
+++ b/src/ophyd_async/core/async_status.py
@@ -78,8 +78,8 @@ class AsyncStatus(Status):
     @classmethod
     def wrap(cls, f: Callable[P, Coroutine]) -> Callable[P, "AsyncStatus"]:
         @functools.wraps(f)
-        def wrap_f(self) -> AsyncStatus:
-            return AsyncStatus(f(self))
+        def wrap_f(self, *args, **kwargs) -> AsyncStatus:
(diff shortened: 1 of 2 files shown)


coretl:
Just to let you know there is also https://github.com/bluesky/ophyd-async/issues/117, not sure if they should be 2 separate PRs or do them both in one?

dperl-dls:
@coretl Oh, cool, thanks for letting me know. Happy to do that in here as well since they're such small related changes. I'll test it closer to the time it becomes actually relevant.

PR #157, opened 2024-03-25
Pin to bluesky alpha release rather than git branch

callumforrester:
Enables a new alpha release of `ophyd-async`


callumforrester:
Another alpha release of bluesky is needed to make the build pass

coretl (approved):

PR #159, opened 2024-03-27
Tidy up core detector module docs

callumforrester:
Tidy up core detector module:

- Standardise docstrings and update parmaters
- Fix public/private methods
- Include generic in standard detector mro so the generic argument to
  prepare() is properly checked

### DIFF
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -10,6 +10,7 @@ from typing import (
     AsyncIterator,
     Callable,
     Dict,
+    Generic,
     List,
     Optional,
     Sequence,
@@ -39,6 +40,8 @@ T = TypeVar("T")
 
 
 class DetectorTrigger(str, Enum):
+    """Type of mechanism for triggering a detector to take frames"""
+
     #: Detector generates internal trigger for given rate
     internal = "internal"
     #: Expect a series of arbitrary length trigger signals
@@ -51,6 +54,8 @@ class DetectorTrigger(str, Enum):
 
 @dataclass(frozen=True)
 class TriggerInfo:
+    """Minimal set of information required to setup triggering on a detector"""
+
     #: Number of triggers that will be sent
     num: int
     #: Sort of triggers that will be sent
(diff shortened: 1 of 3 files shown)


src/ophyd_async/core/detector.py
```diff
@@ -73,17 +83,31 @@ async def arm(
         trigger: DetectorTrigger = DetectorTrigger.internal,
         exposure: Optional[float] = None,
     ) -> AsyncStatus:
-        """Arm the detector and return AsyncStatus.
+        """Arm detector, do all necessary steps to prepare detector for triggers.
 
-        Awaiting the return value will wait for num frames to be written.
+        Args:
+            num: Expected number of frames
```
callumforrester, line 89: Any idea what the user should do if the number of frames is unknown upfront? Would be good to include something about that here

coretl: I think we should support `0` as "go forever", but detectors that don't support it should raise

callumforrester: Hmm, how about `-1`? A more traditional stand-in for infinity.
Although I would still rather two separate APIs if that's possible.

callumforrester: Continued in #162

callumforrester:
CI Fix: https://github.com/bluesky/ophyd-async/pull/164

abbiemery (approved):
I have a query but nothing worth blocking for

src/ophyd_async/core/detector.py
```diff
@@ -307,11 +336,9 @@ async def describe_collect(self) -> Dict[str, Descriptor]:
     async def collect_asset_docs(
         self, index: Optional[int] = None
     ) -> AsyncIterator[StreamAsset]:
-        """Collect stream datum documents for all indices written.
-
-        The index is optional, and provided for flyscans, however this needs to be
-        retrieved for stepscans.
-        """
+        # Collect stream datum documents for all indices written.
```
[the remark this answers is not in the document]
callumforrester: The protocol provides an identical docstring, so its not worth repeating. Usually not worth bothering with docstrings for override methods because it's just pointless duplication. Your IDE will work it out.

PR #164, opened 2024-04-03
Include Type Stubs for PySide6

callumforrester:
The new release of PySide6 (6.6.3.1) seems to have confused mypy and
made it think the library has bad intendation, despite the fact it
imports. Installing the type stubs package appears to fix the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dev = [
     "pydata-sphinx-theme>=0.12",
     "pyepics>=3.4.2",
     "pyside6",
+    "pyside6-stubs",
     "pytest",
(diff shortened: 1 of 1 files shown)


evalott100 (approved):

stan-dot:
`    "pyside6!=6.6.3.1,!=6.6.3",`
just excluding the version from `pyproject.toml` would work fine too
https://github.com/bluesky/ophyd-async/pull/144/commits/f1113fa7ae5d621f4a6b3b32554edc334c12fb6c

callumforrester:
@stan-dot I can imagine this will be needed for all future versions too

PR #171, opened 2024-04-04
Drop python 3.9 support

dperl-dls:
According to [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) we can now drop python 3.9 :tada: 
This makes many nice 3.10 features available, such as `ParamSpec` and getting rid of `Optional` and `Union` for `|`.


evalott100:
> @evalott100 What's with these tests which never run? Are they expected by some repo settings?

They're for the different situations - i.e on a "release" the release ones will be ran...

It's strange you have two pending tests not categorised like the others, not sure why this would be happening

evalott100 (changes requested):
Somehow this has introduced these:

image

dperl-dls:
I think they are required by some repo settings not in the code - as you can see with the latest commit, running the 3.10 tests with editable install satisfied the expectation for that - while the 3.9 one is not satisfied because it is no longer being run, because it is removed from the actions. I don't have permissions to update the branch protection rules here, some admin needs to do that

coretl:
done

coretl (approved):

Issue #168, opened 2024-04-03
Write the planstub which creates the `SeqTable`, `TriggerInfo` and prepares the flyer and detectors with them.


abbiemery:
Read #165  for the context of this Issue first.

Currently the HardwareTriggeredFlyable acts predominantly as a wrapper for a `TriggerLogic`. The only such logic we have is `PandARepeatedTriggerLogic` which is used  in I22 and is stored [here](https://github.com/DiamondLightSource/i22-bluesky/blob/main/src/i22_bluesky/panda/fly_scanning.py)

The `RepeatedTrigger` is currently used in `PandARepeatedTriggerLogic` to created the `SeqTable` in prepare, and to populate the trigger_info.

Instead of doing this inside the TriggerLogic we instead want to pass the things that make up the `RepeatedTrigger` into a planstub as arguments. 
```
@dataclass
class RepeatedTrigger:
    num: int
    width: float
    deadtime: float
    repeats: int = 1
    period: float = 0.0
```
So we want the arguments num, width etc. We can then use these variables to make the `SeqTable` and the `TriggerInfo` in the planstub. Then we want to prepare the TriggerLogic with the `SeqTable` and prepare the detectors with `TriggerInfo`.

Many of these planstubs will need to be made as they each provide very specific instruction, but the first one will be a simple one with a single flyer and a single detector which it will need to take as arguments along with the old `RepeatedTrigger` values.

The new trigger logic for this will be `StaticSeqTableTriggerLogic` and the issue for writing that is [here](https://github.com/bluesky/ophyd-async/issues/167).

This planstub should look something like this:

```
 def prepare_static_seq_table_flyer_and_detector( 
          flyer: HardwareTriggeredFlyable[SeqTable], 
          detector : StandardDetector, 
          num: int,
          width: float,
          deadtime: float,
          repeats: int  = 1, 
          period: float = 0.0
         ):
         
         trigger_info = TriggerInfo( 
                          num = num * repeats, 
                          trigger = DetectorTrigger.constant_gate,
                          deadtime = deadline, 
                          livetime = livetime
                         )

         trigger_time = ...
         pre_delay = ...
         table: SeqTable = seq_table_from_rows(
                          SeqTableRow(...),
                          SeqTableRow(...),
                        )

         yield from bps.prepare(detector, trigger_info, wait=false, group = "thing")
         yield from bps.prepare(flyer, table, wait=false, group = "thing)
         yield from bps.wait(group = "thing"

Issue #167, opened 2024-04-03
Write the `StaticSeqTableTriggerLogic`


abbiemery:
Read #165  for the context of this Issue first.

So far the only place we have been using this is in I22, and we have been using a specific `PandARepeatedTriggerLogic` which is stored in i22-bluesky in the PandA module. This can be seen [here](https://github.com/DiamondLightSource/i22-bluesky/blob/main/src/i22_bluesky/panda/fly_scanning.py).

This TriggerLogic is instantiated when making the flyer in our flyscanning plans like this:

```
flyer = HardwareTriggeredFlyable(
    ...,
    PandARepeatedTriggerLogic(panda.seq[1], shutter_time=0.004),
    ...,
)
```
The TriggerLogic is instantiated with a PandA sequence block, which has a blank sequence table. 

```
class PandARepeatedTriggerLogic(TriggerLogic[RepeatedTrigger]):
    trigger = DetectorTrigger.constant_gate

    def __init__(self, seq: SeqBlock, shutter_time: float = 0) -> None:
        self.seq = seq
        self.shutter_time = shutter_time

    ...

```
When we call `prepare` on the TriggerLogic it creates a SeqTable and sets that Table into the block

```
async def prepare(self, value: RepeatedTrigger):
        ...
        table = seq_table_from_rows(
            SeqTableRow(
                time1=in_micros(pre_delay),
                time2=in_micros(self.shutter_time),
                outa2=True,
            ),
            ...

        await asyncio.gather(
            self.seq.table.set(table),
        )

```
The actual code can be seen [here](https://github.com/DiamondLightSource/i22-bluesky/blob/main/src/i22_bluesky/panda/fly_scanning.py#L33-L63)

What we would like instead
------------------------------------
We would like a new TriggerLogic called `StaticSeqTableTriggerLogic`. For now make this in opyhd-async but it may be moved in the future. This will still be initialised with a blank block but will no longer need a `trigger_info` method and prepare no longer takes a `RepeatedTrigger` it instead takes a `SeqTable` which is set into the block. A guide to the structure is below:

```
class StaticSeqTableTriggerLogic(TriggerLogic[SeqTable]):
    trigger = DetectorTrigger.constant_gate
    repeats: int = 1

    def __init__(self, seq: SeqBlock, shutter_time: float = 0) -> None:
        self.seq = seq
        ...

    async def prepare(self, value: SeqTable):
        
        ...

        await asyncio.gather(
            ...
[...]

Issue #166, opened 2024-04-03
Remove trigger info from the TriggerLogic and the HardwareTriggeredFlyer


abbiemery:
Read #165 for the context of this Issue first.

For this issue we need to remove all cases of TriggerInfo from the TriggerLogic, and from the HardwareTriggeredFlyer. The trigger_info property needs removing from the flyer and the trigger_info method needs removing from the TriggerLogic ABC.

The trigger logic should look like this:
https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/flyer.py#L15-L30

```
class TriggerLogic(ABC, Generic[T]):
    @abstractmethod
    async def prepare(self, value: T):
        """Move to the start of the flyscan"""

    @abstractmethod
    async def start(self):
        """Start the flyscan"""

    @abstractmethod
    async def stop(self):
        """Stop flying and wait everything to be stopped"""
```
[Link to the code](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/flyer.py#L15-L30)

In the HardwareTriggeredFlyable it needs removing from the [init](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/flyer.py#L50) and removing as a [property](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/flyer.py#L57-L59) and finally removing from [prepare](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/flyer.py#L74)

PR #175, opened 2024-04-05
Move trigger logic into dedicated plan stub

### LINKED ISSUE #165: Change the preparation of the HardwareTriggeredFlyer and the Detectors
Currently to prepare the detectors we have to prepare the flyer first. The flyer is composed with `TriggerLogic` and this `TriggerLogic` provides the `TriggerInfo` needed for the detectors.

The `HardwareTriggeredFlyer` currently is only usable in a very specific use case where it is only making one set of triggers that are consumed by all the detectors. The `TriggerLogic` is responsible for making SequenceTable and the `TriggerInfo` that describes the Triggers that will come out of that table.

We now instead want a plan stub to create both the sequence table, `SeqTable` and the `TriggerInfo`. It will prepare the `TriggerLogic` with the `SeqTable`. The `TriggerLogic` no longer wants responsibility over `TriggerInfo`. This `TriggerInfo` wants to then be used to prepare the detectors.

Each of these plans will be specific for the Type of `TriggerLogic`, the combination of detectors and the things it wants to do.

In order to do this we need to:

To remove all cases of `TriggerInfo` from the `TriggerLogic`, and from the `HardwareTriggeredFlyer`. The `trigger_info` property needs removing from the flyer and the `trigger_info` method needs removing from the `TriggerLogic` ABC.

At the moment the `HardwareTriggeredFlyable` is generic on T which is the same as the T the `TriggerLogic` is generic on. We want to retain this behaviour. By retaining this behaviour we can make a new TriggerLogic `StaticSeqTableTriggerLogic(TriggerLogic[SeqTable])` which will be used by a `HardwareTriggeredFlyable[SeqTable]` .
We therefore need to write a `SaticSeqTableTriggerLogic(TriggerLogic[SeqTable])`, taking the `PandARepeatedTriggerLogic` in I22 as an example with two main changes:
1. The SequenceTable is no longer created in prepare and instead that prepare needs to take a sequence table as an argument as it will be passed in from the plan.
2. It will not have a  trigger_info method as we removed this from the ABC.
We can argue later where TriggerLogic's should live.

Make a specific planstub which takes our new `SaticSeqTableTriggerLogic(TriggerLogic[SeqTable])` , a `HardwareTriggeredFlyable[SeqTable]` and a single detector. The planstub needs to create a `SeqTable` and a `TriggerInfo`, prepares the flyer with the sequence table and prepares the detector with the `TriggerInfo`, then waits for them all to be prepared.

To do
-------

- [ ] #166 
- [ ] #167
- [ ] #168

abbiemery:
Fixes #165 #166 #167 #168


abbiemery:
I have a few issues i'd appreciate opinions on:

1. I don't think the planstubs directory should be in ophyd-async, but i'm open to where we should put it.
2. I have the same thoughts on TriggerLogics, but these are not as strongly felt.
3. I don't really like the requirement to have a the typed structure `value` as the only argument for prepare in the flyer. I understand it ends up with us making data structures to group things together which are not very general

src/ophyd_async/core/flyer.py
```diff
@@ -71,17 +65,19 @@ def prepare(self, value: T) -> AsyncStatus:
...
     async def _prepare(self, value: T) -> None:
-        self._trigger_info = self._trigger_logic.trigger_info(value)
         # Move to start and setup the flyscan
         await self._trigger_logic.prepare(value)
 
     @AsyncStatus.wrap
     async def kickoff(self) -> None:
-        self._fly_status = AsyncStatus(self._trigger_logic.start())
+        self._fly_status = AsyncStatus(self._trigger_logic.kickoff())
+        await self._fly_status
```
coretl, line 74:
```suggestion
        await self._trigger_logic.kickoff()
```

src/ophyd_async/core/flyer.py
```diff
@@ -71,17 +65,19 @@ def prepare(self, value: T) -> AsyncStatus:
...
 
-    def complete(self) -> AsyncStatus:
+    @AsyncStatus.wrap
+    async def complete(self) -> None:
         assert self._fly_status, "Kickoff not run"
-        return self._fly_status
+        await self._fly_status
```
coretl, line 79: I don't think we should bother checking that kickoff hasn't been done, at least not at the moment, it requires too much holding of state for too little benefit...
```suggestion
```

src/ophyd_async/triggers/static_seq_table_trigger.py
coretl, line 1: I think this belongs in the `panda` module

tests/panda/db/panda.db
coretl, line 1: I think after @evalott100's pvi PR is merged then we can probably get rid of this database and the corresponding test, and replace it with individual tests in the pvi functions

tests/test_flyer_with_panda.py
```diff
@@ -0,0 +1,221 @@
...
+            except TimeoutError:
+                pass
+            else:
+                done = True
+            yield from bps.collect(
+                *detector_list,
+                return_payload=False,
+                name="main_stream",
+            )
+            yield from bps.sleep(0.01)
```
coretl, line 202: What is this sleep needed for?

tests/test_flyer_with_panda.py
```diff
@@ -0,0 +1,221 @@
...
+        yield from bps.wait(group="complete")
```
coretl, line 203: I think this section should be turned into a plan stub which I think there is a ticket for already, maybe in ophyd-async to start then move into bluesky later

coretl:
>     1. I don't think the planstubs directory should be in ophyd-async, but i'm open to where we should put it.

I think ophyd-async is where we are doing flyscanning, so if there are plan stubs that do flyscanning they should live here. As all flyscanning is done with PandAs then I'd be happy to put it in the `panda` module?

>     2. I have the same thoughts on TriggerLogics, but these are not as strongly felt.

Same, should be in `panda` somewhere

>     3. I don't really like the requirement to have a the typed structure `value` as the only argument for prepare in the flyer. I understand it ends up with us making data structures to group things together which are not very general

This is because `ParamSpec` did not exist when we designed the bluesky protocols, so you could not make something `Generic` on a set of arguments. If this now works then we can reconsider this for `Preparable`...

src/ophyd_async/core/flyer.py
abbiemery: Yes I agree

tests/panda/db/panda.db
abbiemery: I agree, but I think we should put that in a different PR.

tests/test_flyer_with_panda.py
abbiemery: I think it is lingering from an old issue that has since been sorted. I just copied this plan format from the old stuff Rose did.

abbiemery: I've removed it and the one in the flyer.

tests/test_flyer_with_panda.py
abbiemery: I would prefer to leave it as it is now, and make a separate ticket. Preferably for when I do the fixing of all the plans that use our flyscanning to use the new structure which I will do before making the release but once this sprint has ended.

abbiemery:
> I think ophyd-async is where we are doing flyscanning, so if there are plan stubs that do flyscanning they should live here. As all flyscanning is done with PandAs then I'd be happy to put it in the `panda` module?

In which case fine happy for it to stay. We can always move it later.


> Same, should be in `panda` somewhere

I've moved those into panda.


> This is because `ParamSpec` did not exist when we designed the bluesky protocols, so you could not make something `Generic` on a set of arguments. If this now works then we can reconsider this for `Preparable`...

This is a bigger Issue, but one I'd potentially like to consider in the future.

src/ophyd_async/planstubs/prepare_trigger_and_dets.py
```diff
@@ -0,0 +1,58 @@
...
+
+from ophyd_async.core.detector import DetectorTrigger, StandardDetector, TriggerInfo
+from ophyd_async.core.flyer import HardwareTriggeredFlyable
+from ophyd_async.core.utils import in_micros
+from ophyd_async.panda.table import SeqTable, SeqTableRow, seq_table_from_rows
+from ophyd_async.panda.trigger import SequenceTableInfo
+
+
+def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
+    flyer: HardwareTriggeredFlyable[SeqTable],
```
coretl, line 13: This type hint seems wrong, shouldn't it be
```suggestion
    flyer: HardwareTriggeredFlyable[SequenceTableInfo],
```
Can you find out why type checking didn't catch this?

src/ophyd_async/core/flyer.py
```diff
@@ -47,17 +46,12 @@ def __init__(
         self._configuration_signals = tuple(configuration_signals)
         self._describe: Dict[str, Descriptor] = {}
```
coretl, line 47: As can `_describe`

src/ophyd_async/planstubs/prepare_trigger_and_dets.py
abbiemery: Yes it should, also can you think of a better name for `SequenceTableInfo`?

abbiemery: Also It didn't catch it because it's not technically wrong here, `HardwareTriggeredFlyable[T]` is generic on T, so its got no reason to know it shouldn't be that.

abbiemery: I've corrected it and for now changed the name to `SeqTableInfo`.

src/ophyd_async/core/flyer.py
```diff
@@ -47,17 +46,12 @@ def __init__(
...
         self._fly_status: Optional[AsyncStatus] = None
```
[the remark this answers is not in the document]
abbiemery: Done.

src/ophyd_async/core/flyer.py
abbiemery: Done.

src/ophyd_async/planstubs/prepare_trigger_and_dets.py
coretl: > Yes it should, also can you think of a better name for `SequenceTableInfo`?

`SeqTableAndFriends`? Naming is hard... Maybe `RepeatedSeqTable` is ok?

coretl (approved):

src/ophyd_async/planstubs/prepare_trigger_and_dets.py
abbiemery: I can make up a test script tomorrow and check it out

PR #176, opened 2024-04-08
Add a Watchable AsyncStatus and extend the wrap decorator

### LINKED ISSUE #45: limited functionality of AsyncStatus wrap

### LINKED ISSUE #117: Add support for watchers into AsyncStatus.wrap

dperl-dls:
Fixes #117, fixes #45
As according to discussion in https://github.com/bluesky/ophyd-async/issues/117, adds a `WatchableAsyncStatus`, and improves the wrappers so that they can be applied to any function returning an `Awaitable` or an `AsyncIterator`

Will break anything previously passing watchers to `AsyncStatus` which will need to update to `WatchableAsyncStatus` (but there hopefully isn't much out there yet)


src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,59 @@ def __repr__(self) -> str:
...
+        )
+        super().__init__(awaitable)
+
+    async def _notify_watchers_from(self, iterator: AsyncIterator[WatcherUpdate[T]]):
+        async for self._last_update in iterator:
+            for watcher in self._watchers:
+                self._update_watcher(watcher, self._last_update)
+
+    def _update_watcher(self, watcher: Watcher, update: WatcherUpdate[T]):
+        watcher(replace(update, time_elapsed_s=time.monotonic() - self._start))
```
coretl, line 127: Given https://blueskyproject.io/bluesky/hardware.html#watch we need parameters to be passed down individually, so it would be nice if we could make the dataclass args the same as the signature so we could do:
```suggestion
        watcher(**update.dict())
```

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,59 @@ def __repr__(self) -> str:
...
+        self._last_update: WatcherUpdate[T] | None = None
+        awaitable = (
+            iterator_or_awaitable
+            if isinstance(iterator_or_awaitable, Awaitable)
+            else self._notify_watchers_from(iterator_or_awaitable)
+        )
+        super().__init__(awaitable)
+
+    async def _notify_watchers_from(self, iterator: AsyncIterator[WatcherUpdate[T]]):
+        async for self._last_update in iterator:
```
coretl, line 122: Nit: I would also probably apply time elapsed here so late watchers get exactly the same update as existing watchers
```suggestion
        async for update in iterator:
            self._last_update = replace(update, time_elapsed=time.monotonic() - self._start)
```

coretl: Or maybe only if is not None?

coretl (changes requested):

src/ophyd_async/core/utils.py
```diff
@@ -77,6 +83,40 @@ def __str__(self) -> str:
...
+
+    name: str
+    current: T
+    initial: T
+    target: T
+    units: str
+    precision: float
+    fraction: float
+    time_elapsed: float
+    time_remaining: float
```
coretl, line 98: Looking at https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/utils/__init__.py#L1301-L1343 then we should probably make these optional, then we don't have to calculate them every time
```suggestion
    fraction: Optional[float] = None
    time_elapsed: Optional[float] = None
    time_remaining: Optional[float] = None
```

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +77,56 @@ def __repr__(self) -> str:
...
+        self._watchers: list[Watcher] = watchers
+        self._start = time.monotonic()
+        self._last_update: WatcherUpdate[T] | None = None
+        super().__init__(self._notify_watchers_from(iterator))
+
+    async def _notify_watchers_from(self, iterator: AsyncIterator[WatcherUpdate[T]]):
+        async for update in iterator:
+            self._last_update = replace(
+                update, time_elapsed=time.monotonic() - self._start
+            )
```
coretl, line 111: We could probably also update fraction and time_remaining here, but given they are not used in https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/utils/__init__.py#L1301-L1343 then I'd say it isn't worth it.

Maybe we only update if time_elapsed is not None though?

```suggestion
            self._last_update = update if update.time_elapsed is not None else replace(
                update, time_elapsed=time.monotonic() - self._start
            )
```

src/ophyd_async/core/utils.py
dperl-dls: according to https://blueskyproject.io/bluesky/hardware.html#watch it kind of sounds like all/many of them should be optional

coretl: This is true, but I suspect the minimum needed for a progress bar is `current`, `initial` and `target` as per https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/utils/__init__.py#L1315C40-L1315C64 so if people have opted into a `WatchableAsyncStatus` then they should at least provide those (otherwise they should just use `AsyncStatus`). The rest should probably be optional as you say.

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,75 @@ def __repr__(self) -> str:
...
+        self._last_update: WatcherUpdate[T] | None = None
+        super().__init__(self._notify_watchers_from(iterator))
+
+    async def _notify_watchers_from(self, iterator: AsyncIterator[WatcherUpdate[T]]):
+        async for update in iterator:
+            if self._timeout and time.monotonic() > self._timeout:
+                raise TimeoutError()
+            self._last_update = (
+                update
+                if update.time_elapsed is None
```
coretl, line 123:
```suggestion
                if update.time_elapsed is not None
```

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,75 @@ def __repr__(self) -> str:
...
+        self._watchers: list[Watcher] = []
+        self._start = time.monotonic()
+        self._timeout = self._start + timeout_s if timeout_s else None
+        self._last_update: WatcherUpdate[T] | None = None
+        super().__init__(self._notify_watchers_from(iterator))
+
+    async def _notify_watchers_from(self, iterator: AsyncIterator[WatcherUpdate[T]]):
+        async for update in iterator:
+            if self._timeout and time.monotonic() > self._timeout:
+                raise TimeoutError()
```
coretl, line 120: Rather than doing the timeout here...

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,75 @@ def __repr__(self) -> str:
...
+    def __init__(
+        self,
+        iterator: AsyncIterator[WatcherUpdate[T]],
+        timeout_s: float = 0.0,
+    ):
+        self._watchers: list[Watcher] = []
+        self._start = time.monotonic()
+        self._timeout = self._start + timeout_s if timeout_s else None
+        self._last_update: WatcherUpdate[T] | None = None
+        super().__init__(self._notify_watchers_from(iterator))
```
coretl, line 115: Could we do it here?
```suggestion
        super().__init__(asyncio.wait_for(self._notify_watchers_from(iterator), timeout_s)
```

src/ophyd_async/core/async_status.py
```diff
@@ -94,3 +86,75 @@ def __repr__(self) -> str:
...
+        cls: Type[WAS],
+        f: Callable[P, AsyncIterator[WatcherUpdate[T]]],
+        timeout_s: float = 0.0,
+    ) -> Callable[P, WAS]:
+        """Wrap an AsyncIterator in a WatchableAsyncStatus. If it takes
+        'timeout_s' as an argument, this must be a float and it will be propagated
+        to the status."""
+
+        @functools.wraps(f)
+        def wrap_f(*args: P.args, **kwargs: P.kwargs) -> WAS:
```
coretl, line 151: Does it work if we
```suggestion
        def wrap_f(*args: P.args, timeout: Optional[float] = None, **kwargs: P.kwargs) -> WAS:
```

coretl (commented):
I have a general preference for `timeout` rather than `timeout_s` as it matches all the asyncio arguments, and also the ophyd convention

src/ophyd_async/core/async_status.py
dperl-dls: I couldn't get this to work - besides `ParamSpec` you can only have positional arguments at the beginning, which kind of defeats the purpose of having the paramspec at all if we don't want to enforce timeout being the first argument for all wrappable methods

dperl-dls: Screenshot from 2024-04-18 16-29-50

coretl: Having read [the spec](https://peps.python.org/pep-0612/#concatenating-keyword-parameters) I now understand the reasoning, so this makes sense

src/ophyd_async/core/utils.py
```diff
@@ -77,6 +84,59 @@ def __str__(self) -> str:
...
+        name: str | None,
+        unit: str | None,
+        precision: float | None,
+        fraction: float | None,
+        time_elapsed: float | None,
+        time_remaining: float | None,
+    ) -> Any: ...
+
+
+Watcher: TypeAlias = _ClsWatcher | _InsWatcher
```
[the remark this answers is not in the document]
dperl-dls: Because I had problems with calling instance methods Watchers in the tests, but they seem to have gone away so I removed it for now

src/ophyd_async/epics/demo/__init__.py
```diff
@@ -88,32 +89,29 @@ def set_name(self, name: str):
...
-                    name=self.name,
-                    current=current_position,
-                    initial=old_position,
-                    target=new_position,
-                    unit=units,
-                    precision=precision,
-                    time_elapsed=time.monotonic() - start,
-                )
-            if np.isclose(current_position, new_position):
-                break
```
coretl, line 114: Actually, this change with using `move_status` only applies to the epics motor record, this demo motor is actually backed by database logic, so the existing logic of `np.isclose` inside an `observe_value` should apply here rather than waiting for `move_status`

src/ophyd_async/epics/motion/motor.py
```diff
@@ -41,34 +45,28 @@ def set_name(self, name: str):
...
+                initial=old_position,
+                current=old_position,
+                target=new_position,
+                unit=units,
+                precision=precision,
+            ),
+            move_status,
+        )
 
     def move(self, new_position: float, timeout: Optional[float] = None):
```
coretl, line 71: Ditto for removing `move`

src/ophyd_async/epics/motion/motor.py
```diff
@@ -78,14 +76,26 @@ def move(self, new_position: float, timeout: Optional[float] = None):
...
+        async for current_position in observe_value(
+            self.user_readback, done_status=move_status
+        ):
+            if not self._set_success:
+                raise RuntimeError("Motor was stopped")
+            yield replace(
+                update,
+                name=self.name,
+                current=current_position,
+                time_elapsed=time.monotonic() - start,
```
coretl, line 92: And time elapsed

src/ophyd_async/epics/motion/motor.py
```diff
@@ -78,14 +76,26 @@ def move(self, new_position: float, timeout: Optional[float] = None):
             raise RuntimeError("Will deadlock run engine if run in a plan")
         call_in_bluesky_event_loop(self._move(new_position), timeout)  # type: ignore
 
-    def set(self, new_position: float, timeout: Optional[float] = None) -> AsyncStatus:
-        watchers: List[Callable] = []
-        coro = asyncio.wait_for(self._move(new_position, watchers), timeout=timeout)
-        return AsyncStatus(coro, watchers)
+    @WatchableAsyncStatus.wrap
+    async def set(self, new_position: float, timeout: float = 0.0):
+        update, move_status = await self._move(new_position)
```
coretl, line 81: and inlining `_move`

src/ophyd_async/epics/motion/motor.py
```diff
@@ -78,14 +76,26 @@ def move(self, new_position: float, timeout: Optional[float] = None):
...
+        start = time.monotonic()
+        async for current_position in observe_value(
+            self.user_readback, done_status=move_status
```
coretl, line 84: But here the `done_status=move_status` is correct as we are talking to a motor record

src/ophyd_async/sim/demo/sim_motor.py
```diff
@@ -54,7 +54,7 @@ def set(self, new_position: float, timeout: Optional[float] = None) -> AsyncStat
         """
         watchers: List[Callable] = []
         coro = asyncio.wait_for(self._move(new_position, watchers), timeout=timeout)
-        return AsyncStatus(coro, watchers)
+        return AsyncStatus(coro)
 
     async def _move(self, new_position: float, watchers: List[Callable] = []):
```
coretl, line 59: Please can we use the same `wrap` logic here? Something like

```suggestion
    @WatchableAsyncStatus.wrap
    async def set(self, new_position: float) -> :
```

then yielding `WatcherUpdate` instead of calling watchers

src/ophyd_async/core/async_status.py
```diff
@@ -2,28 +2,41 @@
...
+    def __init__(self, awaitable: Awaitable, timeout: SupportsFloat | None = None):
+        if isinstance(timeout, SupportsFloat):
+            timeout = float(timeout)
         if isinstance(awaitable, asyncio.Task):
             self.task = awaitable
         else:
-            self.task = asyncio.create_task(awaitable)  # type: ignore
+            self.task = asyncio.create_task(
+                asyncio.wait_for(awaitable, timeout=timeout)
+            )
```
coretl, line 37: ... there is race condition between creating the `wait_for` coro and scheduling it as a task. Does this fix it?
```suggestion
            async def inner():
                    async with asyncio.timeout(timeout):
                            await awaitable
            self.task = asyncio.create_task(inner())
```

tests/core/test_watchable_async_status.py
```diff
@@ -0,0 +1,200 @@
...
+        time_elapsed: float | None,
+        time_remaining: float | None,
+    ) -> None:
+        self.updates.append(current or -1)
+
+
+class ASTestDevice(StandardReadable, Movable):
+    def __init__(self, name: str = "") -> None:
+        self._staged: bool = False
+        self.sig = SignalR(backend=SimSignalBackend(datatype=int))
```
coretl, line 69: Could be `soft_signal_r_and_backend`

coretl (commented):
Just spotted this...

src/ophyd_async/epics/demo/__init__.py
```diff
@@ -88,32 +89,29 @@ def set_name(self, name: str):
...
+                initial=old_position,
+                current=old_position,
+                target=new_position,
+                unit=units,
+                precision=precision,
+            ),
+            move_status,
+        )
 
     def move(self, new_position: float, timeout: Optional[float] = None):
```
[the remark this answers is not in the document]
dperl-dls: The one from `motor` as well?

tests/core/test_async_status.py
```diff
@@ -50,13 +50,17 @@ async def test_async_status_has_no_exception_if_coroutine_successful(normal_coro
...
+    status = AsyncStatus(coro)
     assert status.exception() is None
     status.task.cancel()
     with pytest.raises(asyncio.CancelledError):
         await status
     assert status.success is False
     assert isinstance(status.exception(), asyncio.CancelledError)
+    # asyncio will RuntimeWarning us about this never being awaited if we don't.
+    # RunEngine handled this as a special case
+    await coro
```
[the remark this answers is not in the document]
dperl-dls: Unfortunately not really applicable in this case, see: https://github.com/bluesky/ophyd-async/tree/try_with_asyncio_timeout which has the same problem but is much more complicated

src/ophyd_async/core/async_status.py
dperl-dls: I don't know - it's hard to see how this actually differs from the above... does whether you call `inner` or `wait_for` to create the coro change anything?

src/ophyd_async/core/detector.py
```diff
@@ -299,28 +299,26 @@ async def _prepare(self, value: T) -> None:
             exposure=self._trigger_info.livetime,
         )
 
-    @AsyncStatus.wrap
-    async def kickoff(self) -> None:
-        self._fly_status = AsyncStatus(self._fly(), self._watchers)
+    def kickoff(self, timeout: Optional[float] = None):
```
coretl, line 302: The original code kicked off `self._observe_writer_indices` and returned immediately, then returned this status in `complete`. This code now actually awaits this Status, which is wrong.

However, I think a simplification of the code is the better fix:

```python
    @AsyncStatus.wrap
    async def kickoff(self) -> None:
        # Nothing to do in kickoff, we already armed in prepare
        pass

    @WatchableAsyncStatus.wrap
    async def complete(self):
        async for index in self.writer.observe_indices_written(self._frame_writing_timeout):    
            yield WatcherUpdate(
                name=self.name,
                current=index,
                initial=self._initial_frame,
                target=end_observation,
                unit="",
                precision=0,
                time_elapsed=time.monotonic() - self._fly_start,
            )
            if index >= end_observation:
                break
```
Then can delete `_fly` and `_observe_writer_indices`...

src/ophyd_async/epics/demo/__init__.py
```diff
@@ -85,46 +85,41 @@ def set_name(self, name: str):
...
-            raise RuntimeError("Will deadlock run engine if run in a plan")
-        call_in_bluesky_event_loop(self._move(new_position), timeout)  # type: ignore
-
-    # TODO: this fails if we call from the cli, but works if we "ipython await" it
-    def set(self, new_position: float, timeout: Optional[float] = None) -> AsyncStatus:
-        watchers: List[Callable] = []
-        coro = asyncio.wait_for(self._move(new_position, watchers), timeout=timeout)
-        return AsyncStatus(coro, watchers)
+    @WatchableAsyncStatus.wrap  # uses the timeout argument from the function it wraps
+    async def set(self, new_position: float, timeout: float | None = None):
```
coretl, line 113: Can't see timeout being used below...
```suggestion
    @WatchableAsyncStatus.wrap
    async def set(self, new_position: float):
```

tests/core/test_device_collector.py
```diff
@@ -85,6 +86,7 @@ def test_async_device_connector_run_engine_same_event_loop():
     async def set_up_device():
         async with DeviceCollector(sim=True):
             sim_motor = motor.Motor("BLxxI-MO-TABLE-01:X")
+            sim_motor.set
```
coretl, line 89: What is this?
```suggestion
```

tests/core/test_device_collector.py
```diff
@@ -95,6 +97,7 @@ async def set_up_device():
         RE = RunEngine(call_returns_result=True, loop=loop)
 
         def my_plan():
+            sim_motor.motor_done_move._backend._set_value(True)  # type: ignore
```
coretl, line 100:
```suggestion
            set_sim_value(sim_motor.motor_done_move, True)
```

tests/epics/demo/test_demo.py
```diff
@@ -76,8 +100,9 @@ async def wait_for_call(self, *args, **kwargs):
 
 
 async def test_mover_moving_well(sim_mover: demo.Mover) -> None:
+    set_sim_put_proceeds(sim_mover.setpoint, False)
```
coretl, line 103: These shouldn't be needed
```suggestion
```

tests/epics/demo/test_demo.py
```diff
@@ -106,6 +131,7 @@ async def test_mover_moving_well(sim_mover: demo.Mover) -> None:
         time_elapsed=pytest.approx(0.1, abs=0.05),
     )
     set_sim_value(sim_mover.readback, 0.5499999)
+    set_sim_put_proceeds(sim_mover.setpoint, True)
```
coretl, line 134:
```suggestion
```

src/ophyd_async/epics/motion/motor.py
```diff
@@ -40,51 +47,47 @@ def set_name(self, name: str):
...
 
     async def stop(self, success=False):
         self._set_success = success
         # Put with completion will never complete as we are waiting for completion on
         # the move above, so need to pass wait=False
-        status = self.motor_stop.trigger(wait=False)
-        await status
+        await self.motor_stop.trigger(wait=False)
+        # Trigger any callbacks
+        await self.user_readback._backend.put(await self.user_readback.get_value())
```
coretl, line 93: This shouldn't be needed now we added the `done_status` argument above?
```suggestion
```

tests/epics/motion/test_motor.py
```diff
@@ -18,23 +21,38 @@ async def sim_motor():
...
 
+async def wait_for_eq(item, attribute, comparison, timeout):
+    timeout_time = time.monotonic() + timeout
+    while getattr(item, attribute) != comparison:
+        await asyncio.sleep(A_BIT)
+        if time.monotonic() > timeout_time:
+            raise TimeoutError
+
+
 async def test_motor_moving_well(sim_motor: motor.Motor) -> None:
```
coretl, line 47: Please could you revert this test and check it still passes?

coretl (changes requested):

src/ophyd_async/epics/demo/__init__.py
dperl-dls: No, the `WatchableAsyncStatus` wrapper uses the timeout argument from functions that it wraps, as we discussed in detail earlier in this PR

src/ophyd_async/core/detector.py
dperl-dls: `Flyable` tells us that `kickoff` and `complete` must both return statuses.

coretl: That's fine, they're both being wrapped by `*AsyncStatus` so return a `Status`. It's just that `kickoff` returns one that is immediately done...

src/ophyd_async/epics/demo/__init__.py
coretl: Yes, but doesn't it add that to the signature itself? So the function doesn't need to mention timeout at all...

dperl-dls: It pulls it from `kwargs`, so the function still needs to either take `timeout` or `**kwargs`

tests/epics/motion/test_motor.py
dperl-dls: It doesn't the way it is in main - it needs a sleep after the `set_mock_value(readback ...)` to process. This slight alteration does pass:



```python
async def test_motor_moving_well_2(sim_motor: motor.Motor) -> None:
    set_mock_put_proceeds(sim_motor.user_setpoint, False)
    s = sim_motor.set(0.55)
    watcher = Mock()
    s.watch(watcher)
    done = Mock()
    s.add_callback(done)
    await asyncio.sleep(A_BIT)
    assert watcher.call_count == 1
    assert watcher.call_args == call(
        name="sim_motor",
        current=0.0,
        initial=0.0,
        target=0.55,
        unit="mm",
        precision=3,
        time_elapsed=pytest.approx(0.0, abs=0.05),
    )
    watcher.reset_mock()
    assert 0.55 == await sim_motor.user_setpoint.get_value()
    assert not s.done
    await asyncio.sleep(0.1)
    set_mock_value(sim_motor.user_readback, 0.1)
    await asyncio.sleep(0.1)
    assert watcher.call_count == 1
    assert watcher.call_args == call(
        name="sim_motor",
        current=0.1,
        initial=0.0,
        target=0.55,
        unit="mm",
        precision=3,
        time_elapsed=pytest.approx(0.11, abs=0.05),
    )
    set_mock_put_proceeds(sim_motor.user_setpoint, True)
    await asyncio.sleep(A_BIT)
    assert s.done
    done.assert_called_once_with(s)
```

coretl (approved):

Issue #177, opened 2024-04-08
Document and expose functionality from AD plugins


DiamondJoseph:
To enable the HDFWriter to make use of signals exposed by e.g. the StatsPlugin, the plugins should be documented and their possible fields exposed. This would make using them as ScalarSignals easier

callumforrester:
Duplicate of #213

PR #178, opened 2024-04-08
Ad plugin stats

DiamondJoseph:
Initial implementation of potential solution to #177 

Adds a pair of example enums to the NDStatsPlugin, allows configuring appropriate fields for that plugin.
Allows configuring an ADBase with a dict of {suffix: PluginClass} that will then later enable configurable use of ScalarSigs in the HDFWriter

Submitting mostly as a RFC and first attempt: From the i22 pilatus and AdAravis cameras, the Stats plugin was hardcoded, without enabling the Statistics calculation, or allowing other fields. Ideally, this behaviour would be configurable from here?

### DIFF
--- a/src/ophyd_async/epics/areadetector/drivers/ad_base.py
+++ b/src/ophyd_async/epics/areadetector/drivers/ad_base.py
@@ -1,6 +1,6 @@
 import asyncio
 from enum import Enum
-from typing import FrozenSet, Sequence, Set
+from typing import Dict, FrozenSet, Optional, Sequence, Set, Type
 
 from ophyd_async.core import (
     DEFAULT_TIMEOUT,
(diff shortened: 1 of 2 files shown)


DiamondJoseph:
I don't know if this has already been discussed for ophyd-async, just something I stumbled across while porting the i22 cameras.

DiamondJoseph:
I see that SingleTriggerDet is already taking a kwargs of Plugins

```
class SingleTriggerDet(StandardReadable, Triggerable):
    def __init__(
        self,
        drv: ADBase,
        read_uncached: Sequence[SignalR] = (),
        name="",
        **plugins: NDPluginBase,
    ) -> None:
        self.drv = drv
        self.__dict__.update(plugins)
    ....
```

coretl:
What we need of this is done in https://github.com/bluesky/ophyd-async/pull/500

PR #183, opened 2024-04-09
Add docstring for DeviceVector and document how to group devices

callumforrester:
Add docstring for DeviceVector and document How to Group Devices
  
Move initial how-to on assembling devices into its own page and add a section about using DeviceVector to make arbitrary-length groups. Add an example to the epics demo with tests.

### DIFF
--- a/src/ophyd_async/epics/demo/__init__.py
+++ b/src/ophyd_async/epics/demo/__init__.py
@@ -154,11 +154,22 @@ def start_ioc_subprocess() -> str:
     pv_prefix = "".join(random.choice(string.ascii_uppercase) for _ in range(12)) + ":"
     here = Path(__file__).absolute().parent
     args = [sys.executable, "-m", "epicscorelibs.ioc"]
+
+    # Create standalone sensor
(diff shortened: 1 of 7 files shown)


src/ophyd_async/epics/demo/__init__.py
```diff
@@ -43,6 +49,19 @@ def __init__(self, prefix: str, name="") -> None:
         super().__init__(name=name)
 
 
+class SensorGroup(StandardReadable):
```
coretl, line 52: Does the ioc subprocess command at the bottom of the file actually contain all these sensors? I'd like to keep this file and the tutorials in step so that you can always run up an IOC to test the example code...

callumforrester: It does now, although I think we should document that process/make it easier. Maybe even add tests to prove that the IOC contains the demo objects. I'll make an issue or two.

callumforrester: Done: https://github.com/bluesky/ophyd-async/issues/193

coretl (approved):

PR #26, opened 2023-10-11
Restructure repository to solidify public interface and document it

coretl:
We are moving towards a 1.0 release, so should solidify our public interface. This PR documents what it should be, and then rearranges the repository accordingly.

PR #185, opened 2024-04-10
Make HDFPandA StandardDetector

### LINKED ISSUE #127: Rename PandA to HDFPandA deriving from CommonPandABlocks and StandardDetector
We want the PandA to be unified under the Standard Detector format. To do this we want to have a HDFPandA which derives from Standard detector and CommonPandABlocks. We want this since PandA can make triggers and use triggers. The CommonPandABlocks provides the interface. The Standard detector provides the behaviour that allows you use these triggers and reads in the larger ecosystem: for example to kickoff, collect (during the time frame from trigger, what have my encoders done).

Depends on #126

evalott100:
Closes #127 

Requires https://github.com/bluesky/ophyd-async/pull/198


src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -80,10 +82,15 @@ def _verify_common_blocks(entry: PVIEntry, common_device: Type[Device]):
...
-            )
+        device_t, is_optional = _strip_union(sub_device)
+        if sub_name not in entry.sub_entries:
+            if is_optional:
+                continue
+            else:
+                raise RuntimeError(
+                    f"sub device `{sub_name}:{type(sub_device)}` "
+                    "was not provided by pvi"
+                )
```
coretl, line 93:
```suggestion
        if sub_name not in entry.sub_entries and not is_optional:
            raise RuntimeError(
                f"sub device `{sub_name}:{type(sub_device)}` "
                "was not provided by pvi"
            )
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -185,7 +192,7 @@ def _sim_common_blocks(device: Device, stripped_type: Optional[Type] = None):
             signal_type = args[0] if (args := get_args(sub_device_t)) else None
             sub_device = sub_device_t(SimSignalBackend(signal_type, sub_name))
         else:
-            sub_device = sub_device_t()
+            sub_device = getattr(device, sub_name, sub_device_t())
```
coretl, line 195: What are the cases when this might or might not exist as an attribute?

src/ophyd_async/panda/hdf_panda.py
```diff
@@ -0,0 +1,49 @@
...
+class HDFPanda(CommonPandaBlocks, StandardDetector):
+    def __init__(
+        self,
+        prefix: str,
+        directory_provider: DirectoryProvider,
+        config_sigs: Sequence[SignalR] = (),
+        name: str = "",
+    ):
+        self._prefix = prefix
+        self.set_name(name)
```
coretl, line 27: This line is done by the super call
```suggestion
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -296,3 +306,30 @@ async def fill_pvi_entries(
     # We call set name now the parent field has been set in all of the
     # introspect-initialized devices. This will recursively set the names.
     device.set_name(device.name)
+
+
+def pre_initialize_blocks(
```
coretl, line 311: blocks is a PandA specific term, please could we change this to something like:
```suggestion
def create_children_from_annotations(
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -296,3 +306,30 @@ async def fill_pvi_entries(
...
+    device: Device, included_optional_fields: Optional[Tuple[str, ...]] = None
```
coretl, line 312: Tuple is not mutable so can simplify
```suggestion
    device: Device, included_optional_fields: Tuple[str, ...] = ()
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -296,3 +306,30 @@ async def fill_pvi_entries(
...
+    """For intializing blocks at __init__ of ``device``."""
+    for name, device_type in get_type_hints(type(device)).items():
+        if name in ("_name", "parent"):
+            continue
+        device_type, is_optional = _strip_union(device_type)
+        if (
+            is_optional
+            and included_optional_fields
+            and name not in included_optional_fields
+        ):
```
coretl, line 323: So we can make this:
```suggestion
        if is_optional and name not in included_optional_fields:
```

src/ophyd_async/panda/hdf_panda.py
```diff
@@ -0,0 +1,49 @@
...
+
+        pre_initialize_blocks(self, included_optional_fields=("data",))
```
coretl, line 29: What is the purpose of `included_optional_fields`?

coretl (changes requested):
Please could you make all of the files in the `panda` module start with an underscore

src/ophyd_async/epics/pvi/pvi.py
evalott100: This just gets the object created in `pre_initialize_blocks`:

If the device was made in `pre_initialize_blocks` in `HdfPanda.__init__`, then we can't create a new object, since those references will have been given to devices that `HdfPanda` has initialized.

src/ophyd_async/panda/hdf_panda.py
evalott100: Leftover from when we made `data` `Optional` in `CommonPandaBlocks`, we don't need this arg anymore.

The idea of the argument is that we won't want to initialize most optional fields at `__init__` time. If the Panda didn't send such a block then you'd have blocks without signals at connect time. This is the argument used to specify that `"data"` is an optional field we do want to `pre_initialize`.

src/ophyd_async/epics/pvi/pvi.py
evalott100: I couldn't do this without adding another `sub_name not in entry.sub_entries and is_optional`, else it will hit one of the other conditions below?

abbiemery:
@evalott100 docs are breaking for some reason

src/ophyd_async/panda/_common_panda.py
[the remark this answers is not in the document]
evalott100: Yea good point, it's already in panda :thinking:

src/ophyd_async/panda/_hdf_panda.py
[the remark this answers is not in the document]
abbiemery: I agree with you, it feels like if this is the only part we want user to access it should be from an non-underscored file, and if this if the only implementation of panda we want then it should be in `.panda` not `.hdf_panda`. That being said, if there are technical reasons for it i don't know about, or if the opinions are strong the other way I am not going to fight it.

src/ophyd_async/epics/pvi/pvi.py
coretl: Makes sense

src/ophyd_async/panda/_hdf_panda.py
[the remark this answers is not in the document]
coretl: I will cover this in #26 so leave it as is until then

tests/panda/test_hdf_panda.py
```diff
@@ -0,0 +1,235 @@
...
+        yield from bps.wait(group="complete")
+        yield from bps.close_run()
+
+        yield from bps.unstage_all(flyer, sim_hdf_panda)
+        # assert sim_hdf_panda.controller.disarm.called  # type: ignore
+
+    # fly scan
+    RE(flying_plan())
+
+    assert names == [
```
coretl, line 161: Please can you use the pattern for https://github.com/bluesky/ophyd-async/issues/206. You can copy the `assert_emitted` function into this file until #199 is done

coretl (commented):
Almost there

coretl (approved):

PR #187, opened 2024-04-10
Pin event model to pre StreamResource Changes

### LINKED ISSUE #170: Pin event model < 1.21.0
To prepare for a possible breaking changes with the new major event model release we should pin to below that.

abbiemery:
Fixes #170


evalott100 (approved):
Nice

PR #190, opened 2024-04-10
Move ADAravis detectors to ophyd-async

DiamondJoseph:
As discussed with @coretl, standard detectors should exist in ophyd-async instead of dodal.

Implementation of ADAravis OAV for i22, p38 at Diamond.

Fixes https://github.com/DiamondLightSource/dodal/issues/409 (after a release of ophyd-async)

### DIFF
--- a/src/ophyd_async/epics/areadetector/aravis.py
+++ b/src/ophyd_async/epics/areadetector/aravis.py
@@ -1,0 +1,69 @@
+from typing import get_args
+
+from bluesky.protocols import HasHints, Hints
+
+from ophyd_async.core import DirectoryProvider, StandardDetector, TriggerInfo
+from ophyd_async.epics.areadetector.controllers.aravis_controller import (
+    AravisController,
+)
+from ophyd_async.epics.areadetector.drivers import ADBaseShapeProvider
+from ophyd_async.epics.areadetector.drivers.aravis_driver import AravisDriver
+from ophyd_async.epics.areadetector.writers import HDFWriter, NDFileHDF
+
+
+class AravisDetector(StandardDetector, HasHints):
+    """
+    Ophyd-async implementation of an ADAravis Detector.
+    The detector may be configured for an external trigger on a GPIO port,
+    which must be done prior to preparing the detector
+    """
+
+    _controller: AravisController
+    _writer: HDFWriter
+
+    def __init__(
+        self,
+        name: str,
+        directory_provider: DirectoryProvider,
+        driver: AravisDriver,
+        hdf: NDFileHDF,
+        gpio_number: AravisController.GPIO_NUMBER = 1,
+        **scalar_sigs: str,
+    ):
+        # Must be child of Detector to pick up connect()
+        self.drv = driver
+        self.hdf = hdf
+
+        super().__init__(
+            AravisController(self.drv, gpio_number=gpio_number),
+            HDFWriter(
+                self.hdf,
+                directory_provider,
+                lambda: self.name,
+                ADBaseShapeProvider(self.drv),
+                **scalar_sigs,
+            ),
+            config_sigs=(self.drv.acquire_time, self.drv.acquire),
+            name=name,
+        )
+
+    async def _prepare(self, value: TriggerInfo) -> None:
+        await self.drv.fetch_deadtime()
+        await super()._prepare(value)
+
+    def get_external_trigger_gpio(self):
+        return self._controller.gpio_number
+
+    def set_external_trigger_gpio(self, gpio_number: AravisController.GPIO_NUMBER):
+        supported_gpio_numbers = get_args(AravisController.GPIO_NUMBER)
+        if gpio_number not in supported_gpio_numbers:
+            raise ValueError(
+                f"{self.__class__.__name__} only supports the following GPIO "
+                f"indices: {supported_gpio_numbers} but was asked to "
+                f"use {gpio_number}"
+            )
+        self._controller.gpio_number = gpio_number
+
+    @property
+    def hints(self) -> Hints:
+        return self._writer.hints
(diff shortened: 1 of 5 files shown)


callumforrester:
@DiamondJoseph @coretl I'm not overjoyed at the separation of detector components across libraries. The fact that I have to have the writer/controller over in ophyd-async but assemble them in dodal is a bit cumbersome, especially if I want to update my code, I may have to do two PRs and an alpha release etc.

One thing that I think would help is to have an example assembly of these components in ophyd-async. In other words a more generic version of `ADAaravisDetector` in here https://github.com/DiamondLightSource/dodal/pull/421/files#diff-6b371b1b87d0470e413dd5e1fc0a50233e042ddf494229e8a7e22fc70d673c64R147-R201

To be clear, that is _as well as_ rather than _instead of_. What do you think?

coretl:
> One thing that I think would help is to have an example assembly of these components in ophyd-async. In other words a more generic version of `ADAaravisDetector` in here https://github.com/DiamondLightSource/dodal/pull/421/files#diff-6b371b1b87d0470e413dd5e1fc0a50233e042ddf494229e8a7e22fc70d673c64R147-R201
> 
> To be clear, that is _as well as_ rather than _instead of_. What do you think?

That sounds like a good idea, the only reason I don't want to put it in ophyd-async in the first place is because we don't want to put anything that relies on our naming conventions there.

Actually we could extend that idea even further, and make the `HDFAravis` class not know anything about naming conventions, e.g.

```python
class HDFAravis(StandardDetector):
    def __init__(
        self,
        prefix: str,
        driver_suffix: str,
        hdf_suffix: str,
        ...
    ):
        self.drv = ADAravis(prefix + driver_suffix)
        self.hdf = NDFileHDF(prefix + hdf_suffix)
        super().__init__(
            controller=ADAravisController(self.drv, ...),
            writer=HDFWriter(self.hdf, ...),
            ...
        )
```
Then in dodal:
```python
def oav():
    return HDFAravis("BL22I-EA-DET-01:", driver_suffix="DRV:", hdf_suffix="HDF:")
```
More typing, but clearer how the PV names are constructed...

What do you reckon?

DiamondJoseph:
At that point why not just 

`class HDFAravis(StandardDetector):
    def __init__(
        self,
        driver: ADAravisDriver,
        hdf: NDFileHDF,
        ...
    ):
        self.drv = driver
        self.hdf = hdf
        super().__init__(
            controller=ADAravisController(driver),
            writer=HDFWriter(hdf, ...),
            ...
        )
`

coretl:
Hmm, so then dodal would have:
```python
def oav():
    prefix = "BL22I-EA-DET-01:"
    return HDFAravis(ADAravis(prefix + "DRV:"), NDFileHDF(prefix + "HDF:"))
```
I guess that extends to extra plugins reasonably nicely too...

DiamondJoseph:
device_instantiation still requires prefix, name as arguments, but with those passed too

DiamondJoseph:
tests/epics/demo/test_demo.py::test_mover_moving_well seems to be reliably slower than expected on Python 3.10...

src/ophyd_async/epics/areadetector/aravis.py
```diff
@@ -0,0 +1,70 @@
...
+                lambda: self.name,
+                ADBaseShapeProvider(self.drv),
+                **scalar_sigs,
+            ),
+            config_sigs=(self.drv.acquire_time, self.drv.acquire),
+            name=name,
+        )
+
+    async def _prepare(self, value: TriggerInfo) -> None:
+        await self.drv._fetch_deadtime()
```
callumforrester, line 52: Can we do this without private access? To be honest could just make `_fetch_deadtime()` a public method

callumforrester: Or make a property

callumforrester (changes requested):

src/ophyd_async/epics/areadetector/drivers/aravis_driver.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def _reverse_lookup(
+    model_deadtimes: Dict[float, Tuple[str, ...]]
+) -> Callable[[str], float]:
+    def inner(pixel_format: str) -> float:
+        for deadtime, pixel_formats in model_deadtimes.items():
+            if pixel_format in pixel_formats:
+                return deadtime
+        return nan
```
Tom-Willemsen, line 40: I think I'd probably prefer to `raise ValueError` or something similar rather than return a `NaN` here.

The reason I prefer this is that `NaN` propagates quietly through maths operations and then is more likely to show an obscure problem later on, while really what I want to know is that it got an invalid/unknown pixel format.

Please can we also write a unit test for the invalid/unknown pixel format case.

src/ophyd_async/epics/areadetector/drivers/aravis_driver.py
```diff
@@ -0,0 +1,157 @@
...
+        self.model = ad_r(str, prefix + "Model")
+        self.pixel_format = ad_rw(str, prefix + "PixelFormat")
+        self.dead_time: Optional[float] = None
+        super().__init__(prefix, name=name)
+
+    async def _fetch_deadtime(self) -> None:
+        # All known in-use version B/C have same deadtime as non-B/C
+        model: str = (await self.model.get_value()).removesuffix("B").removesuffix("C")
+        pixel_format: str = await self.pixel_format.get_value()
+        self.dead_time = _deadtimes.get(model, lambda _: nan)(pixel_format)
```
Tom-Willemsen, line 157: As above, suspicious about use of `NaN` here, think I would rather it causes an immediate descriptive error.

DiamondJoseph:
This change no longer requires #249 and instead passes AravisTriggerSource as a String. #249 will be handled later

abbiemery (approved):
LGTM. Having it merged allows us make a pre-release and test before commuting to a full release so I'm favouring ship and iterate here.

callumforrester (approved):

Issue #210, opened 2024-04-15
Add driver, controller classes for Kinetix, Vimba cameras


jwlodek:
As a part of developing flyscanning at the HEX beamline at NSLS-II, we have developed controller/driver classes for Kinetix and Vimba cameras. I would like to submit these via a PR.

PR #216, opened 2024-04-15
Adding driver and controller classes for Kinetix and Vimba cameras

jwlodek:
Addresses #210


src/ophyd_async/epics/areadetector/controllers/kinetix_controller.py
```diff
@@ -0,0 +1,55 @@
...
+
+    async def arm(
+        self,
+        num: int,
+        trigger: DetectorTrigger = DetectorTrigger.internal,
+        exposure: Optional[float] = None,
+    ) -> AsyncStatus:
+        await asyncio.gather(
+            self.driver.trigger_mode.set(KINETIX_TRIGGER_MODE_MAP[trigger]),
+            self.driver.num_images.set(999_999 if num == 0 else num),
```
coretl, line 42: The `999999` number is the max images that Pilatus can make, is there a similar limit for the kinetix?

jwlodek: I think the limit for the Kinetix is just the max value of an asynInt32 record, but I felt like 999999 is probably large enough so I kept it in. I have no issue with removing this (although the Kinetix is fast enough with a large enough frame size that the filesaving plugin is almost certainly not going to keep up, so you will run out of memory before you reach 999999 images).

src/ophyd_async/epics/areadetector/drivers/kinetix_driver.py
```diff
@@ -0,0 +1,24 @@
...
+    speed = 2
+    dynamic_range = 3
+
+
+class KinetixDriver(ADBase):
+    def __init__(self, prefix: str) -> None:
+        super().__init__(prefix)
+        # self.pixel_format = ad_rw(PixelFormat, prefix + "PixelFormat")
+        self.trigger_mode = ad_rw(KinetixTriggerMode, prefix + "TriggerMode")
+        self.mode = ad_rw(KinetixReadoutMode, prefix + "ReadoutPortIdx")
```
coretl, line 24: For areaDetector I've tended to name the attribute after the PV suffix, is there a reason to name this `mode` rather than `readout_port_idx` or `readout_mode`?

coretl (commented):
Please could we have some tests too?

A subset of https://github.com/bluesky/ophyd-async/blob/main/tests/epics/areadetector/test_aravis.py would be good

I note that the vimba driver should probably use the deadtime calculations from aravis, and the aravis driver should use the gated modes from vimba, but I'll make a separate issue for that

src/ophyd_async/epics/areadetector/controllers/kinetix_controller.py
```diff
@@ -0,0 +1,55 @@
...
+    DetectorTrigger.variable_gate: KinetixTriggerMode.gate,
+    DetectorTrigger.edge_trigger: KinetixTriggerMode.edge,
+}
+
+
+class KinetixController(DetectorControl):
+    def __init__(
+        self,
+        driver: KinetixDriver,
+        good_states: Set[DetectorState] = set(DEFAULT_GOOD_STATES),
```
DiamondJoseph, line 26: I don't know enough about AD DetectorStates, so I ignored this param for the Pilatus and Aravis- do we think it's important to expose? I can go back and adjust those detectors before there's any pickup.

coretl: I see no reason to have to override the good states when instantiating the Controller, I think we know up front what are the good states for a given detector.

@jwlodek any reason you added it as an init arg? If not then I recommend we:

```suggestion
```

coretl:
> For the Pilatus and the Aravis we've also created fooDetector types- partly because we didn't want to split the implementation of the device between two repositories- should we stick with that format (and have facility specific handling limited to calling the constructor fo the StandardDetector), or should the StandardDetectorImplementation be in facility specific code?

Having thought about this a bit more, I think I would prefer this pattern:
```python
class FooDetector(StandardDetector):
    def __init__(
        self,
        drv_prefix: str,
        hdf_prefix: str,
        ...,
        name: str = "",
    ):
        self.drv = ADFoo(drv_prefix)
        self.hdf = NDFileHDF(hdf_prefix)
        super().__init__(
            FooController(self.drv),
            HDFWriter(self.hdf, ...),
            ...,
            name=name
        )     
```
Then we instantiate as `FooDetector(f"{prefix}DRV:", f"{prefix}HDF:")` which makes no assumptions about PV naming convention, but does tie the detector to use the correct class of template class and controller class

@jwlodek @DiamondJoseph what do you think?

src/ophyd_async/epics/areadetector/drivers/vimba_driver.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class VimbaTriggerSource(str, Enum):
+    freerun = "Freerun"
+    line1 = "Line1"
+    line2 = "Line2"
+    fixed_rate = "FixedRate"
+    software = "Software"
+    action0 = "Action0"
+    action1 = "Action1"
```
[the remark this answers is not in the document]
jwlodek: The list is exhaustive, and I believe applicable to all Vimba cameras. Since this parameter comes from the Vimba API rather than the GenICam generated xml from Aravis, we can assume these are universally the same.

src/ophyd_async/epics/areadetector/controllers/kinetix_controller.py
jwlodek: The only reason was that it was done this way for the Pilatus when I was writing these classes.

src/ophyd_async/epics/areadetector/drivers/kinetix_driver.py
jwlodek: I had `ReadoutMode` originally, but changed it because I couldn't make the linter happy actually...

I didn't want to use `ReadoutModeIdx` because I think treating it as a `ReadoutMode` selection rather than an index in the list is better.

coretl: Could you point me to the corresponding record in the template please?

jwlodek:
@coretl could we merge this PR in before we make further changes to the format of the various classes? I'd like to be able to close out this branch and re-sync with main.

src/ophyd_async/epics/areadetector/drivers/kinetix_driver.py
jwlodek: https://github.com/NSLS-II/ADKinetix/blob/ee1359dbb02d99e74bd442e228a760f4f9e069b7/kinetixApp/Db/ADKinetix.template#L116

DiamondJoseph (changes requested):
Pilatus is now using the default GOOD_STATES from the ADBaseDriver, I'd like to see that consistent until we find a detector that isn't just taking GOOD_STATES. The exact form of FooDetector we can continue to quibble about.

DiamondJoseph:
eek pushed the big rebase button prematurely

DiamondJoseph:
I've made a new issue from the "What form should StandardDetector.__init__ take"

https://github.com/bluesky/ophyd-async/issues/263

jwlodek:
Removed the states set kwarg.

DiamondJoseph (approved):

src/ophyd_async/epics/areadetector/drivers/kinetix_driver.py
coretl: Could these enum strings be pushed down to the template? Then having a `ReadoutMode` PV would make more sense...

I've tried not to add any translation in the "bucket of PVs" layer, and put the logic in the `Controller`, but I guess that is difficult here as the attribute isn't used in the controller. 

Either way, I think `readout_mode` for the attribute and `KinetixReadoutMode` are reasonable names, is the linter happy with those?

coretl (approved):
I'd prefer to get the `mode` -> `readout_mode` change (or similar) in soon, but happy to merge this and do a separate PR later if that is better

PR #217, opened 2024-04-15
Add create_soft_signal_r and creates_soft_signal_rw methods

jsouter:
Moved from dodal, have placed in core.signals as seemed logical to prevent circular imports, could be moved to sim_signal_backends.py

### DIFF
--- a/src/ophyd_async/epics/signal/signal.py
+++ b/src/ophyd_async/epics/signal/signal.py
@@ -58,6 +58,25 @@ def epics_signal_rw(
     return SignalRW(backend)
 
 
+def epics_signal_rw_rbv(
+    datatype: Type[T],
+    write_pv: str,
+    read_suffix: str = "_RBV",
+) -> SignalRW[T]:
+    """Create a `SignalRW` backed by 1 or 2 EPICS PVs, with a suffix on the readback pv
+
+    Parameters
+    ----------
+    datatype:
+        Check that the PV is of this type
+    write_pv:
+        The PV to write to
+    read_suffix:
+        Append this suffix to the write pv to create the readback pv
+    """
+    return epics_signal_rw(datatype, f"{write_pv}{read_suffix}", write_pv)
+
+
 def epics_signal_r(datatype: Type[T], read_pv: str) -> SignalR[T]:
     """Create a `SignalR` backed by 1 EPICS PV
 
--- a/tests/epics/test_signals.py
+++ b/tests/epics/test_signals.py
@@ -31,7 +31,14 @@ from bluesky.protocols import Reading
 from ophyd_async.core import SignalBackend, T, get_dtype, load_from_yaml, save_to_yaml
 from ophyd_async.core.utils import NotConnected
 from ophyd_async.epics.signal._epics_transport import EpicsTransport
-from ophyd_async.epics.signal.signal import _make_backend
+from ophyd_async.epics.signal.signal import (
+    _make_backend,
+    epics_signal_r,
+    epics_signal_rw,
+    epics_signal_rw_rbv,
+    epics_signal_w,
+    epics_signal_x,
+)
 
 RECORDS = str(Path(__file__).parent / "test_records.db")
 PV_PREFIX = "".join(random.choice(string.ascii_lowercase) for _ in range(12))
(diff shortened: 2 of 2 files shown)


tests/core/test_signal.py
```diff
@@ -119,3 +122,18 @@ async def test_set_and_wait_for_value():
     assert not st.done
     set_sim_put_proceeds(sim_signal, True)
     assert await time_taken_by(st) < 0.1
+
+
+async def test_create_soft_signal():
```
DominicOram, line 127: Could: I would suggest splitting these into separate tests by either:
* Using `pytest.mark.parametrize`
* Have a helper function that is the test and call it with each case

src/ophyd_async/core/signal.py
coretl, line 1: @DominicOram you mentioned you needed a way to get a value into a `SignalR`, should we change `create_soft_signal_r` to return `Tuple[SignalR[T], SimSignalBackend]` instead? Should we also do the same for `create_soft_signal_rw`?

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +253,18 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
     return _sim_backends[signal].set_callback(callback)
 
 
+def create_soft_signal_rw(
```
coretl, line 256: Slight preference for:
```suggestion
def soft_signal_rw(
```

src/ophyd_async/core/signal.py
jsouter: Okay, have pushed with only soft_signal_r returning the tuple, can change it if we prefer to have the same interface.

coretl: > Do we need it for create_soft_signal_rw? Can't you just do a usual set in that case?

No real need for it here, only if we want to make the 2 functions have the same return type.

coretl: I guess we could call them `soft_signal_rw` and `soft_signal_r_and_backend` (or something similar) if we care about being explicit?

DominicOram: Sounds good to me, I have no strong opinions

jsouter:
@gilesknap @coretl Add in option to pass an initial value to SimSignalBackend in the init method, and also supporting this in the soft_signal_* methods. self._initial_value gets set to the initial_value argument in SimSignalBackend's init, but the converter only converts it when connect is called, though I think this is fine if users are only accessing it through the public methods. soft_signal_r_and_backend still returns the backend, though this may not be required now.

gilesknap:
> @gilesknap @coretl Add in option to pass an initial value to SimSignalBackend in the init method, and also supporting this in the soft_signal_* methods. self._initial_value gets set to the initial_value argument in SimSignalBackend's init, but the converter only converts it when connect is called, though I think this is fine if users are only accessing it through the public methods. soft_signal_r_and_backend still returns the backend, though this may not be required now.

Thanks James, I have rebased and used the new feature successfully.

jsouter:
Following discussion with @coretl, SimSignalBackends no longer take a source/name at init, and all backends now provide source as a method, where the Signal classes pass a name to the backend. source is available from signals as a property.
This shouldn't affect the CA/PVA backend much, but for devices using SimSignalBackend, the source descriptor entry is now based on the device's name, so the source may no longer resemble proper PV names, and will be more like "soft://my-sim-motor" instead of "soft://MY-SIM-MOTOR:X_RBV" e.g.

evalott100 (approved):
Looks good!

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +253,34 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+def soft_signal_rw(
+    datatype: Optional[Type[T]],
+    name: str,
+    initial_value: Optional[T] = None,
+) -> SignalRW[T]:
+    """Creates a read-writable Signal with a SimSignalBackend"""
+    signal = SignalRW(
+        SimSignalBackend(datatype, initial_value)
+    )
+    signal.set_name(name)
```
coretl: Please can we make `name=None` be passed to these helpers, and if it isn't already there then add it as an optional arg to `Signal.__init__`

coretl: And also add this to `epics_signal_*` helpers

jsouter: Ah! Missed this comment somehow, might have to be a new PR.

PR #218, opened 2024-04-15
197 add long name to device

### LINKED ISSUE #197: Add long_name to Device
Should be optionally passed to `__init__`, but I don't think we'll add a setter until there is a clear need for one.

See https://github.com/bluesky/ophyd/pull/1187 for implementation in ophyd sync

Relm-Arrowny:
Fix #197 

Added long_name to __init__. for device, I have also added get/setter in-line with ophydObj implementation [#1187](https://github.com/bluesky/ophyd/pull/1187)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ dev = [
     "tox-direct",
     "types-mock",
     "types-pyyaml",
+    "bluesky",
 ]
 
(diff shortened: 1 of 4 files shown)


coretl:
I think we will hold off merging this until the ophyd implementation is merged

Issue #208, opened 2024-04-15
Create async versions of all the protocols we use in ophyd-async


coretl:
At the moment we import `from bluesky.protocols import Readable, Configurable, ...`. This has 2 issues:
- If we extend the protocol in ophyd-async (e.g. add `Collectable.get_index`) then we need a release of bluesky before we can use them
- If we type narrow (e.g. we don't have `SyncOrAsync` returns, we write `async def` instead) then the editor picks up the extended version when we override it, so we have to edit the text to make it match our narrowed version

I suggest we make `ophyd_async.protocols`, where any protocol that has `SyncOrAsync` in it is narrowed for use in ophyd-async. At least:
- `AsyncReadable`
- `AsyncConfigurable`
- `AsyncPausable`

then update any imports to use them.

This also gives us a place to put things like `Watchable` from #176 until they can migrate into bluesky

coretl:
For example, this:
https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/protocols.py#L263-L303

Would turn into this:
```python
from bluesky.protocols import HasName, Reading, DataKey, ...

@runtime_checkable
class AsyncReadable(HasName, Protocol):
    """Readable narrowed to be async only"""
    @abstractmethod
    async def read(self) -> Dict[str, Reading]:
        ...

    @abstractmethod
    async def describe(self) -> Dict[str, DataKey]:
        ...
```

callumforrester:
Reviewers: @coretl

callumforrester:
Fixed in #219

Issue #57, opened 2023-11-02
Create an instantly moving mock motor


DominicOram:
It may be useful for tests to have a mock motor that instantly moves it's readback to it's setpoint. This mock should leave near the `Motor` object. Maybe it should be a function that mocks out any `Motor` children of a given device?

This will be easier done when https://github.com/bluesky/ophyd-async/issues/49 is done.

callumforrester:
@DominicOram is this part of https://github.com/bluesky/ophyd-async/issues/94?

DominicOram:
> @DominicOram is this part of https://github.com/bluesky/ophyd-async/issues/94?

Unrelated I think. This issue is to make it easier to spin up mock motors for unit tests. #94 is I think to make more generic soft signal backends.

coretl:
I think there should be 2 sorts of "not real" motors:
1. A motor that looks realistic from the outside, i.e. takes time to move. There is `ophyd_async.epics.demo.Mover` that does this with a supporting IOC, and we should make `ophyd_async.sim.demo.SimMotor` that does the same thing and works with the `PatternDetector` created in #131 
2. A motor that has no logic behind it, just a bunch of PVs that immediately accept the value that has been given to it. This will work for `ophyd_async.epics.motion.Motor` if `connect(sim=True)` is passed, with a slight issue, the readback will not update. This is what we want for tests, but I suspect not for this issue.

@DominicOram will 1. be sufficient to solve this? In which case I suggest we turn this ticket into "Make a slightly realistic SimMotor"

callumforrester:
@coretl does 1. solve other use cases? If so @DominicOram, could you have a 1-style motor with the time to move set to 0?

gilesknap:
If nobody objects I'll work on creating a 1 style motor with a default time to move of 0. In full sim mode it would use the velocity to work out how long the move would take and update it throughout the move (probably), Could also take acceleration into account if there is any application for that.

DominicOram:
Yh, that's fine for me. Thank you!

coretl:
> @coretl does 1. solve other use cases? If so @DominicOram, could you have a 1-style motor with the time to move set to 0?

The primary use case is docs and tutorials, something you can poke in iPython to play around with bluesky plans

callumforrester:
Does #224 provide everything this issue requires? If so we can close it

@DominicOram @coretl @gilesknap

coretl:
I think it does

DominicOram:
[...]

PR #219, opened 2024-04-16
Add Async Protocols

OCopping:
Replacing imports of `bluesky.protocols` with new async counterparts.
Relates to #208


tests/protocols/test_protocols.py
```diff
@@ -0,0 +1,50 @@
...
+    "type_, hardware, pass_",
+    [
+        ("Readable", "ABDetector(name='hi')", True),
+        ("Readable", "SynAxis(name='motor1')", True),
+        ("Readable", "TrivialFlyer()", False),
+        ("Configurable", "ABDetector(name='hi')", True),
+        ("Pausable", "ABDetector(name='hi')", True),
+    ],
+)
+def test_mypy(type_, hardware, pass_):
```
callumforrester, line 32: This is already covered by the separate CI job to run mypy, so a test for it isn't needed

tests/protocols/test_protocols.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_readable():
+    assert isinstance(sim.motor1, bs_protocols.AsyncReadable)
+    assert isinstance(sim.det1, bs_protocols.AsyncReadable)
+    assert not isinstance(sim.flyer1, bs_protocols.AsyncReadable)
+
+
+def test_pausable():
+    assert isinstance(sim.det1, bs_protocols.AsyncPausable)
```
callumforrester, line 17: I'm concerned that these pass, since the ophyd (not async) sim objects are explicitly _not_ `AsyncReadable`, `AsyncPausable` etc.

The test cases should look something like this:
| Library used to write device | `AsyncReadable` | `Readable` |
|------------------------------|-----------------|------------|
| Original ophyd               | No              | Yes        |
| ophyd-async                  | Yes             | Yes        |

But we should figure out why these pass, ophyd objects are not async, so they shouldn't register as async

tests/protocols/test_protocols.py
callumforrester, line 1: Is it worth having tests that check each protocol subclasses it's `SyncOrAsync` protocol? e.g. `assert issubclass(AsyncReadable, Readable)`. I'm not sure that's even allowed for protocols, but should have some way of checking they are in sync.

src/ophyd_async/protocols/protocols.py
callumforrester, line 1: Happy with this, but could the classes have docstrings? Can more or less copy/paste from their equivalents in `bluesky.protocols`

callumforrester (changes requested):

tests/protocols/test_protocols.py
coretl: They pass because `isinstance` checks on protocols are essentially `hasattr` checks

tests/protocols/test_protocols.py
```diff
@@ -0,0 +1,50 @@
+import tempfile
+
+import mypy.api
+import pytest
+from ophyd import sim
```
coretl, line 5: We definitely don't want to import `ophyd` inside `ophyd_async`, even for tests. We'll have to find another way to test these. There will is an `ophyd_async.sim` that you can use, but it might not cover everything at the moment. Maybe after #57 there will be enough

tests/protocols/test_protocols.py
callumforrester: Is `check_supports` more comprehensive?

coretl: nope

tests/protocols/test_protocols.py
OCopping: I have rebased this branch onto @gilesknap 's sim_motor branch to get a motor sim, however there doesn't appear to be any detector or flyer sim

src/ophyd_async/protocols/protocols.py
OCopping: The protocols in bluesky only have docstrings on their methods
e.g. for `Readable`:
https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/protocols.py#L263-L303

tests/protocols/test_protocols.py
OCopping: Should this go into a separate issue?

callumforrester: Yes, if it's worth doing at all, @coretl any opinions?

src/ophyd_async/protocols/protocols.py
callumforrester: Nevermind then, that's to be sorted separately

coretl, line 1: Please can you move this file to `ophyd_async/protocols.py`

coretl (approved):

PR #221, opened 2024-04-16
Add SignalRW helpers

### LINKED ISSUE #139: Make _RBV helper externally visible
The pattern of having `_RBV` at the end of a PV for the corresponding readback is common outside just AD. `ophyd-async` should provide a helper function for it. One already exists at https://github.com/bluesky/ophyd-async/blob/c42afcb1f801d5f3f6ae88fd6a7fa8e0c7043893/src/ophyd_async/epics/areadetector/utils.py#L11 but it feels weird for my device to depend on area detector just for this.

GDYendell:
Fixes #139

### DIFF
--- a/src/ophyd_async/epics/signal/signal.py
+++ b/src/ophyd_async/epics/signal/signal.py
@@ -58,6 +58,25 @@ def epics_signal_rw(
     return SignalRW(backend)
 
 
+def epics_signal_rw_rbv(
+    datatype: Type[T],
+    write_pv: str,
+    read_suffix: str = "_RBV",
+) -> SignalRW[T]:
+    """Create a `SignalRW` backed by 1 or 2 EPICS PVs, with a suffix on the readback pv
+
+    Parameters
+    ----------
+    datatype:
+        Check that the PV is of this type
+    write_pv:
+        The PV to write to
+    read_suffix:
+        Append this suffix to the write pv to create the readback pv
+    """
+    return epics_signal_rw(datatype, f"{write_pv}{read_suffix}", write_pv)
+
+
 def epics_signal_r(datatype: Type[T], read_pv: str) -> SignalR[T]:
     """Create a `SignalR` backed by 1 EPICS PV
 
--- a/tests/epics/test_signals.py
+++ b/tests/epics/test_signals.py
@@ -31,7 +31,14 @@ from bluesky.protocols import Reading
 from ophyd_async.core import SignalBackend, T, get_dtype, load_from_yaml, save_to_yaml
 from ophyd_async.core.utils import NotConnected
 from ophyd_async.epics.signal._epics_transport import EpicsTransport
-from ophyd_async.epics.signal.signal import _make_backend
+from ophyd_async.epics.signal.signal import (
+    _make_backend,
+    epics_signal_r,
+    epics_signal_rw,
+    epics_signal_rw_rbv,
+    epics_signal_w,
+    epics_signal_x,
+)
 
 RECORDS = str(Path(__file__).parent / "test_records.db")
 PV_PREFIX = "".join(random.choice(string.ascii_lowercase) for _ in range(12))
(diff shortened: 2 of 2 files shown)


src/ophyd_async/epics/signal/signal.py
coretl, line 1: I might have got the wrong end of the stick, but I thought we were keeping `epics_signal_rw` as it is, and adding `epics_signal_rw_rbv(datatype, write_pv, read_suffix="_RBV")`...

@DominicOram what did you think?

olliesilvester: Yeah that's what is currently in dodal, I like it the most

callumforrester: I agree, would prefer a single function with a default argument

GDYendell: Updated to just add a new function that allows overriding the readback suffix.

src/ophyd_async/epics/signal/signal.py
```diff
@@ -58,6 +58,28 @@ def epics_signal_rw(
     return SignalRW(backend)
 
 
+def epics_signal_rw_rbv(
```
callumforrester, line 61: I think we still want it to be called `epics_signal_rw`

coretl: nope, this is the additional function to specifically do `prefix`, `prefix + "_RBV"` PV pairs

src/ophyd_async/epics/signal/signal.py
```diff
@@ -58,6 +58,28 @@ def epics_signal_rw(
...
+    datatype: Type[T],
+    read_pv: str,
+    write_pv: Optional[str] = None,
+    read_suffix: str = "_RBV",
```
coretl, line 65: We never need to provide a separate read pv, so suggest we change signature to:
```suggestion
    write_pv: str,
    read_suffix: str = "_RBV",
```

src/ophyd_async/epics/signal/signal.py
```diff
@@ -58,6 +58,28 @@ def epics_signal_rw(
...
+    datatype:
+        Check that the PV is of this type
+    read_pv:
+        The PV to read and monitor
+    write_pv:
+        If given, use this PV to write to, otherwise use read_pv
+    read_suffix:
+        Append this suffix to the readback pv
+    """
+    return epics_signal_rw(datatype, f"{read_pv}{read_suffix}", write_pv or read_pv)
```
coretl, line 80: so this becomes
```suggestion
    return epics_signal_rw(datatype, f"{write_pv}{read_suffix}", write_pv)
```

coretl (changes requested):

src/ophyd_async/epics/signal/signal.py
coretl: nope, we already have one of these, this is a second function for when we have a `prefix`, `prefix + "_RBV"` record pair

src/ophyd_async/epics/signal/signal.py
DominicOram: Yes, new function, behaviour of `epics_signal_rw` need not change. This is literally just a small helper function to tidy up some devices here and there

src/ophyd_async/epics/signal/signal.py
GDYendell: Done

src/ophyd_async/epics/signal/signal.py
GDYendell: Done

src/ophyd_async/epics/signal/signal.py
```diff
@@ -58,6 +58,25 @@ def epics_signal_rw(
     return SignalRW(backend)
 
 
+def epics_signal_rw_rbv(
+    datatype: Type[T],
+    write_pv: Optional[str] = None,
```
coretl, line 63:
```suggestion
    write_pv: str,
```

tests/core/test_signal.py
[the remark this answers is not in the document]
GDYendell: Done

coretl (approved):

DominicOram:
Does it make sense to update the signals in AD with this e.g. https://github.com/bluesky/ophyd-async/blob/c42afcb1f801d5f3f6ae88fd6a7fa8e0c7043893/src/ophyd_async/epics/areadetector/utils.py#L11?

PR #224, opened 2024-04-16
Add Simulation motor

gilesknap:
Add a simulation motor class that can move instantly (default) or do a timed move according to the velocity and distance.

@coretl I'm doing a draft PR so you can give me some early feedback. Does this look about right? 

At the moment I'm fighting flake8 vs my installed ruff extension. Perhaps we should update to latest copier?
edit: copier adoption is under way at #207


coretl:
#207 is progress so far with copier

src/ophyd_async/epics/motion/sim_motor.py
coretl, line 1: Can this be in `ophyd_async.sim.demo` please?

coretl: Also we don't want to depend on anything from `ophyd_async.epics`, we should make soft signals with #217

gilesknap: `ophyd_asyc.sim` I get, but what is with the `demo`?

gilesknap:
@coretl How is this now?

CI is broken due to linting - awaiting https://github.com/bluesky/ophyd-async/pull/207

gilesknap:
This is a fix for #57

src/ophyd_async/sim/demo/sim_motor.py
```diff
@@ -0,0 +1,122 @@
...
+        - name: str: name of device
+        - instant: bool: whether to move instantly, or with a delay
+        """
+        self._instant = instant
+        self._move_task: Optional[asyncio.Task] = None
+
+        # Define some signals
+        self.user_setpoint = soft_signal_rw(
+            float, "user_setpoint", prefix + ".setpoint", initial_value=0
+        )
```
coretl, line 28: After @jsouter makes the changes then this will be:
```suggestion
        self.user_setpoint = soft_signal_rw(float, initial_value=0)
```

src/ophyd_async/sim/demo/sim_motor.py
```diff
@@ -0,0 +1,122 @@
...
+        def update_watchers(current_position: float):
+            for watcher in watchers:
+                watcher(
+                    name=self.name,
+                    current=current_position,
+                    initial=old_position,
+                    target=new_position,
+                    unit=units,
+                    time_elapsed=time.monotonic() - start,
+                )
```
coretl, line 109: I would put this updating of watchers in `update_positions`

src/ophyd_async/sim/demo/sim_motor.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # set up a task that updates the motor position at 10hz
+        self._move_task = asyncio.create_task(update_position())
+
+        # set up watchers to be called when the motor position changes
+        self.user_readback.subscribe_value(update_watchers)
```
coretl, line 115: So then don't need this subscribe

coretl (changes requested):

gilesknap:
Awaiting the branch 215_soft_signals to be merged and then will rebase off of main.
(so at present this does not pass CI)

Will address the change requests on Fri.

src/ophyd_async/sim/demo/sim_motor.py
gilesknap: OK -  i already did this and checked it works against the latest of James branch - just not rebased this PR yet.

However - why is there no name at all? I also added the prefix to the name - if this is not needed should I drop prefix from the constructor?

src/ophyd_async/sim/demo/sim_motor.py
gilesknap: Makes sense - will apply on Fri.

jsouter:
> Awaiting the branch 215_soft_signals to be merged and then will rebase off of main. (so at present this does not pass CI)
> 
> Will address the change requests on Fri.

Have yet to push the changes which support numpy integer/floating types in the SimConverter, can push that on top of my branch or make a separate PR (should just be a few lines)

src/ophyd_async/sim/demo/sim_motor.py
gilesknap: James has now removed the required argument 'name' and I have dropped prefix from the SimMotor constructor.

src/ophyd_async/sim/demo/sim_motor.py
gilesknap: This is done

gilesknap:
rebased over https://github.com/bluesky/ophyd-async/tree/215_soft_signals.

gilesknap:
@callumforrester would you like to take a look?

gilesknap:
@callumforrester because this is rebased it looks like lots of changes but you just need to look at:

- src/ophyd_async/sim/demo/sim_motor.py
- tests/sim/demo/test_sim_motor.py

coretl (approved):

Issue #206, opened 2024-04-15
Write docs on how to do testing


coretl:
Need to document 2 ways:
1. Exercise the device directly, i.e. call `device.read()` and inspect the output. Should use helpers made in #199 
2. Run a plan in the RunEngine and examine the docs. Use a pattern like:
```python
from collections import defaultdict

def assert_emitted(docs: Dict[str, list], **numbers: int):
    assert list(docs) == list(numbers)
    assert {name: len(d) for name, d in docs.items()} == numbers

def test_plan():
    docs = defaultdict(list)
    RE(plan(), lambda name, doc: docs[name].append(doc))
    assert_emitted(docs, start=1, descriptor=1, resource=1, datum=1, event=1, stop=1)
```

callumforrester:
There is an existing docs page on testing that we can add to: https://blueskyproject.io/ophyd-async/main/user/how-to/write-tests-for-devices.html

PR #226, opened 2024-04-17
199 test helpers for assering value reading and configuration

### LINKED ISSUE #199: Test helpers for assering value, reading and configuration
[Dodal](https://github.com/DiamondLightSource/dodal) developers have produced various useful ways to test ophyd-async devices such as [`verify()`](https://github.com/DiamondLightSource/dodal/blob/6b43b48fcb388c43f919b7d0866fb44d8f8199ee/tests/devices/unit_tests/test_synchrotron.py#L201). Some of these should probably make it back into `main` and sit alongside other helper functions like `set_sim_signal`.

I think we need 3 helper functions:
- `assert_value(signal: SignalR[T], value: T)`
- `assert_reading(readable: Readable, reading: Reading)`
- `assert_configuration(readable: Readable, configuration: Configuration)`
 
## Acceptance Criteria
- Functions in ophyd-async
- Tests provided
- Docs provided where appropriate
- Issues raised in dodal to replace `verefy()` and any other old functions that can be superseded

Relm-Arrowny:
This is to fix #199 and possibly #206 

Acceptance Criteria

- [x]  Functions in ophyd-async:
- assert_reading
- assert_configuration
- assert_value
- assert_emitted
- [x]  Tests provided
- [x]  Docs provided where appropriate
- [x]  Issues raised in dodal to replace verefy() and any other old functions that can be superseded

For #206 Need to document 2 ways:
- [x] Exercise the device directly, i.e. call device.read() and inspect the output. Should use helpers made in #199 
- This is already convered in test_sensor_reading_shows_value so I only changed it to use helpers made here.
- [x] Run a plan in the RunEngine and examine the docs. Use a pattern like:
- Added test_sensor_in_plan in demo.py together with changes in the documentation.


src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
     return _sim_backends[signal].set_callback(callback)
 
 
+async def verify_readable(
```
callumforrester, line 260: I think this function should be made explicitly private with an _, I'm also not sure about the name, maybe `_verify_async_output`?

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+        result = await func()
+    else:
+        result = func()
+
+    for signal in expectation:
+        for field in expectation[signal]:
+            if field == "timestamp":
+                assert isinstance(result[signal]["timestamp"], float)
+            else:
+                assert result[signal][field] == expectation[signal][field]
```
callumforrester, line 276: I think it's fine to just do 

```suggestion
    assert result == expectation
```

And expect people to do things like

```python
from unittest.mock import ANY

await assert_reading(my_detector, {"value": 0.1, "timestamp": ANY})
```

It also gives them more power to make other attributes `ANY` if they don't care about the values

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+
+
+async def assert_value(signal: SignalR[T], value: T) -> None:
+    """Assert a signal value"""
+    assert await signal.get_value() == value
+
+
+async def assert_reading(
+    readable: Readable, reading: Dict[str, Reading] | dict[str, dict[str, Any]]
```
callumforrester, line 285: I think we should explicitly require readings and not allow the `dict[str, Any]` path, ditto for `assert_configuration`. Also, I recommend using the `Mapping` type rather than `Dict` if you do not expect the dictionary to be mutated

src/ophyd_async/core/signal.py
```diff
@@ -48,16 +58,10 @@ def __init__(
         self, backend: SignalBackend[T], timeout: Optional[float] = DEFAULT_TIMEOUT
     ) -> None:
         self._name = ""
+        self._long_name = None
```
coretl, line 61: This line doesn't belong in this PR does it?

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+
+async def assert_reading(
+    readable: Readable, reading: Dict[str, Reading] | dict[str, dict[str, Any]]
+) -> None:
+    """Assert readable by calling verify_readable"""
+    await verify_readable(readable.read, reading)
+
+
+async def assert_configuration(
+    configurable: Configurable,
```
coretl, line 292: When #219 is done then this can be `AsyncConfigurable` then we know the result is always awaitable

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+async def verify_readable(
+    func: Callable[[], dict[str, Any] | Awaitable[dict[str, Any]]],
+    expectation: Dict[str, Any],
+) -> None:
+    """Loop thorough dictionary and verify it against
+    a expected dictionary (readable)"""
+    if asyncio.iscoroutinefunction(func):
+        result = await func()
+    else:
+        result = func()
```
coretl, line 269: ... so this becomes
```suggestion
    result = await func()
```

coretl (commented):
Looks like this is on the right track, thanks!

src/ophyd_async/core/signal.py
```diff
@@ -253,6 +257,52 @@ def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> No
...
+async def assert_configuration(
+    configurable: Configurable,
+    configuration: Dict[str, Reading] | dict[str, dict[str, Any]],
+) -> None:
+    """Assert readable generated by configurable.read_configuration
+    using verify_readable"""
+    await verify_readable(configurable.read_configuration, configuration)
+
+
+def assert_emitted(docs: Dict[str, list], **numbers: int):
```
callumforrester, line 300: You can import `DocumentType` from event model

```suggestion
def assert_emitted(docs: DocumentType, **numbers: int):
```

src/ophyd_async/core/signal.py
callumforrester, line 1: Unsure if you were already doing this but I think the docstrings should have a bit more detail, including parameters

callumforrester (changes requested):

src/ophyd_async/core/signal.py
coretl: Although docs is a dict of document name to list of document instances, so I think that should be:
```suggestion
def assert_emitted(docs: Dict[str, list[DocumentType]], **numbers: int):
```
Except `DocumentType` is actually the types, which is not what we want, we would quite like a similar `Document` union in event model...

coretl: so in the absence of that it's probably as correct as it could be...

src/ophyd_async/core/signal.py
coretl: This checks the output of `read` or `read_configuration` which is `Dict[str, Reading]` so how about `_verify_readings`?

src/ophyd_async/core/signal.py
Relm-Arrowny: I have pretty much made all the changes suggested, just want to make a note of things i noticed making this change, By forcing Mapping t[str,Reading], mypy will fail unless Reading is in the correct order of {value, timestamp, alarm_severity, message} e.g. 
```
 {
                "value": 5.0,
                "timestamp": ANY,
                "alarm_severity": 0,
                "message": "str",
}
```
This also make [varify](https://github.com/DiamondLightSource/dodal/blob/6b43b48fcb388c43f919b7d0866fb44d8f8199ee/tests/devices/unit_tests/test_synchrotron.py#L201) function seem less versatile, It will still behave in the same way but mypy will fail if a function does not return Mapping[str,Reading].

callumforrester (approved):

src/ophyd_async/core/signal.py
coretl: I'd put it back to dict then, we don't want to enforce this ordering...

src/ophyd_async/core/signal.py
```diff
@@ -2,9 +2,20 @@
...
+    Callable,
+    Dict,
+    Generic,
+    Mapping,
+    Optional,
+    Union,
+)
 
 from bluesky.protocols import (
+    Configurable,
```
coretl, line 18: Since #219 is merged, please can you update this to use `ophyd_async.protocols.AsyncReadable` and `AsyncConfigurable`

src/ophyd_async/core/signal.py
coretl: Then you can apply this change

coretl:
Please could you also add to the "writing tests for devices" doc the information on how to use `assert_emitted` from https://github.com/bluesky/ophyd-async/issues/206

src/ophyd_async/core/signal.py
callumforrester: This seems like weird behaviour from `TypedDict` and I can't reproduce it, maybe something else is going on here? https://mypy-play.net/?mypy=latest&python=3.12&gist=57a93d787c3cbfe8ab39c061e398edde

Relm-Arrowny: After merging with main I can no longer create this behaviour, so nothing to worry about.

Relm-Arrowny:
I think I got everything, but please have a quick check if it all make sense and I have covered everything.

docs/how-to/write-tests-for-devices.rst
```diff
@@ -43,3 +45,12 @@ There is another utility function, ``set_sim_callback``, for hooking in logic wh
 
 .. literalinclude:: ../../tests/epics/demo/test_demo.py
    :pyobject: test_mover_stopped
+
+
+Testing device in a plan with RunEngine
```
callumforrester, line 50: Title case
```suggestion
Testing a Device in a Plan with the RunEngine
```

docs/how-to/write-tests-for-devices.rst
```diff
@@ -43,3 +45,12 @@ There is another utility function, ``set_sim_callback``, for hooking in logic wh
...
+---------------------------------------
+.. literalinclude:: ../../tests/epics/demo/test_demo.py
+   :pyobject: test_sensor_in_plan
+
+
+This test verifies that the sim_sensor behaves as expected within a plan. The plan we use here is a count, which takes a specified number of readings from the sim_sensor. Since we set the repeat to two in this test, the sensor should emit two "event" documents along with "start", "stop" and "descriptor" documents. Finally, we use the helper function ``assert_emitted`` to confirm that the emitted documents match our expectations.
```
callumforrester, line 56: Some extra formatting

```suggestion
This test verifies that the sim_sensor behaves as expected within a plan. The plan we use here is a ``count``, which takes a specified number of readings from the ``sim_sensor``. Since we set the ``repeat`` to two in this test, the sensor should emit two "event" documents along with "start", "stop" and "descriptor" documents. Finally, we use the helper function ``assert_emitted`` to confirm that the emitted documents match our expectations.
```

callumforrester (changes requested):
Docs look good, only a couple of formatting points

src/ophyd_async/core/signal.py
```diff
@@ -272,6 +285,119 @@ def soft_signal_r_and_backend(
...
+    configuration:
+        The expected readings from configurable.
+
+    Notes
+    -----
+    Example usage::
+        await assert_configuration(configurable configuration)
+
+    """
+    await _verify_readings(configurable.read_configuration, configuration)
```
coretl, line 377: Please can we now ditch `_verify_readings` and just write:
```suggestion
    assert await configurable.read_configuration() == configuration
```

src/ophyd_async/core/signal.py
```diff
@@ -272,6 +285,119 @@ def soft_signal_r_and_backend(
...
+    reading:
+        The expected readings from the readable.
+
+    Notes
+    -----
+    Example usage::
+        await assert_reading(readable, reading)
+
+    """
+    await _verify_readings(readable.read, reading)
```
coretl, line 354: Ditto

coretl (changes requested):

src/ophyd_async/core/signal.py
Relm-Arrowny: done, so simple, only took 23 attempt...

callumforrester (approved):
LGTM, thanks!

coretl (changes requested):
Sorry, one last thing

src/ophyd_async/core/signal.py
```diff
@@ -272,6 +284,91 @@ def soft_signal_r_and_backend(
...
+    Notes
+    -----
+    Example usage::
+        await assert_configuration(configurable configuration)
+
+    """
+    assert await configurable.read_configuration() == configuration
+
+
+def assert_emitted(docs: Mapping[str, list[DocumentType]], **numbers: int):
```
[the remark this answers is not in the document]
callumforrester: Good point, @coretl do you think something like this would work?
 
```python
from typing import Union
from event_model import ALL_DOCUMENTS

def assert_emitted(docs: Mapping[str, list[Union[ALL_DOCUMENTS]]], **numbers: int):
````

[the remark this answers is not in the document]
coretl: Yes, but I think that should be a future PR...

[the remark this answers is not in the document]
Relm-Arrowny: changed it back and checked it does not affect anything else.

coretl (approved):

PR #232, opened 2024-04-18
Add logging to signals

### LINKED ISSUE #195: Log everytime a signal is updated or sends an update
To aid in debugging beamline issues it would be good to know the following:

* When a signal gets updated from an external source (e.g. a PV). In this case we will want to know:
    * The name of the signal
    * The name of the PV
    * The value we got
    * The old value (if possible)
    * Metadata about the update e.g. the EPICS timestamp
* When a signal has been written to we would also like to log:
    * The name of the signal
    * The name of the PV
    * The value we got
    * The old value (if possible)
 
It should be obviously easy from the logs to differentiate between these two cases. It should be logged as debug as it will be a lot of info

olliesilvester:
Fixes #195 

Adapts the logging from original `ophyd`, puts debug messages into signal subscriptions, value changing and value reading. Logs include the device name, timestamp, old value and new value

### DIFF
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -194,4 +194,3 @@ class DeviceCollector:
                 "user/explanations/event-loop-choice.html for more info."
             )
         return fut
-        return fut
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import sys
+from logging import LoggerAdapter, getLogger
 from typing import (
     Any,
     Coroutine,
@@ -33,6 +34,9 @@ class Device(HasName):
 
     def __init__(self, name: str = "") -> None:
         self.set_name(name)
+        self.log = LoggerAdapter(
+            getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
+        )
 
     @property
     def name(self) -> str:
@@ -189,4 +193,5 @@ class DeviceCollector:
                 "https://blueskyproject.io/ophyd-async/main/"
                 "user/explanations/event-loop-choice.html for more info."
             )
+        return fut
         return fut
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -64,6 +64,7 @@ class Signal(Device, Generic[T]):
         super().__init__(name)
         self._timeout = timeout
         self._init_backend = self._backend = backend
+        super().__init__(name)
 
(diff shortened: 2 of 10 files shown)


src/ophyd_async/core/signal.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         super().__init__(name)
         self._timeout = timeout
         self._init_backend = self._backend = backend
+        super().__init__(name)
```
coretl, line 57: I think I agree with this being at the bottom of the function, please could you remove the `super` call at the top of the function?

src/ophyd_async/core/device.py
```diff
@@ -190,3 +194,4 @@ def __exit__(self, type_, value, traceback):
                 "user/explanations/event-loop-choice.html for more info."
             )
         return fut
+        return fut
```
coretl, line 197: Duplicate return function

coretl (changes requested):
Please could we switch to using https://pypi.org/project/colorlog/ for the colour part? That should shrink log.py into just the `config_ophyd_async_logging` function using the coloured log formatter from colorlog

src/ophyd_async/core/signal.py
ZLLentz: I'm not sure I agree, is there a specific reason why the timeout and backend need to be set prior to the super call? To me this looks like an out-of-place change in the context of the rest of the PR.

src/ophyd_async/epics/_backend/_p4p.py
```diff
@@ -238,9 +238,8 @@ def __init__(self, datatype: Optional[Type[T]], read_pv: str, write_pv: str):
         self.converter: PvaConverter = DisconnectedPvaConverter()
         self.subscription: Optional[Subscription] = None
 
-    @property
-    def source(self, name: str):
-        return f"pva://{self.read_pv}"
+    def source(self, name: str = "") -> str:
+        return f"ca://{self.read_pv}"
```
ZLLentz, line 242: How is this property -> method API change in the p4p backend related to signal logging?
Why are p4p sources being marked as ca?

ZLLentz (commented):
The overall inclusion of logging and the addition to the test suite are both good. I had a few questions about things that looked odd to me.

src/ophyd_async/epics/_backend/_p4p.py
olliesilvester: Ah thanks for spotting that, the 'ca' was a typo.

olliesilvester:
As discussed with @coretl , this PR has been split off into https://github.com/bluesky/ophyd-async/pull/275 and https://github.com/bluesky/ophyd-async/pull/277

PR #241, opened 2024-04-18
Allow PVI devices to be initialized before connection 

### LINKED ISSUE #194: Allow common pvi devices to be initialized pre-connection and allow `Signal.connect` to take a backend
In https://github.com/bluesky/ophyd-async/pull/185, we allow top level pvi blocks to be initialized during `Panda.__init__`:

https://github.com/bluesky/ophyd-async/blob/1caca7afaadcb365545c757447ac676489316dac/src/ophyd_async/epics/pvi/pvi.py#L304-L319

Then they can be filled with signals in `Panda.connect` and any device intialized in `Panda.__init__` (such as `PandaPcapController`) can be given references to blocks which will contain signals later.

We would like to generalize this so that any devices which depend on these pvi filled signals can be initialized with them, before we've received the PVs through introspection.

To do this we'll need to change `pre_initialize_blocks` and `Signal`.

* `Signal`s should be able to be intialized without a backend, backend will be optional here:
https://github.com/bluesky/ophyd-async/blob/e54c91ce826f76a7398178b93243b980f796c102/src/ophyd_async/core/signal.py#L44-L52
then during `connect` we'll have a kwarg `backend`. If this differs against one passed in during `__init__` we can fail. This will allow signals to be set up and passed to devices before they've been filled through PVI.
* For `DeviceVector`s, we should be able to provide an argument to `pre_initialize_blocks`: `Dict[str, Tuple[int,...]]`. This will be the indices we'll require for a given `DeviceVector` block. If a controller requires `seq[1], seq[2]` then we'll call `pre_initialize_blocks({"seq": (1,2)})` and those blocks, along with any non-`DeviceVector` blocks, will be initalized. Then `seq` can be passed into the controller and the signals can be used after introspection.

noemifrisina:
Fixes #194

### DIFF
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -66,18 +66,18 @@ class Signal(Device, Generic[T]):
         timeout=DEFAULT_TIMEOUT,
         backend: Optional[SignalBackend[T]] = None,
     ):
+        if backend:
+            if self._init_backend and backend != self._init_backend:
+                raise ValueError(
+                    "Backend at connection different from initialised one."
+                )
+            self._init_backend = backend
         if sim:
             self._backend = SimSignalBackend(
                 datatype=self._init_backend.datatype, source=self._init_backend.source
             )
             _sim_backends[self] = self._backend
         else:
-            if backend:
-                if self._init_backend and backend != self._init_backend:
-                    raise ValueError(
-                        "Backend at connection different from initialised one."
-                    )
-                self._init_backend = backend
             self._backend = self._init_backend
             _sim_backends.pop(self, None)
         await self._backend.connect(timeout=timeout)
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -45,7 +45,7 @@ class Signal(Device, Generic[T]):
     """A Device with the concept of a value, with R, RW, W and X flavours"""
 
     def __init__(
-        self, backend: SignalBackend[T], timeout: Optional[float] = DEFAULT_TIMEOUT
+        self, backend: Optional[SignalBackend[T]] = None, timeout: Optional[float] = DEFAULT_TIMEOUT
     ) -> None:
         self._name = ""
         self._timeout = timeout
@@ -58,13 +58,19 @@ class Signal(Device, Generic[T]):
     def set_name(self, name: str = ""):
         self._name = name
 
-    async def connect(self, sim=False, timeout=DEFAULT_TIMEOUT):
+    async def connect(self, sim=False, timeout=DEFAULT_TIMEOUT, backend: Optional[SignalBackend[T]] = None):
         if sim:
             self._backend = SimSignalBackend(
                 datatype=self._init_backend.datatype, source=self._init_backend.source
             )
             _sim_backends[self] = self._backend
         else:
+            if backend:
+                if self._init_backend and backend != self._init_backend:
+                    raise ValueError(
+                        "Backend at connection different from initialised one."
+                    )
+                self._init_backend = backend
             self._backend = self._init_backend
             _sim_backends.pop(self, None)
         await self._backend.connect(timeout=timeout)
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -45,7 +45,9 @@ class Signal(Device, Generic[T]):
     """A Device with the concept of a value, with R, RW, W and X flavours"""
 
     def __init__(
-        self, backend: Optional[SignalBackend[T]] = None, timeout: Optional[float] = DEFAULT_TIMEOUT
+        self,
+        backend: Optional[SignalBackend[T]] = None,
+        timeout: Optional[float] = DEFAULT_TIMEOUT,
     ) -> None:
         self._name = ""
         self._timeout = timeout
@@ -58,7 +60,12 @@ class Signal(Device, Generic[T]):
     def set_name(self, name: str = ""):
         self._name = name
 
-    async def connect(self, sim=False, timeout=DEFAULT_TIMEOUT, backend: Optional[SignalBackend[T]] = None):
+    async def connect(
+        self,
+        sim=False,
+        timeout=DEFAULT_TIMEOUT,
+        backend: Optional[SignalBackend[T]] = None,
+    ):
         if sim:
             self._backend = SimSignalBackend(
                 datatype=self._init_backend.datatype, source=self._init_backend.source
--- a/tests/core/test_signal.py
+++ b/tests/core/test_signal.py
@@ -56,6 +56,13 @@ async def test_set_sim_put_proceeds():
     assert sim_signal._backend.put_proceeds.is_set() is True
 
 
+async def test_signal_connect_fails_with_different_backend_on_connection():
+    sim_signal = Signal(SimSignalBackend(str, "test"))
+
+    with pytest.raises(ValueError):
+        await sim_signal.connect(sim=True, backend=SimSignalBackend(int, "test"))
+
+
 async def time_taken_by(coro) -> float:
     start = time.monotonic()
     await coro


evalott100:
@noemifrisina where are we with this? Should someone else take over?

noemifrisina:
I was waiting for #185 to be merged - and didn't notice it had been. Can pick it up by end of the week if you don't need it sooner...

evalott100:
Is this waiting for a review?

noemifrisina:
> Is this waiting for a review? I'm happy to take over - we need the connect time signal backends pretty soon.

The backend part of this is ready for review. 
The change to the blocks initialisation in `create_children_from_annotations` I'm not sure as it's been awhile. 
I committed some change I  had stashed and merged things from main. I think it would be good to start a review also to see what is still missing?

src/ophyd_async/core/signal.py
```diff
@@ -66,13 +66,21 @@ def __init__(
...
+                )
+            self._initial_backend = backend
         if mock and not isinstance(self._backend, MockSignalBackend):
             # Using a soft backend, look to the initial value
-            self._backend = MockSignalBackend(
-                initial_backend=self._initial_backend,
-            )
+            self._backend = MockSignalBackend(initial_backend=self._initial_backend)
         self.log.debug(f"Connecting to {self.source}")
         await self._backend.connect(timeout=timeout)
```
evalott100, line 85: In the case where no backend was passed on init and it's not in mock, won't `self._backend` still be `None` here?

``` python
await self._backend.connect(timeout=timeout)
```

I think it would be better to keep `_initial_backend` as `None` and change `_backend` here.

``` python
            self._backend = backend
        if mock and not isinstance(self._backend, MockSignalBackend):
            # Perhaps a rename of
            # MockSignalBackend.initial_backend -> MockSignalBackend.real_backend?
            self._backend = MockSignalBackend(initial_backend=self._backend)
        self.log.debug(f"Connecting to {self.source}")
        await self._backend.connect(timeout=timeout)
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -296,7 +301,9 @@ async def fill_pvi_entries(
 
 
 def create_children_from_annotations(
-    device: Device, included_optional_fields: Tuple[str, ...] = ()
+    device: Device,
+    included_optional_fields: Tuple[str, ...] = (),
+    device_vectors: Dict[str, int] = {},
 ):
     """For intializing blocks at __init__ of ``device``."""
```
evalott100, line 308:
```suggestion
    device: Device,
    included_optional_fields: Tuple[str, ...] = (),
    device_vectors: Optional[Dict[str, int]] = None,
):
    """For intializing blocks at __init__ of ``device``."""
```

Avoiding mutable defaults.

evalott100 (changes requested):
A couple more things :pray:

src/ophyd_async/core/signal.py
noemifrisina: Yh good point, changed now

tests/core/test_signal.py
```diff
@@ -50,6 +52,25 @@ async def test_signals_equality_raises():
...
+async def test_set_mock_put_proceeds():
+    sim_signal = Signal(MockSignalBackend(str))
+    await sim_signal.connect(mock=True)
+
+    assert sim_signal._backend.put_proceeds.is_set() is True
+
+    set_mock_put_proceeds(sim_signal, False)
+    assert sim_signal._backend.put_proceeds.is_set() is False
+    set_mock_put_proceeds(sim_signal, True)
+    assert sim_signal._backend.put_proceeds.is_set() is True
```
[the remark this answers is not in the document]
noemifrisina: Ah entirely possible yes... fixing now

tests/core/test_signal.py
```diff
@@ -50,6 +52,29 @@ async def test_signals_equality_raises():
...
+
+
+async def test_signal_connect_fails_if_different_backend_but_same_by_value():
+    initial_backend = MockSignalBackend(str)
+    sim_signal = Signal(initial_backend)
+
+    with pytest.raises(ValueError):
+        await sim_signal.connect(mock=True, backend=MockSignalBackend(str))
+
+    await sim_signal.connect(mock=True, backend=initial_backend)
```
evalott100, line 75: We should skip the `mock` arg in this since we're doing it here, also assert the error string is what we expect.

```suggestion
    with pytest.raises(ValueError) as exc:
        await sim_signal.connect(mock=False, backend=MockSignalBackend(str))
    assert str(exc.value) == "<whatever output string>" 

    await sim_signal.connect(mock=False, backend=initial_backend)
```

tests/core/test_signal.py
```diff
@@ -50,6 +52,29 @@ async def test_signals_equality_raises():
         s1 > 4
```
evalott100, line 53: A test that we can do what we expect too:

``` python
async def test_signal_can_be_given_backend_on_connect():
    sim_signal = Signal()
    backend = MockSignalBackend(int)
    assert sim_signal._backend is None
    await sim_signal.connect(mock=False, backend=backend)
    assert await sim_signal.get_value() == 0
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -296,7 +301,9 @@ async def fill_pvi_entries(
 
 
 def create_children_from_annotations(
```
evalott100, line 303: We definitely need a PVI test that `included_optional_fields` and `device_vectors` args are functioning. They should:
    * check we get the attributes (including device vectors) after `create_children_from_annotations`, but before `connect`
    * check that the device vectors are of the specified size
    * check that after `connect` (mock for test), the signals and devices remain the same by ref that they were after `create_children_from_annotations`

evalott100 (changes requested):
Nice work, two more tests required!

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -179,7 +179,15 @@ def _mock_common_blocks(device: Device, stripped_type: Optional[Type] = None):
...
+                else:
+                    sub_device = DeviceVector(
+                        {
+                            1: getattr(device, device_name, device_cls()),
+                            2: getattr(device, device_name, device_cls()),
+                        }
+                    )
 
                 for sub_device_in_vector in sub_device.values():
                     _mock_common_blocks(sub_device_in_vector, stripped_type=device_cls)
```
evalott100, line 193: You've already checked if `device` has the `device_name` attr.

We should also still recurse on this if if pre-declare-blocks was called:

```suggestion
                if hasattr(device, device_name):
                    sub_device = getattr(device, device_name)
                else:
                    sub_device = DeviceVector(
                        {
                            1: device_cls(),
                            2: device_cls()
                        }
                    )

                for sub_device_in_vector in sub_device.values():
                    _mock_common_blocks(sub_device_in_vector, stripped_type=device_cls)
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -306,13 +316,25 @@ def create_children_from_annotations(
         if is_optional and name not in included_optional_fields:
             continue
         is_device_vector, device_type = _strip_device_vector(device_type)
+        if is_device_vector and not device_vectors:
+            continue
```
evalott100, line 320:
```suggestion
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -306,13 +316,25 @@ def create_children_from_annotations(
...
         if (
-            is_device_vector
+            (is_device_vector and name not in device_vectors.keys())
```
evalott100, line 322:
```suggestion
            (is_device_vector and (not device_vectors or name not in device_vectors))
```

src/ophyd_async/epics/pvi/pvi.py
```diff
@@ -306,13 +316,25 @@ def create_children_from_annotations(
...
 
-        sub_device = device_type()
-        setattr(device, name, sub_device)
-        create_children_from_annotations(sub_device)
+        if is_device_vector:
+            n_device_vector = DeviceVector(
+                {i: device_type() for i in range(1, device_vectors[name] + 1)}
+            )
+            setattr(device, name, n_device_vector)
+            for sub_device in list(n_device_vector.values()):
```
evalott100, line 333: Unnecessary list initialisation.

```suggestion
            for sub_device in n_device_vector.values():
```

tests/core/test_signal.py
```diff
@@ -50,6 +52,38 @@ async def test_signals_equality_raises():
...
+    sim_signal = Signal(MockSignalBackend(str))
+
+    with pytest.raises(ValueError):
+        await sim_signal.connect(mock=True, backend=MockSignalBackend(int))
+
+    with pytest.raises(ValueError):
+        await sim_signal.connect(mock=True, backend=SoftSignalBackend(str))
+
+    with pytest.raises(ValueError):
+        await sim_signal.connect(mock=False, backend=MockSignalBackend(str))
```
evalott100, line 73: This is the same test case as below

evalott100 (changes requested):

tests/epics/test_pvi.py
```diff
@@ -144,3 +144,51 @@ async def test_device_create_children_from_annotations(
...
+
+
+@pytest.fixture
+def pvi_test_device_with_device_vectors_t():
+    """A fixture since pytest discourages init in test case classes"""
+
+    class TestDevice(Block2):
+        def __init__(self, prefix: str, name: str = ""):
+            self._prefix = prefix
+            create_children_from_annotations(self, device_vectors={"device_vector": 2})
```
evalott100, line 156: Add an `optional_signal_rw: Optional[SignalRW[int]]` to `Block2` and check
```suggestion
            create_children_from_annotations(
                self,
                included_optional_fields=("optional_signal_rw",),
                device_vectors={"device_vector": 2}
            )
                
```

It doesn't seem like we're testing this functionality currently.

tests/epics/test_pvi.py
```diff
@@ -144,3 +144,51 @@ async def test_device_create_children_from_annotations(
...
+
+    yield TestDevice
+
+
+async def test_device_create_children_from_annotations_with_device_vectors(
+    pvi_test_device_with_device_vectors_t,
+):
+    async with DeviceCollector(mock=True):
+        device = pvi_test_device_with_device_vectors_t("PREFIX:", name="test_device")
+
```
evalott100, line 176:
```suggestion
    device = pvi_test_device_with_device_vectors_t("PREFIX:", name="test_device")
```

I believe from the rest of the code you probably don't want device to be connected here, only initialized?

evalott100 (changes requested):

tests/epics/test_pvi.py
noemifrisina: It doesn't look like that functionality has ever been added at all tbh, or at least not for signals, which would get caught by this if in any case and not initialised in any case...
```if (
            (is_device_vector and (not device_vectors or name not in device_vectors))
            or ((origin := get_origin(device_type)) and issubclass(origin, Signal))
            or (isclass(device_type) and issubclass(device_type, Signal))
        ):
```
I thought from looking at the code that signals should not pre-initialised anyway? Or did I get that wrong?

evalott100: > I thought from looking at the code that signals should not pre-initialised anyway? Or did I get that wrong?

Ah yes, my mistake. It would be
``` python
                included_optional_fields=("optional_device_with_a_signal",),
```

evalott100 (approved):
Fab, thanks very much!

Issue #214, opened 2024-04-15
Create PathProvider and FilenameProviders that plugin to it


jwlodek:
Depends on #209 .

We use UUID filenames at NSLS-II, and we typically use y/m/d w/ some root directory for directory names.

abbiemery:
Closed as done by #245 .

PR #245, opened 2024-04-18
Split out FilenameProvider, rename DirectoryProvider to PathProvider

jwlodek:
Creating this as a draft PR first to get some feedback before I apply fixes to the filewriters to account for this change.

Addresses #214


src/ophyd_async/core/_providers.py
```diff
@@ -17,41 +20,105 @@ class DirectoryInfo:
...
-        return self._directory_info
+
+class YMDDirectoryProvider(DirectoryProvider):
+    def __init__(
+        self,
+        filename_provider: FilenameProvider,
+        directory_path: Union[str, Path],
+        filename_prefix: str = "",
+        filename_suffix: str = "",
+        create_dir_depth: int = 0,
```
coretl, line 105: Probably want to default this to `-3` if you want Y/M/D directories to be made?

src/ophyd_async/core/_providers.py
```diff
@@ -17,41 +20,105 @@ class DirectoryInfo:
...
+            self._directory_path = Path(directory_path)
+        self._resource_dir = resource_dir
+        self._prefix = filename_prefix
+        self._suffix = filename_suffix
+        self._create_dir_depth = create_dir_depth
+
+    def __call__(self, device_name=None) -> DirectoryInfo:
+        return DirectoryInfo(
+            root=self._directory_path,
+            resource_dir=self._resource_dir,
```
coretl, line 92: @DiamondJoseph do we need control about which bit of the path is root and which bit is the resource dir?

src/ophyd_async/core/_providers.py
```diff
@@ -17,41 +20,105 @@ class DirectoryInfo:
...
+        self._static_filename = filename
+
+    def __call__(self, prefix="", suffix="", device_name=None) -> str:
+        return f'{prefix}{self._static_filename}{suffix}'
+
+
+class DeviceNameFilenameProvider(FilenameProvider):
+
+    def __call__(self, prefix="", suffix="", device_name=None) -> str:
+        return f'{prefix}{device_name}{suffix}'
```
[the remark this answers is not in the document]
jwlodek: Is `WritesStreamAssets` a part of ophyd-async or dodal? I quite like using the device itself instead of a bunch of args here, then you can add any number of modifiers to the filename based on device specific attributes (even like naming files based on a signal reading perhaps?)

src/ophyd_async/core/_providers.py
jwlodek: I think that makes sense, will do.

src/ophyd_async/core/_providers.py
DiamondJoseph: Currently we're writing root = /data/YYYY/visit, resource_dir = ".", but we probably want it to be resource_dir=dataCollection imminently, so all our h5s are neatly in a folder with the same name as the nexus file they are linked from.

coretl: This is probably the job of the directory provider then to decide which is root and which is resource dir

src/ophyd_async/core/_providers.py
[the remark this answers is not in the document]
coretl: Do we need anything from the Device apart from name?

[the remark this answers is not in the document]
DiamondJoseph: Just name <beamline>-<device.name>-<dataCollectionId>.h5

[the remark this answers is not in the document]
coretl: So just passing `device_name` rather than `device` would be sufficient? Or do we think we will need anything else in the future from the object?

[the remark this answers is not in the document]
DiamondJoseph: If we're passing in a Device, all we can reliably expect is device.name and device.children AFAIK? device_name should be fine I guess? Each child that needs the DirectoryProvider will have the ability to call the NameProvider themselves

jwlodek:
Before I continue, I'd like to hear if you have any suggestions.

src/ophyd_async/core/_providers.py
coretl: Had a conversation with @danielballan and @skarakuzu who say that root is going away, so I suggest we don't expose it in the `PathProvider` interface at all, then hardcode it to `/` in the old `StreamResource` until June when we can get rid of it completely

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +20,155 @@ class DirectoryInfo:
...
+
+        self._current_value += self._increment
+        return filename
+
+
+class StaticPathProvider(PathProvider):
     def __init__(
         self,
+        filename_provider: FilenameProvider,
         directory_path: Union[str, Path],
```
DiamondJoseph, line 112: Must: I'm surprised this isn't raising an exception in the CI, directory_path is str | Path but then it's passed into the PathInfo as a Path without any handling.

src/ophyd_async/panda/writers/_hdf_writer.py
```diff
@@ -119,16 +119,17 @@ async def open(self, multiplier: int = 1) -> Dict[str, DataKey]:
...
         await asyncio.gather(
             self.panda_device.data.hdf_directory.set(
                 str(info.root / info.resource_dir)
             ),
             self.panda_device.data.hdf_file_name.set(
-                f"{info.prefix}{self.panda_device.name}{info.suffix}",
+                info.filename,
             ),
             self.panda_device.data.num_capture.set(0),
+            # TODO: Set create_dir_depth once available
```
DiamondJoseph, line 132: Make an issue for this and link it here

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +20,155 @@ class DirectoryInfo:
...
+class StaticFilenameProvider(FilenameProvider):
+    def __init__(self, filename: str):
+        self._static_filename = filename
+
+    def __call__(self) -> str:
+        return self._static_filename
+
+
+class UUIDFilenameProvider(FilenameProvider):
+    def __init__(self, uuid_call_func: callable = uuid.uuid4):
```
DiamondJoseph, line 56: May be able to make use of a ParamSpec here to have Callable[P], *calling_args: Optional[P] = None

Not sure how complex it may be however

Then __call__ = callable(calling_args) if calling_args else callable()

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +20,155 @@ class DirectoryInfo:
...
+    :param filename: Base filename to use generated by FilenameProvider
+    :param create_dir_depth: Optional depth of dirs to create if they do not exist
     """
 
     root: Path
     resource_dir: Path
-    prefix: Optional[str] = ""
-    suffix: Optional[str] = ""
+    filename: str
+    create_dir_depth: Optional[int] = -3
```
DiamondJoseph, line 30: Does a negative number have specific meaning? Must document: negatives, 0, positives, None. Is None the same as 0, if so can be non-Optional but default to 0.

coretl (commented):
I can think of 2 usecases for us:
- For testing we would like all the files for a given detector to be placed in a predictable `tmp_path`. I guess we could either use a different `StaticPathProvider` and `AutoIncrementingFilenameProvider` for each detector, or we could pass down `device_name` to all `FilenameProvider`s and allow `AutoIncrementingFilenameProvider` to use that in its format string.
- In production we will probably make a `PathProvider` that doesn't use a `FilenameProvider` at all, just generates the names according to DLS conventions

So feel free to delete all the `prefix` and `suffix` code that you wouldn't use yourself.

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +20,155 @@ class DirectoryInfo:
...
-            suffix=filename_suffix,
+        self._filename_provider = filename_provider
+        self._directory_path = directory_path
+        self._resource_dir = resource_dir
+        self._create_dir_depth = create_dir_depth
+        self._prefix = filename_prefix
+        self._suffix = filename_suffix
+
+    def __call__(self, device_name=None) -> PathInfo:
+        filename = f"{self._prefix}{self._filename_provider()}{self._suffix}"
```
DiamondJoseph, line 126: filenameprovider should be handling prefixing/suffixing, StaticFileNameProvider -> StaticFormatFileNamer? init(prefix: Optional[str], suffix: Optional[str]) ?

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +20,155 @@ class DirectoryInfo:
...
+            )
+        else:
+            resource_dir = os.path.join(
+                device_name,
+                str(current_date.year),
+                str(current_date.month),
+                str(current_date.day),
+            )
+
+        filename = f"{self._prefix}{self._filename_provider()}{self._suffix}"
```
DiamondJoseph, line 165: As above, prefix, suffix should be on Filename Provider

DiamondJoseph (changes requested):

src/ophyd_async/core/_providers.py
jwlodek: Should I keep it as `resource_dir` or rename to `resource_path`?

DiamondJoseph: I oppose that change: our NeXus writing assumes that resource_dir is a relative path from the visit directory to do relative linking of h5 files.

mrakitin:
I'll review once rebased :)

src/ophyd_async/core/_providers.py
danielballan: I know that Mark Rivers' AD uses a negative number here. At some point, @jwlodek, I think we brainstormed how best to handle this and agreed that it might be best to accept positive or negative here and use `abs` to pass the right thing to EPICS.

I would vote for making this not `Optional`.

jwlodek: You can see my comment on the PR in the PandA IOC about how this number works: https://github.com/PandABlocks/PandABlocks-ioc/pull/102#issuecomment-2117552752

jwlodek:
@mrakitin @danielballan @coretl I think this is ready for review and potentially merge

src/ophyd_async/core/_providers.py
DiamondJoseph: This is now `create_dir_depth: int = 0` which looks correct with the ad fileplugin docs.

src/ophyd_async/core/_providers.py
mrakitin: @DiamondJoseph, what is ParamSpec?

mrakitin: Found that: https://docs.python.org/3/library/typing.html#typing.ParamSpec.

mrakitin (commented):
Looks great to me! Proposed some non-blocking suggestions, most of them are optional.

mrakitin:
I have some fixes from yesterday's and today's work at HEX in https://github.com/jwlodek/ophyd-async/pull/1. Please merge, then we can rebase.

src/ophyd_async/epics/areadetector/writers/_hdffile.py
mrakitin, line 1: Not sure if it's a leftover from my rebasing...
Is it all now implemented in https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/areadetector/writers/general_hdffile.py?

src/ophyd_async/epics/areadetector/writers/nd_file_hdf.py
```diff
@@ -37,4 +37,7 @@ def __init__(self, prefix: str, name="") -> None:
         self.capture = epics_signal_rw_rbv(bool, prefix + "Capture")
         self.flush_now = epics_signal_rw(bool, prefix + "FlushNow")
         self.xml_file_name = epics_signal_rw_rbv(str, prefix + "XMLFileName")
+        self.array_size0 = epics_signal_r(int, prefix + "ArraySize0")
+        self.array_size1 = epics_signal_r(int, prefix + "ArraySize1")
```
mrakitin, line 41: A note for the future - we may need all 3 dimensions in general (width, height, depth). Consider adding `ArraySize2`.

src/ophyd_async/panda/writers/_panda_hdf_file.py
mrakitin, line 1: I guess we'll also need https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/areadetector/writers/general_hdffile.py, not this file, right?

mrakitin:
A comment https://github.com/bluesky/ophyd-async/pull/245#pullrequestreview-2173340570 was intended to be a part of my review. Sorry for posting it separately.

src/ophyd_async/panda/writers/_panda_hdf_file.py
```diff
@@ -0,0 +1,57 @@
+from dataclasses import dataclass
```
DiamondJoseph, line 1: As above RE: multiple implementations of HDFFile

tests/conftest.py
```diff
@@ -124,5 +128,21 @@ async def inner_coroutine():
...
+
+    return create_static_dir_provider_given_fp
+
+
+@pytest.fixture
+def static_path_provider(
+    static_path_provider_factory: callable,
+    static_filename_provider: FilenameProvider,
+):
+    return static_path_provider_factory(static_filename_provider)
```
DiamondJoseph, line 148:
```suggestion
@pytest.fixture
def static_path_provider(tmp_path: Path, static_filename_provider: StaticFilenameProvider):
        return StaticPathProvider(static_filename_provider, tmp_path)
```

Do you need this factory function elsewhere? It doesn't seem to be required.

DiamondJoseph (changes requested):

DiamondJoseph:
I think all HDFWriting should be being done with the unified HDFFile now?
https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/areadetector/writers/general_hdffile.py

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +21,199 @@ class DirectoryInfo:
...
+        self._static_filename = filename
+
+    def __call__(self) -> str:
+        return self._static_filename
+
+
+class UUIDFilenameProvider(FilenameProvider):
+    def __init__(
+        self, uuid_call_func: Callable = uuid.uuid4, uuid_call_args: Optional[List] = []
+    ):
```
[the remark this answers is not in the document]
DiamondJoseph: Or make `call_func: Callable[[P], str], call_args: P` where `P = ParamSpec("P")`

src/ophyd_async/panda/writers/_panda_hdf_file.py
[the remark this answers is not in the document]
evalott100: Same with areadetector.

src/ophyd_async/sim/sim_pattern_generator.py
```diff
@@ -17,15 +22,16 @@ def __init__(
         config_sigs: Sequence[AsyncReadable] = [],
```
evalott100, line 22: Seeing unnoticed mutable defaults in the codebase during this review :hand_over_mouth:

```suggestion
        config_sigs: Sequence[AsyncReadable] = (),
```

evalott100 (changes requested):
Lots there, good work all!

src/ophyd_async/epics/areadetector/writers/nd_file_hdf.py
mrakitin: https://github.com/bluesky/ophyd-async/issues/454 to address it later.

src/ophyd_async/epics/areadetector/drivers/kinetix_driver.py
```diff
@@ -15,6 +15,7 @@ class KinetixReadoutMode(str, Enum):
     sensitivity = 1
     speed = 2
     dynamic_range = 3
+    sub_electron = 4
```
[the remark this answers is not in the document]
mrakitin: https://github.com/bluesky/ophyd-async/issues/456

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +21,199 @@ class DirectoryInfo:
...
+        self._uuid_call_func = uuid_call_func
+        self._uuid_call_args = uuid_call_args
```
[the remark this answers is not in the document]
mrakitin: Implemented via https://github.com/bluesky/ophyd-async/pull/245/commits/0d9d9eff344b78b95cef35d9b1ff4afb8e8ac412.

src/ophyd_async/core/_providers.py
```diff
@@ -17,43 +21,199 @@ class DirectoryInfo:
...
+        directory_path: Path,
+        create_dir_depth: int = -3,  # Default to -3 to create YMD dirs
+        device_name_as_base_dir: bool = False,
+    ) -> None:
+        self._filename_provider = filename_provider
+        self._directory_path = Path(directory_path)
+        self._create_dir_depth = create_dir_depth
+        self._device_name_as_base_dir = device_name_as_base_dir
+
+    def __call__(self, device_name=None) -> PathInfo:
```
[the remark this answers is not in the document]
mrakitin: Addressed via 2e305e00.

src/ophyd_async/panda/writers/_panda_hdf_file.py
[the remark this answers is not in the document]
mrakitin: Done via https://github.com/bluesky/ophyd-async/pull/245/commits/a26d6245a2e8b55c0a53435d18a8aa5bc8fd193d.

src/ophyd_async/panda/writers/_panda_hdf_file.py
mrakitin: Cleaned up via https://github.com/bluesky/ophyd-async/pull/245/commits/a26d6245a2e8b55c0a53435d18a8aa5bc8fd193d.

src/ophyd_async/epics/areadetector/writers/_hdffile.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
mrakitin: Cleaned up via https://github.com/bluesky/ophyd-async/pull/245/commits/a26d6245a2e8b55c0a53435d18a8aa5bc8fd193d.

evalott100 (approved):

PR #247, opened 2024-04-19
Add name argument to Signal, epics_signal_* and soft_signal_*

jsouter:
Allows names to be passed to Signal on init as the last argument, empty string by default.

### DIFF
--- a/src/ophyd_async/core/signal.py
+++ b/src/ophyd_async/core/signal.py
@@ -50,16 +50,9 @@ class Signal(Device, Generic[T]):
         timeout: Optional[float] = DEFAULT_TIMEOUT,
         name: str = "",
     ) -> None:
-        self._name = name
+        super().__init__(name)
         self._timeout = timeout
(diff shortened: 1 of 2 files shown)


src/ophyd_async/core/signal.py
```diff
@@ -45,17 +45,20 @@ class Signal(Device, Generic[T]):
...
         self._timeout = timeout
         self._init_backend = self._backend = backend
 
     @property
     def name(self) -> str:
         return self._name
 
-    def set_name(self, name: str = ""):
+    def set_name(self, name: str):
         self._name = name
```
coretl, line 62: Device already has `set_name` and `name`, so you can delete these and do `super().__init__(name)`

coretl (approved):

src/ophyd_async/core/signal.py
jsouter: Okay, have made that change

PR #251, opened 2024-04-19
Create mock signal backend

### LINKED ISSUE #94: Rename Sim -> Mock and add SoftSignalBackend
Currently, you can set up ophyd-async devices in "sim" mode, which actually just mock out the control layer underneath (i.e. EPICs) to let someone test the stuff above it, i.e. plans and filewriting, works as expected. As such, we need to rename the "sim" interface to "mock" to prevent confusion.

evalott100:
Closes #94

### DIFF
--- a/src/ophyd_async/core/mock_signal_backend.py
+++ b/src/ophyd_async/core/mock_signal_backend.py
@@ -1,0 +1,66 @@
+import asyncio
+from typing import Optional
+from unittest.mock import MagicMock
+
+from bluesky.protocols import Descriptor, Reading
+
+from ophyd_async.core.signal_backend import SignalBackend
+from ophyd_async.core.soft_signal_backend import SoftSignalBackend
+from ophyd_async.core.utils import DEFAULT_TIMEOUT, ReadingValueCallback, T
+
+
+class MockSignalBackend(SoftSignalBackend[T]):
+    def __init__(
+        self,
+        datatype: Optional[type[T]],
+        initial_value: Optional[T] = None,
+        init_backend: Optional[SignalBackend] = None,
+    ) -> None:
+        super().__init__(datatype, initial_value=initial_value)
+        self.mock = MagicMock()
+        self.init_backend = init_backend
+
+        self.put_proceeds = asyncio.Event()
+        self.put_proceeds.set()
+
+    def source(self, name: str) -> str:
+        self.mock.source(name)
+        if self.init_backend:
+            return f"mock+{self.init_backend.source(name)}"
+        return f"mock+{name}"
+
+    async def connect(self, timeout: float = DEFAULT_TIMEOUT) -> None:
+        self.mock.connect(timeout=timeout)
+        await super().connect(timeout=timeout)
+
+    async def put(self, value: Optional[T], wait=True, timeout=None):
+        self.mock.put(value, wait=wait, timeout=timeout)
+        await super().put(value, wait=wait, timeout=timeout)
+
+        if wait:
+            await asyncio.wait_for(self.put_proceeds.wait(), timeout=timeout)
+
+    def set_value(self, value: T):
+        self.mock.set_value(value)
+        super().set_value(value)
+
+    async def get_descriptor(self, source: str) -> Descriptor:
+        self.mock.get_descriptor(source)
+        return await super().get_descriptor(source)
+
+    async def get_reading(self) -> Reading:
+        self.mock.get_reading()
+        return await super().get_reading()
+
+    async def get_value(self) -> T:
+        self.mock.get_value()
+        return await super().get_value()
+
+    async def get_setpoint(self) -> T:
+        """For a soft signal, the setpoint and readback values are the same."""
+        self.mock.get_setpoint()
+        return await super().get_setpoint()
+
+    def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
+        self.mock.set_callback(callback)
+        super().set_callback(callback)
--- a/src/ophyd_async/core/mock_signal_backend.py
+++ b/src/ophyd_async/core/mock_signal_backend.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Optional
+from typing import Optional, Type
 from unittest.mock import MagicMock
 
 from bluesky.protocols import Descriptor, Reading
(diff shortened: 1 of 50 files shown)


src/ophyd_async/core/mock_signal_utils.py
```diff
@@ -0,0 +1,183 @@
...
+class _SetValuesIterator:
+    # Garbage collected by the time __del__ is called unless we put it as a
+    # global attrbute here.
+    require_all_consumed: bool = False
+
+    def __init__(
+        self,
+        signal: Signal,
+        *values: List[Any],
+        repeat: bool = False,
```
coretl, line 81: I can't actually think of a usecase for `repeat` now, as we will be consuming from within a for loop we will always repeat the same value. Please can you delete it?

src/ophyd_async/core/mock_signal_utils.py
```diff
@@ -0,0 +1,183 @@
...
+    """Allow or block a put with wait=True from proceeding"""
+    backend = _get_mock_signal_backend(signal)
+
+    if proceeds:
+        backend.put_proceeds.set()
+    else:
+        backend.put_proceeds.clear()
+
+
+def set_mock_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> None:
```
coretl, line 37: Deletion and name changes as per https://github.com/bluesky/ophyd-async/issues/94

src/ophyd_async/core/mock_signal_utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+@asynccontextmanager
+async def block_puts_during(*signals: List[Signal]):
+    for signal in signals:
+        set_mock_put_proceeds(signal, False)
+    yield
+    for signal in signals:
+        set_mock_put_proceeds(signal, True)
+
+    await asyncio.sleep(_A_BIT)
```
coretl, line 56: Please can we get rid of this sleep and put it in the tests that need it instead, I'm happy with tests having a sleep in them, but not the core code

src/ophyd_async/core/mock_signal_utils.py
```diff
@@ -0,0 +1,183 @@
...
+    ----------
+    signal:
+        A signal with a `MockSignalBackend` backend.
+    callback:
+        The callback to call when the backend is put to during the context.
+    """
+    backend = _get_mock_signal_backend(signal)
+
+    # Save the original put method
+    original_put = backend.put
```
coretl, line 170: Could we do `backend.mock.put.side_effect = callback` instead?

src/ophyd_async/core/mock_signal_utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+@asynccontextmanager
+async def set_signals_after(signals_to_value: Dict[Signal, Any]):
+    async with block_puts_during(*signals_to_value.keys()):
+        yield
+    for signal, value in signals_to_value.items():
+        await signal.set(value)
+
+
+def assert_mock_put_called_with(signal: Signal, value: Any, wait=ANY, timeout=ANY):
```
coretl, line 67: Please can we also have `reset_mock_put_calls`?

src/ophyd_async/core/mock_signal_backend.py
```diff
@@ -0,0 +1,52 @@
...
+class MockSignalBackend(SoftSignalBackend[T]):
+    def __init__(
+        self, datatype: type[T] | None, initial_value: T | None = None
+    ) -> None:
+        super().__init__(datatype, initial_value=initial_value)
+        self.mock = MagicMock()
+
+    def source(self, name: str) -> str:
+        self.mock.source(name)
+        return f"mock+{name}"
```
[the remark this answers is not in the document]
evalott100: I was thinking something like this, but we'd have to change the `Signal.source` to use the `init_backend` if the backend is a `MockSignalBackend`... I'll do it to see how we like it.

src/ophyd_async/core/mock_signal_utils.py
evalott100: You don't think it's a little nicer to have a dedicated callback in `MockSignalBackend`?

``` python
    def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
        if callback:
            assert not self.callback, "Cannot set a callback when one is already set"
            reading: Reading = self.converter.reading(
                self._value, self._timestamp, self._severity
            )
            callback(reading, self._value)
        self.callback = callback
```

This allows for easier reading metadata in tests:
``` python
    assert calls == [
        (
            {
                "alarm_severity": 0,
                "timestamp": ANY,
                "value": 0.0,
            },
            0.0,
        ),
    ]
```

src/ophyd_async/core/mock_signal_utils.py
evalott100: https://github.com/bluesky/ophyd-async/issues/94#issuecomment-2088428517

tests/epics/motion/test_motor.py
coretl, line 1: The test logic seems to have changed here, was that required? If not, then please can you put it back?

coretl (changes requested):

src/ophyd_async/core/signal.py
```diff
@@ -65,13 +64,20 @@ def __init__(
...
 
-    async def connect(self, sim=False, timeout=DEFAULT_TIMEOUT):
-        if sim:
-            self._backend = SimSignalBackend(datatype=self._init_backend.datatype)
-            _sim_backends[self] = self._backend
-        else:
-            self._backend = self._init_backend
-            _sim_backends.pop(self, None)
+    async def connect(self, mock=False, timeout=DEFAULT_TIMEOUT):
+        if mock and not isinstance(self._backend, MockSignalBackend):
```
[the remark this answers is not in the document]
evalott100: Because if the user has done a `SignalRW(MockSignalBackend(...), ...)` we don't need to replace their `MockSignalBackend`.

tests/epics/motion/test_motor.py
evalott100: Previously, the `AsyncStatus` for `sim_motor.set` wasn't being awaited. After the set point has been set, I wanted to check, and then await the task.

src/ophyd_async/core/mock_signal_backend.py
```diff
@@ -0,0 +1,87 @@
...
+        datatype: Optional[Type[T]] = None,
+        initial_backend: Optional[SignalBackend[T]] = None,
+    ) -> None:
+        if isinstance(initial_backend, MockSignalBackend):
+            raise ValueError("Cannot make a MockSignalBackend for a MockSignalBackends")
+
+        self.initial_backend = initial_backend
+
+        if not datatype:
+            datatype = self.initial_backend.datatype if self.initial_backend else None
```
coretl, line 24:
```suggestion
        if datatype is not None:
            assert self.initial_backend, "Must supply either initial_backend or datatype"
            datatype = self.initial_backend.datatype
```

src/ophyd_async/core/mock_signal_backend.py
```diff
@@ -0,0 +1,87 @@
...
+        self.datatype = datatype
+
+        if not isinstance(self.initial_backend, SoftSignalBackend):
+            # If the backend is a hard signal backend, or not provided,
+            # then we create a soft signal to mimick it
+
+            self.soft_backend = SoftSignalBackend(datatype=datatype)
+        else:
+            self.soft_backend = initial_backend
```
coretl, line 33:
```suggestion
        if self.initial_backend and isinstance(self.initial_backend, SoftSignalBackend):
            # We have a SoftSignalBackend already, so use it
            self.soft_backend = initial_backend
        else:
            # Our initial backend is not soft, so make one to hold our value
            self.soft_backend = SoftSignalBackend(datatype=datatype)
```

src/ophyd_async/core/mock_signal_backend.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.mock = MagicMock()
+
+        self.put_proceeds = asyncio.Event()
+        self.put_proceeds.set()
+
+    @property
+    def _value(self) -> T:
+        return self.soft_backend._value
```
coretl, line 42: Where do we use this?

src/ophyd_async/core/soft_signal_backend.py
```diff
@@ -122,17 +116,14 @@ def __init__(
...
         self._initial_value = initial_value
-        self.put_proceeds = asyncio.Event()
-        self.put_proceeds.set()
         self.callback: Optional[ReadingValueCallback[T]] = None
 
     def source(self, name: str) -> str:
         return f"soft://{name}"
 
     async def connect(self, timeout: float = DEFAULT_TIMEOUT) -> None:
-        self.converter = make_converter(self.datatype)
```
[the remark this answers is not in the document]
evalott100: I rebased and did the following:

``` python
    def __init__(
        self,
        datatype: Optional[Type[T]],
        initial_value: Optional[T] = None,
    ) -> None:
        self.datatype = datatype
        self._initial_value = initial_value
        self.converter: SoftConverter = make_converter(datatype)
        if self._initial_value is None:
            self._initial_value = self.converter.make_initial_value(self.datatype)
        else:
            self._initial_value = self.converter.write_value(self._initial_value)

        self.callback: Optional[ReadingValueCallback[T]] = None
        self._severity = 0
        self.set_value(self._initial_value)

  
    async def connect(self, timeout: float = DEFAULT_TIMEOUT) -> None:
        pass
```

src/ophyd_async/core/signal.py
```diff
@@ -62,16 +61,15 @@ def __init__(
...
+            # Using a soft backend, look to the initial value
+            self._backend = MockSignalBackend(
+                initial_backend=self._initial_backend,
+            )
```
coretl, line 72: We need the else clause!
```suggestion
            )
        else:
            self._backend = self._initial_backend
```

src/ophyd_async/core/signal.py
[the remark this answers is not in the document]
coretl: Ok, but I don't think there's a problem with replacing it...

src/ophyd_async/core/signal.py
```diff
@@ -255,47 +253,32 @@ def trigger(self, wait=True, timeout=USE_DEFAULT_TIMEOUT) -> AsyncStatus:
...
     """
-    backend = SimSignalBackend(datatype, initial_value)
+    backend = SoftSignalBackend(datatype, initial_value)
     signal = SignalR(backend, name=name)
-    return (signal, backend)
+
+    def backend_put(value: T) -> None:
+        signal._backend.set_value(value)
+
+    return (signal, backend_put)
```
coretl, line 281: Can now be simply:
```suggestion
    return (signal, backend.set_value)
```

coretl (changes requested):
Almost there...

src/ophyd_async/core/mock_signal_backend.py
coretl: Ignore this

src/ophyd_async/core/signal.py
coretl: And this

tests/epics/motion/test_motor.py
evalott100: test_motor.py back to how it was!

coretl (approved):

PR #252, opened 2024-04-19
Pyright

gilesknap:
This PR re-adopts the copier, making sure that pyproject.toml is correct for pyright adoption.

It also has started to fix pyright errors in the source. BUT there seem to be many errors in here that someone with a better understanding of the code should look at.

- For the moment I have removed the tests from pyright checks - this should also eventually be restored.
- There are many cases of incompatible function signatures in derived classes that should be properly addressed
- Also many argument / parameter type disagreements
- There is too much to even just apply `type: ignore` to everything at 5PM on Friday and I also don't think that is the right way to go.

@coretl I'll let you decide what to do with this branch, I'm OK to carry on with it next week but imagine it is a few days work to do right.


.devcontainer/devcontainer.json
coretl, line 1: These are intentional, so we get reasonable settings without a `.vscode` dir

catalog-info.yaml
coretl, line 1: This should also be deleted

tests/epics/demo/test_demo.py
```diff
@@ -170,20 +170,20 @@ async def test_mover_disconncted():
     with pytest.raises(NotConnected):
         async with DeviceCollector(timeout=0.1):
             m = demo.Mover("ca://PRE:", name="mover")
-    assert m.name == "mover"
+        assert m.name == "mover"
```
coretl, line 173: These should be in their original position

coretl (commented):
I think this is worth spending some time one, but I think we should spend half an hour going through the errors in person

tests/epics/demo/test_demo.py
gilesknap: pyright did not like that. it said that m may not be instantiated. 
Looking at it again I see that this is not going to behave as required with my change (because the expected exception would jump over final assert)
pyright is correct in that if the DeviceCollector constructor throws NotConnected then m would not be set.

But this is rather pedantic and should probably be ignored.

docs/examples/epics_demo.py
```diff
@@ -1,9 +1,9 @@
 # Import bluesky and ophyd
 import matplotlib.pyplot as plt
-from bluesky import RunEngine
 from bluesky.callbacks.best_effort import BestEffortCallback
 from bluesky.plan_stubs import mov, movr, rd  # noqa
 from bluesky.plans import grid_scan  # noqa
+from bluesky.run_engine import RunEngine
```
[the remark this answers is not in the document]
gilesknap: Here I was just doing what pyright asked me to -  It seemed that it did not like the module level declaration in __init__.py.

tests/epics/demo/test_demo.py
gilesknap: or pre declare m.

.devcontainer/devcontainer.json
gilesknap: noted. I feel like was set up to fail on this task!

catalog-info.yaml
gilesknap: right

gilesknap:
> I think this is worth spending some time one, but I think we should spend half an hour going through the errors in person

OK lets do this. But you realize 30 mins is less than 18 seconds per issue?

PR #255, opened 2024-04-22
Remove extraneous Pilatus constructor arg

### LINKED ISSUE #254: Remove extraneous prefix from Pilatus constructor
Required for dodal instantiation of device, to be moved into dodal specific handling.

DiamondJoseph:
Fixes #254


callumforrester (approved):
Will this work with `device_instantiation` in dodal?

PR #275, opened 2024-05-02
Add logging messages to signals

olliesilvester:
Fixes part of #https://github.com/bluesky/ophyd-async/issues/195

### DIFF
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -45,9 +45,7 @@ class Device(HasName):
     @lru_cache(1)
     def log(self):
         return LoggerAdapter(
-            LoggerAdapter(
-                getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
-            )
+            getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
         )
 
     def children(self) -> Iterator[Tuple[str, Device]]:
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import sys
-from functools import lru_cache
+from functools import cached_property
 from logging import LoggerAdapter, getLogger
 from typing import (
     Any,
@@ -41,8 +41,7 @@ class Device(HasName):
         """Return the name of the Device"""
         return self._name
 
-    @property
-    @lru_cache(1)
+    @cached_property
     def log(self):
         return LoggerAdapter(
             getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
@@ -61,6 +60,9 @@ class Device(HasName):
         name:
             New name to set
         """
+        if hasattr(self, "log"):
+            del self.log
+
         self._name = name
         for attr_name, child in self.children():
             child_name = f"{name}-{attr_name.rstrip('_')}" if name else ""
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -60,6 +60,8 @@ class Device(HasName):
         name:
             New name to set
         """
+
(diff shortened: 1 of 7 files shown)


src/ophyd_async/epics/_backend/_p4p.py
```diff
@@ -238,7 +238,6 @@ def __init__(self, datatype: Optional[Type[T]], read_pv: str, write_pv: str):
         self.converter: PvaConverter = DisconnectedPvaConverter()
         self.subscription: Optional[Subscription] = None
 
-    @property
```
olliesilvester, line 241: I don't think this should ever have been a property?

coretl: nope, @jsouter's changes turned this into a method, so not sure how it ever worked with the property decorator?

tests/core/test_signal.py
```diff
@@ -39,6 +42,24 @@ async def connect(self, sim=False, timeout=DEFAULT_TIMEOUT):
...
+    def source(self) -> str:
+        return "source"
+
+    async def connect(self):
+        pass
+
+
+@pytest.fixture
+def mock_signal_rw():
+    return
```
coretl, line 60: Empty fixture? Delete?

src/ophyd_async/core/device.py
```diff
@@ -33,6 +34,9 @@ class Device(HasName):
 
     def __init__(self, name: str = "") -> None:
         self.set_name(name)
+        self.log = LoggerAdapter(
+            getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
```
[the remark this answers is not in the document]
ZLLentz: LoggerAdapter doesn't normally accept a lamba function, but we could subclass LoggerAdapter and pull the name at logtime.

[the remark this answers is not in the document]
coretl: Or not use the logging adapter and pass it into each `self.log` call...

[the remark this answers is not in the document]
olliesilvester: I think I'm in favour of subclassing `LoggerAdapter`. The other logging PR (https://github.com/bluesky/ophyd-async/pull/277/files#diff-ac5bd0cd80af7a87523435d81101c0716ee2ff076e747265d91d2517463dfe5e) used the `ophyd_async_device_name` info to format a message

src/ophyd_async/core/device.py
```diff
@@ -39,6 +41,13 @@ def name(self) -> str:
         """Return the name of the Device"""
         return self._name
 
+    @property
+    @lru_cache(1)
+    def log(self):
+        return LoggerAdapter(
+            getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
```
ZLLentz, line 48: Is this guaranteed to be run for the first time after the name has been finalized? Naively I'd assume it's possible for some code to access this prior to name finalization (without any way to check here) but I'm not familiar enough with the initialization flow to know that for sure.

If this is guaranteed to not be run prior to name finalization, then this PR looks good to me.

olliesilvester: Thanks, good point. If we clear the cache during `set_name`, it should guarantee that we have the latest version for `name` in the log

ZLLentz (approved):
This looks good to me!

src/ophyd_async/core/device.py
```diff
@@ -52,6 +60,9 @@ def set_name(self, name: str):
         name:
             New name to set
         """
+        if hasattr(self, "log"):
+            del self.log
```
coretl, line 64: A comment as to intent might be good here
```suggestion
        # If changing name, delete self.log so a new one with the right name will be made by cached_property above
        if hasattr(self, "log"):
            del self.log
```

PR #276, opened 2024-05-02
Fix p4p converter bug

evalott100:
p4p will fail if it tries to use datatype int for a float value on a PV.

The panda currently has some fields which are analogue Out with precision 0, we should let these be ints in ophyd async even if the values come over as floats.


evalott100:
I made an issue for checking PREC: https://github.com/bluesky/ophyd-async/issues/280

Imo we should merge this to get panda working for the next sprint.

PR #277, opened 2024-05-02
Add log formatter with colour

olliesilvester:
Color or colour?! 
Fixes part of #https://github.com/bluesky/ophyd-async/issues/195


abbiemery (changes requested):
Colour :) 

It looks good to me from a brief poke. The only thing I think I would prefer is the log.py to be in ophyd_async/log.py, and not in core. It doesn't contribute to the core constituent parts of the framework, its a nice tool. 

(also remember to move the test relative, it's something I frequently forget).

Issue #212, opened 2024-04-15
Move signals common to all areaDetector plugins up to nd_plugin class


jwlodek:
Certain signals that are shared between all plugins are currently included in nd_file_hdf, they should be moved up to the higher level class.

jwlodek:
Those signals cannot be shared, because the size of the image coming out of the camera does not necessarily need to be the same as the size of the image that the plugin receives, either due to cropping via the ROI plugin or binning, or something similar. The `ArraySize0` etc signals do however use the same names for all plugins, and so can be moved up from NDFileHDF5 to NDPluginBase.

For reference, any signals that appear as records in: https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/NDPluginBase.template

should be moved to the `NDPluginBase` class.

PR #286, opened 2024-05-03
Move areaDetector plugins from child classes to parent classes

jmaruland:
This PR solves the issue presented in #212 which looks to give more visibility to some nd_plugins.
In order to decide which plugins need to be moved up, the NDArrayBase template (https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/NDArrayBase.template) was used. Plugins mentioned in this template were moved to the immediate parent class (NDArrayBase or NDPluginBase)


jmaruland:
Closes #212

jmaruland:
This PR got a bit behind of the current stat of the main branch and there was a conflict with `src/ophyd_async/epics/areadetector/writers/nd_file_hdf.py`.
I had to apply a rebase on this branch in order to fix the conflict.

PR #289, opened 2024-05-08
Fix support for Str datatype Enum signals

DiamondJoseph:
Allows for signals that are backed by enums but have a Signal datatype of str.

e.g. for the Aravis TriggerMode

### DIFF
--- a/src/ophyd_async/epics/_backend/_aioca.py
+++ b/src/ophyd_async/epics/_backend/_aioca.py
@@ -2,7 +2,7 @@ import logging
 import sys
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, Optional, Sequence, Tuple, Type, Union
+from typing import Any, Dict, Optional, Sequence, Type, Union
 
 from aioca import (
     FORMAT_CTRL,
@@ -79,7 +79,7 @@ class CaArrayConverter(CaConverter):
 
 @dataclass
 class CaEnumConverter(CaConverter):
-    choices: Tuple[Any, ...]
+    choices: dict[str, str]
 
     def write_value(self, value: Union[Enum, str]):
         if isinstance(value, Enum):
@@ -88,7 +88,7 @@ class CaEnumConverter(CaConverter):
             return value
 
     def value(self, value: AugmentedValue):
-        return value
+        return self.choices[value]
 
     def get_datakey(self, source: str, value: AugmentedValue) -> DataKey:
         return {
@@ -142,8 +142,8 @@ def make_converter(
         pv_choices = get_unique(
             {k: tuple(v.enums) for k, v in values.items()}, "choices"
         )
-        enum_class = get_supported_values(pv, datatype, pv_choices)
-        return CaEnumConverter(dbr.DBR_STRING, None, enum_class)
+        supported_values = get_supported_values(pv, datatype, pv_choices)
+        return CaEnumConverter(dbr.DBR_STRING, None, supported_values)
     else:
         value = list(values.values())[0]
         # Done the dbr check, so enough to check one of the values
--- a/src/ophyd_async/epics/_backend/_aioca.py
+++ b/src/ophyd_async/epics/_backend/_aioca.py
@@ -2,7 +2,7 @@ import logging
 import sys
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, Optional, Sequence, Type, Union
+from typing import Any, Dict, Optional, Sequence, Tuple, Type, Union
 
 from aioca import (
     FORMAT_CTRL,
@@ -28,7 +28,7 @@ from ophyd_async.core import (
 )
 from ophyd_async.core.utils import DEFAULT_TIMEOUT, NotConnected
 
-from .common import get_supported_enum_class
+from .common import get_supported_values
 
 dbr_to_dtype: Dict[Dbr, Dtype] = {
     dbr.DBR_STRING: "string",
@@ -79,7 +79,7 @@ class CaArrayConverter(CaConverter):
 
 @dataclass
 class CaEnumConverter(CaConverter):
-    enum_class: Type[Enum]
+    choices: Tuple[Any, ...]
 
     def write_value(self, value: Union[Enum, str]):
         if isinstance(value, Enum):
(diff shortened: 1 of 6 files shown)


Tom-Willemsen (commented):
The code, as far as I understand it, looks fine.

Can we add some documentation about exactly how enums are supported and which combinations of `SomeEnum(str, Enum)`, `SomeEnum(Enum)` etc are supported?

e.g. [some other places](https://github.com/DiamondLightSource/dodal/pull/467#discussion_r1582881725) seem to believe that str enums are always required. If that's no longer true can we say so in the docs somewhere.

The testing for this is also getting a little bit hard to follow and check that it's really checking all the cases we care about. Maybe out of scope for this issue though.

DiamondJoseph:
> Can we add some documentation about exactly how enums are supported and which combinations of `SomeEnum(str, Enum)`, `SomeEnum(Enum)` etc are supported?
> 
> e.g. [some other places](https://github.com/DiamondLightSource/dodal/pull/467#discussion_r1582881725) seem to believe that str enums are always required. If that's no longer true can we say so in the docs somewhere.

`SomeEnum(str, Enum)`, and `str` are valid datatypes for a signal with an enumeration as a backend. `SomeEnum(int, Enum)`, `SomeEnum(float, Enum)`, `SomeEnum(Enum)`, `float`, `int` etc. are not.

> The testing for this is also getting a little bit hard to follow and check that it's really checking all the cases we care about. Maybe out of scope for this issue though.

I think https://github.com/bluesky/ophyd-async/pull/137 simplifies the tests by making all of the expected descriptors much simpler to generate, I started following it up again today.

Tom-Willemsen:
> SomeEnum(str, Enum), and str are valid datatypes for a signal with an enumeration as a backend. SomeEnum(int, Enum), SomeEnum(float, Enum), SomeEnum(Enum), float, int etc. are not.

Perfect. So [where the docs currently talk about using `Enum`](https://blueskyproject.io/ophyd-async/main/how-to/make-a-simple-device.html#readable) can you add this sort of clarification? Happy for it to be a link to a new/separate docs page if that's easiest. The relationship between datatype and backend probably also could do with explaining for an unfamiliar user.

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -88,11 +88,15 @@ def write_value(self, value: Union[Enum, str]):
             return value
 
     def value(self, value: AugmentedValue):
-        return self.enum_class(value)
+        return value
```
coretl, line 91: This means that:
```python
class MyEnum(str, Enum):
   aaa = "AAA"
   bbb = "BBB"

sig = epics_signal_rw(MyEnum, pv)
await sig.connect()
assert (await sig.get_value()) == MyEnum.aaa
assert (await sig.get_value()) == "AAA"
```
Unfortunately
```python
print(await sig.get_value())  # prints "AAA" not <MyEnum.aaa: "AAA">
```
@DominicOram is this ok?

coretl (commented):
I'm not convinced that we want to start returning `str` from `SignalRW[Enum].get_value()`, which I think this does. Can we leave the enum conversion in, but make it optional? Or make a new converter for it?

src/ophyd_async/epics/_backend/_aioca.py
DiamondJoseph: Added some tests to ensure that:
for signals backed by an enum:

Signal[str].get_value() is not Enum.val
Signal[Enum].get_value() is Enum.val

in each case, Signal.get_value() == "val" and Signal.get_value() == Enum.val

src/ophyd_async/epics/_backend/common.py
```diff
@@ -1,25 +1,25 @@
...
+            raise TypeError(
+                (
+                    f"{pv} has choices {pv_choices}, "
+                    f"which do not match {datatype}, which has {choices}"
+                )
             )
-        )
-    return datatype
+        return {x: datatype(x) for x in pv_choices}
+    return {x: x for x in pv_choices}
```
coretl, line 25: I literally have no idea how this is working, here you drop the reference to the Enum, and keep only its string value...

tests/epics/test_signals.py
```diff
@@ -535,3 +536,26 @@ def test_signal_helpers():
...
+
+
+async def test_str_enum_returns_enum(ioc: IOC):
+    await ioc.make_backend(MyEnum, "enum")
+    pv_name = f"{ioc.protocol}://{PV_PREFIX}:{ioc.protocol}:enum"
+
+    sig = epics_signal_rw(MyEnum, pv_name)
+    await sig.connect()
+    val = await sig.get_value()
+    assert val is MyEnum.b
```
coretl, line 548: ... but here the value `is` the Enum. How does that work?

tests/epics/test_signals.py
```diff
@@ -535,3 +536,26 @@ def test_signal_helpers():
...
     assert execute._backend.write_pv == "Execute"
+
+
+async def test_str_enum_returns_enum(ioc: IOC):
+    await ioc.make_backend(MyEnum, "enum")
+    pv_name = f"{ioc.protocol}://{PV_PREFIX}:{ioc.protocol}:enum"
+
+    sig = epics_signal_rw(MyEnum, pv_name)
+    await sig.connect()
+    val = await sig.get_value()
```
coretl, line 547: For my sanity please could we also add this test:
```suggestion
    val = await sig.get_value()
    assert repr(val) == "<MyEnum.b: 'Bbb'>"
```

src/ophyd_async/epics/_backend/common.py
DiamondJoseph:
```
if issubclass(datatype, Enum):
        choices = tuple(v.value for v in datatype)
        if choices != pv_choices:
            raise TypeError(
                (
                    f"{pv} has choices {pv_choices}, "
                    f"which do not match {datatype}, which has {choices}"
                )
            )
        return {x: datatype(x) for x in pv_choices}  # <- This is if the datatype is enum, we make a map of str : Enum
    return {x: x for x in pv_choices}  # <- This is if the datatype is not None, not an Enum, is str or some other str extending type. We make a map of str: str
```

tests/epics/test_signals.py
DiamondJoseph: This is where datatype is an enum, so get_supported_values was map{str:enum} and the EnumConverter has the Enum values keyed by name (or index, appropriately)

coretl (approved):
Happy to merge when the ordering issue is sorted

PR #290, opened 2024-05-08
Factor stream resource producing code into a separate class

### LINKED ISSUE #173: Factor StreamResource producing code into a separate class
This seems generally good, from a DRY point of view, and would specifically be helpful in managing the transition from old-style to new-style StreamResource

skarakuzu:
Closes #173


src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -1,11 +1,31 @@
...
+    StreamDatum,
+    StreamResource,
+)
 
 from ophyd_async.core import DirectoryInfo
 
-from ._hdfdataset import _HDFDataset
+
+@dataclass
+class _HDFDataset:
```
danielballan, line 16: Per @coretl:

This should be slimmed to only what's needed to describe a file, not what's needed to _write_ it.

The pattern detector is different from the others because it is actually _writing_ the HDF5 file, not describing a file written by an external program (IOC). For the pattern detector, separate dataclass should _extend_ this to add `maxshape` and `fillvalue`.

With that separtion, we can make all of these fields required fields.

---

Some of this information is used to construct the EventDescriptor (`shape`, `dtype`, `name` ?, `multiplier`). Other information is used to construct the StreamResource (`path`, meaning the "path" of the dataset inside the layout of the HDF5 file, distinct from the file; and optionally `smwr`).

Add `dtype_str`, the numpy-style dtype like `"<i4"` or `"<f8"`.

This should also include the `uri`. (For old-style StreamResource we can convert this into `root` and `resource_path` by ignoring the hostname, hard-coding `root` to `"/"` as discussed, and using the rest of the path as `resource_path`.)

https://github.com/bluesky/bluesky/blob/bc8222be2e099a00baafcede1d761d61b213bf19/src/bluesky/callbacks/tiled_writer.py#L167-L178

pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
     "packaging",
     "pint",
     "bluesky>=1.13.0a3",
-    "event-model<1.21.0",
```
danielballan, line 20: If DLS needs this pinned somehow for now, to keep the original draft of StreamResource, could it be pinned in a DLS-specific package? Maybe dodal?

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -0,0 +1,129 @@
...
+    StreamDatum,
+    StreamResource,
+)
+
+from ophyd_async.core import DirectoryInfo
+
+
+@dataclass
+class _HDFDataset:
+    name: str
```
coretl, line 19: I think this is clearer if we call it:
```suggestion
    #: Name of the data_key within the Descriptor document
    data_key: str
```

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -0,0 +1,129 @@
...
+    dtype_numpy: Optional[str] = None
+    swmr: bool = False
+    shape: Optional[List[int]] = None
+    multiplier: Optional[int] = 1
+    path: Optional[str] = None
```
coretl, line 24: And this one has also confused us in the past, I suggest:
```suggestion
    #: Name of the dataset within the HDF file
    dataset: str
```

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -0,0 +1,129 @@
...
+                    "",
+                    None,
+                )
+            )
+
+            self._bundles: List[ComposeStreamResourceBundle] = [
+                bundler_composer(
+                    mimetype="application/x-hdf5",
+                    uri=uri,
+                    data_key=ds.name.replace("/", "_"),
```
coretl, line 100: Which would make this:
```suggestion
                    data_key=ds.data_key,
```
I would make data_keys with `/` in them an error, not a silent replacement

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -0,0 +1,129 @@
...
+                    parameters={
+                        "path": ds.path,
```
coretl, line 102: And this would be:
```suggestion
                        "dataset": ds.dataset,
```

coretl (changes requested):

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
skarakuzu: I have a question regarding the removal of `replace("/", "_")` . If I remove this I have to replace `DATA_PATH = /entry/data/data` `DATA_PATH = /entry/data/data` and so on. If not I get a  `regex not matching error` from the bluesky runengine . I tentatively replaced the paths this way in the code but would like to ask for suggestions.

coretl: They are different things, data_key is the name within the descriptor, and dataset is the path within the HDF file. So for pattern generator I think we want something like:
```python
            _HDFDataset(
                data_key=f"{detector_name}-data,
                dataset=DATA_PATH,
            )
```
I'm not sure quite how you pass the detector name through and into the pattern generator though...

skarakuzu: In `pattern_generator.py`, PATH parameters were assigned to `name` (now `data_key`). What I meant is that, deleting the `replace("/", "_")` resulted in errors so I had to modify the PATH parameters. The other thing is that, I am not sure I understand passing detector name to `_HDFDataset` since I am not experienced with the usage of this class. Is the device going to assign the parameters in this class? I would be happy to have more context.

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -0,0 +1,132 @@
...
+    swmr: bool = False
+    shape: Optional[List[int]] = None
+    multiplier: Optional[int] = 1
+    #: Name of the dataset within the HDF file
+    dataset: Optional[str] = None
+    device_name: Optional[str] = None
+    block: Optional[str] = None
+    maxshape: tuple[Any, ...] = (None,)
+    dtype: Optional[Any] = None
+    fillvalue: Optional[int] = None
```
DiamondJoseph, line 31:
```suggestion
    #: Name of the data_key within the Descriptor document
    data_key: str
    # Numpy dtype representation of the type
    dtype_numpy: str
    # Is the h5 file written in SingleWriter, MultipleRead mode
    swmr: bool = False
    # Shape of a frame of the h5 file
    shape: List[int] = []
    # How many frames per Stream Datum index
    multiplier: int = 1
    #: Name of the dataset within the HDF file
    dataset: str = None
```

DiamondJoseph: I think?

src/ophyd_async/panda/writers/_hdf_writer.py
```diff
@@ -17,10 +17,9 @@
     wait_for_value,
 )
 from ophyd_async.core.signal import observe_value
+from ophyd_async.epics.areadetector.writers.general_hdffile import _HDFDataset, _HDFFile
 from ophyd_async.panda import CommonPandaBlocks
 
-from ._panda_hdf_file import _HDFDataset, _HDFFile
-
 
 class Capture(str, Enum):
```
DiamondJoseph, line 24:
```suggestion

@dataclass
class PandaDataset(_HDFDataset):
  block: str
```

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -1,66 +1,45 @@
...
-DATA_PATH = "/entry/data/data"
+DATA_PATH = "_entry_data_data"
 
 # pixel sum path
-SUM_PATH = "/entry/sum"
+SUM_PATH = "_entry_sum"
 
 MAX_UINT8_VALUE = np.iinfo(np.uint8).max
 
 SLICE_NAME = "AD_HDF5_SWMR_SLICE"
```
DiamondJoseph, line 29:
```suggestion
@dataclass
class PatternDataset(_HDFDataset):
    maxshape: tuple[Any, ...] = (None,)
    fillvalue: Optional[int] = None
    dtype: Optional[type] = None # Or whatever this was on the previous class
```

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -1,66 +1,45 @@
...
-    maxshape: tuple[Any, ...] = (None,)
-    path: Optional[str] = None
-    multiplier: Optional[int] = 1
-    dtype: Optional[Any] = None
-    fillvalue: Optional[int] = None
-
-
 def get_full_file_description(
-    datasets: List[DatasetConfig], outer_shape: tuple[int, ...]
+    datasets: List[_HDFDataset], outer_shape: tuple[int, ...]
```
DiamondJoseph, line 32:
```suggestion
    datasets: List[PatternDataset], outer_shape: tuple[int, ...]
```

src/ophyd_async/panda/writers/_hdf_writer.py
```diff
@@ -152,17 +151,17 @@ async def open(self, multiplier: int = 1) -> Dict[str, DataKey]:
             for suffix in capture_signal.capture_type.split(" "):
                 self._datasets.append(
                     _HDFDataset(
```
DiamondJoseph, line 153:
```suggestion
                    PandaDataset(
```

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -255,24 +173,24 @@ def _get_new_path(self, directory: DirectoryProvider) -> Path:
         new_path: Path = info.root / info.resource_dir / filename
         return new_path
 
-    def _get_datasets(self) -> List[DatasetConfig]:
-        raw_dataset = DatasetConfig(
+    def _get_datasets(self) -> List[_HDFDataset]:
```
DiamondJoseph, line 176:
```suggestion
    def _get_datasets(self) -> List[PatternDataset]:
```

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -255,24 +173,24 @@ def _get_new_path(self, directory: DirectoryProvider) -> Path:
...
+        raw_dataset = _HDFDataset(
```
DiamondJoseph, line 177:
```suggestion
        raw_dataset = PatternDataset(
```

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
coretl: that looks right to me

subinsaji: thanks for explaining!

coretl:
This was done in https://github.com/bluesky/ophyd-async/pull/429

PR #292, opened 2024-05-09
Move colorlog to a regular dependency

olliesilvester:
In https://github.com/bluesky/ophyd-async/pull/277 I accidentally put `colorlog` as a dev dependency


evalott100 (approved):
lgtm

PR #293, opened 2024-05-09
Add fly scan plan stub

### LINKED ISSUE #278: Write planstub for the current standard fly-scan
We currently have a unified way to stage, prepare, collect etc with a set of detectors and a flyer. It is worth having a plan stub for this within ophyd-async so we don't keep re-writing this in each of the beamline-bluesky repos.

abbiemery:
Fixes #278.

This adds a plan stub for the current fly scanning flow we have. That being:

    - Stage devices and open a run
    - Set up the flyer with a static sequence table and detectors with a trigger
    - Declare the stream and kickoff the scan
    - Collect while completing
    - Close run and unstage devices.

This does the middle three points, so all you should need to do outside this is instantiate devices, stage and open a run.


src/ophyd_async/planstubs/fly.py
```diff
@@ -0,0 +1,113 @@
...
+import bluesky.plan_stubs as bps
+
+from ophyd_async.core.detector import DetectorTrigger, StandardDetector, TriggerInfo
+from ophyd_async.core.flyer import HardwareTriggeredFlyable
+from ophyd_async.core.utils import in_micros
+from ophyd_async.panda._table import SeqTable, SeqTableRow, seq_table_from_rows
+from ophyd_async.panda._trigger import SeqTableInfo
+
+
+def fly_and_collect(
```
coretl, line 12: This should probably have a more specific name.

Something like `time_resolved_trigger_detectors`?

coretl:
Requires:
- https://github.com/bluesky/ophyd-async/pull/305
- https://github.com/bluesky/bluesky/pull/1720

Test script:
```python
# Import bluesky and ophyd
import bluesky.plans as bp
import bluesky.preprocessors as bpp
from bluesky import RunEngine
from bluesky.utils import ProgressBarManager, register_transform
from dodal.beamlines import p45
from dodal.beamlines.beamline_utils import set_directory_provider

from ophyd_async.core import HardwareTriggeredFlyable, StaticDirectoryProvider
from ophyd_async.panda._trigger import StaticSeqTableTriggerLogic
from ophyd_async.planstubs import fly_and_collect

# Create a run engine, with plotting, progressbar and transform
RE = RunEngine({}, call_returns_result=True)
RE.waiting_hook = ProgressBarManager()
register_transform("RE", prefix="<")

# Create ophyd-async devices
set_directory_provider(StaticDirectoryProvider("/dls/p45/data/2024/cm37283-2/tmp"))
det = p45.diff()
panda = p45.panda1()
flyer = HardwareTriggeredFlyable(StaticSeqTableTriggerLogic(panda.seq[1]))


@bpp.stage_decorator([panda, det])
@bpp.run_decorator()
def my_custom_plan():
    yield from fly_and_collect(
        "primary", [panda, det], flyer, 10, 0.1, 0.5, repeats=2, period=1
    )


RE(my_custom_plan(), lambda n, d: print(n, d))
```

src/ophyd_async/planstubs/fly.py
abbiemery: I agree, I just wanted to call it something and figured i'd ask for opinions before merging.

abbiemery:
@coretl I have had to change the `collect_while_completing` stub call back to the code used for the i22 experiment. An issue that was raised back then has not been addressed and as such the new stub is not compatible yet. The issue and following discussion is here https://github.com/bluesky/bluesky/issues/1630.

abbiemery:
I have also had to change back the `complete_all` and `kickoff_all` to just bare `kickoff` and `complete` as this are not in the latest alpha of bluesky. I will make an issue to go through ophyd-async and update these once we get the next bluesky alpha.

abbiemery:
I have also dropped the patch to 90% as it was up at 98 and was starting to get prohibitive. This will not effect the catching of overall coverage.

abbiemery:
Made an issue to investgate the stub. https://github.com/bluesky/ophyd-async/issues/329

Issue #265, opened 2024-04-29
Features to allow Devices to be connected up front, or lazily connected in plans


coretl:
https://github.com/DiamondLightSource/dodal/issues/415#issuecomment-2079651830 summarizes a discussion on Device connection. The strategy we would like to support in ophyd-async is:
- Device connection can be kicked off at any point, but we would like to ensure a Device is connected before we use it in a plan
- If we keep track of the connect task then we can make subsequent calls to connect return immediately if a successful connect has been done
- This means we can write an `ensure_connected(*devices)` plan stub that can be used in plans to run that connect method, which will do no I/O if the Device is already connected
- We need a method of doing a reconnect if we want to make sure the Device is still there

This ticket is for the following:

## New ADR

Describing strategies for Device connection

## `Device.connect`

Change so it looks like:
```python
class Device:
    # None if connect hasn't started, an Event if it has, a set Event if it's done
    _connect_task: Optional[asyncio.Task] = None
    async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT, force_reconnect=False):
        previous_connect_ok = self._connect_task and self._connect_task.done() and not self._connect_task.exception()
        if force_reconnect or not previous_connect_ok:
            # Kick off a connection
            coros = {
                name: child_device.connect(mock, timeout=timeout)
                for name, child_device in self.children()
            }
            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
        # Wait for it to complete
        await self._connect_task
```

## New `ophyd_async.planstubs.ensure_connected`
```python
def ensure_connected(*devices: Device, mock: bool = False, timeout: float = DEFAULT_TIMEOUT, force_reconnect=False):
    coros = {device.name: device.connect(mock, timeout, force_reconnect) for device in devices}
    bps.wait_for([wait_for_connection(**coros)])
```

jwlodek:
This would be a very helpful feature for us at NSLS-II. Currently the strategy has been "comment out the devices you are not using..."

PR #294, opened 2024-05-09
Device connect lazily

ZohebShaikh:
closes https://github.com/bluesky/ophyd-async/issues/265

### DIFF
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import sys
 from functools import cached_property
 from logging import LoggerAdapter, getLogger
@@ -32,6 +33,10 @@ class Device(HasName):
     _name: str = ""
     #: The parent Device if it exists
     parent: Optional[Device] = None
+
+    # Previous connect was successful, False on initialization
+    # since there was no previous connect
+    _previous_connect_success: bool = False
 
     def __init__(self, name: str = "") -> None:
         self.set_name(name)
@@ -71,7 +76,12 @@ class Device(HasName):
             child.set_name(child_name)
             child.parent = self
 
-    async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
+    async def connect(
+        self,
+        sim: bool = False,
+        timeout: float = DEFAULT_TIMEOUT,
+        force_reconnect=False,
+    ):
         """Connect self and all child Devices.
 
         Contains a timeout that gets propagated to child.connect methods.
@@ -83,12 +93,18 @@ class Device(HasName):
         timeout:
             Time to wait before failing with a TimeoutError.
         """
-        coros = {
-            name: child_device.connect(sim, timeout=timeout)
-            for name, child_device in self.children()
-        }
-        if coros:
-            await wait_for_connection(**coros)
+
+        if force_reconnect or not self._previous_connect_success:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(sim, timeout=timeout)
+                for name, child_device in self.children()
+            }
+            connect_task = asyncio.create_task(wait_for_connection(**coros))
+
+            # Wait for it to complete
+            await connect_task
+            self._previous_connect_success = not connect_task.exception()
 
 
 VT = TypeVar("VT", bound=Device)
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -78,7 +78,7 @@ class Device(HasName):
         self,
         mock: bool = False,
         timeout: float = DEFAULT_TIMEOUT,
-        force_reconnect=False,
+        force_reconnect: bool = False,
     ):
         """Connect self and all child Devices.
 
@@ -100,7 +100,9 @@ class Device(HasName):
         if force_reconnect or not previous_connect_ok:
             # Kick off a connection
             coros = {
-                name: child_device.connect(mock, timeout=timeout)
+                name: child_device.connect(
+                    mock, timeout=timeout, force_reconnect=force_reconnect
+                )
                 for name, child_device in self.children()
             }
             self._connect_task = asyncio.create_task(wait_for_connection(**coros))
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -33,6 +33,8 @@ class Device(HasName):
     _name: str = ""
     #: The parent Device if it exists
     parent: Optional[Device] = None
+    # None if connect hasn't started, an Event if it has, a set Event if it's done
+    _connect_task: Optional[asyncio.Task] = None
 
     # Previous connect was successful, False on initialization
     # since there was no previous connect
@@ -94,17 +96,25 @@ class Device(HasName):
             Time to wait before failing with a TimeoutError.
         """
 
+        self._previous_connect_success = (
+            self._connect_task is not None
+            and self._connect_task.done()
+            and not self._connect_task.exception()
+        )
         if force_reconnect or not self._previous_connect_success:
             # Kick off a connection
             coros = {
                 name: child_device.connect(mock, timeout=timeout)
                 for name, child_device in self.children()
             }
-            connect_task = asyncio.create_task(wait_for_connection(**coros))
+            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
 
-            # Wait for it to complete
-            await connect_task
-            self._previous_connect_success = not connect_task.exception()
+        assert self._connect_task, "Connect task not created, this shouldn't happen"
+        # Wait for it to complete
+        await self._connect_task
+        self._previous_connect_success = (
+            self._connect_task.done() and not self._connect_task.exception()
+        )
 
 
 VT = TypeVar("VT", bound=Device)
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -33,8 +33,9 @@ class Device(HasName):
     _name: str = ""
     #: The parent Device if it exists
     parent: Optional[Device] = None
-    # None if connect hasn't started, an Event if it has, a set Event if it's done
+    # None if connect hasn't started, a Task if it has
     _connect_task: Optional[asyncio.Task] = None
+    _connect_mock_arg: bool = False
 
     def __init__(self, name: str = "") -> None:
         self.set_name(name)
@@ -96,6 +97,7 @@ class Device(HasName):
             self._connect_task
             and self._connect_task.done()
             and not self._connect_task.exception()
+            and self._connect_mock_arg == mock
         )
         if force_reconnect or not previous_connect_ok:
             # Kick off a connection
@@ -106,6 +108,7 @@ class Device(HasName):
                 for name, child_device in self.children()
             }
             self._connect_task = asyncio.create_task(wait_for_connection(**coros))
+            self._connect_mock_arg = mock
 
         assert self._connect_task, "Connect task not created, this shouldn't happen"
         # Wait for it to complete
--- a/src/ophyd_async/core/device.py
+++ b/src/ophyd_async/core/device.py
@@ -24,8 +24,6 @@ from bluesky.run_engine import call_in_bluesky_event_loop
 from .utils import DEFAULT_TIMEOUT, NotConnected, wait_for_connection
 
 
-_background_connect_task = set()
-
 class Device(HasName):
     """Common base class for all Ophyd Async Devices.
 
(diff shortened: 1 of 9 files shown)


src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,21 @@ async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+        )
+        if force_reconnect or not previous_connect_ok:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(sim, timeout=timeout)
+                for name, child_device in self.children()
+            }
+            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
+        # Wait for it to complete
+        assert self._connect_task
```
coretl, line 107:
```suggestion
        assert self._connect_task, "Connect task not created, this shouldn't happen"
```

src/ophyd_async/planstubs/ensure_connected.py
```diff
@@ -0,0 +1,22 @@
...
+
+def ensure_connected(
+    *devices: Device,
+    sim: bool = False,
+    timeout: float = DEFAULT_TIMEOUT,
+    force_reconnect=False,
+):
+    yield from bps.wait_for(
+        [
+            lambda: wait_for_connection(
```
coretl, line 15: Why does this have to be a lambda?

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,21 @@ async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+            and self._connect_task.done()
+            and not self._connect_task.exception()
+        )
+        if force_reconnect or not previous_connect_ok:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(sim, timeout=timeout)
+                for name, child_device in self.children()
+            }
+            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
```
evalott100, line 105: Could we please have a test running this without the plan stub:

1. We try to connect a `Device` which doesn't work (giving one of its signals a hard backend should work)
2. It doesn't work so then we allow the signal to have a sim backend
3. We try to connect again now that the sim backend is there, so it works

tests/core/test_device.py
```diff
@@ -117,3 +121,66 @@ async def test_device_log_has_correct_name():
...
+        parentMotor._connect_task
+        and parentMotor._connect_task.done()
+        and not parentMotor._connect_task.exception()
+    )
+    for vector in parentMotor.V.values():
+        assert (
+            vector._connect_task
+            and vector._connect_task.done()
+            and not vector._connect_task.exception()
+        )
```
coretl, line 186: In general I try and avoid copy paste, even in tests...
```suggestion
    for device in [parentMotor, parentMotor.X, parentMotor.Y] + list(parentMotor.V.values()):
        assert (
            device._connect_task
            and device._connect_task.done()
            and not device._connect_task.exception()
        )
```

evalott100 (changes requested):
Nice work, nearly there.

src/ophyd_async/planstubs/ensure_connected.py
evalott100: Discussed here:

https://blueskyproject.io/bluesky/main/msg.html#wait-for

The `wait_for` planstub is weird.

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,21 @@ async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+        # Wait for it to complete
+        assert self._connect_task
+        await self._connect_task
```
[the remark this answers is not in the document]
ZohebShaikh: It so that we don't get this issue 
`"None" is not awaitable
  "None" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
    "__await__" is not presentPylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)`

src/ophyd_async/core/device.py
```diff
@@ -32,6 +33,8 @@ class Device(HasName):
     _name: str = ""
     #: The parent Device if it exists
     parent: Optional[Device] = None
+    # None if connect hasn't started, an Event if it has, a set Event if it's done
+    _connect_task: Optional[asyncio.Task] = None
```
evalott100, line 37:
```suggestion
    # Previous connect was successful, False on initialisation since there was
    # no previous connect
    _previous_connect_ok: bool = False
```

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,21 @@ async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+        if force_reconnect or not previous_connect_ok:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(sim, timeout=timeout)
+                for name, child_device in self.children()
+            }
+            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
+        # Wait for it to complete
+        assert self._connect_task, "Connect task not created, this shouldn't happen"
+        await self._connect_task
```
evalott100, line 108:
```suggestion
        if force_reconnect or not self._previous_connect_ok:
            # Kick off a connection
            coros = {
                name: child_device.connect(sim, timeout=timeout)
                for name, child_device in self.children()
            }
            connect_task = asyncio.create_task(wait_for_connection(**coros))
            # Wait for it to complete
            await connect_task
            self._previous_connect_ok = not connect_task.exception()
```

Since we're awaiting the task within the method anyway, I think it's probably nicer to do this...

tests/core/test_device.py
```diff
@@ -117,3 +121,51 @@ async def test_device_log_has_correct_name():
...
+    )
+
+    RE(ensure_connected(parentMotor, sim=True))
+
+    for device in [parentMotor, parentMotor.X, parentMotor.Y] + list(parentMotor.V.values()):
+        assert (
+            device._connect_task
+            and device._connect_task.done()
+            and not device._connect_task.exception()
+        )
```
evalott100, line 171:
```suggestion
    parentMotor = MotorBundle("parentMotor")

    assert not parentMotor._previous_connect_ok
    assert not parentMotor.X._previous_connect_ok
    assert not parentMotor.Y._previous_connect_ok
    RE(ensure_connected(parentMotor, sim=True))

    for device in [parentMotor, parentMotor.X, parentMotor.Y] + list(
        parentMotor.V.values()
    ):
        assert device._previous_connect_ok
```

tests/core/test_device.py
```diff
@@ -117,3 +121,51 @@ async def test_device_log_has_correct_name():
...
+    assert sim_motor._connect_task is None
+
+    # When ready to connect
+    RE(ensure_connected(sim_motor, sim=True))
+
+    assert (
+        sim_motor._connect_task
+        and sim_motor._connect_task.done()
+        and not sim_motor._connect_task.exception()
+    )
```
evalott100, line 139:
```suggestion
    assert not sim_motor._previous_connect_ok

    # When ready to connect
    RE(ensure_connected(sim_motor, sim=True))

    assert sim_motor._previous_connect_ok
```

evalott100 (changes requested):
Had the idea that the following is probably nicer :thinking:

Since we're awaiting the task in init anyway, we only need to store whether or not it was successful

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,21 @@ async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
-            name: child_device.connect(sim, timeout=timeout)
-            for name, child_device in self.children()
-        }
-        if coros:
-            await wait_for_connection(**coros)
+        previous_connect_ok = (
+            self._connect_task
+            and self._connect_task.done()
+            and not self._connect_task.exception()
+        )
```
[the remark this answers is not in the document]
ZohebShaikh: There is a connected property added in the code that is similar to the `previous_connect_ok`

src/ophyd_async/core/device.py
```diff
@@ -83,12 +93,18 @@ async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
             Time to wait before failing with a TimeoutError.
         """
-        coros = {
-            name: child_device.connect(mock=mock, timeout=timeout)
-            for name, child_device in self.children()
-        }
-        if coros:
-            await wait_for_connection(**coros)
+
+        if force_reconnect or not self._previous_connect_success:
```
coretl, line 97: Unfortunately this misses a requirement of the original ticket, if we `connect()` and there is already another `connect()` running then we should `await connect_task` from the previous connect.

@evalott100 we should also think about what happens when switching between `connect(mock=True)` and `connect(mock=False)`. Should we ban it, or do a `force_reconnect` each time we change the argument `mock`?

tests/epics/motion/test_motor.py
```diff
@@ -59,7 +59,7 @@ async def test_motor_moving_well(sim_motor: motor.Motor) -> None:
         target=0.55,
         unit="mm",
         precision=3,
-        time_elapsed=pytest.approx(0.1, abs=0.05),
+        time_elapsed=pytest.approx(0.1, abs=0.2),
```
evalott100, line 62:
``` suggestion
        time_elapsed=pytest.approx(0.1, abs=0.05),
```
The tolerance should never be more than the approximate value in my opinion - this should be taking a little more or less that 100ms every single time. With this change, any value `0 <= x <= 0.3` would be allowed...

We should work out why this takes ~ 0.2 seconds on that one CI build instead of ~ 0.1. This doesn't seem to happen on main.

evalott100 (dismissed):

src/ophyd_async/core/device.py
evalott100: In my opinion it would be quite useful during testing to be able to make a signal fail to connect and then swap in `mock=True` on the next connect. In the other rarer case of a test using an ioc/softioc, I don't think it's neccessary to prohibit reconnection with mock.

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,23 @@ async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+
+        previous_connect_ok = (
+            self._connect_task
+            and self._connect_task.done()
+            and not self._connect_task.exception()
+        )
+        if force_reconnect or not previous_connect_ok:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(mock, timeout=timeout)
```
coretl, line 103:
```suggestion
                name: child_device.connect(mock, timeout=timeout, force_reconnect=force_reconnect)
```

src/ophyd_async/core/device.py
```diff
@@ -71,7 +74,12 @@ def set_name(self, name: str):
             child.set_name(child_name)
             child.parent = self
 
-    async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT):
+    async def connect(
+        self,
+        mock: bool = False,
+        timeout: float = DEFAULT_TIMEOUT,
+        force_reconnect=False,
```
coretl, line 81:
```suggestion
        force_reconnect: bool = False,
```

coretl (changes requested):
Please could you also update `Signal.connect` to take the same arguments as `Device.connect`?

ZohebShaikh:
> Please could you also update `Signal.connect` to take the same arguments as `Device.connect`?

I have added it to Signal.connect which has mockbackend and backend but they don't use the param force_reconnect

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,25 @@ async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
-            for name, child_device in self.children()
-        }
-        if coros:
-            await wait_for_connection(**coros)
+
+        previous_connect_ok = (
+            self._connect_task
+            and self._connect_task.done()
+            and not self._connect_task.exception()
+        )
```
coretl, line 99: Please can we support switching between `mock=True` and `mock=False`.

Something like:
```suggestion
            and self._connect_mock_arg == mock
        )
```

src/ophyd_async/core/device.py
```diff
@@ -32,6 +33,8 @@ class Device(HasName):
     _name: str = ""
     #: The parent Device if it exists
     parent: Optional[Device] = None
+    # None if connect hasn't started, an Event if it has, a set Event if it's done
```
coretl, line 36:
```suggestion
    # None if connect hasn't started, a Task if it has
```

src/ophyd_async/core/device.py
coretl, line 37: ...
```suggestion
    _connect_task: Optional[asyncio.Task] = None
    _connect_mock_arg: bool = False
```

src/ophyd_async/core/device.py
```diff
@@ -83,12 +91,25 @@ async def connect(self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT):
...
+        )
+        if force_reconnect or not previous_connect_ok:
+            # Kick off a connection
+            coros = {
+                name: child_device.connect(
+                    mock, timeout=timeout, force_reconnect=force_reconnect
+                )
+                for name, child_device in self.children()
+            }
+            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
```
coretl, line 108: ...
```suggestion
            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
            self._connect_mock_arg = mock
```

coretl (changes requested):
Thought of a bug...

coretl (approved):

PR #302, opened 2024-05-13
remove things that go into the cookbook

stan-dot:


.devcontainer/devcontainer.json
callumforrester, line 1: Please remove/raise as a separate PR

docs/explanations/flyscanning.rst
```diff
@@ -1,42 +1,6 @@
 Flyscanning
 ===========
 
-Flyscanning (also known as hardware triggered scanning, asynchronous acquisition, and hardware synchronized scanning) is the practice of accelerating the rate of data collection by handing control over to an external hardware system that can control and synchronize the triggering of detectors with other signals and/or commands. Flyscans take many forms.
```
callumforrester, line 4: Will have to wait until after https://github.com/bluesky/bluesky-cookbook/pull/12 is merged, but I think we should then link to that page here

callumforrester (changes requested):

stan-dot:
`It should enable `motor trajectory scanning` and `multiple triggering rates<detectorsync_>` based around a base rate, and pausing/resuming scans. Scans should be modelled using scanspec_, which serves as a universal language for defining trajectory and time-resolved scans, and converted to the underlying format of the given motion controller. It should also be possible to define an `outer scan <outerscan_>`.`

docs/explanations/design-goals.rst contain a reference to those places. not sure how to handle this.

callumforrester:
For design goals, can also link to the new docs when they are published

callumforrester:
@stan-dot After https://github.com/bluesky/bluesky-cookbook/pull/12 is merged and the docs are published, they will be on Github pages. You can then link to them from here in order to fill missing pieces of information.

stan-dot:
what I mean @callumforrester  is that the design-goals.rst link to that removed flyscanning document
image

stan-dot:
from discussion: two step, publish one commit, then publish that new thing and make a new url-only (not markdown) link

callumforrester (approved):

PR #306, opened 2024-05-14
Export epics_signal_rw_rbv

callumforrester:
Export epics_signal_rw_rbv so a user can import it from
ophyd_async.epics.signal rather than ophyd_async.epics.signal.signal


evalott100 (approved):
Strange we weren't doing this before.

lgtm

GDYendell:
@evalott100 it was only just added in #221 and should have been done then

PR #309, opened 2024-05-15
Move plugins from child classes to parent classes

ZohebShaikh:
Closes https://github.com/bluesky/ophyd-async/issues/212


coretl (approved):
This is correct as of ADCore 3-8, so I think this is fine

PR #316, opened 2024-05-17
Improve performance of MockSignalBackend

### LINKED ISSUE #312: Fix the slow down in tests observed since the mock signal backend was added
Since the addition of https://github.com/bluesky/ophyd-async/pull/251, there has been a slow down in tests. `test_motor_moving_well` takes 0.2s longer on certain runs - it's unclear as to why.

coretl:
Fixes #312


callumforrester (approved):
LGTM!

PR #318, opened 2024-05-20
Move timeout handling to functions wrapped with AsyncStatus

coretl:
I think the best place to apply timeouts for motion is in the `*Motor.set` function, rather than the caller to this function.

This PR shows what this looks like:
- Removing `timeout` handling from `*AsyncStatus`
- Calculating the timeout from `velocity` in `*Motor.set` and adding `DEFAULT_TIMEOUT`
- Making a mechanism to apply a timeout via `observe_value` for `Mover.set`

I think this makes things easier to use as you are not forced to `bps.mv(motor, value, timeout=<something I have to calculate>)`, it fills in a sensible value for timeout itself. The only question I have is whether we should allow this timeout to be overridden.

@dperl-dls @callumforrester thoughts?

### DIFF
--- a/tests/epics/motion/test_motor.py
+++ b/tests/epics/motion/test_motor.py
@@ -11,6 +11,7 @@ from ophyd_async.core import (
     set_mock_put_proceeds,
     set_mock_value,
 )
+from ophyd_async.core.mock_signal_utils import callback_on_mock_put
 from ophyd_async.epics.motion import motor
 
 # Long enough for multiple asyncio event loop cycles to run so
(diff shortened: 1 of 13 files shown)


tests/epics/motion/test_motor.py
```diff
@@ -118,6 +119,29 @@ async def test_motor_moving_well_2(sim_motor: motor.Motor) -> None:
     done.assert_called_once_with(s)
 
 
+async def test_motor_move_timeout(sim_motor: motor.Motor):
+    def do_timeout(value, wait=False, timeout=None):
+        # Check we were given the right timeout of move_time + DEFAULT_TIMEOUT
+        assert timeout == 10.3
+        raise ValueError("I Timed Out")
```
callumforrester, line 126: Nit: Can this be a `TmeoutError`? Makes more sense

callumforrester (commented):
@coretl This seems very nice for motors but how does it affect your common garden signal?

> The only question I have is whether we should allow this timeout to be overridden.

I can see the use case of "I want this motor to reach position `a` in `n` seconds otherwise my beamline is broken and I want to know about it" being a thing

tests/epics/motion/test_motor.py
coretl: Just making it a non-obvious error to make sure it bubbles up correctly, could make a custom Exception subclass for this instead to make it extra obvious

coretl:
> @coretl This seems very nice for motors but how does it affect your common garden signal?

Not at all, they will still take `set(value, timeout)` which is handled by the underlying control system. It is up to the caller of `Signal.set` to supply a reasonable titmout. It is only the case of `MyDevice.set` which we are talking about here.

> 
> > The only question I have is whether we should allow this timeout to be overridden.
> 
> I can see the use case of "I want this motor to reach position `a` in `n` seconds otherwise my beamline is broken and I want to know about it" being a thing

So "I want this to move within 2 seconds no matter where it was to begin with"? That's fairly easy to add as an override `Motor.set(value, timeout=2)`. I'll do that.

dperl-dls (approved):
lgtm

callumforrester (approved):

PR #321, opened 2024-05-20
Fix typing in `callback_on_mock_put`

dperl-dls:
`callback_on_mock_put` had an incorrect type annotation which was causing some errors, removing it lets the type checker correctly note it as:

```python
(function) def callback_on_mock_put(
    signal: Signal[Unknown],
    callback: (T@callback_on_mock_put) -> None
) -> _GeneratorContextManager[None]
```


evalott100 (approved):
lgtm

PR #326, opened 2024-05-22
added a test for mock side effect with misaligned args

### LINKED ISSUE #319: Errors in `MockSignalBackend` `callback`/`side_effect` are not reported.
If you have something like:

```python
from ophyd_async.epics.motion import Motor
from ophyd_async.core import (callback_on_mock_put)


def patch_motor(motor: Motor, initial_position=0):
    return callback_on_mock_put(
        motor.user_setpoint, lambda v: set_mock_value(motor.user_readback, v)
    )


async def test_patched_motor():
    m = Motor("")
    await m.connect(mock=True)
    with patch_motor(m):
        await m.set(10)
        assert await m.user_readback.get_value() == 10
```

This will fail, because the implicit kwargs `wait` and `timeout` from `.set()` are passed to the lambda, which doesn't take them, but this failure is not reported, and the callback simply doesn't happen.

If you instead do:

```python
def patch_motor(motor: Motor, initial_position=0):
    return callback_on_mock_put(
        motor.user_setpoint, lambda v, **_: set_mock_value(motor.user_readback, v)
    )
```

then there is no problem.

This is quite mysterious and could be at least better documented in the docstring and maybe type bound of the callback argument, and it would be better if the exception from the mock was propagated to the user.

evalott100:
Closes #319

I couldn't reproduce the behaviour in 319 for whatever reason. Maybe it's been indirectly fixed since now that we use a MagicMock instead in the mock signal backend.

This PR just adds a test to show that we do get a nice error.

### DIFF
--- a/tests/core/test_mock_signal_backend.py
+++ b/tests/core/test_mock_signal_backend.py
@@ -231,6 +231,21 @@ async def test_callback_on_mock_put_no_ctx():
     ]
 
 
+async def test_callback_on_mock_put_fails_if_args_are_not_correct():
+    mock_signal = SignalRW(SoftSignalBackend(float))
+    await mock_signal.connect(mock=True)
+
+    def some_function_without_kwargs(arg):
+        pass
+
+    callback_on_mock_put(mock_signal, some_function_without_kwargs)
+    with pytest.raises(TypeError) as exc:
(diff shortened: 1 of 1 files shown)


dperl-dls (approved):
Great, thanks for looking at this!

PR #333, opened 2024-05-23
Exception for no detectors

### LINKED ISSUE #332: Assert that there must be at least one detector passed to the fly plan
You can currently pass an empty list into the fly-plan which shouldn't be allowed. We should raise an exception in this case.

abbiemery:
Fixes #332

### DIFF
--- a/src/ophyd_async/plan_stubs/fly.py
+++ b/src/ophyd_async/plan_stubs/fly.py
@@ -32,6 +32,10 @@ def time_resolved_fly_and_collect_with_static_seq_table(
     stages/unstages the devices, and opens and closes the run.
 
     """
+    # Ensure there is at least on detector
+    if len(detectors) == 0:
+        raise ValueError
+
     # Set up scan and prepare trigger
     deadtime = max(det.controller.get_deadtime(exposure) for det in detectors)
     yield from prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
--- a/src/ophyd_async/plan_stubs/fly.py
+++ b/src/ophyd_async/plan_stubs/fly.py
@@ -32,9 +32,8 @@ def time_resolved_fly_and_collect_with_static_seq_table(
     stages/unstages the devices, and opens and closes the run.
 
     """
-    # Ensure there is at least on detector
-    if len(detectors) == 0:
-        raise ValueError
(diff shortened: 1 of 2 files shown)


src/ophyd_async/plan_stubs/fly.py
```diff
@@ -32,6 +32,10 @@ def time_resolved_fly_and_collect_with_static_seq_table(
     stages/unstages the devices, and opens and closes the run.
 
     """
+    # Ensure there is at least on detector
+    if len(detectors) == 0:
```
DiamondJoseph, line 36: Can also 

`if detectors:` as a list's truthiness is whether it is empty or not, and it catches `detectors = None`

src/ophyd_async/plan_stubs/fly.py
```diff
@@ -32,6 +32,10 @@ def time_resolved_fly_and_collect_with_static_seq_table(
...
+        raise ValueError
```
DiamondJoseph, line 37: Want some detail in here, so you don't have to see the line to understand

tests/plan_stubs/test_fly.py
```diff
@@ -349,3 +349,25 @@ def fly():
         "stream_datum",
         "stop",
     ]
+
+
+async def test_at_least_one_detector_in_fly_plan(
+    RE: RunEngine, flyer, detector_list=[]
```
DiamondJoseph, line 355: test detector_list = None as well?

DiamondJoseph (changes requested):

src/ophyd_async/plan_stubs/fly.py
abbiemery: OOOh nice

tests/plan_stubs/test_fly.py
abbiemery: Done

DiamondJoseph: Hasn't been pushed if so

abbiemery: Whoops

abbiemery: Pushed now.

DiamondJoseph (approved):

PR #341, opened 2024-05-28
added runtime enum class, metaclass, and tests

### LINKED ISSUE #238: Support extensible enums
For the Aravis detector, we need to support a dynamically extensible enumeration of values for some signals (e.g. the AravisTriggerSource, which we rely on the values Free_Run and Line_1 being present, but there may also be Line_2, Line_3...).

This may be accomplished by a new class of Signal that has a write_pv of type Enum and a read_pv of type str, which would require SignalBackend to take two generic arguments. The existing Signal implementations then have a single Generic bound which is passed into both generic bounds of the backend and the new implementation has bounds of [str, enum_class] for read, write respectively.

This ticket is necessary for #190

evalott100:
Closes #238


src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +52,38 @@ async def get_setpoint(self) -> T:
...
+    def choices(self) -> Tuple[str]:
+        return tuple(self._choices.keys())
+
+    def __getitem__(self, choice: str) -> "RuntimeEnum":
+        if choice not in self.choices:
+            raise AttributeError(f"{str(self)} has no attribute '{choice}'")
+        return choice
+
+    def __str__(self):
+        return f"{self.__class__.__name__}{list(self._choices.keys())}"
```
evalott100, line 89: I did some weird metaclassing to allow type annotations and runtime enum classes to be the same... abandoning the `if TYPE_CHECKING` originally planned: https://github.com/bluesky/ophyd-async/issues/238#issuecomment-2112976003

src/ophyd_async/epics/_backend/common.py
```diff
@@ -10,9 +12,11 @@ def get_supported_values(
     if not datatype:
         return {x: x or "_" for x in pv_choices}
 
-    if not issubclass(datatype, str):
+    if issubclass(datatype, RuntimeEnum):
+        pv_choices = datatype.choices
```
coretl, line 16: Here we need to check that `pv_choices` are a strict subset of `datatype.choices`, then we can discard datatype, we don't need to make `pv_choices` from it

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +54,51 @@ async def get_setpoint(self) -> T:
...
+
+            choices_frozenset = frozenset(choices)
+
+            # If the enum has already been created, return it (ignoring order)
+            if choices_frozenset in _RuntimeEnumMeta.__enum_classes_created:
+                return _RuntimeEnumMeta.__enum_classes_created[choices_frozenset]
+
+            # Create a new enum subclass
+            class _RuntimeEnum(cls):
+                _choices = {choice: choice for choice in choices}
```
coretl, line 97: I would prefer that you kept choices as the frozenset, then you can do away with the choices property above
```suggestion
                choices = choices_frozenset
```

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +54,51 @@ async def get_setpoint(self) -> T:
...
+
+            _RuntimeEnumMeta.__enum_classes_created[choices_frozenset] = _RuntimeEnum
+            return _RuntimeEnum
+
+    class RuntimeEnum(metaclass=_RuntimeEnumMeta):
```
coretl, line 102: Then we probably want a typehint here
```suggestion
    class RuntimeEnum(metaclass=_RuntimeEnumMeta):
        choices: ClassVar[FrozenSet[str]]
```

coretl (changes requested):

coretl:
Please can we also have some new entries in the matrix in `tests/epics/test_signals.py` using `RuntimeEnum` on a subset of the entries of the `enum` record?

src/ophyd_async/epics/_backend/common.py
evalott100: I'm a little confused by what you're saying here...

Surely we check `datatype.choices ⊆ pv_choices`. There could be `pv_choices = {"A", "B", ..., "Z"}` on the epics record, and the user might have wrote code to deal with `datatype = RuntimeEnum["A", "B"]`.

If `pv_choices ⊆ datatype.choices` then we'd be allowing `RuntimeEnum["A", ..., "Z"]` for a record with only `pv_choices = {"A"}`.

coretl: Sorry, I got subset and superset mixed up!

src/ophyd_async/epics/_backend/common.py
```diff
@@ -10,9 +12,15 @@ def get_supported_values(
...
+    if issubclass(datatype, RuntimeEnum):
+        if not datatype.choices.issubset(frozenset(pv_choices)):
+            raise TypeError(
+                f"{pv} has choices {pv_choices}, "
+                f"which do not match RuntimeEnum, which has {datatype.choices}"
+            )
+    elif not issubclass(datatype, str):
         raise TypeError(f"{pv} is type Enum but doesn't inherit from String")
-    if issubclass(datatype, Enum):
+    elif issubclass(datatype, Enum):
```
coretl, line 23: The original was correct with the `if`, but a little confusing.

I suggest:
```python
    if issubclass(datatype, Enum):
        # Do an exhaustive check
        if not issubclass(datatype, str):
            raise TypeError(f"{pv} is type Enum but doesn't inherit from String")
        choices = tuple(v.value for v in datatype)
        if set(choices) != set(pv_choices):
            raise TypeError(
                (
                    f"{pv} has choices {pv_choices}, "
                    f"which do not match {datatype}, which has {choices}"
                )
            )
        return {x: datatype(x) for x in pv_choices}
    elif issubclass(datatype, RuntimeEnum):
        # Do a subset check
        if not datatype.choices.issubset(frozenset(pv_choices)):
            raise TypeError(
                f"{pv} has choices {pv_choices}, "
                f"which do not match RuntimeEnum, which has {datatype.choices}"
            )
    return {x: x for x in pv_choices}
```

coretl (commented):
Having just written this, I'm wondering if `SubsetEnum` is a better name than `RuntimeEnum`?

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +54,52 @@ async def get_setpoint(self) -> T:
...
+            choices_frozenset = frozenset(choices)
+
+            # If the enum has already been created, return it (ignoring order)
+            if choices_frozenset in _RuntimeEnumMeta.__enum_classes_created:
+                return _RuntimeEnumMeta.__enum_classes_created[choices_frozenset]
+
+            # Create a new enum subclass
+            class _RuntimeEnum(cls):
+                choices = choices_frozenset
+                default_choice = default_choice_local
```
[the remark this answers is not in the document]
evalott100: Yea, this is what get's returned from the `RuntimeEnum[]`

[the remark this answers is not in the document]
coretl: I mean `default_choice`, I can't immediately see a reference to it. If we don't then we should delete it

src/ophyd_async/epics/_backend/common.py
evalott100: Don't we want to check it's a string even if it's not an Enum? This change would allow:

```
        get_supported_values("", MyEnum, ("test",))
```

coretl: I don't understand. We should support 3 cases:
- `class MyEnum(str, Enum)`: exhaustive check, `signal.get_value()` returns `MyEnum`
- `MyRuntimeEnum = RuntimeEnum["one", "two"]` (or `SubsetEnum` if we rename it): subset check, `signal.get_value()` returns str
- None: no check, `signal.get_value()` returns str

evalott100: Hmmm I'm inclined to think we should have it fail outright here if it's not a `RuntimeEnum` or `Enum`.

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +53,43 @@ async def get_setpoint(self) -> T:
...
+            ):
+                raise TypeError(
+                    f"Choices must be a str or a tuple of str, not {type(_choices)}."
+                )
+            if len(set(_choices)) != len(_choices):
+                raise TypeError("Duplicate elements in runtime enum choices.")
+
+        # If the enum has already been created (with this order)
+        if _choices in _SubsetEnumMeta.__enum_classes_created:
+            return _SubsetEnumMeta.__enum_classes_created[_choices]
```
coretl, line 82: I don't think we need to track the classes already created, can we just delete this and the associated class variable? It seems harmless to make 2 identical SubsetEnums?

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +53,43 @@ async def get_setpoint(self) -> T:
...
+
+        class _SubsetEnum(cls):
+            choices = _choices
+
+        _SubsetEnumMeta.__enum_classes_created[_choices] = _SubsetEnum
+        return _SubsetEnum
+
+
+class SubsetEnum(metaclass=_SubsetEnumMeta):
+    choices: ClassVar[FrozenSet[str]]
```
coretl, line 92:
```suggestion
    choices: ClassVar[Tuple[str, ...]]
```

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +53,43 @@ async def get_setpoint(self) -> T:
...
+            return _SubsetEnumMeta.__enum_classes_created[_choices]
+
+        class _SubsetEnum(cls):
+            choices = _choices
+
+        _SubsetEnumMeta.__enum_classes_created[_choices] = _SubsetEnum
+        return _SubsetEnum
+
+
+class SubsetEnum(metaclass=_SubsetEnumMeta):
```
coretl, line 91: To make typing work we need to alias this to `Literal` if `typing.TYPE_CHECKING`, otherwise the type checker will think `await SignalRW[SubsetEnum["one", "two"]].get_value()` will be a `SubsetEnum` instance, rather than `"one" | "two"` as desired

coretl (changes requested):

src/ophyd_async/core/signal_backend.py
evalott100: Yea I took it out so I could have linting as I programmed and forgot to put it back in :sweat_smile:

src/ophyd_async/core/signal_backend.py
```diff
@@ -45,3 +53,39 @@ async def get_setpoint(self) -> T:
     @abstractmethod
     def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
         """Observe changes to the current value, timestamp and severity"""
+
+
+if TYPE_CHECKING:
+    SubsetEnum = Literal
+else:
+
```
coretl, line 61: Actually, could we unindent the class definition and call it `class RuntimeSubsetEnum`, then do
```suggestion
if TYPE_CHECKING:
    SubsetEnum = Literal
else:
    SubsetEnum = RuntimeSubsetEnum
```
...

src/ophyd_async/core/soft_signal_backend.py
```diff
@@ -74,37 +83,45 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
         return cast(T, datatype(shape=0))  # type: ignore
 
 
-@dataclass
 class SoftEnumConverter(SoftConverter):
-    enum_class: Type[Enum]
+    choices: Tuple[str, ...]
 
-    def write_value(self, value: Union[Enum, str]) -> Enum:
+    def __init__(self, datatype: Union[SubsetEnum, Enum]):  # type: ignore
```
coretl, line 89: ...then this could be:
```suggestion
    def __init__(self, datatype: Union[RuntimeSubsetEnum, Enum]):
```
and `datatype.choices` will be type checked

src/ophyd_async/core/soft_signal_backend.py
```diff
@@ -74,37 +83,45 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
...
+            return cast(T, list(datatype.__members__.values())[0])  # type: ignore
+        return cast(T, self.choices[0])
 
 
 def make_converter(datatype):
     is_array = get_dtype(datatype) is not None
     is_sequence = get_origin(datatype) == abc.Sequence
-    is_enum = issubclass(datatype, Enum) if inspect.isclass(datatype) else False
+    is_enum = inspect.isclass(datatype) and (
+        issubclass(datatype, Enum) or issubclass(datatype, SubsetEnum)  # type: ignore
```
coretl, line 123: and
```suggestion
        issubclass(datatype, Enum) or issubclass(datatype, RuntimeSubsetEnum)
```

src/ophyd_async/core/signal_backend.py
evalott100: Nice, stops the code being grayed out in the editor!

coretl (approved):

PR #344, opened 2024-05-29
added a `_detector_timeout` to `StandardDetector`

### LINKED ISSUE #327: Deal with HDF stall timeout in a better way
This is the code that applies the timeout:
https://github.com/bluesky/ophyd-async/blob/9f332e389da01b3841c7d000ee7ddfa13c7b3502/src/ophyd_async/core/detector.py#L311-L322

It defaults to DEFAULT_TIMEOUT=10s, which is not enough time for long exposures or gappy hardware triggered scans. Only the trigger source will know a reasonable value for this, so I suggest we don't specify it on detector init, but supply it in `StandardDetector.prepare`, and default it to frame exposure time + DEFAULT_TIMEOUT.

_Originally posted by @coretl in https://github.com/DiamondLightSource/i22-bluesky/issues/13#issuecomment-2125084495_

evalott100:
Closes #327

### DIFF
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -173,8 +173,8 @@ class StandardDetector(
             config_sigs: Signals to read when describe and read
             configuration are called. Defaults to ().
             name: Device name. Defaults to "".
-            writer_timeout: Timeout for frame writing to start, if the
-            timeout is reached, ophyd-async assumes the detector
+            writer_timeout: Length of time after exposure for frame writing to
+            start, if the timeout is reached, ophyd-async assumes the detector
             has a problem and raises an error.
             Defaults to DEFAULT_TIMEOUT.
         """
@@ -182,6 +182,7 @@ class StandardDetector(
         self._writer = writer
         self._describe: Dict[str, DataKey] = {}
         self._config_sigs = list(config_sigs)
+
         self._frame_writing_timeout = writer_timeout
         # For prepare
         self._arm_status: Optional[AsyncStatus] = None
@@ -254,6 +255,7 @@ class StandardDetector(
         await written_status
         end_observation = indices_written + 1
 
+        # we don't have to consider livetime for step scan timeout
         async for index in self.writer.observe_indices_written(
             self._frame_writing_timeout
         ):
@@ -309,7 +311,7 @@ class StandardDetector(
         assert self._arm_status, "Prepare not run"
         assert self._trigger_info
         async for index in self.writer.observe_indices_written(
-            self._frame_writing_timeout
+            self._frame_writing_timeout + self._trigger_info.livetime
         ):
             yield WatcherUpdate(
                 name=self.name,


src/ophyd_async/core/detector.py
```diff
@@ -309,7 +311,7 @@ async def complete(self):
         assert self._arm_status, "Prepare not run"
         assert self._trigger_info
         async for index in self.writer.observe_indices_written(
-            self._frame_writing_timeout
+            self._frame_writing_timeout + self._trigger_info.livetime
```
callumforrester, line 314: This does not cover the use case of trigger -> trigger -> trigger -> long wait -> trigger -> trigger -> trigger, as the long wait may still be longer than `frame_writing_timeout + livetime`. I think we need a new variable inside `trigger_info`, unless the frame writing timeout is set per call to prepare

callumforrester (changes requested):

src/ophyd_async/core/detector.py
callumforrester: Re-reading #327 I can see the confusion. 
For _most_ use cases, we want the timeout to be `DEFAULT_TIMEOUT + self._trigger_info.livetime + self._trigger_info.deadtime` (i.e. the exposure time plus a default timeout). This makes sense for a time-resovled acquisition taking x frames at y fps. _However_, in this case we have an arbitrary extra timeout because we are waiting for a user to press a button and we don't know exactly how long that will take. @timsnow suggests 60s is a reasonable timeout. So the suggestion is that we pass in an arbitrary, optional, per-plan timeout via trigger info. If (and only if) we do this, the detector will use that timeout. If we don't pass it the detector defaults to `DEFAULT_TIMEOUT + self._trigger_info.livetime + self._trigger_info.deadtime` (this is probably the majority use case).

So the trigger info should have a new attribute `frame_timeout: float | None = None` (so that shouldn't break anything else in ophyd-async). This line should then be:

```suggestion
            self._trigger_info.frame_timeout or (DEFAULT_TIMEOUT + self._trigger_info.livetime + self._trigger_info.deadtime)
```

And we can remove `self._frame_writing_timeout` altogether because it is no longer used.

Tagging @abbiemery for visibility.

abbiemery: I'll give that a go and see how it looks/.

abbiemery: Could we not do the same thing with just the exisiting `self._frame_writing_timeout`? If we provide a value on instantiation then use it, otherwise, `(DEFAULT_TIMEOUT + self._trigger_info.livetime + self._trigger_info.deadtime)`

abbiemery: I may have missed something, take a look. I do think it needs a specific tests for if the timeout is provided in the `TriggerInfo`. https://github.com/bluesky/ophyd-async/pull/348

abbiemery:
Closed in favour of #348.

PR #348, opened 2024-05-30
Add frame_timeout to trigger info

### LINKED ISSUE #327: Deal with HDF stall timeout in a better way

abbiemery:
This adds the `frame_timeout` to `TriggerInfo` so that it can be passed in via prepare for flyscans. For stepscans a default set of `TriggerInfo` values are provided.

I can't tell if this has made the tests longer or they already have got long.

Fixes #327 .

### DIFF
--- a/src/ophyd_async/core/detector.py
+++ b/src/ophyd_async/core/detector.py
@@ -164,7 +164,6 @@ class StandardDetector(
         writer: DetectorWriter,
         config_sigs: Sequence[AsyncReadable] = (),
         name: str = "",
-        writer_timeout: float = DEFAULT_TIMEOUT,
     ) -> None:
         """
         Constructor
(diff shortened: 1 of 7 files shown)


callumforrester (approved):

callumforrester:
This looks correct, thanks!
Will take a look at adding that test https://github.com/bluesky/ophyd-async/pull/344#issuecomment-2140797344

callumforrester (approved):

src/ophyd_async/core/detector.py
```diff
@@ -245,17 +241,21 @@ async def describe(self) -> Dict[str, DataKey]:
 
     @AsyncStatus.wrap
     async def trigger(self) -> None:
+        # set default trigger_info
+        self._trigger_info = TriggerInfo(
+            num=1, trigger=DetectorTrigger.internal, deadtime=0.0, livetime=0.0
+        )
```
coretl, line 247: What happens if you have previously run `prepare(TriggerInfo(trigger=DetectorTrigger.internal, livetime=0.5))` as a way of setting up for a step scan with 0.5s exposure?

PR #352, opened 2024-05-31
Move panda save load into plan stubs

evalott100:
The device save load is the only run-engine dependent thing outside of plan stubs. We should move it.


coretl:
Please can we do this as part of the big refactor #383 rather than in a separate PR

coretl:
Done in #383

Issue #361, opened 2024-06-03
Remove type conversion in `device_save_loader.py`


DominicOram:
Following https://github.com/bluesky/ophyd-async/issues/360 the types coming out of the signals should be pure `python` types. Therefore, we do not need to do the conversion in `device_save_loader.py`

coretl:
Is this fixed?

DominicOram:
Yes

Issue #360, opened 2024-06-03
Convert signal readings and `get_values` to pure python types


DominicOram:
When you call `get_reading` or `get_value` on a signal it returns an `aioca` channel access type rather than the pure python type e.g. `ca_int` rather than `int`. This has caused issues downstream e.g. https://github.com/DiamondLightSource/hyperion/issues/1417#issuecomment-2144684981. We should do the conversion every time the signal is read or the value is taken from it.

coretl:
Is this fixed?

coretl:
@rtuck99 is this fixed now?

DominicOram:
Yes, I believe this is fixed

PR #364, opened 2024-06-04
361 type conversions for python builtins

rtuck99:
This addresses #360 and associated #361

I have only added conversions for `ca_str` `ca_int` and `ca_float`, as on balance I think that conversions for large numpy arrays have the potential to add more significant overhead and also possibly remove some utility from the returned values due to numpy provision of features such as matrix and vector operations, whereas for more primitive builtin types there is much more scope for values to be passed around in calling code and corresponding type pollution.


coretl:
Having read up on this a bit, I think we can do this more simply:

- Make `CaConverter.value()` return `+value`. This will invoke `value.__pos__` which does the right thing for all the primitive types, including int float and string: 
```
In [14]: signal = await caget("SR-DI-DCCT-01:SIGNAL")

In [15]: type(signal)
Out[15]: epicscorelibs.ca.dbr.ca_float

In [16]: type(+signal)
Out[16]: float

In [17]: +signal
Out[17]: -0.0004269315999000014

In [18]: signal
Out[18]: -0.0004269315999000014
```
- Override `CaArrayConverter.value()` to return `np.array(value, copy=False)` which will make a new wrapper with the right type, but pointing to the underlying block of data:
```
In [19]: arr = await caget("SR-DI-DCCT-01:SIGNAL.INP$", datatype=DBR_CHAR)

In [20]: type(arr)
Out[20]: epicscorelibs.ca.dbr.ca_array

In [21]: type(np.array(arr, copy=False))
Out[21]: numpy.ndarray
```

src/ophyd_async/core/device_save_loader.py
```diff
@@ -1,35 +1,30 @@
...
-
 
 def ndarray_representer(dumper: yaml.Dumper, array: npt.NDArray[Any]) -> yaml.Node:
     return dumper.represent_sequence(
         "tag:yaml.org,2002:seq", array.tolist(), flow_style=True
     )
 
 
-def ca_dbr_representer(dumper: yaml.Dumper, value: CaType) -> yaml.Node:
+def ca_dbr_representer(dumper: yaml.Dumper, value: ca_array) -> yaml.Node:
```
DominicOram, line 23: Should: Do we need this function now? Can't we just remove it for the change suggested below...

src/ophyd_async/core/device_save_loader.py
```diff
@@ -152,9 +147,6 @@ def save_to_yaml(phases: Sequence[Dict[str, Any]], save_path: str) -> None:
 
     yaml.add_representer(np.ndarray, ndarray_representer, Dumper=yaml.Dumper)
 
-    yaml.add_representer(ca_float, ca_dbr_representer, Dumper=yaml.Dumper)
-    yaml.add_representer(ca_int, ca_dbr_representer, Dumper=yaml.Dumper)
-    yaml.add_representer(ca_str, ca_dbr_representer, Dumper=yaml.Dumper)
     yaml.add_representer(ca_array, ca_dbr_representer, Dumper=yaml.Dumper)
```
DominicOram, line 150: Should: I think we can just change this to 
```suggestion
    yaml.add_representer(ca_array, ndarray_representer, Dumper=yaml.Dumper)
```

and remove `ca_dbr_representer`

DominicOram (changes requested):
Thanks. It concerns me that the changes in save/load required no test changes. In fact I can remove the representer for `ca_array` entirely and all the tests pass. Can we add some coverage for it please?

rtuck99:
> Thanks. It concerns me that the changes in save/load required no test changes. In fact I can remove the representer for `ca_array` entirely and all the tests pass. Can we add some coverage for it please?

I can add some additional tests for the yaml serialization, however they won't fail in the event that the wrong types are exported by the backend, as the serialization function has no knowledge of the backend implementation.

Adding something like that would probably have to be a system test.

DominicOram (approved):
Great, thank you!

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -49,7 +50,10 @@ def write_value(self, value) -> Any:
         return value
 
     def value(self, value: AugmentedValue):
-        return value
+        # for channel access ca_xxx classes, this
+        # invokes __pos__ operator to return an instance of
+        # the builtin base class
+        return +value
```
coretl, line 56: This does a copy, which is fine for the scalars, but please can we make `CaArrayConverter.value` do `np.array(value, copy=False)`?
...

src/ophyd_async/core/device_save_loader.py
```diff
@@ -151,11 +135,7 @@ def save_to_yaml(phases: Sequence[Dict[str, Any]], save_path: str) -> None:
     """
 
     yaml.add_representer(np.ndarray, ndarray_representer, Dumper=yaml.Dumper)
-
-    yaml.add_representer(ca_float, ca_dbr_representer, Dumper=yaml.Dumper)
-    yaml.add_representer(ca_int, ca_dbr_representer, Dumper=yaml.Dumper)
-    yaml.add_representer(ca_str, ca_dbr_representer, Dumper=yaml.Dumper)
-    yaml.add_representer(ca_array, ca_dbr_representer, Dumper=yaml.Dumper)
+    yaml.add_representer(ca_array, ndarray_representer, Dumper=yaml.Dumper)
```
coretl, line 138: ...
That should get rid of all the `ca_array`s and turn them into regular numpy arrays, so we can lose this line

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -147,11 +153,12 @@ def make_converter(
...
-            raise TypeError(
-                f"{pv} has type {type(value).__name__.replace('ca_', '')} "
-                + f"not {datatype.__name__}"
-            )
+        if datatype:
+            if not isinstance(value, datatype):
+                raise TypeError(
+                    f"{pv} has type {type(value).__name__.replace('ca_', '')} "
+                    + f"not {datatype.__name__}"
+                )
```
coretl, line 161: Nit: why did this change?

tests/core/test_device_save_loader.py
```diff
@@ -73,6 +107,66 @@ async def test_enum_yaml_formatting(tmp_path):
...
+        device_all_types.pv_array_int8: np.int8,
+        device_all_types.pv_array_uint8: np.uint8,
+        device_all_types.pv_array_int16: np.int16,
+        device_all_types.pv_array_uint16: np.uint16,
+        device_all_types.pv_array_int32: np.int32,
+        device_all_types.pv_array_uint32: np.uint32,
+        device_all_types.pv_array_int64: np.int64,
+        device_all_types.pv_array_uint64: np.uint64,
+    }.items():
+        await pv.set([np.iinfo(dtype).min, np.iinfo(dtype).max, 0, 1, 2, 3, 4])
```
coretl, line 127: IIRC `MockSignalBackend` doesn't do any type conversion, so these should be numpy arrays rather than lists

coretl (changes requested):

src/ophyd_async/epics/_backend/_aioca.py
rtuck99: It's a left over from the first fix attempt, but looks like it wasn't completely reverted, I will revert this also

src/ophyd_async/core/device_save_loader.py
coretl: Can we now lose this line?

coretl (approved):
Looks good, thanks!

mrakitin:
Thanks for the fix! We hit into that problem at NSLS-II HEX, I'll give it a try.

PR #368, opened 2024-06-05
Add signal connection cache

stan-dot:


src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
...
 
     async def connect(
         self, mock=False, timeout=DEFAULT_TIMEOUT, force_reconnect: bool = False
     ):
+        """Connect self with a timeout
+
+        Args:
+            mock (bool, optional): _description_. Defaults to False.
+            timeout (_type_, optional): _description_. Defaults to DEFAULT_TIMEOUT.
+            force_reconnect (bool, optional): _description_. Defaults to False.
```
coretl, line 77: I believe we are using the numpy doc style to match the rest of the org. Please can you make it like https://github.com/bluesky/ophyd-async/blob/0b05b9c0c7c25417cb76432de0eae067fc7a86ce/src/ophyd_async/core/device.py#L76-L92

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
...
+        """
+        can_use_previous_connection: bool = (
+            self._connect_task and not (self._connect_task.done() and self._connect_task.exception())
+        )
```
coretl, line 81:
```suggestion
        can_use_previous_connect = (
            self._connect_task
            and not (self._connect_task.done() and self._connect_task.exception())
            and self._connect_mock_arg == mock
        )
```

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
...
+        if force_reconnect or not can_use_previous_connection:
+            self.log.debug(f"Connecting to {self.source}")
+            self._connect_task = asyncio.create_task(wait_for_connection(self._backend.connect(timeout=timeout)))
+            self.connect_mock_arg = mock
+        assert self._connect_task, "Connect task not created, this shouldn't happen"
+        await self._connect_task
```
coretl, line 87:
```suggestion
        if force_reconnect or not can_use_previous_connect:
            if mock:
                self._backend = MockSignalBackend(
                    initial_backend=self._initial_backend,
                )
            else:
                self._backend = self._initial_backend
            # Kick off a connection
            self._connect_task = asyncio.create_task(self._backend.connect(timeout=timeout))
            self._connect_mock_arg = mock
```

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
...
+
+
+        # todo not sure what to do here
         if mock and not isinstance(self._backend, MockSignalBackend):
             # Using a soft backend, look to the initial value
             self._backend = MockSignalBackend(
                 initial_backend=self._initial_backend,
             )
```
coretl, line 95:
```suggestion
```

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
...
         self.log.debug(f"Connecting to {self.source}")
-        await self._backend.connect(timeout=timeout)
+        # await self._backend.connect(timeout=timeout)
```
coretl, line 97:
```suggestion
        assert self._connect_task, "Connect task not created, this shouldn't happen"
        # Wait for it to complete
        await self._connect_task
```

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +63,38 @@ def __init__(
     ) -> None:
         self._timeout = timeout
         self._initial_backend = self._backend = backend
+
         super().__init__(name)
```
coretl, line 68: Also need to add:
```suggestion
    def children(self) -> Iterator[Tuple[str, "Device"]]:
        yield from ()
```

coretl (changes requested):

src/ophyd_async/core/signal.py
stan-dot: not sure how would the mock work here. the commits pushed were just the draft - tbh I should have marked the PR as draft

coretl: We can use the previous connect if it is in progress or successful, and it used the same mock arg

src/ophyd_async/core/signal.py
```diff
@@ -66,36 +74,49 @@ def __init__(
...
+                MockSignalBackend(initial_backend=self._initial_backend)
+                if mock
+                else self._initial_backend
+            )
+
+            # Kick off a connection
             self.log.debug(f"Connecting to {self.source}")
-            self._connect_task = asyncio.create_task(wait_for_connection(self._backend.connect(timeout=timeout)))
+            self._connect_task = asyncio.create_task(
+                wait_for_connection(self._backend.connect(timeout=timeout))
```
coretl, line 111: This is a single coro, so no need for `wait_for_connection`:
```suggestion
                self._backend.connect(timeout=timeout)
```

src/ophyd_async/core/signal.py
stan-dot: how can I make it automatic?  I used 'generate docstring'

stan-dot: how can I make it automatic?  I used 'generate docstring'

src/ophyd_async/core/signal.py
stan-dot: applied this

src/ophyd_async/core/signal.py
coretl: no idea what 'generate docstring' is, or how it knows what type of docstrings to generate...

src/ophyd_async/core/signal.py
stan-dot: applied this

src/ophyd_async/core/signal.py
stan-dot: applied

src/ophyd_async/core/signal.py
stan-dot: that I understand, what about mocking?

src/ophyd_async/core/signal.py
stan-dot: that's a vscode extension, a pretty popular one with that. https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

I just had to change its settings to use `numpy` style instead of the default `google` style

stan-dot:
re: testing the signal. how should I instantiate the object? with `epics_signal_rw` or the constructor?

@coretl

stan-dot:
@coretl I narrowed down where the knowledge gap is https://github.com/bluesky/ophyd-async/actions/runs/9397709655/job/25881480270?pr=368

src/ophyd_async/core/signal.py
```diff
@@ -63,18 +70,49 @@ def __init__(
...
+
+            # Kick off a connection
+            self.log.debug(f"Connecting to {self.source}")
+            self._connect_task = asyncio.create_task(
+                self._backend.connect(timeout=timeout)
             )
+            self.connect_mock_arg = mock
+        assert self._connect_task, "Connect task not created, this shouldn't happen"
+        # Wait for it to complete
         self.log.debug(f"Connecting to {self.source}")
```
coretl, line 114: Double log, this one can go
```suggestion
```

tests/core/test_signal.py
```diff
@@ -250,3 +252,25 @@ async def test_subscription_logs(caplog):
...
+    RE(ensure_connected(mock_signal_rw, mock=True))
+    assert (
+        mock_signal_rw._connect_task
+        and mock_signal_rw._connect_task.done()
+        and not mock_signal_rw._connect_task.exception()
+    )
+
+
+async def test_signal_mock_and_back_again(RE):
+    mock_signal_rw = epics_signal_rw(int, "pva://mock_signal", name="mock_signal")
```
coretl, line 269: Can you use a `soft_signal_rw` instead please? This will never connect in `mock=False` mode

coretl (commented):
You can also remove the parameterization from https://github.com/bluesky/ophyd-async/blob/0b05b9c0c7c25417cb76432de0eae067fc7a86ce/tests/core/test_mock_signal_backend.py#L24-L37 and set mock=True always to make that test work

tests/core/test_signal.py
stan-dot: ok sure

tests/core/test_mock_signal_backend.py
```diff
@@ -21,12 +21,11 @@
...
 
 
-@pytest.mark.parametrize("connect_mock_mode", [True, False])
-async def test_mock_signal_backend(connect_mock_mode):
+async def test_mock_signal_backend():
     mock_signal = SignalRW(MockSignalBackend(datatype=str))
     # If mock is false it will be handled like a normal signal, otherwise it will
     # initalize a new backend from the one in the line above
-    await mock_signal.connect(mock=connect_mock_mode)
+    # await mock_signal.connect(mock=True)
```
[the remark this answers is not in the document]
stan-dot: I think so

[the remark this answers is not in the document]
stan-dot: because the argument is no longer passed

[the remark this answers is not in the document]
coretl: I don't understand, `Signal.connect(mock=True)` is valid, how does it fail?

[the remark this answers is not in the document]
stan-dot: re-running the tests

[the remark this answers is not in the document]
stan-dot: >   File "/workspaces/ophyd-async/src/ophyd_async/core/signal.py", line 101, in connect
    MockSignalBackend(initial_backend=self._initial_backend)
  File "/workspaces/ophyd-async/src/ophyd_async/core/mock_signal_backend.py", line 20, in __init__
    raise ValueError("Cannot make a MockSignalBackend for a MockSignalBackends")
ValueError: Cannot make a MockSignalBackend for a MockSignalBackends

that is the error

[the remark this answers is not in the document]
stan-dot: it works if I connect without specifying (mock=True)

stan-dot:
the error string does not appear to be present anywhere but in the test. not sure where is it coming from .
image

stan-dot:
`
async def test_signal_can_be_given_backend_on_connect():
    sim_signal = SignalR()
    backend = MockSignalBackend(int)
    assert sim_signal._backend is None
    await sim_signal.connect(mock=False, backend=backend)
    assert await sim_signal.get_value() == 0`
    
 what is the goal of this test? right now it clashes with the caching implementation. we can either stop supporting this use case and drop the test (easy) or somehow change it OR the implementation of the feature (hard)
 
 that test was added 6 months ago by Rose

coretl:
> that test was added 6 months ago by Rose

Looks like it was added quite recently to support supplying backend on connect:
image

We have to keep the test.

I suggest we always force a reconnect if backend is supplied, as this will only be used if the Signal is part of a Device, in which case the caching will happen there.

@evalott100 do you agree?

stan-dot:
> support supplying backend on connect:
I think I am missing some context. what does this do, when it is used and why do we support it?

coretl:
> > support supplying backend on connect:
> 
> I think I am missing some context. what does this do, when it is used and why do we support it?

This is used when we know that a `Signal` will exist, like in https://github.com/bluesky/ophyd-async/blob/cbbb2950ed4c4b0bf4ca4c03729562b842b44b03/src/ophyd_async/panda/_common_blocks.py#L31-L37 but we don't know what PV it will connect to until `connect()`. What we will do is create the given `SignalRW` with `backend=None`, then on `Device.connect()` we actually create the backend and pass it to `Signal.connect()`.

stan-dot:
huh, I'd never have guessed that

stan-dot:
should we have a flag for a case like that? it looks quite unusual. I'm thinking a flag next to 'force_reconnect' and 'mock'. then we can do logic flow around that.

coretl:
> should we have a flag for a case like that? it looks quite unusual. I'm thinking a flag next to 'force_reconnect' and 'mock'. then we can do logic flow around that.

The flag is a supplied `backend`. If `backend is not None` we should force a reconnect, otherwise do the existing behaviour

stan-dot:
this is the error I get, but it looks like expected: ` File "/workspaces/ophyd-async/tests/core/test_signal.py", line 80, in test_signal_connect_fails_if_different_backend_but_same_by_value
    assert str(exc.value) == "Backend at connection different from initialised one."
AssertionError: assert 'Cannot make a MockSignalBackend for a MockSignalBackends' == 'Backend at connection different from initialised one.'
  
  - Backend at connection different from initialised one.
  + Cannot make a MockSignalBackend for a MockSignalBackends
`

stan-dot:
image


these two are too similar, I'd delete the second one or change the expected message. because the value is different.

stan-dot:
image

evalott100:
> what is the goal of this test? right now it clashes with the caching implementation. we can either stop supporting this use case and drop the test (easy) or somehow change it OR the implementation of the feature (hard)

It tests that you can choose to pass the signal backend in at runtime - why would this be hard to support with your changes?

stan-dot:
@evalott100  now the latest version has both tests, just the expected string is different for the second test

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,54 @@ def __init__(
...
+            and self._connect_mock_arg == mock
+        )
+        if force_reconnect or not can_use_previous_connection:
+            self._backend = (
+                MockSignalBackend(initial_backend=self._initial_backend)
+                if mock
+                else backend
+                if backend
+                else self._initial_backend
+            )
```
evalott100, line 110:
```suggestion
            if backend:
                if self._initial_backend and backend is not self._initial_backend:
                    raise ValueError(
                        "Backend at connection different from initialised one."
                    )
                self._backend = backend
            if mock and not isinstance(self._backend, MockSignalBackend):
                # Using a soft backend, look to the initial value
                self._backend = MockSignalBackend(initial_backend=self._backend)

            if self._backend is None:
                raise RuntimeError("`connect` called on signal without backend")
```

I think this section should be put back to how it currently is - this change

* misses out meaningful error messages
* doesn't skip `MockSignalBackend` initialisation if `initial_backend` is a `MockSignalBackend`
* doesn't do any checking for `initial_backend` vs `backend`

evalott100:
> @evalott100 now the latest version has both tests, just the expected string is different for the second test

That makes sense, you got rid of the validation on the backends in the new `connect` logic

src/ophyd_async/core/signal.py
stan-dot: that looks quite right, I'm not sure how it ended up changed so much

stan-dot: ok I know why - because I was emulating the logic from 'Device'

stan-dot:
and now we're back to the test string error:
image

evalott100:
> and now we're back to the test string error

This is because you're trying to mock a `MockSignalBackend`...

https://github.com/bluesky/ophyd-async/pull/368#issuecomment-2165601311

This is why we only mock if `initial_backend isn't a `MockSignalBackend` (in which case it's already mocked.

https://github.com/bluesky/ophyd-async/pull/368#discussion_r1638133666

stan-dot:
> This is because you're trying to mock a MockSignalBackend...

I know that it is what happens in the test.
I am not sure what is the expected behavior, as I do not fully understand the use case (https://github.com/bluesky/ophyd-async/pull/368#issuecomment-2161054867)

I know that the test or code should be changed

evalott100:
> I know that it is what happens in the test. I am not sure what is the expected behavior, as I do not fully understand the use case ([#368 (comment)](https://github.com/bluesky/ophyd-async/pull/368#issuecomment-2161054867))

In those tests - 

This should be what is failing

https://github.com/bluesky/ophyd-async/blob/b59d1e21c91a2697c36e519393f69a29b7601978/src/ophyd_async/core/signal.py#L76-L79

But in your change you're not checking to see if the backend passed in on connect is the same as the one passed into `__init__`.

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+        Connect self with a timeout
+
+        Parameters
+        ----------
+        mock : bool, optional
+            uses the `MockSignalBackend`, by default False
+        timeout : float, optional
+            Time to wait before failing with a TimeoutError, by default DEFAULT_TIMEOUT
+        force_reconnect : bool, optional
+            Always force reconnection, by default False
```
evalott100, line 96: Good idea adding a docstring! Please add `backend` too :pray:

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+        """
+        can_use_previous_connection: bool = (
+            self._connect_task
+            and not (self._connect_task.done() and self._connect_task.exception())
+            and self._connect_mock_arg == mock
```
evalott100, line 101:
```suggestion
            and self._previous_connect_was_mock == mock
```

... but I also question the use of this. There's two scenarios.

``` python
await signal.connect(mock=True)
# some time later
await signal.connect(mock=False)
```
Or
``` python

await signal.connect(mock=False)
# some time later
await signal.connect(mock=True)
```

The second is a little more useful than the first during testing, but I think we should probably fail outright in either scenario (with an error message)? We would only ever use mock in testing and allowing this behaviour introduces unneeded complexity.

@coretl thoughts?

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+        if force_reconnect or not can_use_previous_connection:
+            if overriding_backend:
+                if (
+                    self._initial_backend
+                    and overriding_backend is not self._initial_backend
+                ):
+                    raise ValueError(
+                        "Backend at connection different from initialised one."
+                    )
+                self._backend = overriding_backend
```
evalott100, line 112: Since we can now continue in a state where `signal.connect(backend=backend)` fails, we should ensure the user can't pass in a different backend on the next connection.

```suggestion
                if (
                    self._backend
                    and backend is not self._backend
                ):
                    raise ValueError(
                        "Backend at connection different from previous one."
                    )
                    
                self._backend = backend
```

This also includes the differing `_initial_backend` since if `_initial_backend`, `self._backend is self._initial_backend`

tests/core/test_signal.py
```diff
@@ -281,3 +282,25 @@ async def test_subscription_logs(caplog):
     assert "Making subscription" in caplog.text
     cache.close()
     assert "Closing subscription on source" in caplog.text
+
+
+async def test_signal_lazily_connects(RE):
+    mock_signal_rw = soft_signal_rw(int, "pva://mock_signal", name="mock_signal")
+    await mock_signal_rw.connect(mock=True)
```
evalott100, line 289:
```suggestion
    await mock_signal_rw.connect(mock=False)
```

since it's a soft_signal_backend

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+        force_reconnect : bool, optional
+            Always force reconnection, by default False
+        """
+        can_use_previous_connection: bool = (
+            self._connect_task
+            and not (self._connect_task.done() and self._connect_task.exception())
+            and self._connect_mock_arg == mock
+        )
+        if force_reconnect or not can_use_previous_connection:
+            if overriding_backend:
```
evalott100, line 104: These if conditions should probably live outside 

``` python
if force_reconnect or not can_use_previous_connection:
```

Otherwise we'd be allowing incorrect `signal.connect` args so long as the last connection was successful.

tests/core/test_signal.py
```diff
@@ -281,3 +282,25 @@ async def test_subscription_logs(caplog):
...
+    RE(ensure_connected(mock_signal_rw, mock=True))
+    assert (
+        mock_signal_rw._connect_task
+        and mock_signal_rw._connect_task.done()
+        and not mock_signal_rw._connect_task.exception()
+    )
+
```
evalott100, line 296: It'd be good to see a test where the connect task fails, and then succeeds.

evalott100 (changes requested):

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+            if self._backend is None:
+                raise RuntimeError("`connect` called on a signal without backend")
+
+            # Kick off a connection
+            self.log.debug(f"Connecting to {self.source}")
+            self._connect_task = asyncio.create_task(
+                asyncio.sleep(1) if mock else self._backend.connect(timeout=timeout)
+            )
+            self.connect_mock_arg = mock
+        assert self._connect_task, "Connect task not created, this shouldn't happen"
```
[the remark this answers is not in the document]
coretl: This assert is just for typing, mypy can't work out that it will always be created by the above logic...

src/ophyd_async/core/signal.py
coretl: All the use cases for connecting multiple times with different values for mock could be solved in a more clunky way by restarting the process with a different value for mock. If this is easy to support then I would say allow people to connect with a different value of mock, but if it's more than a couple of lines of code then ban it with a suitable error message

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,61 @@ def __init__(
...
+    def children(self) -> Iterator[Tuple[str, "Device"]]:
+        yield from ()
+
     async def connect(
         self,
         mock=False,
         timeout=DEFAULT_TIMEOUT,
         force_reconnect: bool = False,
-        backend: Optional[SignalBackend[T]] = None,
+        overriding_backend: Optional[SignalBackend[T]] = None,
```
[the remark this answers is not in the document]
stan-dot: ok I still find this bizarre. What if there is no backend passed on init? then it's None.

tests/core/test_mock_signal_backend.py
```diff
@@ -21,12 +21,11 @@
 from ophyd_async.epics.signal.signal import epics_signal_r, epics_signal_rw
 
 
-@pytest.mark.parametrize("connect_mock_mode", [True, False])
-async def test_mock_signal_backend(connect_mock_mode):
+async def test_mock_signal_backend():
```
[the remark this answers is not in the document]
stan-dot: I thought that the mode is redundant.

tests/core/test_signal.py
```diff
@@ -56,30 +57,30 @@ async def test_signal_can_be_given_backend_on_connect():
     sim_signal = SignalR()
     backend = MockSignalBackend(int)
     assert sim_signal._backend is None
-    await sim_signal.connect(mock=False, backend=backend)
+    await sim_signal.connect(mock=False, overriding_backend=backend)
```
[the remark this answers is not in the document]
stan-dot: yep

tests/core/test_signal.py
```diff
@@ -281,3 +282,25 @@ async def test_subscription_logs(caplog):
     assert "Making subscription" in caplog.text
     cache.close()
     assert "Closing subscription on source" in caplog.text
+
+
+async def test_signal_lazily_connects(RE):
+    mock_signal_rw = soft_signal_rw(int, "pva://mock_signal", name="mock_signal")
```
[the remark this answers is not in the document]
stan-dot: ok

tests/core/test_signal.py
stan-dot: ok

stan-dot (commented):
commiting partial response now

evalott100:
For some reason github won't let me respond to your single comments.

https://github.com/bluesky/ophyd-async/pull/368#discussion_r1639598291
> ok I still find this bizarre. What if there is no backend passed on init? then it's None.

We now allow for no backend passed in on init, then one will need to be provided on connect:

``` python
signal = SignalRW()
signal.connect(backend=any_backend_will_do) # passes
# or
signal.connect(backend=any_backend_will_do, mock=True) # passes
```

We don't allow for replacement of a signal passed in on init:

``` python
backend = PvaSignalBackend(int)
signal = SignalRW(backend)

signal.connect(backend=backend) # passes since the backend is the same


signal.connect(backend=PvaSignalBackend(int) # fails since the backend isn't the same by ref
# fails as to be accurate to production
signal.connect(backend=PvaSignalBackend(int), mock=True)
```

https://github.com/bluesky/ophyd-async/pull/368#discussion_r1639616142
> I thought that the mode is redundant.

I believe the comment

```
    # If mock is false it will be handled like a normal signal, otherwise it will
    # initalize a new backend from the one in the line above
```

was from before we added the second condition to

https://github.com/bluesky/ophyd-async/blob/e4ab7d655df5e82d783fb735e100990a82986cb1/src/ophyd_async/core/signal.py#L81

with which, we don't replace the original `MockSignalBackend`.

I believe the test is still useful to show you can use a `MockSignalBackend` for a backend and have `mock=True`, whether you need it or not.

evalott100:
> I suggest we always force a reconnect if backend is supplied, as this will only be used if the Signal is part of a Device, in which case the caching will happen there.

Missed this comment, but I agree

tests/core/test_mock_signal_backend.py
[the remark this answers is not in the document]
stan-dot: then should those 2 comment lines be deleted?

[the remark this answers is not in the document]
evalott100: > then should those 2 comment lines be deleted?

Yep

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,95 @@ def __init__(
...
+                                 +---------------------------+
+                                 |  Continue operations      |
+                                 |  based on backend and     |
+                                 |  mock settings            |
+                                 +---------------------------+
+
+        Exclude conditions:
+            - Switching between mock=True and mock=False (and vice versa)
+                across successive connect calls.
+            - Passing a different backend than the one initially set
```
stan-dot, line 119: @evalott100 I would like to use a clear chart like this as a reference point. could you tell me 
1) if it's not wrong
2) if it's not missing anything crucial?

evalott100: I'd argue we don't need a chart here - and starting to regularly put them in docstrings would be a huge mistake.

All this information could be encapsulated in descriptions of the args - i.e a small sentence `"The value can/cannot be changed between connects.`" on the description for `mock`.

As for what it is missing within the docstring I'd say pretty much just the "mock" explanation above, and specifying the behavior of the connect asyncio task (i.e if `connect` while the connect task isn't active).

tests/core/test_signal.py
stan-dot: could you elaborate please? what would be the reason of connection failure in the test?

src/ophyd_async/core/signal.py
stan-dot: deleted the chart

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,68 @@ def __init__(
...
+        timeout : float, optional
+            Time to wait before failing with a TimeoutError, by default DEFAULT_TIMEOUT
+        force_reconnect : bool, optional
+            Always force reconnection, by default False
+        backend: SignalBackend, optional
+            The backend to use, by default None.
+            If the device initalization has no backend, then this is required.
+            This value cannot be changed between connect() calls.
+            It is either the exact same object, or it fails.
+            User mainly for mocks.
```
evalott100, line 107: It's not really used for mocks.

```suggestion
            If the device was initialized without a backend then this is required.
            Cannot change value between connect() calls.
```

tests/core/test_mock_signal_backend.py
```diff
@@ -21,12 +21,11 @@
 from ophyd_async.epics.signal.signal import epics_signal_r, epics_signal_rw
 
 
-@pytest.mark.parametrize("connect_mock_mode", [True, False])
```
evalott100, line 24: https://github.com/bluesky/ophyd-async/pull/368/files#r1644172513

Please respond to exising review comments before re-requesting a review: either implementing them or leaving a written disagreement. If not it takes more reviews to get to merge-able code.

tests/core/test_signal.py
```diff
@@ -78,8 +81,6 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
         await sim_signal.connect(mock=False, backend=MockSignalBackend(str))
     assert str(exc.value) == "Backend at connection different from initialised one."
 
-    await sim_signal.connect(mock=False, backend=initial_backend)
```
evalott100, line 81: Why remove this line? It's demonstrating you can connect with backeend so long as it's the same as initial_backend.

tests/core/test_signal.py
```diff
@@ -258,14 +259,14 @@ async def test_assert_configuration(mock_readable: DummyReadable):
...
-    await mock_signal_rw.connect(mock=True)
+    await mock_signal_rw.connect(mock=False)
     assert caplog.text.endswith("Connecting to mock+pva://mock_signal\n")
 
 
 async def test_signal_get_and_set_logging(caplog):
     caplog.set_level(logging.DEBUG)
     mock_signal_rw = epics_signal_rw(int, "pva://mock_signal", name="mock_signal")
-    await mock_signal_rw.connect(mock=True)
+    await mock_signal_rw.connect(mock=False)
```
evalott100, line 269: Again, won't this fail? This test needs a mock signal so `mock=True` to mock the real `PvaSignalBackend`.

tests/core/test_signal.py
evalott100: Because with your new connect logic,
``` python
can_use_previous_connection: bool = (
            self._connect_task
            and not (self._connect_task.done() and self._connect_task.exception())
)
```
a signal will reconnect if the previous connect
* doesn't exist (vacuously well tested)
* is `.done()` (we need a test for this)
* is `.exception()` (we need a test for this)
* uses `force_reconnect` (we need a test for this)

tests/core/test_signal.py
```diff
@@ -281,3 +282,39 @@ async def test_subscription_logs(caplog):
...
+    assert isinstance(mock_signal_rw._backend, SoftSignalBackend)
+    assert mock_signal_rw._connect_task
+    # todo is explicit disconnection needed?
+    await mock_signal_rw.connect(mock=True)
+    assert isinstance(mock_signal_rw._backend, MockSignalBackend)
+
+
+async def test_connect_fails_then_next_connect_succeeds():
+    raise AssertionError("Not implemented")
+    mock_signal_rw = soft_signal_rw(int, "pva://mock_signal", name="mock_signal")
```
evalott100, line 313:
```suggestion
    mock_signal_rw = soft_signal_rw(int, 0, name="mock_signal")
```

In a `soft_signal_rw` that is the value, it doesn't take a PV.

If you want this to fail then perhaps you could make a `pva_signal_backend` so the initial connect fails (since the PV won't exist), then afterwards swap out the backend to a mocked one.

evalott100 (changes requested):

tests/core/test_mock_signal_backend.py
stan-dot: there was enough lack of clarity on other points that I decided to focus on the other parts. the truth is that I still don't understand how is the `connect()` logic supposed to go :( 

maybe I should have left a note about the paremeters point

stan-dot (commented):
not sure

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,68 @@ def __init__(
...
+            - Switching between mock=True and mock=False (and vice versa)
+                across successive connect calls.
+            - Passing a different backend than the one initially set
+                or used in the previous successful connection.
+
+        Parameters
+        ----------
+        mock : bool, optional
+            uses the `MockSignalBackend`, by default False
+            If the backend is supplied then this must always be false.
```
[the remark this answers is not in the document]
evalott100: It is perfectly possible to do:

``` python
signal = Signal()
connect_time_backend=PvaSignalBackend(int)
signal.connect(mock=True, backend=connect_time_backend)
```
This logic isn't mutually exclusive:

https://github.com/bluesky/ophyd-async/blob/5f4a8278ce7e519ef8a504881b346cb42f76dca5/src/ophyd_async/core/signal.py#L75-L83

tests/core/test_mock_signal_backend.py
evalott100: > the truth is that I still don't understand how is the connect() logic supposed to go :(

That's totally fine! I'm happy to explain, in person, or over github.

evalott100:
https://github.com/bluesky/ophyd-async/pull/368#discussion_r1646161096
> what is the type of the `self._previous_connect_was_mock` boolean or an object?

boolean, you're running `backend.connect` as a task whether `backend` is a `MockSignalBackend` or not.

evalott100 (commented):
I don't understand why you're making all these changes to the codebase. None of the mock logic should need to change. All you need to do is make the backend connect a task within the signal.

That means:
``` python
    def __init__(
        self,
        ...
    ) -> None:
        ...
        # add these to init
        self._connect_task = None
        self._previous_connect_was_mock = None
        ...

    async def connect(
        self,
        mock=False,
        timeout=DEFAULT_TIMEOUT,
        backend: Optional[SignalBackend[T]] = None,
    ):
        if backend:
            if self._backend and backend is not self._backend:
                raise ValueError(
                    "Backend given at connection is different from previous"
                )
            self._backend = backend
        if self._backend is None:
            raise RuntimeError("`connect` called on signal without backend")
        if (
            self._previous_connect_was_mock is not None and
            self._previous_connect_was_mock != mock
        ):
            raise RuntimeError(
                f"`connect(mock={mock}) called on a function where the previous "
                f"connect was mock={self._previous_connect_was_mock}. Changing mock "
                "value between connects is not permitted."
            )
        if mock and not isinstance(self._backend, MockSignalBackend):
            # Using a soft backend, look to the initial value
            self._backend = MockSignalBackend(initial_backend=self._backend)


        can_use_previous_connection: bool = (
            self._connect_task is not None
            and not (self._connect_task.done() and self._connect_task.exception())
        )

        if force_reconnect or not can_use_previous_connection:        
            self._connect_task = asyncio.create_task(
                self._backend.connect(timeout=timeout)
            )
            self.log.debug(f"Connecting to {self.source}")

        assert self._connect_task, (
            "this assert is for type analysis and will never fail"
        )
        await self._connect_task
```

stan-dot:
@evalott100 that was passing finally, but I added some more tests and there have issues patching the debug method. I think I have a better grasp of what is going on here, now I'd just like to make those 2 tests pass

src/ophyd_async/core/signal.py
```diff
@@ -63,29 +70,68 @@ def __init__(
...
+            The backend to use, by default None.
+            If the device initalization has no backend, then this is required.
+            This value cannot be changed between connect() calls.
+            It is either the exact same object, or it fails.
+            User mainly for mocks.
+        """
+        can_use_previous_connection: bool = (
+            self._connect_task
+            and not (self._connect_task.done() and self._connect_task.exception())
+            and self._connect_mock_arg == mock
```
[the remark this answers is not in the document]
coretl: I can't remember what I said there, but the idea is that if there is an existing task with the right parameters, await it, otherwise spawn a new one. The use case of cancelling an existing one with the wrong parameters is so niche I don't think we need to support it. Just make a new task and ignore the existing one

tests/core/test_signal.py
```diff
@@ -96,6 +96,29 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_type():
...
+    await signal.connect(mock=True)
+    response = f"Reusing previous connection to {signal.source}"
+    mock_logger_instance.assert_called_once_with(response)
+
+
+async def test_signal_connects_with_force_reconnect():
+    signal = Signal(MockSignalBackend(int))
+    await signal.connect(mock=True)
+    assert signal._backend.datatype == int
+    with patch("ophyd_async.core.device.log.debug") as debug:
```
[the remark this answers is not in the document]
stan-dot: didn't realize that. I really should scan the available mocks prior to writing a test.

Zen of Python 13
> There should be one-- and preferably only one --obvious way to do it.[[c]](https://en.wikipedia.org/wiki/Zen_of_Python#cite_note-10)
tbh it's confusing with pytest as patching can be done with inline 'with patch', or test-level patch, or conftest mocks.

tests/core/test_signal.py
```diff
@@ -81,6 +83,99 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+    await signal.connect(mock=first)
+    assert signal._backend.datatype == int
+    await signal.connect(mock=second)
+    response = f"Connecting to {signal.source}"
+    assert caplog.text.__contains__(response)
+
+
+async def test_signal_lazily_connects(RE):
+    mock_signal_rw = soft_signal_rw(
+        int, "pva://this is an initial value not a pva pv", name="mock_signal"
```
evalott100, line 147:
```suggestion
    signal_rw = soft_signal_rw(
        int, 0, name="mock_signal"
```
The dtype is int.

I meant that you're putting a protocol on this like it's a PV, soft signals don't have PVs.

tests/core/test_signal.py
```diff
@@ -81,6 +83,99 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+    )
+    await mock_signal_rw.connect(mock=False)
+    RE(ensure_connected(mock_signal_rw, mock=True))
+    assert (
+        mock_signal_rw._connect_task
+        and mock_signal_rw._connect_task.done()
+        and not mock_signal_rw._connect_task.exception()
+    )
```
evalott100, line 155:
```suggestion
    await signal_rw.connect()
    RE(ensure_connected(signal_rw))
    assert (
        mock_signal_rw._connect_task
        and mock_signal_rw._connect_task.done()
        and not mock_signal_rw._connect_task.exception()
    )
```
This test would have misleadingly failed I assume... The second connect wouldn't have succeeded since `mock` changed value, however it wouldn't have updated the `connect_task` so the assert here would have shown the first connect task was successful and passed.

I'm assuming you're choosing a `soft` signal over a `mock` signal because you want to mimick the performance of real signals within the connect? I'd like to see a version of this test where you use a real pva signal, passing `mock=True` on both connects.

# Edit

Because of the unset `_previous_connect_was_mock` we don't get failure from changing `mocks`.

src/ophyd_async/core/signal.py
```diff
@@ -73,19 +76,49 @@ async def connect(
...
+                )
+
             if self._initial_backend and backend is not self._initial_backend:
                 raise ValueError(
                     "Backend at connection different from initialised one."
                 )
             self._backend = backend
+        if (
+            self._previous_connect_was_mock is not None
+            and self._previous_connect_was_mock != mock
```
evalott100, line 93: You never set `self._previous_connect_was_mock = mock` after this check.

tests/core/test_signal.py
```diff
@@ -81,6 +83,99 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+
+async def test_connect_fails_then_next_connect_succeeds():
+    mock_signal_rw = soft_signal_rw(int, 0, name="mock_signal")
+    assert not mock_signal_rw._connect_task
+    await mock_signal_rw.connect(mock=False)
+    assert isinstance(mock_signal_rw._backend, SoftSignalBackend)
+    assert mock_signal_rw._connect_task
+    # todo is explicit disconnection needed?
+    await mock_signal_rw.connect(mock=True)
+    assert isinstance(mock_signal_rw._backend, MockSignalBackend)
```
evalott100, line 176: These should both fail, they're not because you aren't setting `_previous_connect_was_mock`.

tests/core/test_signal.py
```diff
@@ -81,6 +83,99 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+    [(True, False), (True, False)],
+)
+async def test_rejects_reconnect_when_connects_have_diff_mock_status(
+    caplog, first, second
+):
+    caplog.set_level(logging.DEBUG)
+    signal = Signal(MockSignalBackend(int))
+    await signal.connect(mock=first)
+    assert signal._backend.datatype == int
+    await signal.connect(mock=second)
```
evalott100, line 140:
```suggestion
    with pytest.raises(RuntimeError) as exc:
       await signal.connect(mock=second)
    assert str(exc) == "f"`connect(mock={second}) called on a function where the previous ..."
```

evalott100 (changes requested):

src/ophyd_async/core/signal.py
```diff
@@ -73,19 +76,49 @@ async def connect(
...
+        ):
+            raise RuntimeError(
+                f"`connect(mock={mock}) called on a function where the previous "
+                f"connect was mock={self._previous_connect_was_mock}. Changing mock "
+                "value between connects is not permitted."
```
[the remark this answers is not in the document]
stan-dot: on it

tests/core/test_signal.py
```diff
@@ -81,6 +83,99 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+    signal = Signal()
+    connect_time_backend = MockSignalBackend(int)
+    await signal.connect(backend=connect_time_backend)  # works
+    with pytest.raises(ValueError):
+        await signal.connect(
+            backend=MockSignalBackend(int)
+        )  # fails since backend is different
+
+    # forces reconnect as per , but passes since the backend is the same
+    await signal.connect(connect_time_backend)
```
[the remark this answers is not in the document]
stan-dot: that is correct

tests/core/test_signal.py
stan-dot: ok

tests/core/test_signal.py
stan-dot: yes those are the two remaining ones that fails. 
`test_connect_fails_then_next_connect_succeeds` and `test_signal_mock_and_back_again`

we either need to delete the tests or change their definition in the test name and then the code

evalott100: Well `test_signal_mock_and_back_again` should check that the above behavior raises an error, which I think you've done in a different test - if so we can delete.

As said previously, `test_connect_fails_then_next_connect_succeeds` doesn't test what the title would imply. `soft_signal_rw` will connect fine if `mock=True` or `mock=False`.

You could create a new connect function just for this test which will fail on the first `connect`, but pass after.

``` python
signal = pva_signal_backend(int, "pva://some_pv_that_doesnt_exist", initial_value="")
original_connect_method = signal.connect
connect_attempts = 0
async def connect(*args, **kwargs):
    # Will cause an error on the first connect call, then pass after
    if connect_attempts == 0:
        connect_attempts += 1
        await original_connect(*args, **kwargs)
signal.connect = connect
```

stan-dot: > You could create a new connect function just for this test 


I think that's uphill and excessive

evalott100 (changes requested):
I made a commit causing `Device` to fail on connect if `mock` changes value. Realised it was probably nicer to fail outright for `Devices` too.

A couple more changes left.

src/ophyd_async/core/signal.py
```diff
@@ -73,19 +75,47 @@ async def connect(
         backend: Optional[SignalBackend[T]] = None,
     ):
         if backend:
-            if self._initial_backend and backend is not self._initial_backend:
+            if self._backend and backend is not self._backend:
+                # test if the backend from a previous connection is the same
+                # reject even if there was no initial backend
                 raise ValueError(
                     "Backend at connection different from initialised one."
```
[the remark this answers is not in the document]
stan-dot: I disagree, the comments explain more context

tests/core/test_signal.py
```diff
@@ -81,6 +83,81 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+    async def new_connect(timeout=1):
+        await asyncio.sleep(0.1)
+        await original_connect(timeout=timeout)
+
+    int_mock_backend.connect = new_connect
+    signal = Signal(int_mock_backend)
+    await signal.connect()
+    assert signal._backend.datatype == int
+    await signal.connect()
+    response = f"Reusing previous connection to {signal.source}"
```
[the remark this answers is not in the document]
stan-dot: ok

tests/core/test_signal.py
```diff
@@ -81,6 +83,81 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_value():
...
+
+async def test_signal_lazily_connects(RE):
+    mock_signal_rw = soft_signal_rw(int, 0, name="mock_signal")
+    await mock_signal_rw.connect(mock=False)
+    RE(ensure_connected(mock_signal_rw, mock=True))
+    assert (
+        mock_signal_rw._connect_task
+        and mock_signal_rw._connect_task.done()
+        and not mock_signal_rw._connect_task.exception()
+    )
```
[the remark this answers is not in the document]
stan-dot: that test is the only one that uses ensure_connected, but seems to pass even if I change is as much as here: 
image

src/ophyd_async/core/signal.py
stan-dot: fixed

src/ophyd_async/core/signal.py
```diff
@@ -63,6 +63,9 @@ def __init__(
     ) -> None:
         self._timeout = timeout
         self._initial_backend = self._backend = backend
+        self._connect_mock_arg = None
+        self._connect_task: Optional[asyncio.Task] = None
+        self._previous_connect_was_mock: Optional[bool] = None
```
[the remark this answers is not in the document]
stan-dot: outdated

src/ophyd_async/core/signal.py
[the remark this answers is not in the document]
stan-dot: fixed

tests/core/test_signal.py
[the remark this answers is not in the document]
stan-dot: used caplog with success

src/ophyd_async/core/signal.py
```diff
@@ -86,16 +89,36 @@ async def connect(
...
+        can_use_previous_connection: bool =(
+            self._connect_task is not None
+            and not (self._connect_task.done() and self._connect_task.exception())
+        )
+
+        if force_reconnect or not can_use_previous_connection:
+            self.log.debug(f"Connecting to {self.source}")
+            self._connect_task = asyncio.create_task(
+                asyncio.sleep(0.001) if mock else self._backend.connect(timeout=timeout)
+            )
```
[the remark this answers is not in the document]
stan-dot: fixed

tests/core/test_signal.py
```diff
@@ -96,6 +96,29 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_type():
...
+    response = f"Reusing previous connection to {signal.source}"
+    mock_logger_instance.assert_called_once_with(response)
+
+
+async def test_signal_connects_with_force_reconnect():
+    signal = Signal(MockSignalBackend(int))
+    await signal.connect(mock=True)
+    assert signal._backend.datatype == int
+    with patch("ophyd_async.core.device.log.debug") as debug:
+        await signal.connect(mock=True, force_reconnect=True)
```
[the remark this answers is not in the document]
stan-dot: deleted as requested

tests/core/test_signal.py
```diff
@@ -96,6 +96,29 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_type():
...
 
+@patch("ophyd_async.core.device.getLogger")
+@patch("ophyd_async.core.device.LoggerAdapter")
+async def test_signal_connects_to_previous_backend(mock_logger_adapter, mock_get_logger):
+    mock_logger_instance = Mock()
+    mock_logger_adapter.return_value = mock_logger_instance
+    signal = Signal(MockSignalBackend(int))
+    await signal.connect(mock=True)
+    assert signal._backend.datatype== int
+    await signal.connect(mock=True)
```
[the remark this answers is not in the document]
stan-dot: outdated

tests/core/test_signal.py
```diff
@@ -96,6 +96,29 @@ async def test_signal_connect_fails_if_different_backend_but_same_by_type():
...
+    response = f"Reusing previous connection to {signal.source}"
+    mock_logger_instance.assert_called_once_with(response)
```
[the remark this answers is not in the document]
stan-dot: outdated

tests/core/test_mock_signal_backend.py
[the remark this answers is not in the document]
stan-dot: fixed and outdated

tests/core/test_signal.py
stan-dot: outdated, the tests got deleted

src/ophyd_async/core/signal.py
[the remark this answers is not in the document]
evalott100: But in this case I would argue the comment is misleading:

`"reject even if there was no initial backend"` - if there were no initial backend and it were the first connection then `self._initial_backend = self._backend = None` at this point and the condition in the if statement above would never be true?

I would also say `"test if the backend from a previous connection is the same"` right above a `ValueError("Backend at connection different from previous one.")` doesn't add anything in terms of readability.

Either way, the string in the error needs to changed from `"intitialised"` to `"previous"` because it will be checking against previous backends in general.

tests/core/test_signal.py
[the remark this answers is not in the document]
evalott100: Can you delete this? I don't understand what you mean with `by_type`? The test case is the same as the one above.

tests/core/test_signal.py
evalott100: But this is the entire reason that mock was switching in those tests that you renamed - it wasn't to test the behavior of the connect function when using mock utilities... It was to see if you can try and fail to connect, then try again and have it work.

tests/core/test_signal.py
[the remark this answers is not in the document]
evalott100: https://github.com/bluesky/ophyd-async/pull/368#discussion_r1652513830

Same as above comment. This will work because it's a soft_signal, it shouldn't work on the first connect.

tests/core/test_signal.py
stan-dot: > I'm assuming you're choosing a soft signal over a mock signal because you want to mimick the performance of real signals within the connect? I'd like to see a version of this test where you use a real pva signal, passing mock=True on both connects.


so `mock=True ` creates a 'real pva signal'? I lost the plot

tests/core/test_signal.py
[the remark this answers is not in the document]
stan-dot: ok, I see now - so here we test a scenario where the device rejects at first but later accepts. 

I think the best way to do this is to have a mock device that has its own connect with this 'fail on first connect' functionality, to mimic a real life device that misbehaves when we try to connect to it.  So a similar class to this one:

image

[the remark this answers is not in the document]
evalott100: Do it by swapping out a `MockSignalBackend`'s connect. That way you'll still be using the regular connect method/functionality of Device and Signal.

src/ophyd_async/core/signal.py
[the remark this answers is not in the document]
stan-dot: outdated

tests/core/test_signal.py
```diff
@@ -258,14 +259,14 @@ async def test_assert_configuration(mock_readable: DummyReadable):
 async def test_signal_connect_logs(caplog):
     caplog.set_level(logging.DEBUG)
     mock_signal_rw = epics_signal_rw(int, "pva://mock_signal", name="mock_signal")
-    await mock_signal_rw.connect(mock=True)
+    await mock_signal_rw.connect(mock=False)
```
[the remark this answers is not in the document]
stan-dot: outdated

tests/core/test_signal.py
stan-dot: outdated

src/ophyd_async/core/signal.py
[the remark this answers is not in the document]
stan-dot: outdated

tests/core/test_signal.py
stan-dot: applied

evalott100 (approved):
Nice job! I don't know if we want to wait for @coretl to give this a once over before merging, but I'm happy with it.

src/ophyd_async/core/signal.py
coretl, line 1: I don't think self._initial_backend is used anywhere. If so it can be deleted

coretl (commented):
One minor comment, otherwise good to go

src/ophyd_async/core/signal.py
stan-dot: deleted

coretl (approved):

PR #369, opened 2024-06-07
Derive PandA describe from datasets PV

### LINKED ISSUE #330: PandA dataset naming
At the moment the PandA HDF writer will always write datasets like "INENC1.VAL.Mean", "COUNTER1.OUT.Diff". These are exposed in the descriptor as "panda1-inenc-1-val-Mean" and "panda1-counter-1-out-Diff". These don't make any scientific sense, so we should allow scientific names to be given to the datasets and set the ophyd name accordingly.

The options:
1. Pass in `HDFPanda(prefix, dp, dataset_names = {"inenc-1-val": "sample-x", "counter-1-out": "izero"})`
2. Post init but pre-connect we get the blocks, but don't know how many there are, and things like "INENC" are difficult as they are target specific
3. Post init and post connect we get connected blocks and can do `panda1.set_datsets({panda1.inenc[1].val: "sample-x"})` in the plan

callumforrester:
Fixes #330 

- Make `HDFPAnda` derive `.describe()` output from the new `DATASETS` PV described in #330 
- Remove code that handled panda capture signals, we now expect the IOC to handle that for us and provide a table of datasets to be captured.

### DIFF
--- a/src/ophyd_async/panda/_table.py
+++ b/src/ophyd_async/panda/_table.py
@@ -6,6 +6,16 @@ import numpy as np
 import numpy.typing as npt
 
 
+class PandaHdf5DatasetType(str, Enum):
+    FLOAT_64 = "float64"
+    UINT_32 = "uint32"
+
+
+class DatasetTable(TypedDict):
+    name: npt.NDArray[np.str_]
+    hdf5_type: Sequence[PandaHdf5DatasetType]
+
+
 class SeqTrigger(str, Enum):
     IMMEDIATE = "Immediate"
     BITA_0 = "BITA=0"
--- a/src/ophyd_async/panda/writers/_hdf_writer.py
+++ b/src/ophyd_async/panda/writers/_hdf_writer.py
@@ -1,8 +1,6 @@
 import asyncio
-from dataclasses import dataclass
-from enum import Enum
 from pathlib import Path
-from typing import Any, AsyncGenerator, AsyncIterator, Dict, List, Optional
+from typing import AsyncGenerator, AsyncIterator, Dict, List, Optional
 
 from bluesky.protocols import DataKey, StreamAsset
 from p4p.client.thread import Context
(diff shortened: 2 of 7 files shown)


src/ophyd_async/panda/_table.py
```diff
@@ -6,6 +6,16 @@
...
 
+class PandaHdf5DatasetType(str, Enum):
+    FLOAT_64 = "float64"
+    UINT_32 = "uint32"
+
+
+class DatasetTable(TypedDict):
+    name: npt.NDArray[np.str_]
+    hdf5_type: Sequence[PandaHdf5DatasetType]
+
```
callumforrester, line 17: @evalott100 I'm not married to these names, feel free to change them

src/ophyd_async/panda/writers/_hdf_writer.py
```diff
@@ -172,6 +77,18 @@ async def open(self, multiplier: int = 1) -> Dict[str, DataKey]:
...
 
+    async def _update_datasets(self) -> None:
+        """
+        Load data from the datasets PV on the panda, update internal
+        representation of datasets that the panda will write.
+        """
+
+        capture_table = await self.panda_device.data.datasets.get_value()
+        self._datasets = [
+            _HDFDataset(dataset_name, dataset_name, [1], multiplier=1)
```
callumforrester, line 88: @evalott100 and @coretl I don't think this is quite right, presuming there should be a prefix or something in the internal path but I'm not sure where to get it from?

coretl: nope, the dataset name is the dataset path, you don't need a leading slash or anything.

E.g. in an existing dataset using `{field_name}.{capture}` as the dataset path
https://github.com/PandABlocks/PandABlocks-client/blob/master/examples/plot_counter_hdf.py

DiamondJoseph (approved):

PR #370, opened 2024-06-10
Add handling for propagating limits from underlying record

DiamondJoseph:
Also adds tests for units, precision propagation

### DIFF
--- a/src/ophyd_async/epics/_backend/common.py
+++ b/src/ophyd_async/epics/_backend/common.py
@@ -1,6 +1,6 @@
-from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, Optional, Tuple, Type
+from math import isnan
+from typing import Dict, Optional, Tuple, Type, TypedDict
 
 common_meta = {
     "units",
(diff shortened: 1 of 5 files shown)


tests/epics/test_signals.py
```diff
@@ -614,3 +619,76 @@ async def test_str_returns_enum(ioc: IOC):
...
+    expected_limits = Limits(
+        # LOW, HIGH, not set
+        alarm=LimitPair(),
+        # LOLO, HIHI
+        warning=LimitPair(low=5, high=96),
+        # DRVL, DRVH, not set
+        control=LimitPair(),
+        # LOPR, HOPR
+        display=LimitPair(low=0, high=100),
+    )
```
DiamondJoseph, line 680: @coretl are these mappings correct?

src/ophyd_async/epics/_backend/common.py
```diff
@@ -1,12 +1,33 @@
...
+    def __bool__(self):
+        return self.high is not None or self.low is not None
+
+
+@dataclass
+class Limits:
+    alarm: LimitPair
+    control: LimitPair
+    display: LimitPair
+    warning: LimitPair
```
DiamondJoseph, line 25: @evalott100 does this match with how you want to do limits in the event-model?

evalott100: It can do, though if you plan to validate these against an event-model schema then I think it's better to make `LimitPair` a `TypedDict`.

tests/epics/test_signals.py
DiamondJoseph: For PVA, with LOW, HIGH, LOLO, HIHI set: `alarm.low=alarm.high=warning.low=warning.high=nan`
For VA, with LOW, HIGH, LOLO, HIHI set: `alarm.low=alarm.high=warning.low=warning.high=0`

src/ophyd_async/epics/_backend/common.py
DiamondJoseph: Adjusted LimitPair and Limits to be TypedDict, made high, low float with nan for when not set, so's to be slightly more predictable (?)

abbiemery (approved):
Approving to be merged into `units_signal_metadata` where the tests will be fixed.

PR #383, opened 2024-06-12
Restructure repository based on ADR

abbiemery:
Based on #26 .


abbiemery:
A few questions/notes:
- Where does the log live, does it live in core or outside of it? I have put it in core for now.
- There are a few private classes which I changed to not private as they were accessed outside of the immediate module space, however for things that are accessed within a space imported like `from ._thing import Class` are we happy having private things accessed like this? `from ._thing import _SneakyClass`. Same with methods.
- For the case where these private classes are not imported directly for normal use, but need to be for tests, do we want to keep them private for make them public.
- Do patches need the full module path with the private modules included? This should be clear once we hit the points of checking tests

abbiemery:
Closed in favor of #467

PR #391, opened 2024-06-17
Handle case of aioca returning None for some alarm limits

callumforrester:
Fixes issues we've seen on the floor @DiamondJoseph to comment with bug report for proper investigation


DiamondJoseph (approved):

DiamondJoseph:
This fixes the behaviour of #392 but will require tests

PR #397, opened 2024-06-18
Refactor stream resource producing code

subinsaji:
This is another copy of #173 with PR #290. This is to take it from a fork and to merge into dev, not main.


abbiemery:
Closed as it was merged into main separately.

PR #401, opened 2024-06-18
398 pilatus acquire period

### LINKED ISSUE #398: Pialtus Controller does not set acquire period
The pilatus should set the acquire period to the exposure time plus the deadtime on arming. This is probably a general case, so the logic could go into a utility function.

callumforrester:
Fixes #398 

This is one possible implementation: There are issues we should consider.

### DIFF
--- a/src/ophyd_async/epics/areadetector/drivers/ad_base.py
+++ b/src/ophyd_async/epics/areadetector/drivers/ad_base.py
@@ -5,6 +5,7 @@ from typing import FrozenSet, Sequence, Set
 from ophyd_async.core import (
     DEFAULT_TIMEOUT,
     AsyncStatus,
+    DetectorControl,
     ShapeProvider,
     set_and_wait_for_value,
 )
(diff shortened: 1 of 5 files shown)


src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -52,6 +54,36 @@ def __init__(self, prefix: str, name: str = "") -> None:
         super().__init__(prefix, name=name)
 
 
+async def set_exposure_time_and_acquire_period_if_supplied(
+    controller: DetectorControl,
```
callumforrester, line 58: We pass the controller here to get the detector's deadtime, which we can then add to the exposure time to make an acquire period. This works but may be misleading as a longer deadtime could be passed into the detector via `TriggerInfo`. It is unclear whether that deadtime should be followed here, it will never be smaller than the result of `get_deadtime` and the `TriggerInfo` is not available from the controller, so some refactoring and potential API changes would be required.

Obeying the output of `get_deadtime` gives us the following if `TriggerInfo` supplies a larger value:
image

Which is quite nice as it gives us the freedom to configure the trriger system to delay the frame to the middle of the duty cycle, which is useful in preventing things like snaking:
image

Using the `TriggerInfo` deadtime seems more restrictive but I believe that is how Malcolm works:
image

callumforrester:
@coretl @evalott100 @abbiemery Opinions welcome

src/ophyd_async/epics/areadetector/drivers/ad_base.py
coretl: Let's use `TriggerInfo.deadtime`. This means we need to consider the interface between `StandardDetector` and `DetectorController`. Opened https://github.com/bluesky/ophyd-async/issues/405 to track

src/ophyd_async/epics/areadetector/drivers/ad_base.py
```diff
@@ -52,6 +54,36 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+    exposure:
+        Desired exposure time, this is a noop if it is None.
+    timeout:
+        How long to wait for the exposure time and acquire period to be set.
+    """
+    if exposure is not None:
+        full_frame_time = exposure + controller.get_deadtime(exposure)
+        await asyncio.gather(
+            driver.acquire_time.set(exposure, timeout=timeout),
+            driver.acquire_period.set(full_frame_time, timeout=timeout),
```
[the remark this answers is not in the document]
coretl: I think you're right, the existing behaviour of bluesky allows you to count a detector without touching exposure or acquire time, so I think that it what we should do if `exposure` is `None`.

src/ophyd_async/epics/areadetector/drivers/ad_base.py
callumforrester: The middle diagram does concern me in that case, how do we prevent snaking artefacts in grid scans?

coretl: I believe I have now clarified this in #405

coretl (approved):

PR #412, opened 2024-06-25
Use enum on aravis driver

### LINKED ISSUE #411: ADAravis detector broken by new enum requirements
Changes introduced in https://github.com/bluesky/ophyd-async/pull/341 have broken aravis initialisation:

```shell
ERROR:blueapi.core.context:
oav: NotConnected:
    coros: NotConnected:
        drv: NotConnected:
            trigger_source: TypeError: BL22I-DI-OAV-01:DET:TriggerSource_RBV has choices ('Freerun', 'Line1', 'FixedRate', 'Software'). Use an Enum or SubsetEnum to represent this.
```

callumforrester:
Fixes #411

### DIFF
--- a/src/ophyd_async/epics/areadetector/drivers/aravis_driver.py
+++ b/src/ophyd_async/epics/areadetector/drivers/aravis_driver.py
@@ -22,6 +22,13 @@ class AravisTriggerMode(str, Enum):
 AravisTriggerSource = Literal["Freerun", "Line1", "Line2", "Line3", "Line4"]
 
 
+class AravisTriggerSource(str, Enum):
+    Freerun = "Freerun"
+    Line1 = "Line1"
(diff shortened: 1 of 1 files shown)


coretl (changes requested):

DiamondJoseph:
Pretty sure we should support LineN for N in {1,2,3,4}.
Docstring should be updated above AravisTriggerSource

callumforrester:
`Line1` was the only thing that came back in the error, I presume if we supported `LineN`, then the connection would break again. Need @coretl's input

coretl:
Manta supports `Line1` and `Line2`. Mako supports only `Line1`. We list the lowest common denominator for type checking, then it's typed as a string at runtime so we can do `sig.set(f"Line{n}")`

coretl (approved):

PR #421, opened 2024-06-28
Made `ensure_connected` fail properly

evalott100:
Requires https://github.com/bluesky/bluesky/pull/1758


evalott100:
I'll add tests for `ensure_connected` passing/failing

src/ophyd_async/plan_stubs/ensure_connected.py
```diff
@@ -10,13 +13,18 @@ def ensure_connected(
     timeout: float = DEFAULT_TIMEOUT,
     force_reconnect=False,
 ):
-    yield from bps.wait_for(
+    connect_task_singleton: List[asyncio.Task] = yield from bps.wait_for(
```
coretl, line 16: Nit: could do

```suggestion
    connect_task, = yield from bps.wait_for(
```

src/ophyd_async/plan_stubs/ensure_connected.py
```diff
@@ -10,13 +13,18 @@ def ensure_connected(
...
+                        mock=mock, timeout=timeout, force_reconnect=force_reconnect
+                    )
                     for device in devices
                 }
             )
         ]
     )
+
+    if connect_task_singleton and connect_task_singleton[0].exception() is not None:
+        raise connect_task_singleton[0].exception()
```
coretl, line 30: Wondering why you check the exception and raise rather than just
```suggestion
    connect_task.result()
```

evalott100: I was doing this before, but decided what I have now makes for nicer reading. Can change it back.

coretl: I think the `result()` will give a slightly nicer stack trace, but not sure...

coretl (approved):

Issue #376, opened 2024-06-11
Make a motor record flyer


abbiemery:
Add `prepare`, `kickoff` and `complete` to `ophyd_async.epics.motor.Motor`. These will allow a constant velocity flyscan to be conducted, where the start, end and velocity of the constant velocity section are supplied.

## prepare
This needs to take the following parameters wrapped in a dataclass:
- start: position where the constant velocity section should begin after the run-up
- end: position where the constant velocity section should end before the run-down
- time: the amount of time to spend in the constant velocity section

We then need to:
- Calculate `self.velocity = abs(end - start) / time`, and check it's less than `self.max_velocity`
- The run-up (and run-down) distance, which is `0.5 * self.acceleration_time * self.velocity`
- Check run-up and run-down are within `self.low_limit_travel : self.high_limit_travel`
- Move to start minus run-up in the right direction, and wait for it to complete

This should be detailed in the docstring for the Motor object

## kickoff
Move to end plus run-down in the right direction

## complete
Wait for it to complete

## plan
We will not change `self.acceleration_time` as there is no reliable way to put it back.

This plan will eventually be extended to add the Panda PCOMP flyer too, but for now:
- Grab current motor velocity
- Prepare motor with start, end, time
- Kickoff
- Complete
- Restore motor velocity

## tests
Need to test both positive and negative moves, too high velocities, and run-up out of range of limits

PR #425, opened 2024-07-01
Make motor flyable

olliesilvester:
Closes https://github.com/bluesky/ophyd-async/issues/376


src/ophyd_async/epics/motion/motor.py
```diff
@@ -93,5 +241,62 @@ async def set(
...
+                self.fly_info.prepared_position < motor_lower_limit
+                or self.fly_info.completed_position > motor_upper_limit
+            )
+        ) or (
+            not self.fly_info.direction
+            and (
+                self.fly_info.prepared_position > motor_lower_limit
+                or self.fly_info.completed_position < motor_upper_limit
+            )
+        ):
```
coretl, line 296: Could: if run_up_distance had the sign of the direction of travel you wouldn't need both sets of if statements below...

src/ophyd_async/epics/motion/motor.py
```diff
@@ -43,13 +121,83 @@ def __init__(self, prefix: str, name="") -> None:
...
+                target=self.fly_info.end_position,
+                unit=units,
+                precision=precision,
+                time_elapsed=round(time() - self.fly_info.complete_time_elapsed, 2),
+                time_remaining=self.fly_info.time_for_move
+                - self.fly_info.complete_time_elapsed,
+            )
+
+            if value >= self.fly_info.end_position:
+                break
```
coretl, line 199: I don't think we care that it has actually started, just that we have kicked it off, so this can be simplified to:
```suggestion
    @AsyncStatus.wrap
    async def kickoff(self):
        """Begin moving motor from prepared position to final position. Mark as
        complete once motor reaches start position"""
        assert self._completed_position, "Motor must be prepared before attempting to kickoff"
        self._fly_status = self.set(self._completed_position)

    def complete(self) -> WatchableAsyncStatus:
        """Send periodic updates of the motor's fly progress. Mark as complete once
        motor reaches end position."""
        assert self._fly_status, "kickoff not called"
        return self._fly_status
```

.pre-commit-config.yaml
```diff
@@ -1,4 +1,14 @@
...
+
+  # Automatic source code formatting
+  - repo: https://github.com/psf/black
+    rev: 24.1.1
+    hooks:
+      - id: black
+        args: [--safe, --quiet]
+        files: \.pyi?$|SConscript$|^libtbx_config$
+        types: [file]
+
```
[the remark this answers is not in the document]
olliesilvester: Whoops, I didn't mean to include this commit, was just testing something

src/ophyd_async/epics/motion/motor.py
```diff
@@ -19,7 +23,81 @@
...
+
+    end_position: float
+        Absolute position of the motor once it begins decelerating from desired
+        velocity, in millimetres
+
+    time_for_move: float
+        Time taken for the motor to get from start_position to end_position, in seconds
+
+    complete_timeout: float
+        Maximum time for the motor 'complete' to finish before throwing an error
```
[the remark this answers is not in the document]
olliesilvester: When I wrote this I was imagining this timeout to be just for the `complete` function, so excluding the run-up and run-down. We could have a separate `fly_timeout` too which does an overall timeout for everything? Or just have a `fly_timeout` and no `complete_timeout`?

src/ophyd_async/epics/motion/motor.py
olliesilvester: clever!

src/ophyd_async/epics/motion/motor.py
olliesilvester: Looks good - this also means `complete` will be done once the motor has travelled to `end_position` + `run_up_distance`. Is that fine? The docstrings for `kickoff` and `complete` are written with respect to acquisition, so is a bit ambiguous to what it should mean for motors

src/ophyd_async/epics/motion/motor.py
```diff
@@ -93,5 +184,48 @@ async def set(
     async def stop(self, success=False):
         self._set_success = success
         # Put with completion will never complete as we are waiting for completion on
-        # the move above, so need to pass wait=False
+        # the move above, so need to pass wait=False.
         await self.motor_stop.trigger(wait=False)
+
+    async def _prepare_velocity(
+        self, start_position, end_position, time_for_move
```
stan-dot, line 191: can we please get type hints here?

src/ophyd_async/epics/motion/motor.py
```diff
@@ -93,5 +184,48 @@ async def set(
...
+        self, fly_velocity: float, start_position: float, end_position: float
+    ) -> float:
+        # Distance required for motor to accelerate from stationary to fly_velocity, and
+        # distance required for motor to decelerate from fly_velocity to stationary
+        run_up_distance = (await self.acceleration_time.get_value()) * fly_velocity
+
+        fly_prepared_position = start_position - run_up_distance
+
+        self._fly_completed_position = end_position + run_up_distance
+
```
stan-dot, line 214: on the first look it's not clear why one variable is put into 'self' and another is not

stan-dot (commented):
small requests for clarification only

src/ophyd_async/epics/motion/motor.py
```diff
@@ -43,13 +88,59 @@ def __init__(self, prefix: str, name="") -> None:
...
+            self.fly_info.end_position,
+            self.fly_info.time_for_move,
+        )
+
+        # start_position with run_up_distance added on.
+        fly_prepared_position = await self._prepare_motor_path(
+            abs(fly_velocity), self.fly_info.start_position, self.fly_info.end_position
+        )
+
+        await self.set(fly_prepared_position)
```
olliesilvester, line 126: Prepare shouldn't ever take very long so it might not be needed, but we could make it run in the background instead, like with `kickoff`

src/ophyd_async/epics/motion/motor.py
```diff
@@ -19,7 +22,49 @@
...
+class FlyMotorInfo(BaseModel):
+    """Minimal set of information required to fly a motor:
+
+    start_position : float
+        Absolute position of the motor once it finishes accelerating to desired
+        velocity, in motor EGUs
+
+    end_position: float
+        Absolute position of the motor once it begins decelerating from desired
+        velocity, in millimetres
```
coretl, line 46: EGUs

src/ophyd_async/epics/motion/motor.py
```diff
@@ -19,7 +22,49 @@
...
+
+
+DEFAULT_MOTOR_FLY_TIMEOUT = 60
+DEFAULT_WATCHER_UPDATE_FREQUENCY = 0.2
+
+
+class FlyMotorInfo(BaseModel):
+    """Minimal set of information required to fly a motor:
+
+    start_position : float
```
coretl, line 40: Please can we put these as `#:` comments above the field so they render in the docs, also please check they render in the docs!

src/ophyd_async/epics/motion/motor.py
```diff
@@ -43,13 +88,60 @@ def __init__(self, prefix: str, name="") -> None:
...
         super().set_name(name)
         # Readback should be named the same as its parent in read()
         self.user_readback.set_name(name)
 
+    @AsyncStatus.wrap
+    async def prepare(self, value: FlyMotorInfo):
+        """Calculate required velocity and run-up distance, then if motor limits aren't
+        breached, move to start position minus run-up distance"""
+
+        self.fly_info = value
```
coretl, line 112: Could: store `_fly_timeout` rather than the whole `fly_info`. If you do store the whole thing it should be private.

src/ophyd_async/epics/motion/motor.py
```diff
@@ -95,3 +187,47 @@ async def stop(self, success=False):
         # Put with completion will never complete as we are waiting for completion on
         # the move above, so need to pass wait=False
         await self.motor_stop.trigger(wait=False)
+
+    async def _prepare_velocity(
+        self, start_position: float, end_position: float, time_for_move: float
+    ) -> float:
+        fly_velocity = (start_position - end_position) / time_for_move
+        max_speed = await self.max_velocity.get_value()
+        egu = await self.motor_egu.get_value()
```
coretl, line 196: In general we should do all pv gets in parallel for performance reasons:
```suggestion
        max_speed, egu = await asyncio.gather(self.max_velocity.get_value(), await self.motor_egu.get_value())
```

src/ophyd_async/epics/motion/motor.py
```diff
@@ -95,3 +187,47 @@ async def stop(self, success=False):
...
+        if abs(fly_velocity) > max_speed:
```
coretl, line 197: Nit: The term speed is not the same as the term velocity... But do we care?

src/ophyd_async/epics/motion/motor.py
```diff
@@ -95,3 +187,47 @@ async def stop(self, success=False):
...
+        # distance required for motor to decelerate from fly_velocity to stationary
+        run_up_distance = (await self.acceleration_time.get_value()) * fly_velocity
+
+        self._fly_completed_position = end_position + run_up_distance
+
+        # Prepared position not used after prepare, so no need to store in self
+        fly_prepared_position = start_position - run_up_distance
+
+        motor_lower_limit = await self.low_limit_travel.get_value()
+        motor_upper_limit = await self.high_limit_travel.get_value()
```
coretl, line 218: ditto for gather

src/ophyd_async/epics/motion/motor.py
```diff
@@ -95,3 +187,47 @@ async def stop(self, success=False):
...
+        egu = await self.motor_egu.get_value()
+
+        if (
+            not motor_upper_limit >= fly_prepared_position >= motor_lower_limit
+            or not motor_upper_limit
+            >= self._fly_completed_position
+            >= motor_lower_limit
```
coretl, line 225: Could go in either direction, so need to check:
```python
for position in fly_prepared_position, self._fly_completed_position:
    if not (motor_lower_limit <= position <= motor_upper_limit):
        raise MotorLimitsExecption(...)
```

coretl (changes requested):
Almost there, some minor fixes and performance suggestions

src/ophyd_async/epics/motion/motor.py
olliesilvester: Yeah, the motor signal should really be called `max_speed`, but I didn't think this was worth changing this

coretl (approved):

PR #429, opened 2024-07-03
173 stream resource seperate classes

### LINKED ISSUE #173: Factor StreamResource producing code into a separate class

subinsaji:
Fixes #173


.github/workflows/_test.yml
coretl, line 1: Please can you drop the changes in this file

.github/workflows/ci.yml
coretl, line 1: And in this one

.github/workflows/_test.yml
abbiemery: I was hoping to get the streamresource refactor done here only, then do the removal of the old event model in another issue. The fiddly work for the switching has already been done, so I'm going to help through to get it in. 

It is preferable to have two atomic commits since it is possible and not much more work, one where the internal stuff is changing and the other where a dependency is.

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
```diff
@@ -17,8 +17,9 @@
 class _HDFDataset:
     data_key: str
     dataset: str
-    swmr: bool = False
+    shape: Sequence[int] = field(default_factory=tuple)
     multiplier: int = 1
+    swmr: bool = False
```
subinsaji, line 23: Correction to the commit message - add back shape and move `swmr` to the `bottom`

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
coretl, line 1: Eventually this should move to core as it is not areaDetector specific. Will you be doing that in the refactor PR tho?

src/ophyd_async/panda/writers/_hdf_writer.py
coretl, line 1: Shape and Multipliers are wrong in this file, but I've raised a separate issue to track https://github.com/bluesky/ophyd-async/issues/431

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -231,10 +156,7 @@ async def open_file(
         datasets = self._get_datasets()
         for d in datasets:
             self._handle_for_h5_file.create_dataset(
-                name=d.name,
-                shape=d.shape,
-                dtype=d.dtype,
-                maxshape=d.maxshape,
```
coretl, line 237: Why has maxshape disappeared here?

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -81,74 +64,16 @@ def generate_interesting_pattern(x: float, y: float) -> float:
...
 class PatternGenerator:
     def __init__(
         self,
         saturation_exposure_time: float = 1,
         detector_width: int = 320,
         detector_height: int = 240,
     ) -> None:
+        self.maxshape: tuple[Any, ...]
+        self.fillvalue: int = field(default_factory=int)
+        self.shape: List[int] = field(default_factory=[])
```
coretl, line 76: These look like they were copied from a dataclass, where are their values?

coretl (changes requested):

src/ophyd_async/epics/areadetector/writers/general_hdffile.py
abbiemery: Yes that will be addressed in the refactor

src/ophyd_async/panda/writers/_hdf_writer.py
abbiemery: Noted :).

src/ophyd_async/sim/pattern_generator.py
subinsaji: Is something like this what we want?

```suggestion
        self.maxshape: tuple[Any, ...] = (1,)
        self.fillvalue: int = 1
        self.shape: List[int] = [1, ...]
```

abbiemery: I have pushed this change.

src/ophyd_async/sim/pattern_generator.py
abbiemery: This has since been discussed in a meeting, we are going to address the larger issue of the pattern generator in a following PR. https://github.com/bluesky/ophyd-async/issues/432

src/ophyd_async/sim/pattern_generator.py
```diff
@@ -71,9 +63,9 @@ def __init__(
         detector_width: int = 320,
         detector_height: int = 240,
     ) -> None:
-        self.maxshape: tuple[Any, ...]
-        self.fillvalue: int = field(default_factory=int)
-        self.shape: List[int] = field(default_factory=[])
+        self.maxshape: tuple[Any, ...] = (1,)
+        self.fillvalue: int = 1
+        self.shape: List[int] = [1, 1]
```
coretl, line 68: For the next ticket:
- shape of a frame is `[detector_height, detector_width]`
- shape of the DATA HDF dataset is `[1, detector_height, detector_width]`
- fill value for SUM dataset is -1, not required for the dataset
- maxshape for the DATA HDF dataset = `[None, detector_height, detector_width]`

As per https://github.com/dls-controls/bluefly/blob/0ee7ea837d5f74e3cdc69e2315e9beed28dd7766/bluefly/areadetector_sim.py#L55-L68

coretl (approved):
I imagine this will break the pattern generator, but it's more important to get this in as it's holding other things in. We can fix the pattern generator in another ticket

PR #430, opened 2024-07-03
Adds a panda flyer that is implemented with a pcomp block

evalott100:


src/ophyd_async/panda/_common_blocks.py
```diff
@@ -22,6 +23,39 @@ class PulseBlock(Device):
...
+    label: SignalRW[str]
+    out: SignalRW[bool]
+    pre_start: SignalRW[int]
+    produced: SignalR[int]
+    pulses: SignalRW[int]
+    relative: SignalRW[PcompRelativeOptions]
+    start: SignalRW[int]
+    state: SignalR[str]
+    step: SignalRW[int]
+    width: SignalRW[int]
```
coretl, line 56: We only want to put the things that we set from code here, so start, step, width, not enable_delay, pre_start, etc.

evalott100:
@coretl Should I wait for https://github.com/bluesky/ophyd-async/issues/376 to add a test with flying?

coretl:
> @coretl Should I wait for #376 to add a test with flying?

I think you should be able to do the test and the plan stub without #376, maybe try it and see?

evalott100:
@coretl I've added a test that mirrors this:

https://github.com/bluesky/ophyd-async/blob/dbbcf28c53fc92170adbda114a26628bf3f3e76a/tests/panda/test_trigger.py#L28-L33

But I think we should probably remove them, I'm not sure what the purpose of this is (checking an object has the methods defined in the class).

src/ophyd_async/panda/_trigger.py
```diff
@@ -37,3 +43,38 @@ async def complete(self) -> None:
...
         await self.seq.enable.set("ZERO")
         await wait_for_value(self.seq.active, False, timeout=1)
+
+
+@dataclass
+class PcompInfo:
+    start_postion: int  # start position in counts
+    pulse_width: int  # width of a single pulse in counts
+    rising_edge_step: int  # step between rising edges of pulses in counts
+    number_of_pulses: int
```
olliesilvester, line 53:
```suggestion
    number_of_pulses: int #Number of pulses to send before disarming the PCOMP block. 0 means go on forever
```

src/ophyd_async/panda/_trigger.py
```diff
@@ -37,3 +43,38 @@ async def complete(self) -> None:
...
+    direction: PcompDirectionOptions  # direction positive or negative
```
olliesilvester, line 54:
```suggestion
    direction: PcompDirectionOptions  # Specifies which direction the motor counts should be moving. Pulses won't be sent unless the values are moving in this direction
```

src/ophyd_async/panda/_trigger.py
```diff
@@ -37,3 +43,38 @@ async def complete(self) -> None:
...
+            self.pcomp.pulses.set(value.number_of_pulses),
+            self.pcomp.dir.set(value.direction),
+        )
+
+    async def kickoff(self) -> None:
+        await self.pcomp.enable.set("ONE")
+        await wait_for_value(self.pcomp.active, True, timeout=1)
+
+    async def complete(self) -> None:
+        await wait_for_value(self.pcomp.active, False, timeout=None)
```
olliesilvester, line 76: Do we want an optional timeout here?

evalott100: Sounds good

src/ophyd_async/panda/_trigger.py
olliesilvester: Or maybe

`#Number of pulses to send before the PCOMP block is disarmed. 0 means infinite`

coretl:
> @coretl I've added a test that mirrors this:
> 
> https://github.com/bluesky/ophyd-async/blob/dbbcf28c53fc92170adbda114a26628bf3f3e76a/tests/panda/test_trigger.py#L28-L33
> 
> But I think we should probably remove them, I'm not sure what the purpose of this is (checking an object has the methods defined in the class).

Not much point in tests like that

src/ophyd_async/panda/_trigger.py
```diff
@@ -37,3 +43,38 @@ async def complete(self) -> None:
...
+    direction: PcompDirectionOptions  # direction positive or negative
+
+
+class StaticPcompTriggerLogic(TriggerLogic[PcompInfo]):
+    def __init__(self, pcomp: PcompBlock) -> None:
+        self.pcomp = pcomp
+
+    async def prepare(self, value: PcompInfo):
+        await self.pcomp.enable.set("ZERO")
+        asyncio.gather(
```
coretl, line 63:
```suggestion
        await asyncio.gather(
```

src/ophyd_async/plan_stubs/fly.py
```diff
@@ -113,6 +151,29 @@ def fly_and_collect(
...
+def fly_and_collect_with_static_pcomp(
+    stream_name: str,
+    flyer: HardwareTriggeredFlyable[PcompInfo],
+    detectors: List[StandardDetector],
+    number_of_pulses: int,
+    pulse_width: int,
+    rising_edge_step: int,
+    direction: PcompDirectionOptions,
+
+):
```
coretl, line 163: We need to revisit all of these plan stubs, but not in this ticket...

src/ophyd_async/core/detector.py
```diff
@@ -51,20 +51,25 @@ class DetectorTrigger(str, Enum):
...
+    livetime: float = Field(ge=0)
     #: What is the maximum timeout on waiting for a frame
-    frame_timeout: float | None = None
+    frame_timeout: float | None = Field(None, gt=0)
+
+    @field_validator("trigger", mode="before")
+    def convert_enum_to_string(cls, value):
+        if issubclass(type(value), Enum):
+            return value.value
+        return value
```
coretl, line 72: I'm interested in knowing what this is needed for?

src/ophyd_async/core/detector.py
```diff
@@ -51,20 +51,25 @@ class DetectorTrigger(str, Enum):
...
 
 
-@dataclass(frozen=True)
-class TriggerInfo:
+class TriggerInfo(BaseModel):
     """Minimal set of information required to setup triggering on a detector"""
 
     #: Number of triggers that will be sent
-    num: int
+    number: int = Field(gt=0)
```
coretl, line 58: We should document here that `==0` means go forever as this is the convention areaDetector chose.
```suggestion
    #: Number of triggers that will be sent, 0 means infinite
    number: int = Field(ge=0)
```

src/ophyd_async/panda/_common_blocks.py
```diff
@@ -22,6 +23,30 @@ class PulseBlock(Device):
...
 
+class PcompDirectionOptions(str, Enum):
+    positive = "Positive"
+    negative = "Negative"
+    either = "Either"
+
+
+class PcompRelativeOptions(str, Enum):
+    absolute = "Absolute"
+    relative = "Relative"
```
coretl, line 34: Is this used?

src/ophyd_async/panda/_table.py
```diff
@@ -54,23 +56,23 @@ class SeqTableRow:
...
+    outd1: NotRequired[pnd.Np1DArrayBool]
+    oute1: NotRequired[pnd.Np1DArrayBool]
+    outf1: NotRequired[pnd.Np1DArrayBool]
+    time2: NotRequired[pnd.Np1DArrayUint32]
+    outa2: NotRequired[pnd.Np1DArrayBool]
+    outb2: NotRequired[pnd.Np1DArrayBool]
+    outc2: NotRequired[pnd.Np1DArrayBool]
+    outd2: NotRequired[pnd.Np1DArrayBool]
+    oute2: NotRequired[pnd.Np1DArrayBool]
+    outf2: NotRequired[pnd.Np1DArrayBool]
```
coretl, line 75: I'm not sure why this needs to change. We have 2 use cases for array fields:
- A `TypedDict` with fields `npt.NDArray[dtype]` that corresponds to a PVA table
- A `BaseModel` with fields that are numpy arrays according to how pydantic-numpy wants to define them

I thought this was the first of those, so I'm confused as to why we have pydantic-numpy things in it? Did I miss something?

src/ophyd_async/panda/_trigger.py
```diff
@@ -1,15 +1,23 @@
...
+)
 
 
-@dataclass
-class SeqTableInfo:
-    sequence_table: SeqTable
-    repeats: int
-    prescale_as_us: float = 1  # microseconds
+class SeqTableInfo(BaseModel):
+    sequence_table: SeqTable = Field(strict=True)
```
coretl, line 18: Ah, is it to do the checking here? In which case maybe we need to reconsider the use of `TypedDict` for PVA tables and switch completely to pydantic models... We could define a `Table(BaseModel)` with a suitable validator for lengths.

Although, after the conversation we had with @danielballan we could ditch `TypedDict` in favour of a row-wise numpy structured datatype, which would enforce column lengths and avoid the need to pydantic models in PVA structures...

I think this needs discussion in person

src/ophyd_async/panda/_trigger.py
```diff
@@ -37,3 +45,57 @@ async def complete(self) -> None:
...
+            "moving. Pulses won't be sent unless the values are moving in "
+            "this direction."
+        )
+    )
+
+    @field_validator("direction", mode="before")
+    def convert_enum_to_string(cls, value):
+        if issubclass(type(value), Enum):
+            return value.value
+        return value
```
coretl, line 75: ditto

src/ophyd_async/plan_stubs/fly.py
```diff
@@ -6,8 +6,45 @@
...
+    This prepares all supplied detectors with the same trigger.
+
+    """
+    trigger_info = PcompInfo(
+        start_postion=0,
+        pulse_width=pulse_width,
+        rising_edge_step=rising_edge_step,
+        number_of_pulses=number_of_pulses,
+        direction=direction,
+    )
```
coretl, line 42: This is a `PcompInfo`, not a `TriggerInfo`. The `StandardDetector`s need a `TriggerInfo`, only the flyer needs the `PcompInfo`

src/ophyd_async/panda/_table.py
evalott100: This allows for validation in `SeqTableInfo`:

https://github.com/bluesky/ophyd-async/blob/a8ad9bf27f41e468f22a1c4e2c915d9a67197bca/src/ophyd_async/panda/_trigger.py#L16-L19

Pydantic needs the type annotation from `pydantic_numpy` in the `SeqTable` in order to validate on it. It doesn't seem to check the exact type of the numpy array (`bool`, `uint32`, etc.) but it will validate based on whether it's a numpy array.

`NotRequired` is added since a plan may only pack `SeqTable` with a subset of rows relevant to the experiment.

src/ophyd_async/panda/_trigger.py
evalott100: Yes, I was thinking of bringing up allowing pydantic model classes as signal datatypes.

coretl: I'm now favouring the row-wise numpy structured datatype. I will raise a separate issue to track, so please can you put it back to `TypedDict` for now and skip the checking of the seq table in the `SeqTableInfo` class?

src/ophyd_async/plan_stubs/fly.py
coretl: Thinking about this, it's not exactly clear how we generate a `TriggerInfo` from a `PcompInfo`. `number` is fine, but we don't know the `live_time` or `dead_time` as they are variable. We might have to define a `PULSE` block per detector to give it some time based pulses that it needs. For now then hardcoding these values will do, and we can tackle in another ticket.

src/ophyd_async/panda/_trigger.py
coretl: Created #442 to track

src/ophyd_async/plan_stubs/fly.py
coretl: Actually, your solution of passing the `TriggerInfo` in from the plan neatly solves this. The plan knows how fast the motor is going. The plan knows how many trigger it asked PCOMP to make, so it can create a `TriggerInfo` telling the detector whether to use it as a trigger or a gate signal. We can defer making another ticket until we get to the double buffered sequencer position compare.

src/ophyd_async/panda/_table.py
coretl: As discussed on slack, I think we should put this back to how it was, then disable strict checking of the `sequence_table` field above until #442 is solved

coretl (approved):

PR #436, opened 2024-07-05
432 fix pattern generator tests

### LINKED ISSUE #432: Pattern Generator dataset creation is not correct, and has no tests to catch it.
The changes in #429 broke the pattern generator, and this issue is to get it working again with appropriate system test coverage. 

The tests need to cover its use as it would be run using the demo.

Changes required for the init of `PatternGenerator`:
- shape of a frame is [detector_height, detector_width]
- shape of the DATA HDF dataset is [1, detector_height, detector_width]
- fill value for SUM dataset is -1, not required for the dataset
- maxshape for the DATA HDF dataset = [None, detector_height, detector_width]

As in the original code in bluefly https://github.com/dls-controls/bluefly/blob/0ee7ea837d5f74e3cdc69e2315e9beed28dd7766/bluefly/areadetector_sim.py#L55-L68

subinsaji:
Fixes #432

### DIFF
--- a/src/ophyd_async/sim/pattern_generator.py
+++ b/src/ophyd_async/sim/pattern_generator.py
@@ -63,9 +63,6 @@ class PatternGenerator:
         detector_width: int = 320,
         detector_height: int = 240,
     ) -> None:
-        self.maxshape: tuple[Any, ...] = (1,)
-        self.fillvalue: int = 1
-        self.shape: List[int] = [1, 1]
         self.saturation_exposure_time = saturation_exposure_time
         self.exposure = saturation_exposure_time
         self.x = 0.0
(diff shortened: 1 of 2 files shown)


subinsaji:
There is a lot of overlapping logic with `hdf_writer.py` : https://github.com/bluesky/ophyd-async/blob/0269697acd9de1ff645fdc40aaf7937c6f346bd4/src/ophyd_async/epics/areadetector/writers/hdf_writer.py#L81-L112

and also with panda `_hdf_writer.py` : https://github.com/bluesky/ophyd-async/blob/0269697acd9de1ff645fdc40aaf7937c6f346bd4/src/ophyd_async/panda/writers/_hdf_writer.py#L69-L79

From a DRY point of view, we can reduce this repeating logic

coretl (approved):

PR #437, opened 2024-07-08
Tango support

burkeds:
Enable support for asynchronous control of Tango servers. 

The Tango support backend complies with the Signal object to interface with Tango servers via PyTango device proxies. 

The TangoDevice is an optional device class which inherits from Device. It includes attributes and methods to accommodate PyTango device proxies and to manage device polling for servers that do not support events. TangoReadable is an extension of the StandardReadable class to support Tango.


coretl:
Before I review this in detail, I thought I would try to describe the type hint approach we use for PandA, and how it could be applied to Tango:

```python
    StandardReadable,
    SignalRW,
    SignalX,
    HintedSignal,
    ConfigSignal,
    DEFAULT_TIMEOUT,
    Device,
)

# This import should go in core after the refactor...
from ophyd_async.epics.pvi import create_children_from_annotations
import inflection


# This goes somewhere in ophyd_async.tango
async def fill_tango_signals(
    device: Device, trl: str, timeout=DEFAULT_TIMEOUT, mock=False
):
    if mock:
        # There is some code in ophyd_async.epics.pvi that we can use for this
        ...
    else:
        # Get a tango proxy and some details about the attributes
        device_proxy = await get_tango_trl(trl)
        configs = await get_tango_configs(trl)
        for attr_trl in device_proxy.get_attribute_list():
            # Convert MyAttribute -> my_attribute
            attr_name = inflection.underscore(attr_trl)
            if not hasattr(device, attr_name):
                # Make a blank signal of the right type, not sure how to tell the RW mode of attributes?
                setattr(device, attr_name, SignalRW())
            # Get the signal off the device and set its backend
            signal = getattr(device, attr_name)
            signal.set_backend(TangoTransport(configs[attr_trl].dtype, device_proxy))
        # Now do similar for commands

    # We call set name now the parent field has been set in all of the
    # introspect-initialized devices. This will recursively set the names.
    device.set_name(device.name)


class SIS3820Counter(StandardReadable):
    counts: SignalRW[float]
    offset: SignalRW[float]
    reset: SignalX

    def __init__(self, trl: str, name: str = "") -> None:
        self._trl = trl
        # Create Signal* objects from annotations but don't set their backends yet
        create_children_from_annotations(self)
        # Register particular signals to contribute to read and read_configuration
        self.add_readables([self.counts], HintedSignal)
        self.add_readables([self.offset], ConfigSignal)
        super.__init__(name=name)

    async def connect(
        self, mock: bool = False, timeout: float = DEFAULT_TIMEOUT
    ) -> None:
        # Fill in the tango backend for the signals we made above, and
        # make new signals for every attribute the tango device made that we didn't mention above
        await fill_tango_signals(self, self._trl, timeout=timeout, mock=mock)
        # Now connect them all in parallel
        await super().connect(mock=mock, timeout=timeout)
```

burkeds:
As of now, tango_signals can automatically infer their types if not specified. I am working on changing TangoReadableDevice::register_signals to automatically instantiate the signals.

burkeds:
@coretl Is there a non-protected method of Signal to set the backend? Currently, to set the backend of the signal I simply do `signal._backend = backend` but it conflicts with Ruff SLF001 to access a protected member of Signal.

burkeds:
@coretl tango devices can now use pytango to automatically infer datatype, signal type, and set a backend. Signals can have the usual `SignalR`,` SignalX`, etc. type hints but `TangoReadableDevice` also supports the typehints `Signal` and `None`. If signals are hinted with `Signal` or `None` then the R/W/X character will also be inferred from the device server and used to set the signal frontend along with the backend.

```python
class TangoMover(TangoReadableDevice, Movable, Stoppable):
    # Enter the name and type of the signals you want to use
    # If type is None or Signal, the type will be inferred from the Tango device
    position: SignalRW
    velocity: SignalRW
    state: SignalR
    _stop: SignalX

    def __init__(self, trl: str, name=""):
        super().__init__(trl, name=name)
        self.add_readables([self.position], HintedSignal)
        self.add_readables([self.velocity], ConfigSignal)
```

coretl:
> @coretl Is there a non-protected method of Signal to set the backend? Currently, to set the backend of the signal I simply do `signal._backend = backend` but it conflicts with Ruff SLF001 to access a protected member of Signal.

No, but you can omit it at signal creation, and supply it at connect instead:
https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/core/_signal.py#L61-L67

src/ophyd_async/tango/_backend/_tango_transport.py
```diff
@@ -0,0 +1,736 @@
...
+    def __init__(
+        self,
+        datatype: Optional[Type[T]],
+        read_trl: str,
+        write_trl: str,
+        device_proxy: Optional[DeviceProxy] = None,
+    ):
+        self.datatype = datatype
+        self.read_trl = read_trl
+        self.write_trl = write_trl
```
coretl, line 629: I don't know much about Tango, are the read_trl and write_trl ever different?

src/ophyd_async/tango/_backend/_tango_transport.py
```diff
@@ -0,0 +1,736 @@
...
+        # all pipes can be always accessible with low register
+        all_pipes = [pipe_name.lower() for pipe_name in device_proxy.get_pipe_list()]
+        if trl_name in all_pipes:
+            raise NotImplementedError("Pipes are not supported")
+
+    raise RuntimeError(f"{trl_name} cannot be found in {device_proxy.name()}")
+
+
+# --------------------------------------------------------------------
+class TangoTransport(SignalBackend[T]):
```
coretl, line 619: Please can we call this:
```suggestion
class TangoSignalBackend(SignalBackend[T]):
```

src/ophyd_async/tango/_backend/_tango_transport.py
coretl, line 1: Nit: We restructured `ophyd_async.epics.signal` to contain both the `epics_signal_*` helpers, and the CA and PVA `SignalBackend`s. For consistency it would be nice to do the same here and put this in `ophyd_async.tango.signal`

src/ophyd_async/tango/demo/counter.py
```diff
@@ -0,0 +1,52 @@
...
+        self.add_readables([self.counts], HintedSignal.uncached)
+        self.add_readables([self.sample_time], ConfigSignal)
+
+    def trigger(self):
+        return AsyncStatus(self._trigger())
+
+    async def _trigger(self):
+        sample_time = await self.sample_time.get_value()
+        timeout = sample_time + DEFAULT_TIMEOUT
+        await self.start.trigger(wait=True, timeout=timeout)
```
coretl, line 42:
```suggestion
    @AsyncStatus.wrap
    async def trigger(self) -> None:
        sample_time = await self.sample_time.get_value()
        timeout = sample_time + DEFAULT_TIMEOUT
        await self.start.trigger(wait=True, timeout=timeout)
```

src/ophyd_async/tango/demo/counter.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass
+class TangoCounterConfig:
+    sample_time: Optional[float] = None
+
+
+class TangoCounter(TangoReadableDevice):
+    # Enter the name and type of the signals you want to use
+    # If type is None or Signal, the type will be inferred from the Tango device
+    counts: None
+    sample_time: Signal
```
coretl, line 26: When we do this in PandA, we specify not just the names of the signals, but their types, e.g.:
```suggestion
    sample_time: SignalRW[float]
```
so that when we `await self.sample_time.get_value()` we know we are getting a float. We then check at connection time that the EPICS PV is actually a float, and error if not. This means we can catch things like `32 + await self.my_string_attribute.get_value()` at lint time rather than runtime

Is it possible to do the same here?

src/ophyd_async/tango/base_devices/base_device.py
```diff
@@ -0,0 +1,113 @@
...
+    infer_signal_frontend,
+    make_backend,
+)
+from tango.asyncio import DeviceProxy
+
+__all__ = ("TangoReadableDevice",)
+
+
+# --------------------------------------------------------------------
+class TangoReadableDevice(StandardReadable):
```
coretl, line 26: I'm wondering if it makes sense to make a `class TangoDevice(Device)`, then have `class TangoMover(TangoDevice, StandardReadable)` which would allow you to make classes that are not `StandardReadable`s (maybe like LIMA plugins?) where they are intended to be children of `StandardDetector` rather than being read in a standalone way?

src/ophyd_async/tango/base_devices/base_device.py
```diff
@@ -0,0 +1,113 @@
...
+                elif obj_type is Signal or obj_type is None:
+                    tango_name = attr_name.lstrip("_")
+                    setattr(
+                        self,
+                        attr_name,
+                        infer_signal_frontend(trl=f"{self.trl}/" f"{tango_name}"),
+                    )
+                else:
+                    print(obj_type, type(obj_type))
+                    raise ValueError(f"Invalid signal type {obj_type}")
```
coretl, line 113: Could you use https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/epics/pvi/_pvi.py#L309-L341 if we put it in `ophyd_async.core`?

src/ophyd_async/tango/_backend/_tango_transport.py
coretl: I think `Transport` is a leftover from when this was in `ophyd.v2`...

src/ophyd_async/tango/demo/README.md
```diff
@@ -0,0 +1,56 @@
...
+    await motor1.connect()
+    await counter1.connect()
+    await counter2.connect()
+    
+    # Events are not supported by the test context so we disable them
+    motor1.position._backend.allow_events(False)
+    motor1.state._backend.allow_events(False)
+    # Enable polling for the position and state attributes
+    motor1.position._backend.set_polling(True, 0.1, 0.1)
+    motor1.state._backend.set_polling(True, 0.1)
```
coretl, line 49: Could we wrap this up at the `Device` or `Signal` level? ...

src/ophyd_async/tango/demo/mover.py
```diff
@@ -0,0 +1,111 @@
...
+    DevState,
+)
+
+
+@dataclass
+class TangoMoverConfig:
+    velocity: Optional[float] = None
+
+
+class TangoMover(TangoReadableDevice, Movable, Stoppable):
```
coretl, line 34: ... e.g. I could imagine
```suggestion
@tango_poll_attributes(0.1)
class TangoMover(TangoReadableDevice, Movable, Stoppable):
```
which could set `TangoMover._poll_attribute_rate = 0.1`, then `connect` could pass that information down when it creates the `TangoSignalBackend`.

What do you think?

coretl (commented):
Some suggestions, let me know what you think.

Some more code could be shared after #472, although that should be done after this is merged, so it might change too

src/ophyd_async/tango/_backend/_tango_transport.py
burkeds: I do not know of any examples of this, but yes it is possible. For example, somebody might write a Tango server where `position` is read-only and one must write to `setpoint` or similar to change the position.

burkeds:
@coretl Something that Tango supports is associating Tango commands with typed values. For example, a command like "RunScript" might be a `str` typed command that carries a filepath to a script and returns a string like `Script executed`. Since SignalX only expects `None` typed commands, as a workaround one must treat these commands as SignalRW in ophyd-async. This seems to work ok but it means that ophyd-async will always treat them like attributes. Is there a possibility to introduce typed commands in a future PR?

src/ophyd_async/tango/base_devices/base_device.py
burkeds: @coretl this doesn't seem to work at the moment for StandardReadables. This throws an error for _readables because it is a Tuple.

In principle yes, we could use this but it does not allow for inferring the signal frontend and datatype from tango. This is fine but it means we wouldn't be taking advantage of this Tango feature. My instinct would be to overwrite this with a tango version. Perhaps it could accept a callback method to run if you want the frontend to be inferred.

```python
  File "/home/burkedev/projects/BLUESKY/ophyd-async/src/ophyd_async/tango/base_devices/base_device.py", line 50, in __init__
    create_children_from_annotations(self)
  File "/home/burkedev/projects/BLUESKY/ophyd-async/src/ophyd_async/epics/pvi/_pvi.py", line 342, in create_children_from_annotations
    sub_device = device_type()
                 ^^^^^^^^^^^^^
  File "/usr/lib/python3.11/typing.py", line 1246, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
TypeError: Type Tuple cannot be instantiated; use tuple() instead
```

src/ophyd_async/tango/base_devices/base_device.py
burkeds: This makes sense. As suggested I created `TangoDevice(Device)` which contains most of the cost that was previously in `TangoReadableDevice`. `TangoReadableDevice` was redefined to simply inherit from TangoDevice and StandardReadable so that it calls the TangoDevice init method.

src/ophyd_async/tango/demo/mover.py
burkeds: That seems like a good idea. I created a tango_polling class decorator that will disable events and enable polling in this way.

src/ophyd_async/tango/demo/README.md
burkeds: Per the other thread, this can now be defined using a class decorator.

src/ophyd_async/tango/_backend/_tango_transport.py
burkeds: I moved these methods to tango.signal and renamed many of the modules to better resemble the rest of the codebase.

burkeds:
> > @coretl Is there a non-protected method of Signal to set the backend? Currently, to set the backend of the signal I simply do `signal._backend = backend` but it conflicts with Ruff SLF001 to access a protected member of Signal.
> 
> No, but you can omit it at signal creation, and supply it at connect instead:
> 
> https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/core/_signal.py#L61-L67

The device level connect calls the connect method of child devices but does not offer a way to pass those children new arguments like backend. It looks like I would have to overwrite Device.connect() entirely.

src/ophyd_async/tango/base_devices/_base_device.py
```diff
@@ -0,0 +1,122 @@
...
+from ophyd_async.tango.signal import (
+    infer_python_type,
+    infer_signal_frontend,
+    make_backend,
+)
+from tango import DeviceProxy as SyncDeviceProxy
+from tango.asyncio import DeviceProxy as AsyncDeviceProxy
+
+
+class TangoDevice(Device):
```
callumforrester, line 19: If possible, I think we should keep the general approach to both epics an tango devices in rough step. That does not mean this PR should squeeze a tango-shaped implementation into an epics-shaped hole. 

For example though, there isn't a generic `EpicsDevice` base class, should there be? Is this a result of a difference between the two control systems or developer-preference? Is there some paradigm we can find where the two implementations look roughly similar, and so the codebase is easier to navigate?

I have picked on this particular example but this comment applies to the whole PR.

src/ophyd_async/tango/base_devices/_base_device.py
```diff
@@ -0,0 +1,122 @@
...
+                if issubclass(type(signal), Signal):
+                    tango_name = name.lstrip("_")
+                    read_trl = f"{self.trl}/{tango_name}"
+                    datatype = infer_python_type(read_trl)
+                    backend = make_backend(
+                        datatype=datatype,
+                        read_trl=read_trl,
+                        write_trl=read_trl,
+                        device_proxy=self.proxy,
+                    )
```
callumforrester, line 91: Another example of the difference vs. epics (see comment above). Here we make the signal backends in one go in the base class, rather than in the signals themselves. Is this a fundamental requirement of tango or just developer preference? Is there a way we can bring the two implementations more inline?

src/ophyd_async/tango/demo/_counter.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TangoCounter(TangoReadable):
+    # Enter the name and type of the signals you want to use
+    # If type is None or Signal, the type will be inferred from the Tango device
+    counts: Signal
+    sample_time: Signal
+    state: Signal
+    reset: Signal
+    start: SignalX
```
callumforrester, line 27: I think the requirement for type hints is a strength of ophyd-async in general, where we have control systems that are loosely defined, being able to say `counts = Signal[float]` and knowing I will get a sensible error message at connect time if the control system does not actually return a float is very useful. I wouldn't want to see that requirement lost here. See also my other comments about maintaining consistency with the rest of the codebase.

src/ophyd_async/tango/demo/README.md
callumforrester, line 1: This should go in the `docs` directory.

src/ophyd_async/tango/base_devices/_base_device.py
```diff
@@ -0,0 +1,122 @@
...
+                    trl = f"{self.trl}/{tango_name}"
+                    setattr(
+                        self, attr_name, infer_signal_frontend(trl=trl, name=attr_name)
+                    )
+                elif issubclass(obj_type, Signal):
+                    setattr(self, attr_name, obj_type(name=attr_name))
+
+
+# --------------------------------------------------------------------
+def tango_polling(*args):
```
callumforrester, line 112: I'm tempted to suggest that polling be per-signal rather than per-device, partly for consistency and partly for flexibility (an ophyd-async device comprising signals from multiple tango servers), but I'm too ignorant about tango servers and common use cases for that to be a firm suggestion.

pyproject.toml
```diff
@@ -54,10 +54,12 @@ dev = [
     "pydata-sphinx-theme>=0.12",
     "pyepics>=3.4.2",
     "pyside6==6.7.0",
+    "pytango>=10.0.0rc1",
```
callumforrester, line 57: Tango should probably have its own extra set like `ca`, `pva` and `sim`

src/ophyd_async/tango/signal/_signal.py
```diff
@@ -0,0 +1,215 @@
...
+        The timeout for the command operation
+    name:
+        The name of the Signal
+    """
+    backend = make_backend(None, write_trl, write_trl, device_proxy)
+    return SignalX(backend, timeout=timeout, name=name)
+
+
+# --------------------------------------------------------------------
+def tango_signal_auto(
```
callumforrester, line 147: See my other comment about requiring type hints: I would prefer we kept that in ophyd-async. We should require the user to say "The control system should return this type" and not give them the option of saying "This signal is whatever the control system says it is".

src/ophyd_async/tango/signal/_signal.py
```diff
@@ -0,0 +1,215 @@
...
+    else:
+        raise RuntimeError(f"Cannot find {tr_name} in {device_trl}")
+
+    if py_type is CmdArgType.DevState:
+        py_type = DevState
+
+    return npt.NDArray[py_type] if isarray else py_type
+
+
+# --------------------------------------------------------------------
```
callumforrester, line 189: This is very nitpicky but I'd prefer these dividers were removed for consistency with the rest of the codebase

callumforrester (changes requested):
The logic here looks good, my concern is keeping the design/structure/API consistent with the rest of `ophyd-async`. I've left a few comments highlighting examples but hopefully the general idea comes across.

src/ophyd_async/tango/base_devices/_base_device.py
coretl: There isn't an `EpicsDevice` because EPICS is decentralized, but there should be a `PviDevice` because there is a single introspection PV which configures the rest of the Device. I think we should concentrate on making `PviDevice` and `TangoDevice` look similar, then we need a `FastCSDevice` that uses the connection mechanism of either of these according to the URI

src/ophyd_async/tango/base_devices/base_device.py
coretl: > @coretl this doesn't seem to work at the moment for StandardReadables. This throws an error for _readables because it is a Tuple.

That should be relatively easy to fix, we should do that

> In principle yes, we could use this but it does not allow for inferring the signal frontend and datatype from tango. This is fine but it means we wouldn't be taking advantage of this Tango feature. My instinct would be to overwrite this with a tango version. Perhaps it could accept a callback method to run if you want the frontend to be inferred.

I'm not sure I understand what the signal frontend is?

As for the datatype, we want to be able to support both specified signals that have specified types, and other unspecified signals that take their type from the control system. For example, imagine a PVI Device that has a `specified_sig` that gives you a `float` when you `caget` it, and an `unspecified_sig` that gives you a `str` when you `caget` it. Then you can:

```python
class MyDevice(PviDevice):
   specified_sig: SignalRW[float]

with DeviceCollector():
    device = MyDevice("PV:PREFIX")
# Now device has both specified_sig and unspecified_sig

await device.specified_sig.put(32.0)  # passes at static analysis and runtime
await device.specified_sig.put("bad")  # fails at static analysis and runtime

RE(save_device(device))  # gets specified_sig and unspecified_sig values
await device.unspecified_sig.put(32.0)  # fails at static analysis but passes at runtime
```

This means that:
- Signals we specify the type of will have their usage checked at static analysis time, then at runtime will be check to ensure the control system type matches
- Signals we don't specify will be added to the device using the control system type. This allows generic load/save to include them, and at runtime you can use the attribute but static analysis will fail if you use it in code
- If you start using an unspecified signal in code, then fix its type and make it a specified signal

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: To expand on what Tom said, there is no EpicsDevice because in EPICS, every PV is basically equivalent to every other PV. There is no inherent signal heirarchy. Tango is fundamentally different from EPICS in this respect because instead of individual signals, you interface with a device server which is intended to be a fully functional device containing attributes, functions, etc. Sort of like an IOC if you were to design one for a single device. This means that a device with a Tango signal requires a shared TangoProxy resource for all signals that are connected to the same TangoServer. Currently, Device does not contain an easy way to share a resource across signals so a new TangoDevice is one way to solve this problem.

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: This choice was made because of the existence of a single tango server for many signals. It makes sense to only create one proxy and pass it to the backend from the Device level than it is to make many equivalent proxies at the signal level.

src/ophyd_async/tango/base_devices/_base_device.py
callumforrester: That makes sense to me, the other option is caching of some sort but I leave it to others to decide what is cleaner

src/ophyd_async/tango/base_devices/_base_device.py
callumforrester: Happy with that, see https://github.com/bluesky/ophyd-async/pull/437#discussion_r1740912296

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: I am not certain on this either. In general, Tango servers can publish events so polling isn't necessary. But some facilities (like DESY) do not use Tango events. Many fast polling signals might effectively DDOS the tango server so one might need to be careful not to do that. The decorator can wrap any class to set the polling attribute so it should work at the device level, or the signal backend. If not, its because I have been inconsistent in how I define the polling attribute. I'll make sure it can work at the device or signal level.

src/ophyd_async/tango/signal/_signal.py
burkeds: Done.

pyproject.toml
callumforrester: Yes, you want it so that `pip install ophyd-async[tango]` installs `pytango` as a dependency, see the `ca` or `pva` section above here, which add dependencies for the case of `pip install ophyd-async[ca]` and `pip install ophyd-async[pva]`

burkeds: Following up on our discussion involving pytango and numpy -- pytango version 10.x will support numpy 2. It should be the next major release.

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: I changed tango_polling to enable either device level with a tuple, signal level polling with a dict, or both.

src/ophyd_async/tango/signal/_signal.py
burkeds: Annotated signals will now require type hints unless they are SignalX

src/ophyd_async/tango/demo/README.md
burkeds: Done

src/ophyd_async/tango/demo/mover.py
burkeds: tango_polling can be used as a class decorator to set either device level or signal level polling.

src/ophyd_async/tango/base_devices/base_device.py
burkeds: I write tango-specific solutions to create children from annotations and to fill signals from the tango device server. This will be adapted to a system agnostic solution once one has been finalized.

src/ophyd_async/tango/demo/counter.py
burkeds: Done

src/ophyd_async/tango/demo/_detector.py
```diff
@@ -0,0 +1,40 @@
...
+
+class TangoDetector(TangoReadable):
+    counters: DeviceVector[TangoCounter]
+    mover: TangoMover
+
+    def __init__(self, *args, **kwargs):
+        if "counters_kwargs" in kwargs:
+            self._counters_kwargs = kwargs.pop("counters_kwargs")
+        if "mover_kwargs" in kwargs:
+            self._mover_kwargs = kwargs.pop("mover_kwargs")
```
coretl, line 21: What are these kwargs doing?

I'm adding typing at the moment, and args and kwargs are making things difficult, so if we could remove them that would help...

src/ophyd_async/tango/demo/_mover.py
```diff
@@ -0,0 +1,83 @@
...
+    async def wait_for_idle(self):
+        event = asyncio.Event()
+
+        def _wait(value: dict[str, Reading]):
+            if value[self.state.name]["value"] == DevState.ON:
+                event.set()
+
+        self.state.subscribe(_wait)
+        await event.wait()
+        self.state.clear_sub(_wait)
```
coretl, line 79: Can this be written with `wait_for_value`?
```suggestion
        await wait_for_value(self.state, DevState.ON, timeout=something)
```

coretl (commented):
I think this is pretty much there, just a couple of minor points then we can merge. There will probably be some typing things to fix when I've finished turning the type checker back on...

src/ophyd_async/tango/demo/_detector.py
burkeds: Those were there because tango devices need a trl to connect. I refactored the code so that devices can be initialized without a trl but one will have to be set before connecting.

src/ophyd_async/tango/demo/_mover.py
burkeds: Done, that's a handy function.

coretl (approved):

burkeds:
@callumforrester Are there any other changes you would like to see before merging?

callumforrester (approved):
LGTM, thanks!

coretl:
@burkeds please could you merge main into this branch once more? `pyright` doesn't seem to be running on the last CI run for some reason...

src/ophyd_async/tango/demo/_detector.py
```diff
@@ -0,0 +1,46 @@
...
+    AsyncStatus,
+    DeviceVector,
+)
+from ophyd_async.tango import TangoReadable
+
+from ._counter import TangoCounter
+from ._mover import TangoMover
+
+
+class TangoDetector(TangoReadable):
```
coretl, line 13: If there is no `trl` passed in `__init__`, does it make sense for this to be a `TangoReadable`? It could be a `StandardReadable` instead and not take `trl` in init

src/ophyd_async/tango/signal/_signal.py
```diff
@@ -0,0 +1,223 @@
...
+
+    backend = make_backend(datatype, trl, trl, device_proxy)
+    if signal_character == "RW":
+        return SignalRW(backend=backend, timeout=timeout, name=name)
+    if signal_character == "R":
+        return SignalR(backend=backend, timeout=timeout, name=name)
+    if signal_character == "W":
+        return SignalW(backend=backend, timeout=timeout, name=name)
+    if signal_character == "X":
+        return SignalX(backend=backend, timeout=timeout, name=name)
```
coretl, line 158: I'd prefer this to be private, so that people didn't end up using `tango_signal_auto` everywhere instead of `tango_signal_r(datatype)`. What do you reckon?

src/ophyd_async/tango/base_devices/_tango_readable.py
```diff
@@ -0,0 +1,33 @@
...
+        Tango resource locator, typically of the device server.
+    proxy : AsyncDeviceProxy
+        AsyncDeviceProxy object for the device. This is created when the
+        device is connected.
+    """
+
+    def __init__(
+        self,
+        trl: str | None = None,
+        device_proxy: DeviceProxy | None = None,
```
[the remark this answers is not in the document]
burkeds: While I don't have a specific use case, it is conceivable that you might have different parts of a tango server in different ophyd devices. Tango servers are not necessarily always written to control a single device. Like ophyd, they can abstract away pretty much anything. For example, you might pass control signals to one ophyd device but scripts or monitoring to another. Or you might have a soft controller device that can send signals to many other devices. This allows you to create one shared resource that can exist across many devices.

src/ophyd_async/tango/demo/_detector.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self, trl: str, mover_trl: str, counter_trls: list[str], name=""):
+        super().__init__(trl, name=name)
+
+        # If devices are inferred from type hints, they will be created automatically
+        # during init. If they are created automatically, their trl must be set before
+        # they are connected.
+        self.mover.set_trl(mover_trl)
+        for i, c_trl in enumerate(counter_trls):
+            self.counters[i + 1] = TangoCounter(c_trl)
```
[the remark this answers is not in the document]
burkeds: I think I might have misunderstood the requirements. A single tango device does not carry information about associated devices. For now perhaps we can say that tango will only be able to infer signals.

If your tango database has a device server, its possible that we could infer higher level functionality from that, but a tango server is beyond the scope of this PR.

src/ophyd_async/tango/demo/_detector.py
burkeds: I had kept it as a TangoReadable so it could use the device inference methods in TangoDevice. Since we won't be doing that, I will change it to a StandardReadable

src/ophyd_async/tango/signal/_signal.py
burkeds: We could do that although it may not have the desired effect. By default, device init will infer the existence, types, and RW character of all attributes/commands on a tango device. By not annotating any signals they can simply point a tango device to a TRL and it will automatically do tango_signal_auto for everything. I suspect this will be the case for many users that don't really care about type safety. If they don't care about type safety, they probably also don't care about encapsulation and will just use the method if it is private anyway.

src/ophyd_async/tango/demo/_detector.py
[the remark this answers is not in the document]
burkeds: It is possible to infer the existence of devices from a tango Database server which IS accessible via a device server API but this is not something I had considered so far. Can implementing this be its own PR in the future?

coretl (approved):

coretl:
@burkeds please can you merge with a suitable commit message?

burkeds:
> @burkeds please can you merge with a suitable commit message?

@coretl Done. PyTango 10 was released so I updated pyproject to install pytango 10.0.0 or higher. I also had to comment out the unused Device and DeviceVector inferrence methods to improve converage. I left the comments in as a placeholder for this functionality which can hopefully be added in the future.

Issue #257, opened 2024-04-23
Include specific numpy dtype in output of `describe()`


danielballan:
As required by `event-model`, the `dtype` in `describe()` is a JSONschema data type, such as `"number"` or `"array"`. For Tiled (and perhaps for other consumers) we need to know that actual machine data type, such as `"<f8"` or `"<i4"`. The event-model does not currently specify a place to put this, but databroker+tiled have been using a new key `dtype_str` to place this (optionally) alongside the JSONschema `dtype`.

It has been observed this, coming fresh to `dtype` and `dtype_str`, it's not at all clear what these keys mean. We should evaluate our options, with a view toward both using obvious names and supporting backward-compatibility. Then we update `event-model` as needed and add the numpy-style data types here.
```[tasklist]
### Tasks
- [ ] https://github.com/bluesky/ophyd-async/issues/428
```

danielballan:
This design decision affects the schema, so I started an Issue at https://github.com/bluesky/event-model/issues/308. Once we decide on the spelling, we can implement `describe()` acordingly.

coretl:
- numpy types for arrays
- python types for scalars, but limit soft signals to int64 and float64

coretl:
Add datatype to `ShapeProvider` as part of this

coretl:
Decided on `dtype_numpy` instead of `dtype_str` over in https://github.com/bluesky/event-model/issues/308

abbiemery:
Closed with #438

PR #438, opened 2024-07-08
Dtype str in desc

jwlodek:
Solves #257. ~~Currently a work in progress - the tests don't yet pass. Making a draft PR to track progress.~~


jwlodek:
@coretl @genematx @mrakitin @danielballan 

First pass at this is done, and should be ready for review. All tests are passing now.

jwlodek:
The linter error appears to be related to `ruff` reformatting a file, but I cannot reproduce in my dev environment. Is there something I'm missing? Running `ruff --fix` and `ruff check src tests`. Both give `All checks passed!`

coretl:
Lint runs on the result of the branch merged to main, so maybe that made some linting error?

Please can you update `pyproject.toml` to match this line:
https://github.com/DiamondLightSource/python-copier-template/blob/516ffaa89832a71a2dedf9446a3d284b9011e27b/pyproject.toml#L52
Then you can at least see what it would change!

src/ophyd_async/epics/_backend/_aioca.py
```diff
@@ -66,9 +66,13 @@ def _data_key_from_augmented_value(
     scalar = value.element_count == 1
     dtype = dtype or dbr_to_dtype[value.datatype]
 
+    dtype_str = ""
+    dtype_str = np.dtype(dbr.DbrCodeToType[value.datatype].dtype).descr[0][1]
```
coretl, line 70: Remove the first line here?

src/ophyd_async/core/soft_signal_backend.py
```diff
@@ -71,7 +75,24 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
...
+        dtype_str = ""
+        if isinstance(value, list):
+            if len(value) > 0:
+                dtype_str = np.dtype(type(value)).descr[0][1]
+        else:
+            dtype_str = np.dtype(value.dtype).descr[0][1]
+
+        print(f"{dtype_str = }")
+        print(f"{value = }")
+        print(f"{len(value) = }")
```
coretl, line 87: Please can you remove the prints

src/ophyd_async/epics/areadetector/writers/nd_plugin.py
```diff
@@ -10,13 +10,43 @@ class Callback(str, Enum):
...
+    Int8 = 0
+    UInt8 = 1
+    Int16 = 2
+    UInt16 = 3
+    Int32 = 4
+    UInt32 = 5
+    Int64 = 6
+    UInt64 = 7
+    Float32 = 8
+    Float64 = 9
```
coretl, line 23: I'd prefer this to be `str, Enum` and go on the strings given by EPICS rather than the indexes

coretl (changes requested):
Looks like it's almost there thanks! A few minor points

tests/epics/test_signals.py
```diff
@@ -243,9 +243,39 @@ def get_dtype(suffix: str) -> str:
...
+            elif "32" in suffix:
+                int_str += "4"
+            elif "64" in suffix:
+                int_str += "8"
+            else:
+                int_str += "4"
+            return int_str
+        if "str" in suffix or "enum" in suffix:
+            return "|S40"
+        return ""
```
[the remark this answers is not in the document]
jwlodek: Turns out it's none of them actually. I just removed the line and the tests still pass.

mrakitin (approved):
The implementation looks good to me.

src/ophyd_async/core/soft_signal_backend.py
```diff
@@ -71,7 +75,20 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
 
 class SoftArrayConverter(SoftConverter):
     def get_datakey(self, source: str, value, **metadata) -> DataKey:
-        return {"source": source, "dtype": "array", "shape": [len(value)], **metadata}
+        dtype_numpy = ""
+        if isinstance(value, list):
+            if len(value) > 0:
+                dtype_numpy = np.dtype(type(value)).descr[0][1]
```
genematx, line 81: I'm not sure I understand how this works. Should it be `np.dtype(value[0].dtype)` on line 81 (or something along those lines). Also, is value always either a scalar or a _list_? Would it be safer to check if it's an iterable instead?
Or is it a numpy array or a list?
Maybe adding a typehint to `value` in the function signature could be helpful.

genematx (approved):
Looking good. Thanks, Jakub! I have one potential refactoring comment, but likely it's just may lack of understanding of internals of ophyd-async.
Aside, are we going to add `chunks` field as well? I think it should go into a separate PR, but we will need them in Tiled eventually.

src/ophyd_async/core/soft_signal_backend.py
mrakitin: Good points, @genematx! I overlooked that case during my review.
Maybe it should just be `np.dtype(type(value[0]))` as the `value[0]` will be a number, which won't (or may not?) have `.dtype` method.

To compare two approaches:
```py
In [28]: np.dtype(type([1, 2, 3][0])).descr
Out[28]: [('', '<i8')]

In [29]: np.dtype(type([1, 2, 3])).descr
Out[29]: [('', '|O')]
```

I support adding a type annotation to the signature and a docstring to explain the expected input types.

coretl (approved):

Issue #427, opened 2024-07-02
Make a PMAC flyer that is implemented with a Trajectory scan Device


coretl:
Make a TriggerLogic that will do a Trajectory scan with the following restrictions:
- GPIOs are always start of row trigger
- ScanSpec only includes trajectory scannable axes

Make the matching IO class with the PV interface.
Make a `epics.pmac.PmacMotor(epics.motor.Motor)` that tells it what CS and CS Axis it is in (like Malcolm). There might be a PmacCSMotor and PmacRawMotor, or possibly one that is told at init what sort of thing it is. Either way we want a single method `async def get_cs_number_and_axis()` attached to the motor so we can use it in prepare.

## prepare
Take a `List[scanspec.Frames[PmacMotor]]`, calculate the entire trajectory (this will later be only the first 4000 points), calculate where the start position of the motors is based on first point of the trajectory (as per #376), then move the motors to the start position and send down the trajectory

## kickoff
Hit start

## complete
Observe points scanned and yield `WatcherStatusUpdate` to say how far we have got (like in StandardDetector). When it's done check it did the right number of points and didn't get an error

PR #440, opened 2024-07-08
Implementation of pmac trajectory

tomtrafford:
Implements #427. 

This implements a simple trajectory for a pmac. It is initialised with a PMAC, Coordinate System name and a list of pmac motors involved in the scan. On Prepare it takes a ScanSpec stack which it converts into a position array for each axis in the scan, a velocity array for each axis in the scan and one time array (shared by all axes in the scan). It adjusts the start of the scan so that there is enough time and distance for the motors to achieve their desired velocity for the first point of the scans, and stops the motors at the end of the scan.

A Pmac CS Motor class has also been defined, which takes a motor class and adds information about its Coordinate system. 

It currently doesn't have the capability to turnaround or use velocity calculations. 

The current test gives the pmac a line trajectory, mocking the settings of the SampleX axis on P45. This trajectory has been tested on the p45 hardware.

### DIFF
--- a/src/ophyd_async/epics/pmac/_pmacIO.py
+++ b/src/ophyd_async/epics/pmac/_pmacIO.py
@@ -1,0 +1,39 @@
+from bluesky.protocols import Flyable, Preparable
+
+from ophyd_async.core import StandardReadable
+
+from ..signal.signal import epics_signal_rw
+
+
+class Pmac(StandardReadable, Flyable, Preparable):
+    """Device that moves a PMAC Motor record"""
+
+    def __init__(self, prefix: str, name="") -> None:
+        self.timeArray = epics_signal_rw(str, prefix + ":ProfileTimeArray")
+        self.a = epics_signal_rw(str, prefix + ":A")
+        self.use_a = epics_signal_rw(bool, prefix + ":A:UseAxis")
+        self.a_vel = epics_signal_rw(str, prefix + ":A:Velocities")
+        self.b = epics_signal_rw(str, prefix + ":B")
+        self.use_b = epics_signal_rw(bool, prefix + ":B:UseAxis")
+        self.b_vel = epics_signal_rw(str, prefix + ":B:Velocities")
+        self.c = epics_signal_rw(str, prefix + ":C")
+        self.use_c = epics_signal_rw(bool, prefix + ":C:UseAxis")
+        self.c_vel = epics_signal_rw(str, prefix + ":C:Velocities")
+        self.u = epics_signal_rw(str, prefix + ":U")
+        self.use_u = epics_signal_rw(bool, prefix + ":U:UseAxis")
+        self.u_vel = epics_signal_rw(str, prefix + ":U:Velocities")
+        self.v = epics_signal_rw(str, prefix + ":V")
+        self.use_v = epics_signal_rw(bool, prefix + ":V:UseAxis")
+        self.v_vel = epics_signal_rw(str, prefix + ":V:Velocities")
+        self.w = epics_signal_rw(str, prefix + ":W")
+        self.use_w = epics_signal_rw(bool, prefix + ":W:UseAxis")
+        self.w_vel = epics_signal_rw(str, prefix + ":W:Velocities")
+        self.x = epics_signal_rw(str, prefix + ":X")
+        self.use_x = epics_signal_rw(bool, prefix + ":X:UseAxis")
+        self.x_vel = epics_signal_rw(str, prefix + ":X:Velocities")
+        self.y = epics_signal_rw(str, prefix + ":Y")
+        self.use_y = epics_signal_rw(bool, prefix + ":Y:UseAxis")
+        self.y_vel = epics_signal_rw(str, prefix + ":Y:Velocities")
+        self.z = epics_signal_rw(str, prefix + ":Z")
+        self.use_z = epics_signal_rw(bool, prefix + ":Z:UseAxis")
+        self.z_vel = epics_signal_rw(str, prefix + ":Z:Velocities")
--- a/src/ophyd_async/epics/pmac/_pmacIO.py
+++ b/src/ophyd_async/epics/pmac/_pmacIO.py
@@ -1,3 +1,5 @@
+import numpy as np
+import numpy.typing as npt
 from bluesky.protocols import Flyable, Preparable
 
 from ophyd_async.core import StandardReadable
@@ -9,31 +11,33 @@ class Pmac(StandardReadable, Flyable, Preparable):
     """Device that moves a PMAC Motor record"""
 
     def __init__(self, prefix: str, name="") -> None:
-        self.timeArray = epics_signal_rw(str, prefix + ":ProfileTimeArray")
-        self.a = epics_signal_rw(str, prefix + ":A")
+        self.timeArray = epics_signal_rw(
+            npt.NDArray[np.float64], prefix + ":ProfileTimeArray"
+        )
+        self.a = epics_signal_rw(npt.NDArray[np.float64], prefix + ":A:Positions")
         self.use_a = epics_signal_rw(bool, prefix + ":A:UseAxis")
-        self.a_vel = epics_signal_rw(str, prefix + ":A:Velocities")
-        self.b = epics_signal_rw(str, prefix + ":B")
+        self.a_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":A:Velocities")
+        self.b = epics_signal_rw(npt.NDArray[np.float64], prefix + ":B:Positions")
         self.use_b = epics_signal_rw(bool, prefix + ":B:UseAxis")
-        self.b_vel = epics_signal_rw(str, prefix + ":B:Velocities")
-        self.c = epics_signal_rw(str, prefix + ":C")
+        self.b_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":B:Velocities")
+        self.c = epics_signal_rw(npt.NDArray[np.float64], prefix + ":C:Positions")
         self.use_c = epics_signal_rw(bool, prefix + ":C:UseAxis")
-        self.c_vel = epics_signal_rw(str, prefix + ":C:Velocities")
-        self.u = epics_signal_rw(str, prefix + ":U")
+        self.c_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":C:Velocities")
+        self.u = epics_signal_rw(npt.NDArray[np.float64], prefix + ":U:Positions")
         self.use_u = epics_signal_rw(bool, prefix + ":U:UseAxis")
-        self.u_vel = epics_signal_rw(str, prefix + ":U:Velocities")
-        self.v = epics_signal_rw(str, prefix + ":V")
+        self.u_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":U:Velocities")
+        self.v = epics_signal_rw(npt.NDArray[np.float64], prefix + ":V:Positions")
         self.use_v = epics_signal_rw(bool, prefix + ":V:UseAxis")
-        self.v_vel = epics_signal_rw(str, prefix + ":V:Velocities")
-        self.w = epics_signal_rw(str, prefix + ":W")
+        self.v_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":V:Velocities")
+        self.w = epics_signal_rw(npt.NDArray[np.float64], prefix + ":W:Positions")
         self.use_w = epics_signal_rw(bool, prefix + ":W:UseAxis")
-        self.w_vel = epics_signal_rw(str, prefix + ":W:Velocities")
-        self.x = epics_signal_rw(str, prefix + ":X")
+        self.w_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":W:Velocities")
+        self.x = epics_signal_rw(npt.NDArray[np.float64], prefix + ":X:Positions")
         self.use_x = epics_signal_rw(bool, prefix + ":X:UseAxis")
-        self.x_vel = epics_signal_rw(str, prefix + ":X:Velocities")
-        self.y = epics_signal_rw(str, prefix + ":Y")
+        self.x_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":X:Velocities")
+        self.y = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Y:Positions")
         self.use_y = epics_signal_rw(bool, prefix + ":Y:UseAxis")
-        self.y_vel = epics_signal_rw(str, prefix + ":Y:Velocities")
-        self.z = epics_signal_rw(str, prefix + ":Z")
+        self.y_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Y:Velocities")
+        self.z = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Z:Positions")
         self.use_z = epics_signal_rw(bool, prefix + ":Z:UseAxis")
-        self.z_vel = epics_signal_rw(str, prefix + ":Z:Velocities")
+        self.z_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Z:Velocities")
--- a/src/ophyd_async/epics/pmac/_pmacIO.py
+++ b/src/ophyd_async/epics/pmac/_pmacIO.py
@@ -2,15 +2,15 @@ import numpy as np
 import numpy.typing as npt
 from bluesky.protocols import Flyable, Preparable
 
-from ophyd_async.core import StandardReadable
+from ophyd_async.core import StandardReadable, SubsetEnum
 
-from ..signal.signal import epics_signal_rw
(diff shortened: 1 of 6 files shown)


src/ophyd_async/epics/pmac/_pmacIO.py
```diff
@@ -0,0 +1,49 @@
...
+class Pmac(StandardReadable, Flyable, Preparable):
+    """Device that moves a PMAC Motor record"""
+
+    def __init__(self, prefix: str, cs="", name="") -> None:
+        self.timeArray = epics_signal_rw(
+            npt.NDArray[np.float64], prefix + ":ProfileTimeArray"
+        )
+        self.a = epics_signal_rw(npt.NDArray[np.float64], prefix + ":A:Positions")
+        self.use_a = epics_signal_rw(bool, prefix + ":A:UseAxis")
+        self.a_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":A:Velocities")
```
coretl, line 19: Could we do these as `DeviceVector`s instead?
```python
        cs_letters = "ABCUVWXYZ"
        # 1 indexed CS axes so we can index into them from the compound motor input link
        self.positions = DeviceVector({
            i + 1: epics_signal_rw(npt.NDArray[np.float64], f"{prefix}{letter}:Positions")
            for i, letter in enumerate(cs_letters)
        })
        self.use_axis = ...
        self.velocities = ...
```

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,128 @@
...
+    ):
+        # Assuming ramping to or from 0
+        max_velocity_acceleration_time = await motor.acceleration_time.get_value()
+        max_velocity = await motor.max_velocity.get_value()
+        accl_time = max_velocity_acceleration_time * end_velocity / max_velocity
+        disp = 0.5 * (velocity + end_velocity) * accl_time
+        return [disp, accl_time]
+
+    @AsyncStatus.wrap
+    async def prepare(self, scanSpecStack):
```
coretl, line 37: We can assume that the plan has given us a stack of frames where the axes are Motor objects
```suggestion
    async def prepare(self, stack: List[Frames[Motor]]):
```

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,128 @@
...
+                    self.motors[axis.lower()],
+                    self.profile[axis.lower() + "_velocity"][0],
+                )
+                self.initial_pos[axis] = self.profile[axis.lower()][0] - run_up_disp
+        self.profile["duration"][0] += run_up_time / TICK_S
+
+        # Send trajectory to brick
+        for axis in scanAxes:
+            if axis != "DURATION":
+                self.profile_cs_name.set(self.cs)
```
coretl, line 94: Then if we add to Motor:
```python
    def __init__(...):
        ...
        self.input_link = epics_signal_r(str, prefix + ".INP")
```

Then we can do somewhere above:
```python
input_links = asyncio.gather(*[axis.input_link.get_value() for axis in scanAxes])
cs_ports = set()
cs_axes: Dict[Motor, int] = {}
for axis in scanAxes:
    input_link = await axis.input_link.get_value()
    # Split "@asyn(PORT,num)" into ["PORT", "num"]
    split = input_link.split("(")[1].rstrip(")").split(",")
    cs_ports.add(split[0].strip())
    cs_axes[axis] = int(split[1].strip())
assert len(cs_ports) == 1, "Motors in more than one CS"
```

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,128 @@
...
+                getattr(self, axis.lower() + "_vel").set(
+                    self.profile[axis.lower() + "_velocity"]
+                )
+            else:
+                self.timeArray.set(self.profile["duration"])
+
+        # MOVE TO START
+        for axis in scanAxes:
+            if axis != "DURATION":
+                await self.motors[axis.lower()].set(self.initial_pos[axis])
```
coretl, line 107: In a separate issue we should do this via the direct demands deferred move like https://github.com/dls-controls/pymalcolm/blob/2c3d7048026ef9c5a0fcd69db89bd88d74f62ad9/malcolm/modules/pmac/parts/cspart.py#L53-L85

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,128 @@
...
+        self.build_profile.set(True)
+        self._fly_start = time.monotonic
+
+    @AsyncStatus.wrap
+    async def kickoff(self):
+        await self.execute_profile.set(True)
+
+    @WatchableAsyncStatus.wrap
+    async def complete(self):
+        yield WatcherUpdate(
```
coretl, line 120: This should be:
```python
async for percent in observe_value(self.percent_complete):
    yield WatcherUpdate(..., current=percent, ...)
```
Eventually we will need to use trajectory points rather than percent, but that can come later

tests/epics/pmac/test_pmac.py
```diff
@@ -0,0 +1,71 @@
...
+    yield sim_motor
+
+
+async def test_sim_pmac_simple_trajectory(sim_x_motor) -> None:
+    # Test the generated Trajectory profile from a scanspec
+    async with DeviceCollector(mock=True):
+        prefix = "BLxxI-MO-STEP-01"
+        motors = [sim_x_motor]
+        traj = PmacTrajectory(prefix, "BRICK1.CS3", motors, name="sim_pmac")
+        spec = fly(Line("z", 1, 5, 9), 1)
```
coretl, line 30:
```suggestion
        spec = fly(Line(sim_x_motor, 1, 5, 9), 1)
```

tests/epics/pmac/test_pmac.py
```diff
@@ -0,0 +1,71 @@
...
+
+    yield sim_motor
+
+
+async def test_sim_pmac_simple_trajectory(sim_x_motor) -> None:
+    # Test the generated Trajectory profile from a scanspec
+    async with DeviceCollector(mock=True):
+        prefix = "BLxxI-MO-STEP-01"
+        motors = [sim_x_motor]
+        traj = PmacTrajectory(prefix, "BRICK1.CS3", motors, name="sim_pmac")
```
coretl, line 29: This is the only line we need inside the DeviceCollector, the rest can be dedented

coretl (changes requested):

src/ophyd_async/epics/pmac/_pmacTrajectory.py
tomtrafford: I believe this would only work for compound motors. An Axis with straight through mapping, (p45 Sample X for example) does not have a compound motor and its link is '@asyn(PORT, num)' where PORT is the brick and num refers to its axis number on the brick rather than a CS assignment.

However from the brick and axis number I can then use '{PMAC_PREFIX}:M{num}:CsPort_RBV' and '{PMAC_PREFIX}:M{num}:CsAxis_RBV' to get the port and axis. But this feels quite long-winded...

coretl: > I believe this would only work for compound motors. 

Correct

> An Axis with straight through mapping, (p45 Sample X for example) does not have a compound motor and its link is '@ASYN(PORT, num)' where PORT is the brick and num refers to its axis number on the brick rather than a CS assignment.

I forgot P45 didn't have all compound axes... I guess we'll have to solve this now.

> However from the brick and axis number I can then use '{PMAC_PREFIX}:M{num}:CsPort_RBV' and '{PMAC_PREFIX}:M{num}:CsAxis_RBV' to get the port and axis. But this feels quite long-winded...

Yes, that seems right. However I think I did the wrong thing in Malcolm attaching them to the motor. We should attach them to the generic PMAC object instead, in the same way as the EDM screens have an "Axes" table with deferred move, CS port and CS Axis. 

I think we need to discuss this with the `DeviceVector` part below, but for now could you attach these somewhere to the PMAC object so you can use them in the CS code above?

src/ophyd_async/epics/pmac/_pmacIO.py
coretl: Maybe we leave it as it is for now, then consider this issue along with the direct demands and raw motor port and address issue below

src/ophyd_async/epics/pmac/_pmacIO.py
```diff
@@ -0,0 +1,49 @@
...
+        self.y_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Y:Velocities")
+        self.z = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Z:Positions")
+        self.use_z = epics_signal_rw(bool, prefix + ":Z:UseAxis")
+        self.z_vel = epics_signal_rw(npt.NDArray[np.float64], prefix + ":Z:Velocities")
+        self.points_to_build = epics_signal_rw(int, prefix + ":ProfilePointsToBuild")
+        self.build_profile = epics_signal_rw(bool, prefix + ":ProfileBuild")
+        self.execute_profile = epics_signal_rw(bool, prefix + ":ProfileExecute")
+        self.scan_percent = epics_signal_r(float, prefix + ":TscanPercent_RBV")
+        cs_names_enum = SubsetEnum[cs]
+        self.profile_cs_name = epics_signal_rw(cs_names_enum, prefix + ":ProfileCsName")
```
coretl, line 49:
```suggestion
        self.profile_cs_name = epics_signal_rw(str, prefix + ":ProfileCsName")
```

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,155 @@
...
+class PmacTrajectory(Pmac, Flyable, Preparable):
+    """Device that moves a PMAC Motor record"""
+
+    def __init__(self, prefix: str, cs: int, name="") -> None:
+        # Make a dict of which motors are for which cs axis
+        self._fly_start: float
+        self.cs = cs
+        super().__init__(prefix, cs, name=name)
+
+    async def _ramp_up_velocity_pos(
```
olliesilvester, line 24: nit: On the motor record I refer to this as `run_up_distance `rather than `ramp_up`. I don't have a preference for which is better, but it would be nice if they're the same

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,154 @@
...
+                if axis != "DURATION":
+                    self.profile[cs_axes[axis] + "_velocity"].append(
+                        (stack[0].upper[axis][i] - stack[0].lower[axis][i])
+                        / (stack[0].midpoints["DURATION"][i])
+                    )
+                    self.profile[cs_axes[axis]].append(stack[0].midpoints[axis][i])
+                else:
+                    self.profile[axis.lower()].append(
+                        int(stack[0].midpoints[axis][i] / TICK_S)
+                    )
```
coretl, line 70: Eventually we should make these numpy operations so we get some speed, but this will do for now.

We also need to make sure that we put lower and midpoint points in here, but that can also be a point for later.

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,154 @@
...
+        delta_v = abs(end_velocity - velocity)
+        accl_time = max_velocity_acceleration_time * delta_v / max_velocity
+        disp = 0.5 * (velocity + end_velocity) * accl_time
+        return [disp, accl_time]
+
+    @AsyncStatus.wrap
+    async def prepare(self, stack: list[Frames[Motor]]):
+        # Which Axes are in use?
+
+        scanSize = len(stack[0])
```
coretl, line 39: We should create a `path = Path(stack)` and call `chunk = path.consume(n)` to consume N points from the scan to process. In this case `chunk = path.consume() # consume the whole scan` and `scan_size = len(chunk)`

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,155 @@
...
+from ophyd_async.core.async_status import AsyncStatus, WatchableAsyncStatus
+from ophyd_async.core.signal import observe_value
+from ophyd_async.core.utils import WatcherUpdate
+from ophyd_async.epics.motion import Motor
+from ophyd_async.epics.pmac import Pmac
+
+TICK_S = 0.000001
+
+
+class PmacTrajectory(Pmac, Flyable, Preparable):
```
coretl, line 17: Please could you implement this as a `TriggerLogic` to be nested within a `StandardFlyer`? That will remove a bit of the boilerplate around `kickoff` and `complete`

tomtrafford:
Completes #494

tomtrafford:
With the recent changes using raw motors, I have had to pass to the PMAC which axes are on the brick. '{PMAC_PREFIX}:M{num}:CsAxis_RBV' and '{PMAC_PREFIX}:M{num}:CsPort_RBV' PVs don't exist unless there is a motor in the IOC for the associated M(num).

src/ophyd_async/epics/pmac/_pmacIO.py
```diff
@@ -0,0 +1,56 @@
...
+        self.profile_cs_name = epics_signal_rw(str, prefix + ":ProfileCsName")
+        self.profile_calc_vel = epics_signal_rw(bool, prefix + ":ProfileCalcVel")
+
+
+class PmacMotor(motor.Motor):
+    """Device that moves a PMAC Motor record"""
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.CsAxis = epics_signal_r(str, f"{prefix}:CsAxis_RBV")
+        self.CsPort = epics_signal_r(str, f"{prefix}:CsPort_RBV")
```
coretl, line 55:
```suggestion
        self.cs_axis = epics_signal_r(str, f"{prefix}:CsAxis_RBV")
        self.cs_port = epics_signal_r(str, f"{prefix}:CsPort_RBV")
```

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,488 @@
...
+    FlyerController,
+    wait_for_value,
+)
+from ophyd_async.epics.pmac import Pmac, PmacMotor
+
+TICK_S = 0.000001
+
+
+class PmacTrajInfo(BaseModel):
+    spec: Spec = Field(default=None)
```
coretl, line 20:
```suggestion
    spec: Spec[PmacMotor | Literal["DURATION"]] = Field(default=None)
```
So this makes me think we need to make duration special in scanspec

src/ophyd_async/epics/pmac/_pmacTrajectory.py
```diff
@@ -0,0 +1,488 @@
...
+    min_time: float = 0.002,
+    min_interval: float = 0.002,
+):
+    """Make consistent time and velocity arrays for each axis
+
+    Try to create velocity profiles for all axes that all arrive at
+    'distance' in the same time period. The profiles will contain the
+    following points:-
+
+    in the following description acceleration can be -ve or +ve depending
```
coretl, line 308: Isn't all this in https://github.com/DiamondLightSource/velocity-profile/ now?

src/ophyd_async/epics/pmac/_pmacTrajectory.py
coretl, line 1: Can we make these `_pmac_io.py` and `_pmac_trajectory.py` please?

src/ophyd_async/epics/pmac/_pmacTrajectory.py
tomtrafford: This gives a pydantic error  ` /workspaces/ophyd-async/src/ophyd_async/epics/pmac/_pmacTrajectory.py:20:11 - error: "__getitem__" method not defined on type "type" (reportIndexIssue)`

src/ophyd_async/epics/pmac/_pmacTrajectory.py
tomtrafford: I don't think so. I took this from https://github.com/DiamondLightSource/pymalcolm/blob/264ec4a319092a545853f41f89ddf1d21da56504/malcolm/modules/pmac/util.py#L222 which didn't get added to velocity-profile...

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,560 @@
...
+                await self.pmac.use_axis[cs_axes[axis] + 1].set(True)
+                await self.pmac.positions[cs_axes[axis] + 1].set(
+                    positions[cs_axes[axis]][:profile_length],
+                )
+                await self.pmac.velocities[cs_axes[axis] + 1].set(
+                    velocities[cs_axes[axis]][:profile_length]
+                )
+            else:
+                await self.pmac.time_array.set(time_array[:profile_length])
+                await self.pmac.user_array.set(user_array[:profile_length])
```
shihab-dls, line 241: FYI @gkalua

PR #457, opened 2024-07-16
Fix race condition in set_and_wait_for_value

### LINKED ISSUE #453: Allow wait_for_value to take existing generator
For a couple of devices in `dodal` we have hit potential race conditions where we:
1. Do `x`
2. This causes `x_rbv` to go 0 -> 1 -> 0
3. `wait_for_value(x_rbv, 1)` - this fails as the PV has already done the transition

A solution to this would be for `wait_for_value` to accept an `AsyncGenerator[T]` and so you could do:
```python
monitor = observe_value(x_rbv)
x.set(1)
wait_for_value(x_rbv, 1)
```

DominicOram:
Fixes #453 

* Fixes the original race condition as discussed
* Adds a new `set_and_wait_for_other_value` for monitoring a different value than setting


src/ophyd_async/core/signal.py
```diff
@@ -557,29 +562,77 @@ async def wait_for_value(
...
+        How long to wait for the set to complete
+
+    Notes
+    -----
+    Example usage::
+
+        set_and_wait_for_value(device.acquire, 1, device.acquire_rbv, 1)
+    """
+    # Start monitoring before the set to avoid a race condition
+    values_gen = observe_value(read_signal)
```
coretl, line 600: There is still a tiny race condition here. What we are effectively doing for a CA signal is:
- Create a task that will eventually call `ca_create_subscription`
- Create a task that will eventually call `ca_array_put_callback`
- Wait until a subscription update matches

It is possible (but less likely) that we might send a put which might send an update before the first subscription connects. A possible fix is to consume the first value from `observe_value` before doing `set`, so we are sure we have a live subscription.

tests/core/test_signal.py
```diff
@@ -208,6 +209,32 @@ async def check_set_and_wait():
...
+        st = await set_and_wait_for_other_value(
+            set_signal, 1, read_signal, "test", timeout=100
+        )
+        await st
+        await asyncio.sleep(0.01)
+
+    assert (
+        0.1
+        < await time_taken_by(asyncio.gather(wait_and_set_read(), check_set_and_wait()))
+        < 0.15
```
coretl, line 233: 0.05s is not long for CI, concerned this test might be flaky in the long run...

Is it possible to use some `asyncio.Events` instead of sleeps to ensure the ordering between the two tasks?

coretl (commented):
I spend half an hour playing with typing, and I now see why you made it two functions. I think this is a pretty teachable API, so happy with the signature. Spotted a further possible race condition (which would only manifest with a heavily loaded IOC and network congestion), which we might want to consider...

src/ophyd_async/core/signal.py
```diff
@@ -557,29 +562,77 @@ async def wait_for_value(
...
+        ) from e
+
+    return status
+
+
 async def set_and_wait_for_value(
-    signal: SignalRW[T],
-    value: T,
+    set_signal: SignalRW[T],
+    set_value: T,
```
coretl, line 620: Nit: I preferred the original names...

src/ophyd_async/core/signal.py
```diff
@@ -557,6 +562,63 @@ async def wait_for_value(
...
+    -----
+    Example usage::
+
+        set_and_wait_for_value(device.acquire, 1, device.acquire_rbv, 1)
+    """
+    # Start monitoring before the set to avoid a race condition
+    values_gen = observe_value(read_signal)
+
+    # Get the initial value from the monitor to make sure we've created it
+    await anext(values_gen)
```
coretl, line 603: One last thing to consider. If we have:
- `signal.value == 3`
- `await set_and_wait_for_value(signal, 3)`
- MDEL>=0

Then this code will receive the initial value, discard it, set it to the value it currently is, then timeout waiting for an update that never comes.

Possible fixes:
1. Don't bother setting the signal if it's already at the right value
2. Error if the signal is already at the right value
3. Store `current_value = await anext(values_gen)`, then only call `_wait_for_value` if it doesn't match

I think I favour 3, what do you think?

coretl (commented):
Sorry, just thought of another edge case! Could either merge as is and make a new issue to handle, or fix it here...

src/ophyd_async/core/signal.py
DominicOram: I agree that 3 seems the most broadly correct, small enough that we can do as part of this I think. Will do now

coretl (approved):

PR #466, opened 2024-07-18
314 defer creation of hdf5 file until preparetrigger

ZohebShaikh:


src/ophyd_async/core/detector.py
```diff
@@ -59,11 +59,16 @@ class TriggerInfo(BaseModel):
     #: Sort of triggers that will be sent
     trigger: DetectorTrigger = Field()
     #: What is the minimum deadtime between triggers
-    deadtime: float = Field(ge=0)
+    deadtime: float | None = Field(ge=0)
```
DiamondJoseph, line 62: Would be good to mention that the behaviour when deadtime/livetime is None is that they are unchanged from the previous prepare that set those values.

DiamondJoseph: Could also make the Field(default=None) so that preparing for a step scan is less verbose

src/ophyd_async/core/detector.py
```diff
@@ -240,10 +245,15 @@ async def describe(self) -> Dict[str, DataKey]:
...
-        self._trigger_info = TriggerInfo(
-            number=1, trigger=DetectorTrigger.internal, deadtime=0.0, livetime=0.0
-        )
+        if self._trigger_info is None:
+            await self.prepare(
+                TriggerInfo(
+                    number=1,
+                    trigger=DetectorTrigger.internal,
+                    deadtime=None,
+                    livetime=None,
```
DiamondJoseph, line 254: if livetime, deadtime get a default these lines they aren't required.

tests/conftest.py
```diff
@@ -146,3 +147,14 @@ def static_path_provider(
     static_filename_provider: FilenameProvider,
 ):
     return static_path_provider_factory(static_filename_provider)
+
+
+@pytest.fixture
+def count_scan_trigger_info() -> TriggerInfo:
```
DiamondJoseph, line 153:
```suggestion
def one_shot_trigger_info() -> TriggerInfo:
```

DiamondJoseph: Maybe? Makes it less implementation specific. It's just a TriggerInfo for 1 shot, without changing deadtime or livetime

tests/epics/areadetector/test_pilatus.py
```diff
@@ -121,12 +121,17 @@ async def test_unsupported_trigger_excepts(pilatus: PilatusDetector):
...
+        await pilatus.prepare(
+            TriggerInfo(
+                number=1,
+                trigger=DetectorTrigger.internal,
+                deadtime=1.0,
+                livetime=1.0,
+            )
         )
-    )
+    assert str(exc.value) in f"File path {tmp_path} for hdf plugin does not exist"
```
DiamondJoseph, line 135: Why is this change required?

tests/epics/demo/test_demo_ad_sim_detector.py
```diff
@@ -63,7 +63,7 @@ def count_sim(dets: List[StandardDetector], times: int = 1):
         for det in dets:
             yield from bps.trigger(det, wait=False, group="wait_for_trigger")
 
-        yield from bps.sleep(0.001)
+        yield from bps.sleep(0.1)
```
DiamondJoseph, line 66: Is this still unreliable with the lower sleep or can this be restored?

tests/epics/demo/test_demo_ad_sim_detector.py
```diff
@@ -116,24 +116,60 @@ async def two_detectors(tmp_path: Path):
...
+        livetime=None,
+        frame_timeout=None,
+    )
     docs = defaultdict(list)
 
+    def assert_n_stream_datums(n: int):
+        if n == 0:
+            assert "stream_datum" not in docs
+        else:
+            assert len(docs["stream_datum"]) == n
```
DiamondJoseph, line 132:
```suggestion
    def assert_n_stream_datums(n: int, start: int | None = None, stop: int | None = None):
        if n == 0:
            assert "stream_datum" not in docs
        else:
            assert len(docs["stream_datum"]) == n
            # check both detectors have the correct start/stop
           for detector_index in {n -1, n - 2}:
               seq_nums = docs["stream_datum"][detector_index]["seq_nums"]
               assert seq_nums["start"] == start
               assert seq_nums["stop"] == stop
```"

**Review by DiamondJoseph — DISMISSED** [review]

File: src/ophyd_async/core/detector.py [hunk 7]
```diff
@@ -283,20 +295,29 @@ async def _prepare(self, value: T) -> None:
...
+                required = self.controller.get_deadtime(self._trigger_info.livetime)
+                assert required <= self._trigger_info.deadtime, (
+                    f"Detector {self.controller} needs at least {required}s deadtime, "
+                    f"but trigger logic provides only {self._trigger_info.deadtime}s"
+                )
+                self._arm_status = await self.controller.arm(
+                    num=self._trigger_info.number,
+                    trigger=self._trigger_info.trigger,
+                    exposure=self._trigger_info.livetime,
+                )
```
coretl, line 318: Logic should be:
- if trigger_info.trigger != DetectorTrigger then assert triggerinfo.deadtime
- if triggerinfo.deadtime then check it is >= self.controller.get_deadtime
- then *always* call self.controller.arm

The underlying controllers should cope with being told to arm with livetime/deadtime of None, and leave the exposure time of the detector where it was

src/ophyd_async/core/detector.py
```diff
@@ -293,30 +287,23 @@ async def _prepare(self, value: TriggerInfo) -> None:
...
-                    f"but trigger logic provides only {self._trigger_info.deadtime}s"
-                )
-                self._arm_status = await self.controller.arm(
-                    num=self._trigger_info.number,
-                    trigger=self._trigger_info.trigger,
-                    exposure=self._trigger_info.livetime,
-                )
-            self._fly_start = time.monotonic()
+        if (value.trigger != DetectorTrigger.internal):
+            assert value.deadtime
```
coretl:
```suggestion
            assert value.deadtime, "Deadtime must be supplied when in externally triggered mode"
```

PR #467, opened 2024-07-18
Restructure on ADR 2.0

abbiemery:
This is the up to date restructure.


docs/examples/foo_detector.py
```diff
@@ -3,22 +3,19 @@
...
-from ophyd_async.epics.areadetector.writers.hdf_writer import HDFWriter
-from ophyd_async.epics.areadetector.writers.nd_file_hdf import NDFileHDF
+from ophyd_async.epics import adcore
 
 
-class FooDriver(ADBase):
+class FooDriver(adcore.ADBase):
     def __init__(self, prefix: str, name: str = "") -> None:
-        self.trigger_mode = ad_rw(str, prefix + "TriggerMode")
+        self.trigger_mode = adcore.ad_rw(str, prefix + "TriggerMode")
```
coretl, line 18: #274 deleted `adcore.ad_rw` and replaced with `epics_signal_rw_rbv`. Please can you do this one two? And see if the docs build is giving an error that we are ignoring...

src/ophyd_async/core/_flyer.py
```diff
@@ -29,7 +27,7 @@ async def stop(self):
         """Stop flying and wait everything to be stopped"""
 
 
-class HardwareTriggeredFlyable(
+class StandardFlyer(
```
coretl, line 30: I'm pretty sure `TriggerLogic` is now the wrong name, but can't think of a better one immediately so let's leave it for now

src/ophyd_async/epics/adcore/_ad_base.py
coretl, line 1: Could: Put all of these classes from this file, and `_nd_file_hdf.py` into a single `_adcore_io.py` to follow the pattern from aravis.

src/ophyd_async/epics/adcore/__init__.py
```diff
@@ -0,0 +1,38 @@
+from ._ad_base import (
+    DEFAULT_GOOD_STATES,
+    ADBase,
+    ADBaseShapeProvider,
+    DetectorState,
+    set_exposure_time_and_acquire_period_if_supplied,
+    start_acquiring_driver_and_ensure_status,
+)
+from ._hdf_writer import HDFWriter
```
coretl, line 9: Should: rename to something with `AD` in to distinguish it from `PandaHDFWriter
```suggestion
from ._hdf_writer import ADHDFWriter
```

src/ophyd_async/epics/adcore/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+from ._nd_file_hdf import NDFileHDF
+from ._nd_plugin import ADBaseDataType, NDPluginStats
```
coretl, line 11: Should: be IO classes
```suggestion
from ._nd_file_hdf import NDFileHDFIO
from ._nd_plugin import ADBaseDataType, NDPluginStatsIO
```

src/ophyd_async/epics/adcore/__init__.py
```diff
@@ -0,0 +1,38 @@
+from ._ad_base import (
+    DEFAULT_GOOD_STATES,
+    ADBase,
```
coretl, line 3: Should: be IO class
```suggestion
    ADBaseIO,
```

src/ophyd_async/epics/adcore/_ad_base.py
coretl: Should: split the IO classes (like `ADBase` and `DetectorState`) from the logic (like `set_exposure_and_acquire_time` and `DEFAULT_GOOD_STATES`) into different files

src/ophyd_async/epics/demo/mover.py
coretl, line 1: Should be `_mover.py`

src/ophyd_async/epics/demo/sensor.py
coretl, line 1: Should be `_sensor.py`

src/ophyd_async/epics/signal/_epics_transport.py
coretl, line 1: Could: rename `Transport` to `SignalBackend` in this file

Then we end up with `EpicsSignalBackend.ca` and `EpicsSignalBackend.pva`, and this file can be merged with `_signal.py`

src/ophyd_async/epics/signal/__init__.py
```diff
@@ -1,4 +1,8 @@
-from .signal import (
+from ._common import LimitPair, Limits, get_supported_values
+from ._epics_transport import EpicsTransport
+from ._p4p import PvaSignalBackend
+from ._signal import (
+    _make_backend,
```
coretl, line 5: Should: make this public and give it a different name
```suggestion
    epics_signal_backend,
```

src/ophyd_async/fastcs/panda/__init__.py
```diff
@@ -28,25 +31,27 @@
 
 __all__ = [
     "CommonPandaBlocks",
-    "HDFPanda",
-    "PcompBlock",
-    "PcompInfo",
-    "PcompDirectionOptions",
+    "DataBlock",
```
coretl, line 34: Could: tempted to replace `Block` -> `IO` here. `PcompIO` probably fine, but not sure if I like `DataIO`... Is `DataBlockIO` too much of a mouthful? Or shall we make PandA a special case that has `Block` instead of `IO`?

coretl (commented):
Thanks, this looks pretty close now

docs/examples/foo_detector.py
abbiemery: Fixed this, and also the outdated DirectoryProvider

src/ophyd_async/epics/adcore/__init__.py
abbiemery: Done.

src/ophyd_async/epics/demo/mover.py
abbiemery: Done

src/ophyd_async/epics/demo/sensor.py
abbiemery: Done.

src/ophyd_async/epics/signal/__init__.py
abbiemery: I have changed it as requested but this feels like it should just be a private function, its only used in signal to make the required signal backends.

src/ophyd_async/epics/adcore/_ad_base.py
abbiemery: I have done both of these things. I called the other file `_core_logic.py`, not a very compelling name but I'm not sure what else it should be called.

src/ophyd_async/epics/adcore/__init__.py
abbiemery: Renamed both.

src/ophyd_async/epics/adcore/__init__.py
abbiemery: Renamed.

src/ophyd_async/epics/signal/__init__.py
abbiemery: I would even argue that it should be omited from the init and imported just for the tests with a noqa.

src/ophyd_async/epics/signal/_epics_transport.py
abbiemery: I'm not sure how I feel about this, we have a lot of use of SignalBackend throughout the codebase and I think this rename might just add another thing to be confused by.

abbiemery: Looking at this does this want to be public at all? Should this not be `_EpicsTransport`.

src/ophyd_async/fastcs/panda/__init__.py
abbiemery: I don't have strong opinions on PandA namings, but perhaps this can be done in a subsequent PR after further discussion?

src/ophyd_async/epics/signal/__init__.py
abbiemery: Refactored the test to not use it so it has been renamed but it now private.

src/ophyd_async/epics/signal/_epics_transport.py
abbiemery: I have made this private and not renamed it, it can be changed in the future before it is made public.

src/ophyd_async/fastcs/panda/__init__.py
coretl: Leave as is for now

coretl (approved):

PR #477, opened 2024-07-23
Remove references to root in PathInfo, PathProvider

jwlodek:
Addresses #470


src/ophyd_async/core/_providers.py
```diff
@@ -13,21 +13,13 @@ class PathInfo:
...
 
-    The bluesky event model splits the URI for a resource into two segments to aid in
-    different applications mounting filesystems at different mount points.
-    The portion of this path which is relevant only for the writer is the 'root',
-    while the path from an agreed upon mutual mounting is the resource_path.
-    The resource_dir is used with the filename to construct the resource_path.
-
-    :param root: Path of a root directory, relevant only for the file writer
-    :param resource_dir: Directory into which files should be written, relative to root
+    :param resource_path: Directory into which files should be written, relative to root
```
coretl, line 16: I got confused about whether `resource_path` included the filename or not. Should it be called `directory` instead?

coretl (commented):
Question on naming, apart from that it looks good

src/ophyd_async/core/_providers.py
jwlodek: Does not include the filename. Should I rename to `directory_path` then?

coretl: Yes please

mrakitin (approved):
Looks good to me!

Issue #482, opened 2024-07-24
Adopt SFL001 for ruff for src


abbiemery:
As was decided here https://github.com/DiamondLightSource/python-copier-template/pull/174 we are going to be adding this flag to ruff. We need to adjust the code accordingly to comply.

Issue #299, opened 2024-05-10
Enable `SLF001` ruff check


coretl:
As per https://github.com/DiamondLightSource/python-copier-template/issues/154

stan-dot:
`SLF001` is about private class method/property access

coretl:
Waiting for #383

coretl:
@abbiemery I merged this into the copier template main, please could you try doing a `copier update --vcs-ref=main` and check this works?

abbiemery:
I did the changes to ophyd-async that are JUST the SLF ones, as the other ones on the copier template bring in a lot of errors to fix. I suggest we merge the SLF changes now, and make issues to adopt the other ones. 

https://github.com/bluesky/ophyd-async/pull/483

PR #483, opened 2024-07-24
Add ruff SLF001

abbiemery:
Fixes https://github.com/bluesky/ophyd-async/issues/482. #299 

I'm not sure if we want a method to access `_backend` in signal, or to just make `backend` public, but I went with the former.

### DIFF
--- a/src/ophyd_async/core/_mock_signal_utils.py
+++ b/src/ophyd_async/core/_mock_signal_utils.py
@@ -8,11 +8,12 @@ from ._utils import T
 
 
 def _get_mock_signal_backend(signal: Signal) -> MockSignalBackend:
-    assert isinstance(signal._backend, MockSignalBackend), (
+    backend = signal.get_backend()
+    assert isinstance(backend, MockSignalBackend), (
         "Expected to receive a `MockSignalBackend`, instead "
-        f" received {type(signal._backend)}. "
+        f" received {type(backend)}. "
     )
-    return signal._backend
+    return backend
 
 
 def set_mock_value(signal: Signal[T], value: T):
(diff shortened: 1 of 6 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -110,6 +110,9 @@ async def connect(
         ), "this assert is for type analysis and will never fail"
         await self._connect_task
 
+    def get_backend(self) -> SignalBackend:
+        return self._backend
+
```
coretl, line 115: I explicitly wanted to hide the backend from the public interface, as the only place that uses this is in the mock signal backend access above, so it shouldn't be public for everything else

src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -8,11 +8,12 @@
 
 
 def _get_mock_signal_backend(signal: Signal) -> MockSignalBackend:
-    assert isinstance(signal._backend, MockSignalBackend), (
+    backend = signal.get_backend()
```
coretl, line 11: I've got 2 suggestions here, either:
- Add a `#noqa` and leave the private access
- When a mock signal backend is created, put it in a global dict, and grab it out of that dict in this function

abbiemery: I'll just do a `#noqa`, its only one place.

src/ophyd_async/core/_signal.py
abbiemery: Okay cool.

coretl (approved):

PR #490, opened 2024-07-25
Tests for AreaDetector drivers/plugins to insure attributes are regularly named

### LINKED ISSUE #80: Add test that areaDetector drivers and plugins have regularly named attributes
Expect `inflection.underscore(pv_name) == attribute_name`.

E.g. `inflection.underscore("NDArrayAddress") == "nd_array_address"`

subinsaji:
Closes #80 

Recalling conversations with @coretl, in this PR we made changes to some names of attributes in order for tests to pass. New test that was created in here https://github.com/bluesky/ophyd-async/blob/5f49153e1bede552453183e2fa32a30b1fe50162/tests/epics/test_areadetector_subclass_naming.py#L1-L32

needs to import all subsequent subclasses from from `ophyd_async.epics` for test coverage. If a developer adds more plugins or drivers, the developer must also import them.


src/ophyd_async/epics/adpilatus/_pilatus_io.py
```diff
@@ -19,5 +19,5 @@ def __init__(self, prefix: str, name: str = "") -> None:
         self.trigger_mode = epics_signal_rw_rbv(
             PilatusTriggerMode, prefix + "TriggerMode"
         )
-        self.armed_for_triggers = epics_signal_r(bool, prefix + "Armed")
+        self.armed_for_triggers = epics_signal_r(bool, prefix + "ArmedForTriggers")
```
coretl, line 22: The PV is set in stone, we can only change the attribute name to match...
```suggestion
        self.armed = epics_signal_r(bool, prefix + "Armed")
```

coretl (changes requested):

coretl:
@DiamondJoseph @callumforrester @jwlodek this PR will mean that all areaDetector classes are guaranteed to have predictable attribute names. E.g. the PV `$(P)$(R)MyAttribute` will be called `self.my_attribute`. This is the mapping that PVI will do by default, although if we *really* don't like the PV name we could put an exception in for particular cases.

Are we happy to keep the slightly more obtuse attribute names so we get uniformity and predictability?

DiamondJoseph:
consistency >> all other concerns IMO

jwlodek:
This is OK with me as well.

coretl (approved):

PR #497, opened 2024-07-29
AD HDF plugin should set directory creation value before path

jwlodek:
If the create_dir_depth record is not set first, then it will have no effect, since the callback to create the directories is done when the file path record is processed.


src/ophyd_async/epics/adcore/_hdf_writer.py
```diff
@@ -51,16 +51,15 @@ def __init__(
     async def open(self, multiplier: int = 1) -> Dict[str, DataKey]:
         self._file = None
         info = self._path_provider(device_name=self.hdf.name)
-        file_path = info.directory_path
         await asyncio.gather(
             self.hdf.num_extra_dims.set(0),
             self.hdf.lazy_open.set(True),
             self.hdf.swmr_mode.set(True),
+            self.hdf.create_dir_depth.set(info.create_dir_depth),
```
coretl, line 58: The gather will fire them off but there's no guarantee in the order they will be received. Better to `await self.hdf.create_dir_depth` before doing the `await asyncio.gather` of all the other signals

coretl (changes requested):

coretl (approved):

PR #499, opened 2024-07-30
added support for str

### LINKED ISSUE #451: Support `str` datatype for CA signals that connect to an mbbo/mbbi record
This is so we can say "it's an enum" at runtime and get its values, but not need to know what those values are at connect time

ZohebShaikh:
closes #451


tests/epics/signal/test_signals.py
```diff
@@ -769,6 +769,15 @@ async def test_str_enum_returns_enum(ioc: IOC):
     assert val == "Bbb"
 
 
+async def test_str_datatype_in_CA_signals_mbbo_mbbi(ioc: IOC):
```
DiamondJoseph, line 772: This test seems to be testing for both ioc.protocols, so the name could be adjusted

coretl (approved):

DiamondJoseph (approved):

PR #500, opened 2024-07-30
NDStats docs update

ZohebShaikh:
small PR to fix docs string for NDStats

### DIFF
--- a/src/ophyd_async/epics/adcore/_core_io.py
+++ b/src/ophyd_async/epics/adcore/_core_io.py
@@ -44,9 +44,6 @@ class NDPluginBaseIO(NDArrayBaseIO):
 class NDPluginStatsIO(NDPluginBaseIO):
     """
     Plugin for computing statistics from an image or region of interest within an image.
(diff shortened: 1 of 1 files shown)


src/ophyd_async/epics/adcore/_core_io.py
```diff
@@ -44,9 +44,6 @@ def __init__(self, prefix: str, name: str = "") -> None:
 class NDPluginStatsIO(NDPluginBaseIO):
     """
     Plugin for computing statistics from an image or region of interest within an image.
-    Each boolean signal enables or disables all signals in the appropriate Enum class.
-    The enum signals may used in the ScalarSignals kwargs of a HDFWriter, and are also
-    read-only signals on the plugin.
```
DiamondJoseph, line 49: Still useful to note which signals are enabled/disabled by the bool signals: maybe sort the signals such that each `compute_` is directly above all of the signals it then enables?

coretl (approved):

PR #511, opened 2024-08-07
move h5py into real dependencies

stan-dot:


DiamondJoseph:
Isn't this only required for the SimPatternGenerator? Can we have it as a sim dependency and note that using the SimPatternGenerator requires it to be available?

stan-dot:
sounds like we could do that, I am not familiar with sim dependencies however. Do tell more

callumforrester:
@coretl said to do that before he left. 

@stan-dot, you need to make the sim dependencies. You should make a set of optional dependencies (similar to `dev`) that lists dependencies that are only required for the sim module. Then a user can type `pip install ophyd-async[sim]` to get them.

stan-dot:
ah I see, it's not that complicated then. thanks!

DiamondJoseph:
Would having in the pyproject.toml of your beamline plans module "ophyd-async[sim]" pick up those dependencies through any chaining? Good thing to try out.

stan-dot:
@DiamondJoseph why wouldn't it?

jwlodek:
I am pretty sure that is how it works. I had packages that did dev installs of dependencies before and it worked

callumforrester:
@DiamondJoseph yes, you can have this, if that's what you're asking

```toml
dev = [
    "ophyd-async[sim]",
    ...
]

sim = [
    ...
]
```

DiamondJoseph (approved):

DiamondJoseph:
> ```toml
> dev = [
>     "ophyd-async[sim]",
>     ...
> ]
> ```

I was more thinking:

dodal.dependencies = "ophyd-async"
i22_bluesky.dependencies = "dodal, ophyd-async[sim]"

DiamondJoseph:
@stan-dot please update the branch and when merging squash and reword the commit

Issue #394, opened 2024-06-17
PandaHDFWriter should only take data block, not whole PandA


coretl:
https://github.com/bluesky/ophyd-async/blob/febf9ceab0744a4699ad5d208f7b737cb3b7376d/src/ophyd_async/panda/writers/_hdf_writer.py#L26

The only use of `self.panda_device` that is not the `DATA` block is
https://github.com/bluesky/ophyd-async/blob/febf9ceab0744a4699ad5d208f7b737cb3b7376d/src/ophyd_async/panda/writers/_hdf_writer.py#L49

Which will go away after #245

PR #512, opened 2024-08-07
Only pass data block to init of PandAHDFWriter instead of entire PandA device

jwlodek:
Resolves https://github.com/bluesky/ophyd-async/issues/394


evalott100 (approved):
Good work!

Issue #516, opened 2024-08-10
`bool` signals use invalid `dtype`


Tom-Willemsen:
My attempt at a minimal user-facing example - might be able to simplify this further...

```python
from bluesky.run_engine import RunEngine
from bluesky.plans import count
from ophyd_async.epics.signal import epics_signal_r
from ophyd_async.plan_stubs import ensure_connected        

RE = RunEngine()    
sig = epics_signal_r(
    bool, 
    "TE:NDW2922:CS:SB:mot:RC:INRANGE",   # Any bi/bo pv, over either ca or pva (but not the mock backend)
    name="foo"
)

def plan():
    yield from ensure_connected(sig)
    yield from count([sig])

RE(plan(), print)

```

Currently this raises:

```
jsonschema.exceptions.ValidationError: 'bool' is not one of ['string', 'number', 'array', 'boolean', 'integer']

Failed validating 'enum' in schema['properties']['data_keys']['additionalProperties']['properties']['dtype']:
    {'title': 'Dtype',
     'description': 'The type of the data in the event.',
     'type': 'string',
     'enum': ['string', 'number', 'array', 'boolean', 'integer']}

On instance['data_keys']['foo']['dtype']:
    'bool'
```

Should presumably be using `"boolean"` rather than `"bool"` as the dtype for `bool`-type signals.

PR #517, opened 2024-08-10
Use 'boolean' rather than 'bool' dtype per event-model

Tom-Willemsen:
Fixes https://github.com/bluesky/ophyd-async/issues/516

Would have liked to add a proper system test for this, but the EPICS-level tests aren't working on windows on `main` for some reason - I hope to get time to look at that issue in a separate PR...


jwlodek (approved):
LGTM

coretl (approved):

Issue #310, opened 2024-05-16
Make a concrete superset of types that Signal should allow


callumforrester:
As a developer I would like to solidify the requirement for enum signals to also inherit from `str` so that I can have more confidence my enum signals will connect.

## Possible Solution
We don't want to pollute all the backends in a duplicating way so the check should go in `Signal` (@coretl?)
For static type checking, only support certain data types for signals (via a `Union`)

## Acceptance Criteria
- When a signal is created with an enum type, validate that it inherits from `str` as well
- When instantiating a device with an enum signal that doesn't inherit from `str`, an error is raised
- Tests to validate behavoir

coretl:
One idea is:
- Make `Signal` take an optional datatype on `__init__`
- Have each `SignalBackend` take a union datatype classvar `supported_datatypes`
- When the backend is set on the Signal then check that the `Signal` datatype is contained in the union of the `SignalBackend.supported_datatypes` and fail with a suitable error message
- This should happen in mock mode against the real backend so we can validate types without connecting to the control system

evalott100:
I believe I implemented this during https://github.com/bluesky/ophyd-async/pull/341

coretl:
@burkeds this would make the following changes to `Signal` and `SignalBackend`:

- `Signal` would take an optional `datatype` in `__init__` and do some validation on it
- `EpicsSignalBackend` would no longer take `datatype` in `__init__`
- `SignalBackend.connect` now takes `datatype` instead

Would that be fine for the Tango backend?

burkeds:
@coretl  This shouldn't cause any major issues. I am in the process of doing some major refactoring of the tango backend so now would be a good time to implement a change like this.

coretl:
@evalott100 when the PCOMP change is done, this is higher priority than the layout ticket

coretl:
The array of types we should accept:

1. `str, int, float, bool`
2. `str, Enum`, which should mean we can drop the `write_value` conversion of the `*EnumConverter`s
3. `np.ndarray` of `np.bool_`, `np.[u]int[8,16,32,64]`, `np.float[32,64]`, `str`
4. `np.ndarray` of record array of the above to support tables
5. `pydantic.BaseModel`

For the transport mechanisms:
- CA should accept 1, 2, half of 3
- PVA should accept 1, 2, 3, 4
- Soft should accept all

Depends on #442

coretl:
Query on the usecase for BaseModels

evalott100:
[...]

PR #522, opened 2024-08-15
Introduce PvaAbstractions and use them in SeqTable

### LINKED ISSUE #442: Switch from `TypedDict` to row-wise numpy structured data for tables
_Originally posted by @coretl in https://github.com/bluesky/ophyd-async/pull/430#discussion_r1669932540_

PVA enforces a strict datatype for each of the columns of the table which allows us to use a numpy structured data type for each row, which we can check against. This has the advantage that it is stored row-wise, so will always have columns the same length. It does mean doing a transpose from PVA column-wise to numpy row-wise, but this is done by numpy in C, so is not a big concern.

To do this:
- Construct a numpy structured datatype `SeqTableRow = np.dtype([('repeats', np.uint16), ('trigger', SeqTrigger), ...])`
- Remove `SeqTable`
- Make sequencer block have `table: npt.NDArray[SeqTableRow]`
- Adapt PVA converter for table to check dtypes of the rows against datatype of PVA structure, and transform between them

Enums are expected to be difficult, so you might need to change them to `SubsetEnum` instead, or even drop to `str`

evalott100:
Closes #442 and #310


coretl (commented):
As discussed, please could we have:
```python
SignalType = TypeVar("SignalType", bool, int, ..., BaseModel)

class Signal:
    def __init__(self, datatype: Optional[SignalType] = None):
        check_one_of(datatype, SignalType)
        ...
```

I forgot that we wanted to take `datatype` in `Signal.__init__` so that we could create a `Signal` with a type but no backend in `create_children_from_annotations`, then fill it in on connect.

As per https://github.com/bluesky/ophyd-async/issues/310#issuecomment-2205415110

evalott100:
@coretl 

Can we implement https://github.com/bluesky/ophyd-async/issues/310#issuecomment-2205415110 in another issue/PR? This is already a very big change, and changing init/connect args for `Signal`s/`SignalBackend`s touches so many places in the code-base.

Additionally, I'd like to disallow `arg -> kwarg` inference on signal and signal backend init and connect which we could do in the same PR.

coretl:
> @coretl
> 
> Can we implement [#310 (comment)](https://github.com/bluesky/ophyd-async/issues/310#issuecomment-2205415110) in another issue/PR? This is already a very big change, and changing init/connect args for `Signal`/`SignalBackend` touches so many places in the code-base.
> 
> Additionally, I'd like to disallow `arg -> kwarg` inference on `Signal` and `SignalBackend` init and connect which we could do in the same PR.

Sounds like a good idea, let me know when this needs another review

evalott100:
New issue here. I'm ready for a new review https://github.com/bluesky/ophyd-async/issues/562

src/ophyd_async/epics/signal/_p4p_table_model.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+
+    @model_validator(mode="after")
+    def validate_arrays(self) -> "PvaTable":
+        first_length = len(next(iter(self))[1])
+        assert all(
+            len(field_value) == first_length for _, field_value in self
+        ), "Rows should all be of equal size."
+
+        assert 0 <= first_length < 4096, f"Length {first_length} not in range."
```
coretl, line 47: This is only a limit on the PandA SeqTable, no other tables have a limit of 4096 rows

src/ophyd_async/fastcs/panda/_table.py
```diff
@@ -34,137 +36,88 @@ class SeqTrigger(str, Enum):
...
+    NpArrayPydanticAnnotation.factory(
+        data_type=np.unicode_, dimensions=1, strict_data_typing=False
+    ),
+    Field(default_factory=lambda: np.array([], dtype=np.dtype("<U32"))),
+]
+
+
+class SeqTable(PvaTable):
+    repeats: PydanticNp1DArrayInt32
+    trigger: PydanticNp1DArrayUnicodeString
```
coretl, line 65: Can we make this support `Sequence[Enum]` and `Sequence[str]` too please?
```suggestion
    trigger: Sequence[SeqTrigger]
```

src/ophyd_async/fastcs/panda/_table.py
```diff
@@ -34,137 +36,88 @@ class SeqTrigger(str, Enum):
...
+    outd2: PydanticNp1DArrayBool
+    oute2: PydanticNp1DArrayBool
+    outf2: PydanticNp1DArrayBool
+
+    @classmethod
+    def row(
+        cls,
+        *,
+        repeats: int = 0,
+        trigger: str = "",
```
coretl, line 87:
```suggestion
        trigger: SeqTrigger = SeqTrigger.IMMEDIATE,
```

src/ophyd_async/fastcs/panda/_table.py
```diff
@@ -34,137 +36,88 @@ class SeqTrigger(str, Enum):
...
+    outc2: PydanticNp1DArrayBool
+    outd2: PydanticNp1DArrayBool
+    oute2: PydanticNp1DArrayBool
+    outf2: PydanticNp1DArrayBool
+
+    @classmethod
+    def row(
+        cls,
+        *,
+        repeats: int = 0,
```
coretl, line 86:
```suggestion
        repeats: int = 1,
```

src/ophyd_async/fastcs/panda/_table.py
```diff
@@ -34,137 +36,88 @@ class SeqTrigger(str, Enum):
...
+            oute1=oute1,
+            outf1=outf1,
+            time2=time2,
+            outa2=outa2,
+            outb2=outb2,
+            outc2=outc2,
+            outd2=outd2,
+            oute2=oute2,
+            outf2=outf2,
+        )
```
coretl, line 123: As we are dropping typing here, might as well make the code shorter. If we made `PvaTable.row` a `@staticmethod`, then would this work?
```suggestion
        return PvaTable.row(**locals())
```

src/ophyd_async/fastcs/panda/_table.py
coretl: Although I'm now confused that this doesn't show up in tests...

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -74,24 +72,26 @@ def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
     trigger_time = number_of_frames * (exposure + deadtime)
     pre_delay = max(period - 2 * shutter_time - trigger_time, 0)
 
-    table: SeqTable = seq_table_from_rows(
+    table = (
         # Wait for pre-delay then open shutter
-        SeqTableRow(
+        SeqTable.row(
             time1=in_micros(pre_delay),
```
coretl, line 78: ... as we have an implicitly defined `repeats` field that will have changes defaults from `1` to `0`. Are we lacking a test for this plan stub?

src/ophyd_async/core/_signal_backend.py
```diff
@@ -11,6 +19,14 @@ class SignalBackend(Generic[T]):
     #: Datatype of the signal value
     datatype: Optional[Type[T]] = None
 
+    _ALLOWED_DATATYPES: ClassVar[Tuple[Type]]
+
+    @classmethod
+    @abstractmethod
+    def datatype_allowed(cls, dtype: type):
+        """Check if a given datatype is acceptable for this signal backend."""
+        pass
```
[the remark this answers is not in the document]
evalott100: Ah I left it in

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -122,17 +127,46 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
...
+class SoftPydanticModelConverter(SoftConverter):
+    def __init__(self, datatype: Type[BaseModel]):
+        self.datatype = datatype
+
+    def reading(self, value: T, timestamp: float, severity: int) -> Reading:
+        value = self.value(value)
+        return super().reading(value, timestamp, severity)
+
+    def value(self, value: Any) -> Any:
+        if isinstance(value, dict):
```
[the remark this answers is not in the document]
evalott100: It's for when it's deserialized. When it's serialized we have the yaml dumper calling `model.dump`.

When it's deserialized it loads all the values from file then puts to the backend:

https://github.com/bluesky/ophyd-async/blob/1de860b0cd99258828fe9bd8b536e6f6cb8bbea1/src/ophyd_async/core/_soft_signal_backend.py#L174-L181

The `write_value` here is the loaded dict which we don't need to format:

``` python
    def write_value(self, value):
        if isinstance(value, self.datatype):
            return value.model_dump(mode="python")
        return value
```

But then we set value and we want that to be the model again.

I know conversion isn't necessary on a soft signal, but we probably do want to match the conversion on serialization/deserialization that we have for hard signals?

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -122,17 +127,46 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
...
+            value = self.datatype(**value)
+        return value
+
+    def write_value(self, value):
+        if isinstance(value, self.datatype):
+            return value.model_dump(mode="python")
```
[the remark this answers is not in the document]
evalott100: Yep, all for the functionality outlined above.

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -122,17 +127,46 @@ def make_initial_value(self, datatype: Optional[Type[T]]) -> T:
...
+
 def make_converter(datatype):
     is_array = get_dtype(datatype) is not None
     is_sequence = get_origin(datatype) == abc.Sequence
     is_enum = inspect.isclass(datatype) and (
         issubclass(datatype, Enum) or issubclass(datatype, RuntimeSubsetEnum)
     )
+    is_pydantic_model = (
+        inspect.isclass(datatype)
+        and isinstance(datatype, ABCMeta)
```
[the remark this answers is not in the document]
evalott100: If `datatype` is `numpy.ndarray[typing.Any, numpy.dtype[numpy.int8]]`
Then `inspect.isclass(datatype) is True`, `issubclass(datatype, Enum) is True`, and `issubclass(datatype, BaseModel)` gives

```
  File "<string>", line 1, in <module>
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
```

This will be because abc metaclassing will do strange things with [`__subclasscheck__`](https://docs.python.org/3/reference/datamodel.html#class.__subclasscheck__), namely

https://github.com/python/cpython/blob/fb1b51a58df4315f7ef3171a5abeb74f132b0971/Lib/_py_abc.py#L108-L111
``` python
    def __subclasscheck__(cls, subclass):
        """Override for issubclass(subclass, cls)."""
        if not isinstance(subclass, type):
            raise TypeError('issubclass() arg 1 must be a class')
```

I was initially intrigued too, but it's a pain to debug and find out what's actually happening here with the hooks. That numpy type should pass this test.

src/ophyd_async/epics/signal/_p4p_table_model.py
evalott100: Tbh I was hoping I could specify this in the `Field` with `lt=4096` I'll try and do this again.

src/ophyd_async/core/_soft_signal_backend.py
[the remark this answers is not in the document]
coretl: I agree for `write_value()`, it needs to take a dict or a BaseModel and return a BaseModel. But this is for `value()` which should always take the internal stored `_value` (a BaseModel) and return it, it shouldn't ever get a dict should it?

src/ophyd_async/core/_soft_signal_backend.py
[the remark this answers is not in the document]
coretl: Ok, please could you put a comment in that says this is for npt.NDArray so we don't forget this oddity!

src/ophyd_async/epics/signal/_aioca.py
```diff
@@ -256,8 +259,31 @@ def _use_pyepics_context_if_imported():
...
+
+    @classmethod
+    def datatype_allowed(cls, datatype: Optional[Type]) -> bool:
+        stripped_origin = get_origin(datatype) or datatype
+        if datatype is None:
+            return True
+
+        return inspect.isclass(stripped_origin) and issubclass(
+            stripped_origin, cls._ALLOWED_DATATYPES
+        )
```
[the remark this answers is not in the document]
coretl: Let's can this idea for now, I've got some other ideas and we need to add typing support for Tango's commands too

src/ophyd_async/core/_soft_signal_backend.py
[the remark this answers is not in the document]
evalott100: Ah yea I just have to remove it from `reading`

src/ophyd_async/epics/signal/_p4p_table_model.py
evalott100: Hmmm this doesn't work and I don't really know why :thinking: We could add an abstract property called `max_length` in `SeqTable` which would be used in the check in `Table`?

src/ophyd_async/fastcs/panda/_table.py
evalott100: I suppose only the bare minimum of panda logic is actually simulated within tests

src/ophyd_async/fastcs/panda/_table.py
evalott100: `locals()` here includes a lot more than the kwargs we want. I changed it to the following:
``` python
        sig = inspect.signature(cls.row)
        kwargs = {k: v for k, v in locals().items() if k in sig.parameters}
        return Table.row(cls, **kwargs)
```

src/ophyd_async/fastcs/panda/_table.py
evalott100: I've adjusted it so that a `Sequence` of `SeqTrigger` can be given instead. Internally it will always be a unicode string though.

src/ophyd_async/fastcs/panda/_table.py
coretl: > `locals()` here includes a lot more than the kwargs we want. 

Do you get more than this?

```python
>>> class A:
...     @classmethod
...     def meth(cls, x=1, b=2):
...             print(locals())
... 
>>> A.meth()
{'cls': <class '__main__.A'>, 'x': 1, 'b': 2}
```

src/ophyd_async/plan_stubs/_fly.py
evalott100: We do have tests for this but at no point in them do we check on `repeats`

coretl (approved):

Issue #523, opened 2024-08-16
`scan` fails with an `epics_signal_r` as a detector


Tom-Willemsen:
As a user I expected to be able to use an `epics_signal_r` which is a `Readable` as a detector in a `scan()` plan:

```python
from ophyd_async.epics.signal import epics_signal_rw, epics_signal_r
from bluesky.plans import scan
from bluesky.run_engine import RunEngine
from ophyd_async.plan_stubs import ensure_connected

RE = RunEngine()


def plan():
    sig1 = epics_signal_r(float, "TE:NDW2922:PARS:USER:R0", name="sig1")
    sig2 = epics_signal_rw(float, "TE:NDW2922:PARS:USER:R1", name="sig2")
    yield from ensure_connected(sig1, sig2)
    yield from scan([sig1], sig2, 0, 1, 2)


RE(plan(), print)
```

This currently raises:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\run_engine.py", line 964, in __call__
    plan_return = self._resume_task(init_func=_build_task)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\run_engine.py", line 1100, in _resume_task
    raise exc
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\run_engine.py", line 1734, in _run
    raise err
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\run_engine.py", line 1588, in _run
    msg = self._plan_stack[-1].send(resp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<stdin>", line 6, in plan
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\plans.py", line 1153, in scan
    return (yield from scan_nd(detectors, full_cycler, per_step=per_step, md=_md))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Instrument\dev\ophyd-async\.venv\Lib\site-packages\bluesky\plans.py", line 1042, in scan_nd
[...]

PR #524, opened 2024-08-16
Allow equality comparisons for signal

Tom-Willemsen:
Fixes https://github.com/bluesky/ophyd-async/issues/523

These comparisons are needed by trigger_and_read for example.

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -45,15 +45,6 @@ def _add_timeout(func):
     return wrapper
 
 
-def _fail(self, other, *args, **kwargs):
-    if isinstance(other, Signal):
-        raise TypeError(
-            "Can't compare two Signals, did you mean await signal.get_value() instead?"
-        )
-    else:
-        return NotImplemented
-
-
 class Signal(Device, Generic[T]):
     """A Device with the concept of a value, with R, RW, W and X flavours"""
 
@@ -117,12 +108,6 @@ class Signal(Device, Generic[T]):
     def source(self) -> str:
         """Like ca://PV_PREFIX:SIGNAL, or "" if not set"""
         return self._backend.source(self.name)
-
-    __lt__ = __le__ = __ge__ = __gt__ = _fail
-
-    def __hash__(self):
-        # Restore the default implementation so we can use in a set or dict
-        return hash(id(self))
 
 
 class _SignalCache(Generic[T]):
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -118,7 +118,7 @@ class Signal(Device, Generic[T]):
         """Like ca://PV_PREFIX:SIGNAL, or "" if not set"""
         return self._backend.source(self.name)
 
-    __lt__ = __le__ = __eq__ = __ge__ = __gt__ = __ne__ = _fail
+    __lt__ = __le__ = __ge__ = __gt__ = _fail
 
     def __hash__(self):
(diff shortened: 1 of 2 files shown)


coretl:
Hmm, so I was trying to catch the case when people do `if sig > 4` or if `sig1 != sig2` then it fails with a suitable error message. Unfortunately this breaks `if sig in signals:` which is used in `trigger_and_read`.

@tacaswell @danielballan can you think of any magic that would let us detect the difference between `sig1 == sig2` and `sig1 in signals`?

If not then I would be inclined to also tear out the `_fail` function and the other comparison methods and let people discover themselves that `sig > 4` doesn't do what they thought it did...

tacaswell:
Per https://docs.python.org/3/reference/datamodel.html#object.__hash__ 

> If a class does not define an __eq__() method it should not define a __hash__() operation either

The best path maybe to fail all the ordering related tests and special case  `__eq__` to fail if the `other` is an int or float (and mayb str)? 

It is a bit funny and "you can use this in a set with other objects but _NOT_ with ints" is a sharp edges that I can not come up with a compelling story why that would cause problems, but I am confident will burn a day of someones time in the decade when they discover it 🤣 .  However, that may be worth the trade off for time saved from `if sig == 5` giving a clear error many times.

Tom-Willemsen:
This all feels a bit... too magical for my taste.

---

> let people discover themselves that `sig > 4` doesn't do what they thought it did...

The existing `_fail` implementation only deals with comparing two signals. Both on `main` and with `_fail` removed, you get:

```
>>> sig > 4
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'SignalR' and 'int'
```
That seems clear enough to me. The difference is when you compare two signals...

`main`:
```
>>> sig > sig
  File "<stdin>", line 1, in <module>
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\core\_signal.py", line 50, in _fail
    raise TypeError(
TypeError: Can't compare two Signals, did you mean await signal.get_value() instead?
```
vs without a `_fail` implementation
```
>>> sig > sig
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'SignalR' and 'SignalR'
```
I don't really see a massive difference between these from a user/dev perspective, in both cases it's pretty obvious what the problem is...

Tom-Willemsen:
For the `__eq__` implementation, I think defining a custom `__eq__` might actually be counterproductive, as it prevents type-checkers from emitting useful diagnostics (notice how _without_ the custom `__eq__` implementation, pyright in strict mode reports an error)

```
# On main
(.venv) c:\Instrument\dev\ophyd-async>pyright test.py
0 errors, 0 warnings, 0 informations

# without a custom __eq__
(.venv) c:\Instrument\dev\ophyd-async>pyright test.py
c:\Instrument\dev\ophyd-async\test.py
  c:\Instrument\dev\ophyd-async\test.py:4:4 - error: Condition will always evaluate to False since the types "SignalR[float]" and "Literal[4]" have no overlap (reportUnnecessaryComparison)
1 error, 0 warnings, 0 informations
```
Content of `test.py`:
```python
from ophyd_async.epics.signal import epics_signal_r

sig = epics_signal_r(float, "foo", name="foo")
if sig == 4:
    print("foo")
```

coretl:
> but I am confident will burn a day of someones time in the decade when they discover it 🤣

I'm slightly concerned that person might end up being me, having forgotten about the cleverness I inserted back in the mists of time...

> This all feels a bit... too magical for my taste.

Agreed, I'm happy for you to rip it all out

coretl (approved):

Issue #405, opened 2024-06-19
Decide the interface between `StandardDetector` and `DetectorController` w.r.t. the `TriggerInfo` that is passed down in `prepare`


coretl:
As mentioned in https://github.com/bluesky/ophyd-async/pull/401#discussion_r1644527435 and https://github.com/bluesky/ophyd-async/pull/401#discussion_r1644506670, we need to decide what `StandardDetector` does with the `TriggerInfo` passed down to `prepare`. We should also consider #162 and #357 while we're at it. I propose:
- `async def DetectorController.arm(num, trigger, exposure) -> Status` -> `async def DetectorController.arm(trigger_info) -> None`, where the return value for `acquire.set(True)` is stored locally rather than returned
- Add a new `async def DetectorController.wait_for_disarmed(timeout=DEFAULT_TIMEOUT)` that waits for aquire to be False and does the successful completion check that will fail if the detector did anything apart from normal completion or manual disarming
- `async def DetectorController.disarm()` -> `async def DetectorController.start_disarm()` as it doesn't wait for completion

We should also document the various permutations of `TriggerInfo` and provide any common helpers like the one defined in #401. The cases we need to document:
- trigger=internal, livetime=deadtime=None: Internal trigger mode, with exposure and period untouched. The default for `trigger()` if `prepare()` not called
- trigger=any, livetime=float: Exposure set to livetime and period set to livetime+`get_deadtime(livetime)`
- trigger=any, livetime=float, deadtime=float: Exposure set to livetime and period set to deadtime
- num=0: Go until disarmed, if detector supports it
- num=+ve: Arm for this many frames

We still need to work out if we should provide things like min_exposure and max_framerate as well as `get_deadtime(exposure)`, but this will we tackled in a different issue

coretl:
@callumforrester @abbiemery @evalott100 @DiamondJoseph thoughts?

callumforrester:
I think there is a use case for keeping the detector to minimum deadtime while the _system_ as a whole has a longer deadtime, which is what this diagram covers. Do any your permutations cover that?

![image](https://github.com/bluesky/ophyd-async/assets/29771545/d94177e7-3a6d-4015-8050-d6050cae66f6)

coretl:
[...]

PR #528, opened 2024-08-19
Fastcs eiger support

DominicOram:
See https://github.com/DiamondLightSource/dodal/issues/700. This implements enough of the Eiger/Odin for us to prove that we can actually take some data with it. To completely cover all the capability of the `ophyd` device we need to do a lot, at least:

* https://github.com/bluesky/ophyd-async/issues/529
* https://github.com/bluesky/ophyd-async/issues/530
* https://github.com/bluesky/ophyd-async/issues/531
* https://github.com/bluesky/ophyd-async/issues/532

# To Test
* Confirm unit tests pass
* Run the system tests as described in the system test directory, confirm they pass

# Notes
* I'm particularly interested in someone who knows the intent of the detector architecture to review this to confirm it follows the pattern correctly.


src/ophyd_async/epics/eiger/_odin_io.py
```diff
@@ -0,0 +1,124 @@
...
+    async def open(self, multiplier: int = 1) -> Dict[str, DataKey]:
+        info = self._path_provider(device_name=self._name_provider())
+
+        await asyncio.gather(
+            self._drv.file_path.set(str(info.directory_path)),
+            self._drv.file_name.set(info.filename),
+            self._drv.data_type.set(
+                "uint16"
+            ),  # TODO: Get from eiger https://github.com/bluesky/ophyd-async/issues/529
+            # TODO: Where should I actually be setting this
```
DominicOram, line 86: I'm not sure on the best way to do this to best fit into the existing architecture

system_tests/epics/eiger/start_iocs_and_run_tests.sh
callumforrester, line 1: @gilesknap I wonder if this can be achieved with `docker-compose` or similar, so we can run it up either in the CI or locally in exactly the same way?

pyproject.toml
```diff
@@ -138,7 +138,7 @@ commands =
 
 
 [tool.ruff]
-src = ["src", "tests"]
+src = ["src", "tests", "system_tests"]
```
callumforrester, line 141: Nit: I thought we were going with the same structure as Hyperion

- tests
     - unit_tests
     - system_tests

I think that might be neater at the top level, but I'm not especially fussed.

coretl: As long as the default set of directories for `pytest` excludes the `system_tests` directory I'm not too fussed which of these structures we go with.

I'd like to continue running `pytest` for just unit tests, then `pytest /path/to/system_tests` as a separate CI job for the system tests

system_tests/epics/eiger/start_iocs_and_run_tests.sh
DominicOram: I didn't want to hold this up on the CI stuff, which could get fiddly so made https://github.com/bluesky/ophyd-async/issues/531

pyproject.toml
DominicOram: Yes, I did it like this because it gives the behaviour @coretl describes, which I like. I will see how messy it is to have the same behaviour with the nested structure

src/ophyd_async/epics/eiger/_eiger_io.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, prefix, tolerance=0.1, name: str = "") -> None:
+        self._photon_energy = epics_signal_rw_rbv(float, f"{prefix}PhotonEnergy")
+        self.tolerance = tolerance
+        super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def set(self, value):
+        current_energy = await self._photon_energy.get_value()
+        if abs(current_energy - value) > self.tolerance:
+            await self._photon_energy.set(value)
```
coretl, line 26: Slightly concerned about putting this logic in the IO class. As written this will break load/save as it's not a Signal.

What's the use case for setting within a tolerance? If it's from a custom plan, could the logic be lifted into a plan stub? If not would a `set_energy` method on the detector be acceptable?

src/ophyd_async/epics/eiger/_odin_io.py
coretl, line 1: As soon as PVI datastructures are published over PVA then this file will drastically change to something like:
```python
class EigerDriverIO(Device):
    bit_depth: SignalR[int]
    ...
```

src/ophyd_async/epics/eiger/_odin_io.py
coretl: I'm not sure what "this" is? All the signal sets you mention seem to be in the right place, although I expected `num_to_capture` to be `0` so it would go forever...

src/ophyd_async/epics/eiger/_eiger_io.py
DominicOram: The use-case is that we want to avoid setting it as much as possible because it increases the time it takes to arm the detector and it's scientifically fine for it to be just within the tolerance. At the moment we bundle all the logic required to set up and arm the detector into a single `bps.set` because we need to do it all in parallel to another plan running. If we want to do this via `plan_stubs` then I guess we're blocked on https://github.com/bluesky/bluesky/issues/1652. We could put it in a `set_energy` and call it during the `arm`

src/ophyd_async/epics/eiger/_odin_io.py
DominicOram: > It will also only need to include the signals that are needed at static analysis time, at runtime then all the extras (like beam_centre_x, etc) will be filled in automatically.

Eventually all these signals will be needed at static analysis time. I have added in all the ones that we need to re-implement what we currently have on the `ophyd` `eiger`.

src/ophyd_async/epics/eiger/_odin_io.py
DominicOram: Sorry, you're right that `this` is the number of frames. Ok, so the idea is that we will never run the file writer knowing how many frames it has up front? I think this will have side-effects that we're not used to for example that we will always have N files, where N is the number of processors, though most of the time some of them will be empty but this is probably ok.

src/ophyd_async/epics/eiger/_eiger_io.py
DominicOram: As discussed in person, we put this in `arm`.

src/ophyd_async/epics/eiger/_odin_io.py
GDYendell: PVI structures will be added in https://github.com/DiamondLightSource/FastCS/pull/54

src/ophyd_async/epics/eiger/_eiger_io.py
DominicOram: I have put it in `prepare`. I would rather put it in the `arm` so that we have a better ordering of when to do it but to do so would need to change the `arm` interface quite a bit, probably to take something that inherits from `TriggerInfo`. Are you happy for me to do this refactor?

pyproject.toml
DominicOram: Unless you now have stronger opinions I think leave like this as it gives the cleanest way to only run some of the tests. We can change another time if we want to,

src/ophyd_async/epics/eiger/_eiger.py
```diff
@@ -0,0 +1,43 @@
...
+
+        super().__init__(
+            EigerController(self.drv),
+            OdinWriter(path_provider, lambda: self.name, self.odin),
+            name=name,
+        )
+
+    @AsyncStatus.wrap
+    async def prepare(self, value: EigerTriggerInfo) -> None:
+        await self._controller.set_energy(value.energy_ev)
```
coretl, line 42: After https://github.com/bluesky/ophyd-async/issues/405 then there will be a `prepare` method on the controller that takes a `TriggerInfoT` so this prepare method will become unnecessary. For now this is correct though.

coretl (approved):

src/ophyd_async/epics/eiger/_eiger_io.py
coretl: @ZohebShaikh will be doing this as part of https://github.com/bluesky/ophyd-async/issues/405, so it would be useful if you could review and/or add to his PR to make sure the Eiger usecase is covered

DominicOram:
Given the refactors in https://github.com/bluesky/ophyd-async/issues/405 and https://github.com/bluesky/ophyd-async/issues/551 I'm going to merge this as a MVP and park any further development on the other issues.

PR #536, opened 2024-08-20
initial commit

ZohebShaikh:


pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
     "numpy<2.0.0",
     "packaging",
     "pint",
-    "bluesky>=1.13.0a3",
+    "bluesky>=1.13.0a4",
```
ZohebShaikh, line 19: I have change this to make sure that we are at version the hardware test was conducted

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -129,17 +127,12 @@ def fly_and_collect(
     done = False
     while not done:
         try:
-            yield from bps.wait(group=group, timeout=0.5)
+            yield from bps.wait(group=group, timeout=0.5, move_on=True)
```
evalott100, line 130: Use a new global constant for readability:

``` python
# Collect every 0.5 seconds even if the flyscan is still completing
_SECONDS_TO_FORCE_COLLECT_AFTER = 0.5
```

```suggestion
            yield from bps.wait(
                group=group,
                timeout=_SECONDS_TO_FORCE_COLLECT_AFTER,
                move_on=True
            )
```

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -129,17 +127,12 @@ def fly_and_collect(
...
         except TimeoutError:
```
evalott100, line 131: I actually don't think we need to do this now. It seems timeout is only raised if `move_on=False` I think we can change this whole thing to:

``` python
while (
    yield from bps.wait(
        group=group,
        timeout=_SECONDS_TO_FORCE_COLLECT_AFTER,
        move_on=True
    )
):
    yield from bps.collect(*detectors, name=stream_name)
```

Since done is returned here:
https://github.com/bluesky/bluesky/blob/978995b69c00cc0b9bef5103bdb31b665a6fbed9/src/bluesky/run_engine.py#L2379

evalott100 (changes requested):

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -128,18 +130,10 @@ def fly_and_collect(
...
-        except TimeoutError:
-            pass
-        else:
-            done = True
-        yield from bps.collect(
-            *detectors,
-            return_payload=False,
-            name=stream_name,
+        done = yield from bps.wait(
+            group=group, timeout=_SECONDS_TO_FORCE_COLLECT_AFTER, move_on=True
```
coretl, line 134: We should add `flush_period: float = 0.5` to the function parameters, then use that rather than a global
```suggestion
            group=group, timeout=flush_period, move_on=True
```

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -102,14 +102,11 @@ def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
...
 
-# Collect every 0.5 seconds even if the flyscan is still completing
-_SECONDS_TO_FORCE_COLLECT_AFTER = 0.5
-
-
 def fly_and_collect(
     stream_name: str,
     flyer: StandardFlyer[SeqTableInfo] | StandardFlyer[PcompInfo],
     detectors: List[StandardDetector],
+    flush_period: float = 0.5,
```
evalott100, line 107: Might be nice to
```suggestion
    flush_period: float = DEFAULT_FLUSH_PERIOD,
```

evalott100 (approved):

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -126,18 +125,8 @@ def fly_and_collect(
...
-        else:
-            done = True
-        yield from bps.collect(
-            *detectors,
-            return_payload=False,
-            name=stream_name,
-        )
-    yield from bps.wait(group=group)
+        done = yield from bps.wait(group=group, timeout=flush_period, move_on=True)
+        yield from bps.collect(*detectors, name=stream_name)
```
evalott100, line 129: Is the removal of `return_payload` intentional?

evalott100 (changes requested):

src/ophyd_async/plan_stubs/_fly.py
ZohebShaikh: I have checked this today...This is working ... I'm able to collect data

evalott100: Can't see the full context for this since the github diff is cluttered with all the changes to main since this PR was opened. If you open a new PR after rebasing then add me as a reviewer :pray:

ZohebShaikh: I have created a new PR https://github.com/bluesky/ophyd-async/pull/691 because the branch go messed up...

PR #540, opened 2024-08-22
Add no op status

### LINKED ISSUE #521: Provide a NoOpStatus
`ophyd` has a `NullStatus` (https://github.com/bluesky/ophyd/blob/050b98f34d68bd9577179accfdfca8f635cc0f31/ophyd/sim.py#L51) that immediately returns, this is very useful for testing as it's something you probably want to inject into many devices. We should provide something similar in `ophyd-async` where it basically just does `AsyncStatus(asyncio.sleep(0.0))`. It will also be nice to be able to:
* Be able to provide an identifier (such as the name of the test) so that we can easily identify it later if needed
* Be able to provide a flag that causes the status to be in error

olliesilvester:
Fixes #521

### DIFF
--- a/tests/core/test_status.py
+++ b/tests/core/test_status.py
@@ -7,7 +7,7 @@ import pytest
 from bluesky.protocols import Movable, Status
 from bluesky.utils import FailedStatus
 
-from ophyd_async.core import AsyncStatus, Device
+from ophyd_async.core import AsyncStatus, Device, completed_status
 
 
 async def test_async_status_success():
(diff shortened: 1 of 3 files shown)


tests/core/test_status.py
```diff
@@ -188,3 +188,9 @@ async def coro_status(x: int, y: int, *, z=False):
         assert test_result == 12
 
     loop.run_until_complete(do_test())
+
+
+async def test_completed_status():
+    with pytest.raises(ValueError):
+        await completed_status(ValueError())
+    await completed_status()
```
dperl-dls, line 196: could: assert the completed status has `.success is True` etc.

dperl-dls (approved):
Thanks, one optional comment

PR #541, opened 2024-08-23
make motor locatable

stan-dot:


dperl-dls (changes requested):
Looks good, would you please add a test?

stan-dot:
> Looks good, would you please add a test and fix the inheritance order?


ah yes I forgot it's not a set but an ordered list. how should the inheritance order look like?

stan-dot:
also I realized if sim motor actions are instant I might not get the assert on the difference between the readback and set point. there should be a way to do this, investigating...

dperl-dls:
Not sure, but some way that https://docs.python.org/3/howto/mro.html can get a consistent order out of `HasName, Movable, Stoppable, Flyable, Preparable, Locatable`.

I think the issue is likely that several of these inherit each other already, so I would try:

`StandardReadable, Flyable, Locatable, Preparable, Stoppable` since this covers all the methods.

You can stop the sim motor from moving with
```python
with ophyd_async.core.mock_puts_blocked(motor.signal):
    < do the set, assert, etc. >
```

stan-dot:
thanks, trying this now

stan-dot:
I get this
```python
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/xma12127/projects/forks/ophyd-async/tests/epics/test_motor.py", line 324, in test_locatable
    with mock_puts_blocked(sim_motor.user_setpoint):
TypeError: '_AsyncGeneratorContextManager' object does not support the context manager protocol

```

but it *is* a SignalRW, just like the only other example of the use of this method that I could find - in `tests/core/test_mock_signal_backend.py`

tests/epics/test_motor.py
```diff
@@ -317,3 +318,12 @@ async def test_complete(sim_motor: motor.Motor) -> None:
...
     assert sim_motor._fly_status.done
+
+
+async def test_locatable(sim_motor: motor.Motor) -> None:
+    await sim_motor.set(10)
+    assert (await sim_motor.locate())["readback"] == 10
+    async with mock_puts_blocked(sim_motor.user_setpoint):
+        # sim_motor.set(10)
+        assert (await sim_motor.locate())["readback"] == 0
+    assert (await sim_motor.locate())["readback"] == 10
```
dperl-dls, line 329:
```suggestion
async def test_locatable(sim_motor: motor.Motor) -> None:
    callback_on_mock_put(
        sim_motor.user_setpoint,
        lambda x, *_, **__: set_mock_value(sim_motor.user_readback, x),
    )
    assert (await sim_motor.locate())["readback"] == 0
    async with mock_puts_blocked(sim_motor.user_setpoint):
        move_status = sim_motor.set(10)
        assert (await sim_motor.locate())["readback"] == 0
    await move_status
    assert (await sim_motor.locate())["readback"] == 10
```

dperl-dls (commented):
That sim motor is not the one from the demo

tests/epics/test_motor.py
stan-dot: thanks, it worked!

dperl-dls (approved):
lgtm, thanks

PR #543, opened 2024-08-28
PerSignalConfig dataclass for StandardReadable devices to be used with a common implementation of prepare.

burkeds:
This is one way prepare can be used in a standard fashion across StandardReadables while preserving the benefits of type-hinting.

PerSignalConfig is a dict-like dataclass which will take device signals as a key and a value. This dataclass is then passed to prepare and the signals will be set accordingly. Type checkers should recognize if a value not matching the signal type is assigned.

Example usage:

```python
class MyMotor(StandardReadable):
    velocity: SignalRW[float]
    accl: SignalRW[float]

        super().__init__(name=name)
        with self.add_children_as_readables(ConfigSignal):
            self.velocity = SignalRW(name="velocity",
                                     backend=SoftSignalBackend(datatype=float,
                                                               initial_value=0.0))
            self.accl = SignalRW(name="accl",
                                 backend=SoftSignalBackend(datatype=float,
                                                           initial_value=0.0))


async def main():
    async with DeviceCollector(mock=True):
        device = MyMotor()
    config = PerSignalConfig()
    config[device.velocity] = 1.0 # Pyright throws a type-error if this is a str or something
    config[device.accl] = 2.0
    await device.prepare(config)
    print(await device.velocity.read())
    print(await device.accl.read())
    print("Finished")

if __name__ == "__main__":
    asyncio.run(main())


# OUTPUT
{'device-velocity': {'value': 1.0, 'timestamp': 6675122.751791038, 'alarm_severity': 0}}
{'device-accl': {'value': 2.0, 'timestamp': 6675122.752000068, 'alarm_severity': 0}}
Finished

```


src/ophyd_async/core/_readable_config.py
```diff
@@ -0,0 +1,97 @@
...
+    tuple_signals: Dict[str, Tuple[tuple, Any]] = field(default_factory=dict)
+    dict_signals: Dict[str, Tuple[dict, Any]] = field(default_factory=dict)
+    set_signals: Dict[str, Tuple[set, Any]] = field(default_factory=dict)
+    frozenset_signals: Dict[str, Tuple[frozenset, Any]] = field(default_factory=dict)
+    bytes_signals: Dict[str, Tuple[bytes, Any]] = field(default_factory=dict)
+    bytearray_signals: Dict[str, Tuple[bytearray, Any]] = field(default_factory=dict)
+    complex_signals: Dict[str, Tuple[complex, Any]] = field(default_factory=dict)
+    none_signals: Dict[str, Tuple[type(None), Any]] = field(default_factory=dict)
+    ndarray_signals: Dict[str, Tuple[np.ndarray, Any]] = field(default_factory=dict)
+    signals: Dict[Type[Any], Dict[str, Tuple[Type[Any], Any]]] = field(init=False)
```
callumforrester, line 23: I think we can simplify this while preserving the benefits of type checking. Some ideas:

##### 1: Just use a dictionary

```python
ReadableDeviceConfig = Mapping[Signal, Any]

class MyMotor(StandardReadable):
        self.velocity = SignalRW(name="velocity", backend=SoftSignalBackend(datatype=float, initial_value=1.0))
        self.add_readables([velocity], ConfigSignal)

with DeviceCollector():
    device = MyMotor()

config = {device.velocity: 3.0}
await device.prepare(config) 
```

`prepare()` can still perform a type check, the fact that velocity is a float has just become implicit. The main difference vs. this PR is that nothing checks the type of velocity when we create the config object. However in the PR version we also have to assert that velocity is a `float` via `config.add_attribute("velocity", float)`. There is nothing guaranteeing that the velocity signal is actually a `float` at this point, so I'm not sure how much difference it makes.

##### 2: Use a mock device

```python
class MyMotor(StandardReadable):
        self.velocity = SignalRW(name="velocity", backend=SoftSignalBackend(datatype=float, initial_value=1.0))
        self.add_readables([velocity], ConfigSignal)

with DeviceCollector():
    device = MyMotor()

with DeviceCollector(mock=True):
    desired_device = MyMotor()

await desired_device.velocity.set(3.0)
await device.prepare(desired_device)
```

This is clunky but some of the clunkiness could be hidden with helpers. The main advantage is that you are setting signals with the same dataypes as the _actual_ signals on the real device, so types, structure etc. should _always_ align with no extra effort i.e. no need for `config.add_attribute("velocity", float)`.

##### 3: Use a function to bind the type of a signal to its configuration

```python
T = TypeVar("T")

@dataclass
class SignalConfiguration:
    signal: Signal[T]
    value: T

ReadableDeviceConfig = list[SignalConfiguration]

class MyMotor(StandardReadable):
        self.velocity = SignalRW(name="velocity", backend=SoftSignalBackend(datatype=float, initial_value=1.0))
        self.add_readables([velocity], ConfigSignal)

with DeviceCollector():
    device = MyMotor()

config = [SignalConfiguration(device.velocity, 3.0)]

await device.prepare(config)
```

Because T binds to both the signal and the value, pyright should detect if we haven't provided a value of `velocity`'s type at "compile time", removing the need for extra code for runtime type checking. Again, the objective here is to have nice hinting while I'm typing code into an IDE.

These are just some suggestions to get the discussion going, I'm sure there are more ways of doing this...

callumforrester (commented):
This looks interesting and mirrors the behaviour of `StandardDetector` quite well. I've left some comments to stimulate discussion in the call later.

src/ophyd_async/core/_readable.py
```diff
@@ -210,6 +212,36 @@ def add_readables(
...
+    @AsyncStatus.wrap
+    async def prepare(self, value: ReadableDeviceConfig) -> None:
+        tasks = []
+        for dtype, signals in value.signals.items():
+            for signal_name, (expected_dtype, val) in signals.items():
+                if hasattr(self, signal_name):
+                    attr = getattr(self, signal_name)
+                    if isinstance(attr, (HintedSignal, ConfigSignal)):
+                        attr = attr.signal
+                    if attr._backend.datatype == expected_dtype:  # noqa: SLF001
```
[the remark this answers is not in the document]
coretl: Will be added in #310

src/ophyd_async/core/_readable_config.py
coretl: Or a variation of 3:
```python
T = TypeVar("T")

class StandardReadableConfig:
    def __init__(self):
        self.signal_configuration: Dict[SignalW, Any] = {}
    def __setitem__(self, signal: SignalW[T], value: T):
        self.signal_configuration[signal] = value

class MyMotor(StandardReadable):
        with self.add_children_as_readables(ConfigSignal):
            self.velocity = SignalRW(...)
            self.accl = SignalRW(...)

with DeviceCollector():
    device = MyMotor()

config = StandardReadableConfig()
config[device.velocity] = 3.0
config[device.accl] = 2.5

await device.prepare(config)
```

coretl: Just checked that overriding `__setitem__` plays nice with pyright, it does: https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIANFAOLErEhIDGAUG6JFACbkzkWAG3IBnUcVGZsufHwHCxojgSgBeQiQogAFACICegJQcAAvMEjxbReMIALVGh31GzFgG0CAXSMAuNigg2D9CDlspAAkwIR4mAODeYmAoAH1UiRgkGGIIdJ0JIWAaAGtiOFCCR3QvbxoAN3IhAFdiSv8oADpujnt1KGjYph0Tey9q5wBGI29%2BgCYggGIoBCVJNjGqpx0AZhn%2BvRA9JahgciQhZSA

src/ophyd_async/core/_readable.py
```diff
@@ -210,6 +212,36 @@ def add_readables(
...
+                        if val is not None:
+                            tasks.append(attr.set(val))
+                    else:
+                        raise TypeError(
+                            f"Expected value of type {expected_dtype} for attribute"
+                            f" '{signal_name}',"
+                            f" got {type(attr._backend.datatype)}"  # noqa: SLF001
+                        )
+        await asyncio.gather(*tasks)
```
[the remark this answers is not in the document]
burkeds: I was motivated to do this because there is no obvious way to configure a device. In StandardReadable there is no method that can use the configure protocol. It looks like `prepare` was intended to be a way to configure a device but there is no obvious way to use it. `prepare` only takes a single argument, intended to be a dataclass, so this is one way a dataclass might be used to write to any of the configuration signals using prepare.

To me, it does not at all seem intuitive to "move" a config signal to some value to set it. As a user or beamline scientist, I suspect this might be perceived as "hacky".

One nice application of these dataclasses is that it provides a data structure for configurations one may use to swap configurations easily within a plan or save/load configurations to and from a file or database.

src/ophyd_async/core/_readable_config.py
burkeds: @coretl I have tried to make an example of this but it does not seem to be working and I am not sure why.

```python
    StandardReadable,
    ConfigSignal,
    SignalRW,
    SignalW,
    DeviceCollector,
    SoftSignalBackend,
)
import asyncio
from dataclasses import dataclass, field
from typing import Dict, TypeVar, Generic

T = TypeVar("T")


@dataclass
class StandardReadableConfig(Generic[T]):
    signal_configuration: Dict[SignalW[T], T] = field(default_factory=dict)

    def __setitem__(self, signal: SignalW[T], value: T):
        self.signal_configuration[signal] = value


class MyMotor(StandardReadable):
    velocity: SignalRW[float]
    accl: SignalRW[float]

        super().__init__(name=name)
        with self.add_children_as_readables(ConfigSignal):
            self.velocity = SignalRW(name="velocity",
                                     backend=SoftSignalBackend(datatype=float,
                                                               initial_value=0.0))
            self.accl = SignalRW(name="accl",
                                 backend=SoftSignalBackend(datatype=float,
                                                           initial_value=0.0))


async def main():
    async with DeviceCollector(mock=True):
        device = MyMotor()
    config = StandardReadableConfig[float]()
    config[device.velocity] = 1  # This should throw a type error
    config[device.accl] = 2.0
    print("Finished")

if __name__ == "__main__":
    asyncio.run(main())
```

If I send this script to PyRight it does not return any errors despite me setting the velocity to an integer in config.

callumforrester: @burkeds There's a couple of things going on here:

First, I don't think there's any way to make it reject an `int` when the type is `float` because it is implicitly castable. I've seen that behaviour before and it's annoying but understandable since lots of people type `1` when they really mean `1.0`. The reverse should be possible i.e. getting it to reject a `float` for a `SignalRW[int]`, and obviously things like `str` etc. are fine, that one particular combination is always annoying though.

Second, by binding `T` to the `StandardReadableConfig` class, you've bound every call to `__setitem__` to the _same_ value of `T`, so a particular instance of `StandardReadableConfig` binds to `int`, `str` etc. and cannot accept anything else. You need to bind `T` to the `__setitem__` method rather than the whole class. You can then make the internal dictionary of type `Dict[SignalW[Any], Any]` since you're performing validation at input time rather than while it's stored. It may then be a good idea to make the dictionary private and/or read-only via a property to make sure no one messes with it outside of the validated path you've defined.

I've made a mocked example here that looks a bit like your code, and hopefully shows off the different cases I've explained: https://pyright-play.net/?code=FASwtgDg9gTgLgAgIYGcCeA7AxiKwBmMUYCAJknElgDaooCmKC408ZFVtKKANAviHrVSBIiThoIIDAHNmkWIgAiILHD4AVSfQBqSGHwDi9DPRiq%2BAQQxpgwDQgC8CLRF36AFACINXgJR2NHQIAMogMhhI1B7GpuZYANoaALp%2BAFzACFkIAHR5dgAC5JRB3MClTCGUGOQwpABK9EjkAEbU9ADCUBgCMhnZCCjhkdQA%2BljdvQCuMBS4GGkIKmoJYRFRCdZoyVY2yU78gsIepPT4SFPUcKPnarBojqSqcAGZ2af4CKOjDHAgcPQwN8PAxqPg%2BEN1tRFmsRkkdggAG5RKb0RYadJvAZZUH4HKQkbjSbhGZzboJAlRfbOZHUVF2cpcJgAWTQzKgcFg-WyiKEUBwEhhww2%2BGoUAoySxyCwNCFUISovFcElAxgjDlcMVEqlYCgpw1G2kyrsAw%2BX1G0n%2BwNxfEiYDRgzgMAOXn83OxgymbhgHj8OW%2Bluuo19UoGuJyvLFArQB1hURDHpxQjxVBoseF0VewFQmCwZDOCDASGkvvd%2BcRqnoB1Z7M5PoCAwmPXCseqtQaTVa7S6zZkCeyTd6CVOFaw9AjfOj1IQXhgXiyAGIXAALEBMODLogAdyYSAQEjcCDMRGdLXoWAuDCRk-%2BMZoTRg1BjKGXUEupAwAHJEGfkI7zLIUqDuEw70KO466qc04AEw5AAzAgS4aKu66blAO5-geVbHrACBnheUxXpBVbLqg%2B5QHhVZIBgzAYOof5aogpBQIwX6IEWADWVYMBgDBAcSMigeBORqigMGISua7IBgGBQGg0gyE%2BCBbt0377mhW6YdoR4wCelEEVe0HMEwLDUKo-xKReKCUG0VacgxYoUPxvZCZWOSptQME5AADBJyFSRAdCMPpl5Vv8n67moUxREptIgCIAwQABcDeAAYpaL70KQ-gMiAnzfHa9DfE4zheN8RbSN8Xhljm2C4CJUwYB4FVNX4ARAA

burkeds: Thanks @callumforrester, I had previous versions that did not bind T to the class itself that didn't work either, it was the intrinsic casting of 1 that was getting me. Once I tried setting it to a string, it worked. I'll push an updated version of the code shortly.

src/ophyd_async/core/_readable_config.py
```diff
@@ -0,0 +1,97 @@
+from dataclasses import asdict, dataclass, field
+from typing import Any, Dict, Tuple, Type
+
+import numpy as np
+
+
+@dataclass
+class ReadableDeviceConfig:
```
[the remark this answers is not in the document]
burkeds: That seems fine, I renamed it.

src/ophyd_async/core/_readable.py
```diff
@@ -15,9 +30,35 @@
...
+
+    def __delitem__(self, signal: SignalW[T]):
+        del self._signal_configuration[signal]
+
+    def __iter__(self) -> Iterator[SignalW[Any]]:
+        return iter(self._signal_configuration)
+
+    def __len__(self) -> int:
+        return len(self._signal_configuration)
+
```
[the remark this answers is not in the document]
burkeds: Yeah that's much cleaner.

src/ophyd_async/core/_readable.py
```diff
@@ -15,9 +28,20 @@
...
+T = TypeVar("T")
+
+
+@dataclass
+class PerSignalConfig(Dict[SignalW[T], T]):
+    _signal_configuration: Dict[SignalW[Any], Any] = field(default_factory=dict)
+
+    @property
+    def signal_configuration(self) -> Dict[SignalW[Any], Any]:
+        return self._signal_configuration
```
coretl, line 40: Just found out that type checking is currently disabled, I'm in the process of restoring it, but for now I will be the type checker...

The `signal_configuration` and dataclassness of this class aren't actually being used, you're actually storing everything in the dict that is `self`. Using a plain `dict` would be the same as what you have here. However, we want to support type checkers. If you replace with:
```python
class PerSignalConfig(Dict[SignalW, Any]):
    def __setitem__(self, signal: SignalW[T], value: T):
        super().__setitem__(signal, value)
    def __getitem__(self, signal: SignalW[T]) -> T:
        return super().__getitem__[signal]
```
Then we are effectively narrowing the type information just for the type checker. If we wanted to be even cleverer we could be clear that this is just for type checkers:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # Narrow getitem and setitem so only Signals are valid
    class PerSignalConfig(Dict[SignalW, Any]):
        def __setitem__(self, signal: SignalW[T], value: T):
            super().__setitem__(signal, value)
        def __getitem__(self, signal: SignalW[T]) -> T:
            return super().__getitem__[signal]
else:
    # At runtime this is effectively the same as a dict, so might as well use one
    PerSignalConfig = dict
```

However I don't think it's worth the extra fluff for a few microseconds runtime performance.

burkeds: Ok I made the change without the additional fluff.

burkeds:
@coretl Looks like there is a new linting error regarding an incompatible override of the epics motor with prepare. How do you think we should handle this?

coretl:
> @coretl Looks like there is a new linting error regarding an incompatible override of the epics motor with prepare. How do you think we should handle this?

Does making `FlyMotorConfig` inherit from `PerSignalConfig` fix the issue?

burkeds:
> > @coretl Looks like there is a new linting error regarding an incompatible override of the epics motor with prepare. How do you think we should handle this?
> 
> Does making `FlyMotorConfig` inherit from `PerSignalConfig` fix the issue?

It does not because the attributes of FlyMotorConfig are not SignalW. For this approach to work, any variables required by prepare might need to be soft signals. What do you think?

coretl:
After the collab call, are we still convinced we need this, or can it be replaced by load/save?

burkeds:
> After the collab call, are we still convinced we need this, or can it be replaced by load/save?

I think we can shelve this now. Config via load/save and abs_set seems fine.

PR #547, opened 2024-08-30
Make FilenameProvider optional on PathProvider impl

### LINKED ISSUE #546: PathProvider must take FilenameProvider and FilenameProvider protocol poor fit for DLS usecase
After #214 we now have a split concept of FilenameProvider and PathProvider. The PathProvider takes the device_name as an optional argument, but the FilenameProvider cannot. 

At Diamond we expect the path+filename to be of the form `/dls/ixx/data/YYYY/cm12345-1/56789/ixx-56789-det.h5`, meaning the device_name is required in both parts.

DiamondJoseph:
Closes #546 

Required for Diamond expectation of file names.

### DIFF
--- a/src/ophyd_async/core/_providers.py
+++ b/src/ophyd_async/core/_providers.py
@@ -26,7 +26,7 @@ class PathInfo:
 
 class FilenameProvider(Protocol):
     @abstractmethod
-    def __call__(self) -> str:
+    def __call__(self, device_name: Optional[str] = None) -> str:
         """Get a filename to use for output data, w/o extension"""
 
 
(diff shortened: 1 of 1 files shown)


coretl (commented):
I think there are 2 use cases we need to support:
1. Where Directory and Filename can be mixed together in different combinations, e.g. static directory + datestamp filename or static directory + uuid filename
2. Where Directory and Filename are so site specific they will never be mixed together with other providers

To support 2 I think we need the option for a `PathProvider` that has no `FilenameProvider`, so it can make the whole path itself. For 1 then I think we need specific `PathProviders` (like `StaticPathProvider`) to support taking a `FilenameProvider` to delegate to

I think we can support all this by removing `_filename_provider` from the `PathProvider` protocol

DiamondJoseph:
After discussion IRL, we've opted to do both: remove the requirement for the PathProvider to have a FilenameProvider, but also adjust the signature of the FilenameProvider to match the PathProvider. This can propagate from the PathProvider call down.

coretl (approved):

PR #555, opened 2024-09-03
Make np_datatype property of ShapeProvider rather than final axis

### LINKED ISSUE #554: Split ShapeProvider.dtype from ShapeProvider.__call__ method.
This allows ShapeProvider.__call__ to have a type hint [tuple[int, ...]] and for this to be handled consistently in all cases, where d_type may not be available in the same way as AreaDetectors.

DiamondJoseph:
This caught me out while updating dodal to opyhd-async: while the numpy datatype may be useful for deciding how much space to allocate for a given stack of frames, it is decidedly not part of the shape of the frame.

For a currently not ADBaseIO detector (the TetrAMM monitor) this tripped me up as the datatype was not being passed as part of the shape and hence it was looking for a datatype with key `1000`.

Closes #554


jwlodek (approved):

jwlodek:
Seems like a cleaner approach, I like it.

src/ophyd_async/core/_providers.py
```diff
@@ -213,5 +213,9 @@ def __call__(self) -> str:
 
 class ShapeProvider(Protocol):
     @abstractmethod
-    async def __call__(self) -> tuple:
+    async def np_datatype(self) -> str:
+        """Represents the numpy datatype"""
+
+    @abstractmethod
+    async def __call__(self) -> tuple[int, ...]:
```
coretl, line 220: I'd be tempted to make this more descriptive now we have 2 things we want out of ShapeProvider
```suggestion
    async def shape(self) -> tuple[int, ...]:
```

DiamondJoseph: At which point, really is a ShapeProvider not a DataDescriber or similar?

DiamondJoseph: I made a prospective PR into this PR in case we want to make that change, but I want to get something merged today so's to enable dodal update ASAP. https://github.com/bluesky/ophyd-async/pull/556

PR #556, opened 2024-09-04
Rename ShapeProvider

DiamondJoseph:
This PR is a for-comment suggested addition to #554, and would be merged into that branch prior to merging into main if it is decided upon.

With the additional method being added to the ShapeProvider, it now does more than provide a shape.


coretl (approved):

coretl:
Naming things is hard...

Issue #539, opened 2024-08-22
Make `put_mock` an `AsyncMock`


DominicOram:
As a developer I was expecting to be able to do the following:

```python

async def my_async_function():
    await ...

callback_on_mock_put(my_mocked_signal, my_mocked_function)
```

However, because `put_mock` is a `Mock` not an `AsyncMock` I cannot. Instead I have had to go around the mock and do:
```python
@AsyncStatus.wrap
async def my_async_function(value: float, wait=None, timeout=None):
    await ...

my_mocked_signal.set = my_async_function
```

noemifrisina:
Adding an example from the PMAC device which has this issue:

```python
@AsyncStatus.wrap
async def set(self, value: int, wait=True, timeout=None):
    prog_str = f"&2b{value}r"
    assert isinstance(timeout, SupportsFloat) or (
        timeout is None
    ), f"ProgramRunner does not support calculating timeout itself, {timeout}"
    await self.signal.set(prog_str, wait=wait)
    # First wait for signal to go to 1, then wait for the scan to finish.
    await wait_for_value(
        self.status,
        ScanState.RUNNING,
        timeout=DEFAULT_TIMEOUT,
    )
    await wait_for_value(self.status, ScanState.DONE, timeout)
```
In the test, `callback_on_mock_put` won't take an async function so the workaround is:
```python
async def go_high_then_low():
    set_mock_value(fake_pmac.scanstatus, 1)
    await asyncio.sleep(0.01)
    set_mock_value(fake_pmac.scanstatus, 0)

prog_num = 10
callback_on_mock_put(
    fake_pmac.pmac_string,
    lambda *args, **kwargs: asyncio.create_task(go_high_then_low()),  # type: ignore
)
```

coretl:
Would someone from MX like to work on this? Or should we put it on our backlog?

DominicOram:
If you're happy for us to do it we can, I think it's like a one line change.

coretl:
Yes please

Issue #542, opened 2024-08-27
Detector controller disarmed once in `stage`, but armed in both `trigger` and `prepare`


evalott100:
Currently, detector controller is only disarmed once:

https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/core/_detector.py#L202-L207

However it is armed in both trigger and prepare:

https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/core/_detector.py#L253-L257

https://github.com/bluesky/ophyd-async/blob/d7f07488fdfa976037791cbdfdb21e0f7c5f7430/src/ophyd_async/core/_detector.py#L299-L303

This is the cause of all the test failure's [here](https://github.com/bluesky/ophyd-async/pull/538) once we enforce that background tasks are handled before test end.

Thoughts? @coretl @abbiemery @ZohebShaikh @callumforrester

callumforrester:
I may be out of date but I think this is intentional: You should be calling `trigger` in a software scan and `prepare` in a hardware scan, if a plan is calling both then the plan is wrong.

On top of that though I thought `arm()` was idempotent.

coretl:
> I may be out of date but I think this is intentional: You should be calling `trigger` in a software scan and `prepare` in a hardware scan, if a plan is calling both then the plan is wrong.

That used to be correct, but then we though that `prepare` would be useful for setting exposure time for software scans

> On top of that though I thought `arm()` was idempotent.

Nope, never has been. The state management was done in `StandardDetector` so it didn't have to be managed in every `DetectorControl` class.

> Thoughts? @coretl 

Don't `caput DET:DRV:Acquire 1` until `trigger` or `kickoff`

## Suggestions to fix

We fix this and https://github.com/bluesky/ophyd-async/issues/405 together.

The addition changes we need here:
- Separate out `DetectorControl.arm(trigger_info)` into `DetectorControl.prepare(trigger_info)` and `DetectorControl.arm()`
- Call `DetectorControl.prepare` in `StandardDetector.prepare` (which is also called in `StandardDetector.trigger` if it hasn't already been called
- Check that trigger_mode is correct then call `DetectorControl.arm` in `StandardDetector.trigger` and `StandardDetector.kickoff`

jwlodek:
[...]

Issue #519, opened 2024-08-13
Correctly handle pending tasks in our tests


olliesilvester:
Running unit tests sometimes gives us a few `Task was destroyed but it is pending!` messages. This is because, in our pytests, we either:

1. Use the RE fixture to create an event loop, do the test, then the fixture closes the event loop at the end of the pytest using pytest's `add_finalizer`. This can run before awaiting for pending tasks, so pending tasks can be cancelled.

2. Use an Asyncio event loop outside of the RE and the same as above

In both cases, if we create a task using `asyncio.create_task `and then finish the test before any `await`s are ran, there is a chance that the event loop gets closed with pending tasks, which causes that warning message
To make sure these pending tasks get completed, we can either:

1. Make our event loop fixtures do an await at the end of a pytest, eg
```
def RE(request)
...
if pending:
            loop.run_until_complete(asyncio.gather(*pending, return_exceptions=True))
```

2. Make our event loop fixtures raise an error if there are pending tasks at the end, and tell us which tasks they are

3. Something else clever which I didn't think of

evalott100:
Closing https://github.com/bluesky/ophyd-async/issues/260 in favour of this

jwlodek:
I think #498 is also a duplicate of this. Suggest we close that since this goes into quite a bit more detail.

evalott100:
> That gets my vote. Do you know how we would do that?

We could do this for tests using the `RE` fixture by getting pending.

We could also do something like:
``` python
@pytest.fixture(autouse=True)
async def assert_no_pending_tasks(request):
    # There should be no tasks pending after a test has finished
    try:
        yield
    finally:
        loop = asyncio.get_event_loop()
        pending = asyncio.all_tasks(loop)
        if pending # and some_filtering_perhaps:
            unfinished_tasks = ...
            raise RuntimeError(
                "Not all tasks {unfinished_tasks} closed during test."
            )

        
```

olliesilvester:
> We could do this for tests using the `RE` fixture by getting pending from the run engine loop.
> 
Yep, this is what I initially tried in that other PR. It raises errors but it's not necessarily obvious as to which test an unfinished task came from

evalott100:
> Yep, this is what I initially tried in that other PR. It raises errors but it's not necessarily obvious as to which test an unfinished task came from

I think we can use `traceback` in the standard library and raise a new `UnresolvedTaskError` encapsulating that info.

PR #559, opened 2024-09-05
539 put mock should be async

DominicOram:
Fixes https://github.com/bluesky/ophyd-async/issues/539

To test:
* Confirm new test passes

Note:
It's not 100% clear to me why the tests fail without the second commit. Clearly something to do with the patch leaking between tests but I'm not sure why it would be leaking or why this would only appear given the async mock change. However, the tests seem to pass fine without the patches and in general patching less seems good so I'm inclined to just move on.

### DIFF
--- a/src/ophyd_async/core/_mock_signal_backend.py
+++ b/src/ophyd_async/core/_mock_signal_backend.py
@@ -1,7 +1,7 @@
 import asyncio
 from functools import cached_property
 from typing import Callable, Optional, Type
-from unittest.mock import Mock
+from unittest.mock import AsyncMock
 
 from bluesky.protocols import Descriptor, Reading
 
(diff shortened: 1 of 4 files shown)


evalott100:
> It's not 100% clear to me why the tests fail without the second commit. Clearly something to do with the patch leaking between tests

There's some unsafe status' throughout the tests, I'm working on a PR to fix this and cause failure during tests where these status' are created, with exact info on where the status is, rather than failure in later tests when the weak ref status' happen to be garbage collected.

https://github.com/bluesky/ophyd-async/issues/519

I looked through the call stack on `ad.arm/ad.disarm` and I couldn't tell which status it might be. I think it's probably a consequence of some other status not being cancelled/awaited by the end of another test, the removal of the fixture didn't actually do anything (I reran tests for that commit and there are no errors now).

That being said, it doesn't seem like we need to mock `wait_for_value` here anyway so we should keep that commit.

evalott100 (approved):
Looks good!

DominicOram:
I can't reproduce the new error locally and not sure where it's from. Are you able to reproduce it Eva?

coretl:
I think we are still waiting for #542 before the fix to #519. I suggest we merge this as is, then fix any test failures that remain after those 2 issues are closed.

PR #563, opened 2024-09-09
Update signature of __init__ for simdetector to match other detector classes

jwlodek:
Other detector classes have kwargs for drv and hdf prefixes, and instantiate the drv and hdf device instances at init, rather than having them passed in. This PR makes the simdetector interface the same.

One thing I noticed, is that since we are not passing in the drv or hdf devices, the signals are not available at init to stick in config_sigs. As a result I needed to update them using the private property. Should we have some exposed public interface for adjusting config sigs for standard detectors to avoid needing to touch `_config_sigs`?

### DIFF
--- a/src/ophyd_async/epics/adsimdetector/_sim.py
+++ b/src/ophyd_async/epics/adsimdetector/_sim.py
@@ -12,14 +12,15 @@ class SimDetector(StandardDetector):
 
     def __init__(
         self,
-        drv: adcore.ADBaseIO,
-        hdf: adcore.NDFileHDFIO,
+        prefix: str,
         path_provider: PathProvider,
+        drv_suffix="cam1:",
+        hdf_suffix="HDF1:",
         name: str = "",
         config_sigs: Sequence[SignalR] = (),
     ):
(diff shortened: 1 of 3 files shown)


coretl:
> One thing I noticed, is that since we are not passing in the drv or hdf devices, the signals are not available at init to stick in config_sigs. As a result I needed to update them using the private property. Should we have some exposed public interface for adjusting config sigs for standard detectors to avoid needing to touch `_config_sigs`?

I suggest we stop passing in config sigs, and hardcode them as `self.drv.acquire_time` and `self.drv.acquire_period` in all the relevant `StandardDetector` subclasses. When we have a use-case for per-instance config_sigs then we can revisit this

jwlodek:
That works for me. Presumably any detector specific config sigs like readout mode can be handled by the subclasses as well.

jwlodek:
I suggest that if the changes in this PR are acceptable we merge as is, and I can submit a separate one that addresses config_sigs, since that will affect all the detectors rather than just SimDetector.

coretl (approved):

Issue #549, opened 2024-08-30
Support multi-run in `StandardDetector.prepare`


coretl:
Building on #542 we want to be able to support the following use case:
- Run a 100 point X-Y flyscan and each of 3 energies
- Only arm the detector once for the whole flyscan as it is slow to arm

To do this we want the plan to:
```
det.stage & flyer.stage
for i in range(3):
  if i == 0:
    det.prepare(3x 100)  # Arm on first iteration
  move_to_energy(i) & flyer.prepare(100)  # Move outer and inner motion at the same time as the above line
  det.kickoff() and flyer.kickoff()  # No-op and start motion
  collect_while_completing(det, flyer)  # Until 100 frames collected
```

This means we need a way to:
- Prepare for M times N frames
- On each kickoff and complete collect N frames
- Raise if we get more than M kickoffs

PR #568, opened 2024-09-11
Interface change of StandardDetector and Standard Controller

### LINKED ISSUE #405: Decide the interface between `StandardDetector` and `DetectorController` w.r.t. the `TriggerInfo` that is passed down in `prepare`

ZohebShaikh:
closes #405 #542 https://github.com/bluesky/ophyd-async/issues/549

### DIFF
--- a/src/ophyd_async/epics/adkinetix/_kinetix_controller.py
+++ b/src/ophyd_async/epics/adkinetix/_kinetix_controller.py
@@ -1,7 +1,7 @@
 import asyncio
-from typing import Optional
 
 from ophyd_async.core import AsyncStatus, DetectorControl, DetectorTrigger
+from ophyd_async.core._detector import TriggerInfo
 from ophyd_async.epics import adcore
 
 from ._kinetix_io import KinetixDriverIO, KinetixTriggerMode
@@ -24,22 +24,19 @@ class KinetixController(DetectorControl):
     def get_deadtime(self, exposure: float) -> float:
         return 0.001
 
-    async def arm(
-        self,
-        num: int,
-        trigger: DetectorTrigger = DetectorTrigger.internal,
-        exposure: Optional[float] = None,
-    ) -> AsyncStatus:
+    async def prepare(self, trigger_info: TriggerInfo):
         await asyncio.gather(
-            self._drv.trigger_mode.set(KINETIX_TRIGGER_MODE_MAP[trigger]),
-            self._drv.num_images.set(num),
+            self._drv.trigger_mode.set(KINETIX_TRIGGER_MODE_MAP[trigger_info.trigger]),
+            self._drv.num_images.set(trigger_info.number),
             self._drv.image_mode.set(adcore.ImageMode.multiple),
         )
-        if exposure is not None and trigger not in [
+        if trigger_info.livetime is not None and trigger_info.trigger not in [
             DetectorTrigger.variable_gate,
             DetectorTrigger.constant_gate,
         ]:
-            await self._drv.acquire_time.set(exposure)
+            await self._drv.acquire_time.set(trigger_info.livetime)
+
+    async def arm(self) -> AsyncStatus:
         return await adcore.start_acquiring_driver_and_ensure_status(self._drv)
 
     async def disarm(self):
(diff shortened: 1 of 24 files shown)


src/ophyd_async/epics/adpilatus/_pilatus_controller.py
```diff
@@ -59,8 +63,6 @@ async def arm(
             timeout=DEFAULT_TIMEOUT,
```
coretl, line 63: This needs moving to `arm()`

coretl (changes requested):

src/ophyd_async/epics/adkinetix/_kinetix_controller.py
```diff
@@ -20,27 +21,31 @@ def __init__(
...
             DetectorTrigger.constant_gate,
         ]:
-            await self._drv.acquire_time.set(exposure)
-        return await adcore.start_acquiring_driver_and_ensure_status(self._drv)
+            await self._drv.acquire_time.set(trigger_info.livetime)
+
+    async def arm(self):
+        self._arm_status = await adcore.start_acquiring_driver_and_ensure_status(
+            self._drv
+        )
```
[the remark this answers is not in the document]
ZohebShaikh: https://github.com/bluesky/ophyd-async/pull/568/commits/338f946769e65408bc2dc7a6430fee7548a43df0 did you mean something like this ?

src/ophyd_async/epics/adpilatus/_pilatus_controller.py
```diff
@@ -47,21 +47,22 @@ async def prepare(self, trigger_info: TriggerInfo):
...
+        self._arm_status = await asyncio.gather(
+            adcore.start_acquiring_driver_and_ensure_status(self._drv),
+            # The pilatus has an additional PV that goes True when the camserver
+            # is actually ready. Should wait for that too or we risk dropping
+            # a frame
+            wait_for_value(
+                self._drv.armed,
+                True,
+                timeout=DEFAULT_TIMEOUT,
+            ),
```
coretl: This was fine as it was...

src/ophyd_async/fastcs/panda/_control.py
```diff
@@ -24,8 +26,9 @@ async def prepare(self, trigger_info: TriggerInfo):
         ), "Only constant_gate and variable_gate triggering is supported on the PandA"
 
     async def arm(self):
-        await asyncio.gather(self.pcap.arm.set(True))
-        await wait_for_value(self.pcap.active, True, timeout=1)
+        self._arm_status = await asyncio.gather(
+            self.pcap.arm.set(True), wait_for_value(self.pcap.active, True, timeout=1)
+        )
```
coretl, line 31: Same here

coretl (approved):

PR #579, opened 2024-09-16
Make table subclass enums be sequence enum rather than numpy string

evalott100:


src/ophyd_async/core/_table.py
```diff
@@ -1,42 +1,126 @@
...
-                field_name: np.concatenate(
-                    (getattr(self, field_name), getattr(right, field_name))
+                field_name: _concat(
+                    getattr(self, field_name), getattr(right, field_name)
                 )
                 for field_name in self.model_fields
             }
         )
 
+    @property
```
coretl, line 66: Any particular reason for the `@property`? I generally favour plain methods especially when we do computation behind the scenes...

src/ophyd_async/fastcs/panda/_table.py
```diff
@@ -51,13 +51,7 @@ class SeqTrigger(str, Enum):
...
     Field(default_factory=lambda: np.array([], dtype=np.bool_)),
 ]
-TriggerStr = Annotated[
-    np.ndarray[tuple[int], np.unicode_],
-    NpArrayPydanticAnnotation.factory(
-        data_type=np.unicode_, dimensions=1, strict_data_typing=False
-    ),
-    Field(default_factory=lambda: np.array([], dtype=np.dtype("<U32"))),
-]
+TriggerStr = Annotated[list[SeqTrigger], Field(default_factory=list)]
```
coretl, line 54: We've used `Sequence` rather than `list` in other places because it's [covariant rather than invariant](https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance)

```suggestion
TriggerStr = Annotated[Sequence[SeqTrigger], Field(default_factory=list)]
```

coretl (changes requested):

coretl (approved):

coretl:
I'll fix the merge conflicts then merge

PR #581, opened 2024-09-17
added number_of_frames instead of iterations

### LINKED ISSUE #575: More flexible multi-run specification
It occurs to me that tomography will not be served with #549, as it doesn't follow that pattern of NxM frames.

I suggest instead of `TriggerInfo(num=3, iterations=4)` we do `TriggerInfo(num=[3]*4)`, which leaves us able to do `TriggerInfo(num=[2, 3, 1000, 3])` for tomography (darks, flats, projections, flats)

ZohebShaikh:
closes #575 
This pull request introduces significant changes to the `TriggerInfo` class and related logic in the `ophyd_async` package to support multiple trigger counts. The main changes include renaming the `number` field to `number_of_triggers`, updating associated logic, and modifying tests to accommodate these changes.

### Core changes:
* **TriggerInfo Class:**
  * Renamed `number` to `number_of_triggers` and updated its type to support both single integers and lists of integers. (`src/ophyd_async/core/_detector.py`)
  * Added a `computed_field` and `cached_property` to calculate the total number of triggers. (`src/ophyd_async/core/_detector.py`)

* **Detector Logic:**
  * Updated `prepare`, `kickoff`, and `complete` methods to handle the new `number_of_triggers` logic. (`src/ophyd_async/core/_detector.py`) 
### Controller updates:
* **EPICS Controllers:**
  * Updated `prepare` methods to use `total_number_of_triggers` instead of `number`. (`src/ophyd_async/epics/adaravis/_aravis_controller.py`, `src/ophyd_async/epics/adkinetix/_kinetix_controller.py`, `src/ophyd_async/epics/adpilatus/_pilatus_controller.py`, `src/ophyd_async/epics/adsimdetector/_sim_controller.py`, `src/ophyd_async/epics/advimba/_vimba_controller.py`, `src/ophyd_async/epics/eiger/_eiger_controller.py`)

### Test updates:
* **Test Adjustments:**
  * Modified tests to accommodate the new `number_of_triggers` field. (`tests/core/test_flyer.py`, `tests/conftest.py`, `system_tests/epics/eiger/test_eiger_system.py`) 
### Other changes:
* **Imports and Typing:**
  * Added `cached_property` and updated imports for `NonNegativeInt` and `computed_field`. (`src/ophyd_async/core/_detector.py`)
  * Removed unused `iteration` parameter from `prepare_static_seq_table_flyer_and_detectors_with_same_trigger`. (`src/ophyd_async/plan_stubs/_fly.py`)


src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -38,13 +38,12 @@ def prepare_static_pcomp_flyer_and_detectors(
 def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
     flyer: StandardFlyer[SeqTableInfo],
     detectors: list[StandardDetector],
-    number_of_frames: int,
+    number_of_frames: int | list[int],
```
coretl, line 41: In this plan stub, we are making number_of_frames * repeats frames.
This means number_of_frames should be int

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -60,16 +59,24 @@ def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
...
     deadtime = max(det.controller.get_deadtime(exposure) for det in detectors)
+    if isinstance(number_of_frames, list):
+        assert (
+            [number_of_frames[0]] * len(number_of_frames) == number_of_frames
+        ), "In fly scan number of frames should be same for each iteration"
+        number_of_frames = [frames * repeats for frames in number_of_frames]
+        single_frame = number_of_frames[0]
+    else:
+        number_of_frames = number_of_frames * repeats
+        single_frame = number_of_frames
```
coretl, line 70: Don't need this change

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -60,16 +59,24 @@ def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
...
 
     trigger_info = TriggerInfo(
-        number=number_of_frames * repeats,
+        number=number_of_frames,
```
coretl, line 73: This was correct as it was

src/ophyd_async/plan_stubs/_fly.py
```diff
@@ -60,16 +59,24 @@ def prepare_static_seq_table_flyer_and_detectors_with_same_trigger(
...
         trigger=DetectorTrigger.constant_gate,
         deadtime=deadtime,
         livetime=exposure,
         frame_timeout=frame_timeout,
-        iteration=iteration,
     )
-    trigger_time = number_of_frames * (exposure + deadtime)
+    trigger_time = single_frame * (exposure + deadtime)
```
coretl, line 79: This was right as it was

coretl (commented):
Please can we have a test that looks like the tomography example?

tests/fastcs/panda/test_hdf_panda.py
```diff
@@ -134,134 +134,8 @@ def flying_plan():
...
-        assert stream_datum["indices"] == {
-            "start": 0,
-            "stop": 1,
-        }
-        assert stream_datum["stream_resource"] in [
-            sd["uid"].split("/")[0] for sd in docs["stream_datum"]
-        ]
-
-
-async def test_hdf_panda_hardware_triggered_flyable_with_iterations(
```
ZohebShaikh, line 192: I have deleted `test_hdf_panda_hardware_triggered_flyable_with_iterations`  and moved it to
 ```
@pytest.mark.parametrize("number_of_frames", [[1, 1, 1, 1], [2, 3, 100, 3]])
async def test_hardware_triggered_flyable(
```

As we are removing iterations

File: src/ophyd_async/core/_detector.py [hunk 6]
```diff
@@ -313,12 +331,17 @@ async def prepare(self, value: TriggerInfo) -> None:
...
-                f"{self._trigger_info.iteration} iteration(s)!"
+                f"{self._total_frames} iteration(s)!"
             )
-        self._iterations_completed += 1
+        if isinstance(self._trigger_info.number, list):
+            self._frames_collected += self._trigger_info.number[
+                self._current_number_index
+            ]
+        else:
+            self._frames_collected += self._trigger_info.number
```
evalott100, line 344: Minor preference for the same comprehension as in `complete` 
```suggestion
        self._frames_collected += (
            self._trigger_info.number[self._current_number_index]
            if isinstance(self._trigger_info.number, list)
            else self._trigger_info.number
        )
```

src/ophyd_async/core/_detector.py
```diff
@@ -193,9 +197,12 @@ def __init__(
         self._watchers: list[Callable] = []
         self._fly_status: WatchableAsyncStatus | None = None
         self._fly_start: float
-        self._iterations_completed: int = 0
+        self._frames_collected: int = 0
+        self._current_number_index: int = 0
```
evalott100, line 201: We should call this `_current_[whatever_the_variable_in_trigger_is_called]_index`

```suggestion
        self._current_number_of_triggers_index: int = 0
```

src/ophyd_async/core/_detector.py
```diff
@@ -45,8 +45,16 @@ class DetectorTrigger(str, Enum):
...
+    #: Number of triggers that will be sent, (0 means infinite) Can be:
+    #  - A single integer or
+    #  - A list of integers for multiple triggers
+    # Example for tomography: TriggerInfo(number=[2,3,100,3])
+    #:     This would trigger:
+    #:     - 2 times for dark field images
+    #:     - 3 times for initial flat field images
+    #:     - 100 times for projections
+    #:     - 3 times for final flat field images
+    number: int | list[int]
```
evalott100, line 57: I think `_total_frames` should live here. If the only reason we're putting it in the `DetectorControl` is to calculate it if `number` is a list then we can do:

``` python
    @computed_field # from pydantic
    @property
    def total_frames(self) -> int:
        return (
            sum(self.number)
            if isinstance(self.number, list) else self.number
        )

File: src/ophyd_async/core/_detector.py [hunk 9]
```diff
@@ -331,19 +354,29 @@ async def complete(self):
...
-            self._iterations_completed = 0
+            if isinstance(self._trigger_info.number, list):
+                if index >= self._trigger_info.number[self._current_number_index]:
+                    self._current_number_index += 1
+                    break
+            else:
+                if index >= self._trigger_info.number:
+                    break
+        if self._frames_collected == self._total_frames:
+            self._frames_collected = 0
```
evalott100, line 379: I think if we're breaking from an async generator then it doesn't necessary close? We probably need something like

``` python
indices_written = self.writer.observe_indices_written(
    self._trigger_info.frame_timeout or (
        DEFAULT_TIMEOUT
        + (self._trigger_info.livetime or 0)
        + (self._trigger_info.deadtime or 0)
    )
)
try:
    async for index in indices_written:
        # .... DO YOUR STUFF AND BREAK
finally:
    await indices_written.aclose()
```

src/ophyd_async/core/_detector.py
evalott100: With how many times we're doing:

``` python
       frames: int = (
            sum(trigger_info.number)
            if isinstance(trigger_info.number, list)
            else trigger_info.number
        )
```

in tests, this is seeming even more useful.

tests/core/test_flyer.py
```diff
@@ -138,22 +139,26 @@ async def dummy_arm_2(self=None, trigger=None, num=0, exposure=None):
     return (detector_1, detector_2)
 
 
+@pytest.mark.parametrize("number_of_frames", [[1, 1, 1, 1], [2, 3, 100, 3]])
 async def test_hardware_triggered_flyable(
-    RE: RunEngine, detectors: tuple[StandardDetector]
+    RE: RunEngine, detectors: tuple[StandardDetector], number_of_frames
```
evalott100, line 144:
```suggestion
    RE: RunEngine, detectors: tuple[StandardDetector], number_of_frames: list[int]
```

tests/core/test_flyer.py
```diff
@@ -138,22 +139,26 @@ async def dummy_arm_2(self=None, trigger=None, num=0, exposure=None):
     return (detector_1, detector_2)
 
 
+@pytest.mark.parametrize("number_of_frames", [[1, 1, 1, 1], [2, 3, 100, 3]])
```
evalott100, line 142: I'd like another test of the `trigger_info` basemodel - that it fails if `number_of_frames` is incorrect.

evalott100 (changes requested):

tests/core/test_flyer.py
ZohebShaikh: There is a test like this in the same file 

```python
async def test_hardware_triggered_flyable_too_many_kickoffs(
    RE: RunEngine, detectors: tuple[StandardDetector]
):
    trigger_logic = DummyTriggerLogic()
    flyer = StandardFlyer(trigger_logic, [], name="flyer")
    trigger_info = TriggerInfo(
        number=1, trigger=DetectorTrigger.constant_gate, deadtime=2, livetime=2
    )

    def flying_plan():
        yield from bps.stage_all(*detectors, flyer)
        assert flyer._trigger_logic.state == TriggerState.stopping

        # move the flyer to the correct place, before fly scanning.
        # Prepare the flyer first to get the trigger info for the detectors
        yield from bps.prepare(flyer, 1, wait=True)

        # prepare detectors second.
        for detector in detectors:
            yield from bps.prepare(
                detector,
                trigger_info,
                wait=True,
            )
``` 

Should I make this parametrized ?

src/ophyd_async/core/_detector.py
evalott100: Actually, I think `_current_number_index` would be much better suited to be an iterator:
```suggestion
        self._number_of_triggers_iter: Iterator[int] | None = None
```
Then you can define it when you create the `trigger_info` as

``` python
    iter(
        trigger_info.number_of_triggers
        if isinstance(trigger.number_of_triggers)
        else iter([number_of_triggers])
    )
```

Then

``` python
        if isinstance(self._trigger_info.number, list):
            self._frames_collected += self._trigger_info.number[
                self._current_number_index
            ]
        else:
            self._frames_collected += self._trigger_info.number
```

becomes

``` python
    self.frames_to_complete = next(self.number_of_triggers_iter)
    self._frames_collected += self.frames_to_complete
```

and 

``` python
            if index >= self._trigger_info.number[self._current_number_index]:
                self._current_number_index += 1
                break
            else:
                if index >= self._trigger_info.number:
                    break
```

becomes

``` python
    if index == self._frames_to_complete:
        break
```

evalott100 (changes requested):

tests/core/test_flyer.py
evalott100: I would make another one (which should be parametrized). This one deals with a specific error where the trigger info is fine but you kickoff too many times.

We want a quick one which tests you can't make a `TriggerInfo` with whatever bad input (you don't need to make it a plan, just trying an initialize directly).

evalott100: Oh this test already exists (I forgot I converted this to a pydantic Model).

https://github.com/bluesky/ophyd-async/blob/e7cef5bebce8f079bf44fb842bb8a26bf2554fa3/tests/core/test_flyer.py#L316

Just add one or two with bad list values (non int value in the list and negative int values in the list).

src/ophyd_async/core/_detector.py
```diff
@@ -313,38 +342,55 @@ async def prepare(self, value: TriggerInfo) -> None:
...
+                yield WatcherUpdate(
+                    name=self.name,
+                    current=index,
+                    initial=self._initial_frame,
+                    target=target,
+                    unit="",
+                    precision=0,
+                    time_elapsed=time.monotonic() - self._fly_start
+                    if self._fly_start
+                    else 0,
```
evalott100, line 381:
```suggestion
                    if self._fly_start
                    else None,
```

src/ophyd_async/core/_detector.py
```diff
@@ -192,10 +205,14 @@ def __init__(
         # For kickoff
         self._watchers: list[Callable] = []
         self._fly_status: WatchableAsyncStatus | None = None
-        self._fly_start: float
-        self._iterations_completed: int = 0
-        self._initial_frame: int
-        self._last_frame: int
+        self._fly_start: float | None
```
evalott100, line 208:
```suggestion
        self._fly_start: float | None = None
```

tests/core/test_flyer.py
```diff
@@ -359,6 +393,24 @@ async def test_read_configuration():
             "Input should be 'internal', 'edge_trigger', 'constant_gate' or "
             "'variable_gate' [type=enum, input_value='not_in_enum', input_type=str]",
```
evalott100, line 394: Make your new `error_msg` be specific like this please.

evalott100 (changes requested):

tests/core/test_flyer.py
ZohebShaikh: This is the error message I get by default from pydantic 
`"1 validation error for TriggerInfo\nnumber_of_triggers\n  Field required [type=missing, input_value={'number': -100}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.9/v/missing"`

Did you mean I should create a custom error message using this https://docs.pydantic.dev/latest/errors/errors/#customize-error-messages

src/ophyd_async/core/_detector.py
```diff
@@ -313,38 +342,55 @@ async def prepare(self, value: TriggerInfo) -> None:
...
+                    if index >= self._frames_to_complete:
+                        break
+                else:
+                    if index >= self._trigger_info.number_of_triggers:
+                        break
+        finally:
+            await indices_written.aclose()
+            if self._frames_collected == self._total_frames_to_capture:
+                self._frames_collected = 0
+                await self.controller.wait_for_idle()
```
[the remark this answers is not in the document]
ZohebShaikh: If we increment this by 1 do when need to change the target to be +1 as well not `frames_to_complete`

tests/core/test_flyer.py
ZohebShaikh: I have made the changes ..Thank you :smile:

src/ophyd_async/core/_detector.py
```diff
@@ -312,39 +332,50 @@ async def prepare(self, value: TriggerInfo) -> None:
...
+        ), "Prepare must be called before kickoff!"
+        if self._frames_completed >= self._trigger_info.total_number_of_triggers:
+            raise RuntimeError(
                 f"Kickoff called more than the configured number of "
-                f"{self._trigger_info.iteration} iteration(s)!"
+                f"{self._trigger_info.total_number_of_triggers} iteration(s)!"
             )
-        self._iterations_completed += 1
+        self._frames_to_complete = next(self._number_of_triggers_iter)
+        self._frames_completed += self._frames_to_complete
```
evalott100, line 344: I've changed my mind on this https://github.com/bluesky/ophyd-async/pull/581#discussion_r1777342185

evalott100 (approved):
I'm happy with this, I think we should wait for a once over from @coretl before we merge.

src/ophyd_async/core/_detector.py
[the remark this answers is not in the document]
evalott100: Nah just the `frames_completed`. We remove it's increment from `kickoff`. Currently in kickoff we do:
``` python
        self._frames_completed += self._frames_to_complete
```

Which means it functions more as a variable indicating what the total number of frames completed so far is if we were to call prepare now - which I think is needlessly complex. If we iterate it by one each time then at the end of prepare `frames_to_complete` will be added to it.

evalott100 (changes requested):
I've changed my mind, I'd quite like

https://github.com/bluesky/ophyd-async/pull/581#discussion_r1780745106

Which means either:
### 1)
adding
``` python
                self._frames_completed += 1
```
to `complete` and removing
``` python
        self._frames_completed += self._frames_to_complete
```
from `kickoff`.

Or (and I actually think this is the nicer option)
### 2)
Changing its name to `self._completable_frames` and leaving a comment in `init` that it represents the total number of frames that will have been completed at the end of the next `complete`.

This is nicer because it means that
``` python
if self._completable_frames >= self._trigger_info.total_number_of_triggers:
            raise RuntimeError(
```

will still work - if we changed to increment by 1 then would probably still need this variable to make sure you can't `kickoff` too many times.

evalott100 (changes requested):

evalott100 (approved):
Fab, thank you :pray:

We should wait for a once over from @coretl

coretl (approved):

Issue #402, opened 2024-06-18
Add helper function to observe multiple signals


DominicOram:
In https://github.com/DiamondLightSource/hyperion/issues/1450 we had the usecase where we wanted to monitor multiple signals and raise under one condition or stop monitoring cleanly under another. To aid in this it would be good to have an `ophyd-async` helper that observed multiple signals and gave you any updates that occured e.g.

```python
async def pin_mounted_or_no_pin_found(self):
      async for signal, value in observe_signals_value(self.error_code, self.goinio_pin_sensor):
          if signal is self.error_code and value == self.NO_PIN_ERROR_CODE:
              # We didn't get a pin
              raise RobotLoadFailed(self.NO_PIN_ERROR_CODE, "Pin was not detected")
          elif signal is self.goinio_pin_sensor and value == PinMounted.PIN_MOUNTED:
              # We did get a pin
              break
```

coretl:
> Do we also want to rename `observe_value` -> `observe_signal_value` to make them more consistent?

Actually, I'm going to roll back on this, and say this ticket is just to add `observe_signals_value`

ZohebShaikh:
![image](https://github.com/user-attachments/assets/2e4438c8-4a1e-4e58-958d-63beefc91fc8)
We're deciding between two options for when to return the status:
Option A: Return status when the match signal changes to the expected value.
Option B: Return status when the signal changes to the expected value.
Both signals start at 0 and are expected to change to 1. The question is: Should we trigger the status return based on the match signal (A) or the main signal (B)?

DominicOram:
You're asking in relation to `set_and_wait_for_other_value`, right? Not `observe_signals_value`, which I think this issue is more aimed towards? I didn't think `observe_signals_value` needed to be given any `match` values.

If it's about `set_and_wait_for_other_value` then I think we should return based on `A` as that's what the name implies and what most of my usecases are. An example use case:
* I set `ACQUIRE` to `Do Acquire`
* I want to know when the hardware is acquiring, for this I need to wait on `ACQUIRE_RBV` to go to `Acquiring`
`ACQUIRE` will probably go to `Do Acquire` immediately on set (depending on the callback logic on the hardware side I guess) but I'm only sure the hardware is acquiring when the `AQUIRE_RBV` says so, which could be a while later.

PR #582, opened 2024-09-17
wait_for_value_interface_change

### LINKED ISSUE #461: Rationalise interface for `*wait_for_value*`
After discussions surrounding #453 and #447 I suggest we make the following changes to the following helpers for signal operations:
- Stop returning the `status` from `set_and_wait_for_*value` and instead await it, using `timeout` for it rather than a separate `status_timeout`
- Update all the areaDetector usages from:
```python
arm_status = await set_and_wait_for_value(self.acquire, True)
...
await arm_status
```
to
```python
arm_status = self.acquire.set(True)
await wait_for_value(self.acquire, True)
...
await arm_status
```
- Support using a `match` function in `set_and_wait_for_value` and `set_and_wait_for_other_value`

That would make the interface:
```python
async def observe_value(
    signal: SignalR[T],
    timeout: float | None = None,
    done_status: Status | None = None) -> AsyncGenerator[T, None]:
    ...

@override
async def observe_signals_value(signal1: SignalR[T], signalR[T2]) -> AsyncGenerator[Tuple[SignalR[T], T] | Tuple[SignalR[T2], T2]]:  # expand this for all combinations like asyncio.gather type stub does
    ...

async def observe_signals_value(
    *signals: SignalR[T],
    timeout: float | None = None,
    done_status: Status | None = None) -> AsyncGenerator[Tuple[SignalR[T], T], None]:
    ...

async def wait_for_value(
    signal: SignalR[T], 
    match_value: T | Callable[[T], bool], 
    timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
    ...

async def set_and_wait_for_value(
    signal: SignalRW[T], 
    value: T, 
    match_value: T | Callable[[T], bool] | None = None, 
    timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
    ...

async def set_and_wait_for_other_value(
    signal: SignalW[T],
    value: T, 
    match_signal: SignalR[S],
    match_value: S | Callable[[S], bool], 
    timeout: Optional[float] = DEFAULT_TIMEOUT) -> None:
    ...
```

ZohebShaikh:
closes #461


src/ophyd_async/core/_signal.py
```diff
@@ -582,37 +652,36 @@ async def set_and_wait_for_other_value(
...
     """
     # Start monitoring before the set to avoid a race condition
-    values_gen = observe_value(read_signal)
+    values_gen = observe_value(match_signal)
 
     # Get the initial value from the monitor to make sure we've created it
     current_value = await anext(values_gen)
 
-    status = set_signal.set(set_value, timeout=set_timeout)
+    set_signal.set(set_value, timeout=set_timeout)
```
coretl, line 660: @DominicOram this is what @ZohebShaikh was alluding to in https://github.com/bluesky/ophyd-async/issues/402#issuecomment-2357824352.

If we choose A, then we need to decide what to do with the Status. Something needs to keep track of it an await it, we can't drop it on the floor as we are doing here or we get teardown errors in the tests, so we have to return it. The calling code then looks like:
```python
status = await set_and_wait_for_other_value(dev.acquire, 1, dev.acquire_rbv, 1)
# do something that you can do as soon as the thing is acquiring
await status
# now the device has finished acquiring
```

If we choose B, then we keep a track of `status` here, then await it at the end of the function. The calling code becomes:
```python
await set_and_wait_for_other_value(dev.acquire, 1, dev.acquire_rbv, 1)
# now the device has finished acquiring,
```

What was your actual use case for this? Will the write_signal ever take significantly longer to caput-callback than the read_signal will take to change to the match_value? I know we have this case for the areaDetector acquire PV above, but we decided to not use this function as it was clearer to write:
```python
arm_status = dev.acquire.set(1)
await wait_for_value(dev.acquire_rbv, 1)
# do something we can do when the device is armed
await arm_status
# now the device is disarmed
```

src/ophyd_async/core/_signal.py
```diff
@@ -490,6 +490,75 @@ async def get_value():
...
+        If it raises an exception then this exception will be raised by the iterator.
+
+    Notes
+    -----
+    Example usage::
+
+        async for value in observe_value(sig):
+            do_something_with(value)
+    """
+    q: asyncio.Queue[tuple[SignalR[T], T | Status]] = asyncio.Queue()
```
coretl, line 519:
```suggestion
    q: asyncio.Queue[tuple[SignalR[T], T] | Status] = asyncio.Queue()
```

src/ophyd_async/core/_signal.py
```diff
@@ -490,6 +490,75 @@ async def get_value():
...
+
+        return queue_value, queue_status, clear_signals
+
+    clear_signals = []
+    for signal in signals:
+        queue_value, queue_status, clear_signal = wrapped_signal_put(signal)
+        clear_signals.append(clear_signal)
+        if done_status is not None:
+            done_status.add_callback(queue_status)
+        signal.subscribe_value(queue_value)
```
coretl, line 546:
```suggestion
    cbs: Dict[SignalR, Callback] = {}
    for signal in signals:
        def queue_value(value: T, signal=signal):
            q.put_nowait((signal, value))
        
        cbs[signal] = queue_value
        signal.subscribe_value(queue_value)
        
    if done_status is not None:
        done_status.add_callback(q.put_nowait)
```

src/ophyd_async/core/_signal.py
```diff
@@ -490,6 +490,75 @@ async def get_value():
...
+            if done_status and item is done_status:
+                if exc := done_status.exception():
+                    raise exc
+                else:
+                    break
+            else:
+                yield item  # type: ignore
+    finally:
+        for clear_signal in clear_signals:
+            clear_signal()
```
coretl, line 559:
```suggestion
        for signal, cb in cbs.items():
            signal.clear_sub(cb)
```

src/ophyd_async/core/_signal.py
```diff
@@ -490,6 +490,75 @@ async def get_value():
         signal.clear_sub(q.put_nowait)
 
 
+async def observe_signals_values(
```
coretl, line 493: After the above changes are made, can we share any code with `observe_value` above?

coretl (commented):
Please can we have a test with 2 soft signals updating and making callbacks?

src/ophyd_async/core/_signal.py
DominicOram: My original usecase for this was the actually the areaDetector one. We tried the code as suggested:
```python
arm_status = dev.acquire.set(1)
await wait_for_value(dev.acquire_rbv, 1)
await arm_status
```
But found that the `wait_for_value` was failing. This is because for this particular configuration the detector immediately takes data and does so so quickly that it has finished by the time we start monitoring the RBV. See https://github.com/bluesky/ophyd-async/issues/453#issue-2405971992. My main motivation for putting this into `ophyd-async` was to provide a function that would always protect against the potential race condition so we're less like to see it with people taking the naive approach.

I believe for my use case the callback has already returned on the set before the RBV goes to the expected value but I would have to check. Either way I think what I would expect to happen is that `set_and_wait_for_other_value` returns a gather of both statuses. If you want to do some in between these then you need to do that yourself. Maybe we should have a chat about it?

src/ophyd_async/core/_signal.py
```diff
@@ -540,35 +601,36 @@ async def wait_for_value(
...
 
 async def set_and_wait_for_other_value(
     set_signal: SignalW[T],
     set_value: T,
-    read_signal: SignalR[S],
-    read_value: S,
+    match_signal: SignalR[S],
+    match_value: S | Callable[[S], bool],
     timeout: float = DEFAULT_TIMEOUT,
     set_timeout: float | None = None,
```
coretl, line 618: We decided to add `wait_for_set_completion = True` (or something like that) to both functions and always return the status, but then pass `False` from the areaDetector utility function

src/ophyd_async/core/_signal.py
```diff
@@ -616,6 +616,7 @@ async def set_and_wait_for_other_value(
     match_value: S | Callable[[S], bool],
     timeout: float = DEFAULT_TIMEOUT,
     set_timeout: float | None = None,
+    wait_for_set_completion: bool = True,
 ):
```
coretl, line 620:
```suggestion
) -> AsyncStatus:
```

src/ophyd_async/core/_signal.py
```diff
@@ -660,7 +663,11 @@ async def _wait_for_value():
                     break
 
         try:
-            await asyncio.wait_for(_wait_for_value(), timeout)
+            status = asyncio.wait_for(_wait_for_value(), timeout)
+            if wait_for_set_completion:
+                await status
+            else:
+                return status
```
coretl, line 670:
```suggestion
            return status
```

src/ophyd_async/core/_signal.py
```diff
@@ -705,5 +714,11 @@ async def set_and_wait_for_value(
         match_value = value
 
     await set_and_wait_for_other_value(
```
coretl, line 716:
```suggestion
    return await set_and_wait_for_other_value(
```

src/ophyd_async/core/_signal.py
```diff
@@ -673,6 +680,7 @@ async def set_and_wait_for_value(
     match_value: T | Callable[[T], bool] | None = None,
     timeout: float = DEFAULT_TIMEOUT,
     status_timeout: float | None = None,
+    wait_for_set_completion: bool = True,
 ):
```
coretl, line 684:
```suggestion
) -> AsyncStatus:
```

src/ophyd_async/epics/adcore/_core_logic.py
```diff
@@ -93,7 +93,9 @@ async def start_acquiring_driver_and_ensure_status(
     """
     status = driver.acquire.set(True, timeout=timeout)
     await wait_for_value(driver.acquire, True, timeout=None)
```
coretl:
```suggestion
```

src/ophyd_async/epics/adcore/_core_logic.py
```diff
@@ -93,7 +93,9 @@ async def start_acquiring_driver_and_ensure_status(
...
-    await set_and_wait_for_value(driver.acquire, True, timeout=timeout)
+    await set_and_wait_for_value(
+        driver.acquire, True, timeout=timeout, wait_for_set_completion=False
+    )
```
coretl, line 98:
```suggestion
    status = await set_and_wait_for_value(
        driver.acquire, True, timeout=timeout, wait_for_set_completion=False
    )
```

src/ophyd_async/core/_signal.py
```diff
@@ -663,15 +663,15 @@ async def _wait_for_value():
                     break
 
         try:
-            status = asyncio.wait_for(_wait_for_value(), timeout)
+            status = AsyncStatus(asyncio.wait_for(_wait_for_value(), timeout))
```
coretl, line 666: The `status` I was referring to is the one returned from `set_signal.set`. I was expecting:
```python
    status = set_signal.set(set_value, timeout=set_timeout)
    ...
            if wait_for_set_completion:
                await status
            return status
```

src/ophyd_async/epics/adaravis/_aravis_controller.py
```diff
@@ -5,10 +5,12 @@
     DetectorController,
     DetectorTrigger,
     TriggerInfo,
-    set_and_wait_for_value,
 )
 from ophyd_async.core._status import AsyncStatus
 from ophyd_async.epics import adcore
+from ophyd_async.epics.adcore._core_logic import (
+    start_acquiring_driver_and_ensure_status,
+)
```
coretl, line 13:
```suggestion
```

tests/core/test_signal.py
```diff
@@ -180,32 +182,33 @@ async def wait_and_set_proceeds():
         set_mock_put_proceeds(signal, True)
 
     async def check_set_and_wait():
-        await (await set_and_wait_for_value(signal, 1, timeout=0.1))
+        await set_and_wait_for_value(signal, 1, timeout=0.1)
+        await wait_for_value(signal, 1, timeout=0.1)
```
coretl, line 186: set_and_wait_for_value is doing this wait_for_value already
```suggestion
```

tests/core/test_signal.py
```diff
@@ -279,6 +282,70 @@ def less_than_42(v):
     assert await time_taken_by(wait_for_value(signal, less_than_42, timeout=2)) < 0.1
 
 
+async def test_set_and_wait_for_value(): ...
```
coretl, line 285: blank?

tests/core/test_signal.py
```diff
@@ -297,6 +364,9 @@ async def test_create_soft_signal(signal_method, signal_class):
     assert (await signal.get_value()) == INITIAL_VALUE
 
 
+# write code to add two numbers
+
+
```
coretl, line 369: `1 + 1 = 2`

tests/epics/test_motor.py
```diff
@@ -279,10 +279,13 @@ async def test_prepare(
     set_mock_value(sim_motor.high_limit_travel, 20)
     set_mock_value(sim_motor.max_velocity, 10)
     fake_set_signal = SignalRW(MockSignalBackend(float))
+    fake_other_set_signal = SignalRW(MockSignalBackend(float))
```
coretl, line 282: Ermm, why are you adding this to a motor test? Could it go in `test_signal.py` instead?

coretl (approved):

PR #590, opened 2024-09-19
Reset completed iterations counter to 0 once target iterations are reached and detector is disarmed

jwlodek:
Without this, if you run a plan from an ipython session without re-instantiating the ophyd device, you will only ever be able to fly it once currently, getting the error about kicking off more than the number of iterations.


ZohebShaikh (approved):
Nice adding the test for more kick-off

PR #591, opened 2024-09-19
Refactor ADPilatus tests to use the ad_standard_det_factory

jwlodek:


jwlodek:
Closing in favor of different PR that does this for pilatus + sim detector + more ad standard det test simplification.

Issue #262, opened 2024-04-24
Parametrize tests for detectors based on list of fixtures.


jwlodek:
A lot of the tests used for `ADAravis`, `ADKinetix` etc. should be run for all detector classes. Ideally, we should make a module `test_detectors` that has a list of fixture names for all the different detector fixtures, then runs tests for each one.

See: https://engineeringfordatascience.com/posts/pytest_fixtures_with_parameterize/

PR #592, opened 2024-09-19
Simplify ad standard det tests

jwlodek:
Significantly simplifies tests for ADPilatus and ADSimDetector to use ad_standard_det_factory fixture. Moves some of the mock callback logic to the fixture to simplify other tests as well.


tests/epics/adpilatus/test_pilatus.py
```diff
@@ -110,7 +97,7 @@ async def _trigger(
 
 
 async def test_hints_from_hdf_writer(test_adpilatus: adpilatus.PilatusDetector):
-    assert test_adpilatus.hints == {"fields": ["test_adpilatus"]}
+    assert test_adpilatus.hints == {"fields": ["test_adpilatus1"]}
```
DiamondJoseph, line 100:
```suggestion
    assert test_adpilatus.hints == {"fields": [test_adpilatus.name]}
```

nit: I think the important check here is that the hinted field for the hdfwriter comes from the device name

DiamondJoseph (approved):
Looks good, nice way of simplifying: is this intended to close #262 or is it just a step in that direction?

There's a couple of type-checking bits that might be useful sanity checks [PathProvider is `Callable[[str | None], Path]`; ad_standard_det_factory is a `Callable[[type[T], int], T]`, but I think the Pytest Fixtures are making is clear what's being passed: just if we want to strict pyright this module I think they'll need the hints?

ZohebShaikh (approved):

jwlodek:
> Looks good, nice way of simplifying: is this intended to close #262 or is it just a step in that direction?

Mostly just a step in that direction - I think the tests can be further simplified by just paramterizing over the different standard detector classes, leaving just the tests for the individual controllers. But I'll leave that for another PR since this one is already fairly broad.

Issue #551, opened 2024-09-02
Eiger FastCS: Use PVI information to fill in PVs


coretl:
As soon as PVI datastructures are published over PVA then this file will drastically change to something like:
```python
class EigerDriverIO(Device):
    bit_depth: SignalR[int]
    stale_parameters: SignalR[bool]
    ...
```
It will also only need to include the signals that are needed at static analysis time, at runtime then all the extras (like `beam_centre_x`, etc) will be filled in automatically.

This will also mean that the class structure and ophyd attribute names will have to exactly match those in PVI.

For this reason I suggest we do this soon, and @GDYendell is going to start looking at this now

_Originally posted by @coretl in https://github.com/bluesky/ophyd-async/pull/528#discussion_r1723187325_

DominicOram:
Following https://github.com/bluesky/ophyd-async/issues/557#issuecomment-2357802049 I think we can now do this.

coretl:
Steps to complete this:
- Run up the Eiger simulator (https://github.com/DiamondLightSource/fastcs-eiger/tree/readme for getting started and comment on https://github.com/DiamondLightSource/fastcs-eiger/pull/58 if it doesn't Just Work)
- Make `ophyd_async/fastcs/eiger/` and `ophyd_async/fastcs/odin/` modules
- Move `EigerDriverIO` to the new `eiger` module and write in the declarative style as in the issue description
- Move `EigerController` to the new `eiger` module
- Rename `Odin` to `OdinHdfIO` and move to new `odin` module and write in the declarative style
- Move `OdinWriter` to new `odin` module
- Delete `ophyd_async/epics/eiger/` in favour of this new module
- Fix up the tests

PR #594, opened 2024-09-24
New signal typing

### LINKED ISSUE #373: Only support the formalized pvi structure from PandA
At the moment a pvi structure has a `pvi` field which contains the PVs for that Device. This is unlike any of the other records with have a `value` field instead.

https://github.com/PandABlocks/PandABlocks-ioc/issues/120 will change all of this so that:
- typeid is `epics:nt/NTPVI:1.0`
- entries are in `value` not `pvi`

This means we can:
- Rename `PvaDictConverter` to `PvaPviConverter`, and remove all the methods apart from `value`
- Use `PvaPviConverter` where we see typeid `epics:nt/NTPVI:1.0`
- Remove https://github.com/bluesky/ophyd-async/blob/cbbb2950ed4c4b0bf4ca4c03729562b842b44b03/src/ophyd_async/epics/_backend/_p4p.py#L62-L72 and hardcode them where they are used

This will require PandA IOC to be updated in sync with this change, but this is fine as we are pre-1.0

### LINKED ISSUE #472: Change mechanism of setting a signal backend after creation, and support pvi and tango with less boilerplate
We would like to add the following features to the pvi connection strategy:
- Share code with tango strategy for creating mock signals and devices
- Support caching of connect

To do this I suggest that the "introspectable device" template should now look like this:
```python
from ophyd_async.core import Device, SignalRW, SignalX, create_children_from_annotations

class MyDevice(Device):
    attr_one: SignalRW[float]
    attr_two: SignalRW[str]
    attr_three: SignalX

        self._prefix = prefix
        # Create Signal* objects from annotations
        create_children_from_annotations(self)
        super.__init__(name=name)

    async def pre_connect(self, timeout: float):
        # Connect to real device
        await fill_pvi_signals(self, self._prefix + ":PVI", timeout)
```

This would require the following changes:
- Support `SignalBackend.set_backend` rather than passing it during `connect`
- Calling `SignalBackend.set_backend` during `fill_pvi_signals`
- Adding support to `DeviceVector` to take a datatype and the number of mock devices to make
- Adding `DeviceVector.connect(mock=True)` that makes the correct number of mock devices
- Adding a blank `Device.pre_connect` method that can be overridden to call `fill_pvi_signals` or `fill_tango_signals` when in mock=False mode

### LINKED ISSUE #505: Allow creation of a numpy array soft signal
When I try and do this I get an error of:
```python
ERROR:blueapi.worker.reworker:invalid converter for value of type <class 'numpy.ndarray'>
  File "/venv/lib/python3.11/site-packages/blueapi/worker/reworker.py", line 250, in _cycle
    self._current.task.do_task(self._ctx)
  File "/venv/lib/python3.11/site-packages/blueapi/worker/task.py", line 31, in do_task
    ctx.run_engine(func(**prepared_params.dict()))
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 918, in __call__
    plan_return = self._resume_task(init_func=_build_task)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1057, in _resume_task
    raise exc
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1687, in _run
    raise err
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1547, in _run
    msg = self._plan_stack[-1].send(resp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/blueapi-plugins/scratch/mx_bluesky/src/mx_bluesky/i04/thawing_plan.py", line 95, in thaw_and_center
    yield from _thaw_and_center()
  File "/venv/lib/python3.11/site-packages/bluesky/utils/__init__.py", line 1203, in dec_inner
    return (yield from plan)
            ^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 408, in subs_wrapper
    return (yield from finalize_wrapper(_inner_plan(),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 535, in finalize_wrapper
    ret = yield from plan
          ^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 406, in _inner_plan
    return (yield from plan)
            ^^^^^^^^^^^^^^^
  File "/blueapi-plugins/scratch/mx_bluesky/src/mx_bluesky/i04/thawing_plan.py", line 87, in _thaw_and_center
    yield from bps.monitor(oav.test_signal, name="oav")
  File "/venv/lib/python3.11/site-packages/bluesky/plan_stubs.py", line 146, in monitor
    return (yield Msg('monitor', obj, name=name, **kwargs))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1607, in _run
    new_response = await coro(msg)
                   ^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1950, in _monitor
    await current_run.monitor(msg)
  File "/venv/lib/python3.11/site-packages/bluesky/bundlers.py", line 392, in monitor
    await self._ensure_cached(obj)
  File "/venv/lib/python3.11/site-packages/bluesky/bundlers.py", line 207, in _ensure_cached
    await asyncio.gather(*coros)
  File "/venv/lib/python3.11/site-packages/bluesky/bundlers.py", line 337, in _cache_describe
    self._describe_cache[obj] = await maybe_await(obj.describe())
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/bluesky/utils/__init__.py", line 1852, in maybe_await
    return await ret
           ^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/ophyd_async/core/signal.py", line 41, in wrapper
    return await asyncio.wait_for(func(self, *args, **kwargs), self._timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/tasks.py", line 489, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/ophyd_async/core/signal.py", line 195, in describe
    return {self.name: await self._backend.get_datakey(self.source)}
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/ophyd_async/core/soft_signal_backend.py", line 186, in get_datakey
    return self.converter.get_datakey(source, self._value, **self._metadata)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/ophyd_async/core/soft_signal_backend.py", line 60, in get_datakey
    dtype in primitive_dtypes
AssertionError: invalid converter for value of type <class 'numpy.ndarray'>
```

I can use a list to get round this but `numpy` arrays are valid as normal PV signal types and so should be valid here

### LINKED ISSUE #535: Make `Signal` `reading["value"]` and `value` be the same
After https://github.com/bluesky/bluesky/issues/1800 we have a `Reading[T]` where `value` is of type `T`.
After https://github.com/bluesky/ophyd-async/pull/522 we will have `SignalBackend.get_value()` and `SignalBackend.get_reading()["value"]` being the same for all types.
This ticket is to remove the possibility for them to be different.

## Change
https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_utils.py#L29
to
```python
ReadingCallback = Callable[[Reading[T]], None]
```

https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_signal_backend.py#L45
to
```python
def set_callback(self, callback: Optional[ReadingCallback[T]]) -> None:
``` 

https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_signal.py#L140-L146
to
```python
def _callback(self, reading: Reading[T]): 
    self._signal.log.debug( 
        f"Updated subscription: reading of source {self._signal.source} changed" 
        f"from {self._reading} to {reading}" 
    ) 
    self._reading = reading 
    self._value = reading["value"]
```

Then tidy up all the converters so that they produce `reading`, rather than `reading, value`, which means the `Converter.value` can probably disappear

### LINKED ISSUE #562: Change `Signal`/`SignalBackend` args
We should alter the `kwargs` of `Signal`/`SignalBackend` such that

- `Signal` takes an optional `datatype` in `__init__` and do some validation on it
- `EpicsSignalBackend` takes `datatype` in `__init__`
- `SignalBackend.connect` now takes `datatype` instead
- On both `Signal` and `SignalBackend` `__init__` and `connect ` `arg` to `kwarg` inference is prevented

``` python

# Signal
    def __init__(
        self,
        *,
        backend: Optional[SignalBackend[T]] = None,
        datatype: Optional[SignalDataType] = None,
        timeout: Optional[float] = DEFAULT_TIMEOUT,
        name: str = "",
    ) -> None:
        #: If set, then the datatype of the value is guaranteed to be this type
        self.datatype = datatype
        ....

    async def connect(
        self,
        *,
        mock=False,
        timeout=DEFAULT_TIMEOUT,
        backend: Optional[SignalBackend[T]] = None,
    ):
        ....

# Signal backend

@abstractmethod
async def connect(
    self,
    *,
    datatype: Optional[type] = None,
    timeout: float = DEFAULT_TIMEOUT
):
    ....
```

Part of this change will be getting rid of conversion in mock signal backend.

_Originally posted by @coretl in https://github.com/bluesky/ophyd-async/issues/310#issuecomment-2205415110_

### LINKED ISSUE #601: Make an ADR on signal types and enforce
Following https://github.com/bluesky/ophyd-async/issues/310#issuecomment-2324326509 we have a well defined list of types that signals are allowed. However, there are still old devices in `dodal` that use different types e.g. https://github.com/DiamondLightSource/dodal/blob/f2e63c2c55d49ed35cef73531be7d4bcceb42546/src/dodal/devices/zocalo/zocalo_results.py#L133. To make sure these get phased out and more don't get introduced we should prohibit other types ideally through the typing system. To make it clear why this is the case we should make an ADR too.

## Acceptance Criteria
- An ADR is made on the types that signals are allowed
- If try and make a signal with an unsupported type e.g. `soft_signal_r_and_setter(MyArbitraryType)` the type system will complain

coretl:
Rewrite the type support, and use a plugin `Connector` architecture to support reconnection for PVI and Tango.

Fixes #472, fixes #562, fixes #535, fixes #505, fixes #373, fixes #601 

Required for #551

# Breaking changes
## pvi structure changes
Structure now read from `.value` rather than `.pvi`. Supported in FastCS. Requires at least PandABlocks-ioc 0.10.0
## `StrictEnum` is now requried for all strictly checked `Enums`
```python
# old
from enum import Enum
class MyEnum(str, Enum):
    ONE = "one"
    TWO = "two"
# new
from ophyd_async.core import StrictEnum
class MyEnum(StrictEnum):
    ONE = "one"
    TWO = "two"
```
## `SubsetEnum` is now an `Enum` subclass:
```python
from ophyd_async.core import SubsetEnum
# old
MySubsetEnum = SubsetEnum["one", "two"]
# new
class MySubsetEnum(SubsetEnum):
    ONE = "one"
    TWO = "two"
```
## Use python primitives for scalar types instead of numpy types
```python
# old
import numpy as np
x = epics_signal_rw(np.int32, "PV")
# new
x = epics_signal_rw(int, "PV")
```
## Use `Array1D` for 1D arrays instead of `npt.NDArray`
```python
import numpy as np
# old
import numpy.typing as npt
x = epics_signal_rw(npt.NDArray[np.int32], "PV")
# new
from ophyd_async.core import Array1D
x = epics_signal_rw(Array1D[np.int32], "PV")
```
## Use `Sequence[str]` for arrays of strings instead of `npt.NDArray[np.str_]`
```python
import numpy as np
# old
import numpy.typing as npt
x = epics_signal_rw(npt.NDArray[np.str_], "PV")
# new
from collections.abc import Sequence
x = epics_signal_rw(Sequence[str], "PV")
```
## `MockSignalBackend` requires a real backend
```python
# old
fake_set_signal = SignalRW(MockSignalBackend(float))
# new
fake_set_signal = soft_signal_rw(float)
await fake_set_signal.connect(mock=True)
```
## `get_mock_put` is no longer passed timeout as it is handled in `Signal`
```python
# old
get_mock_put(driver.capture).assert_called_once_with(Writing.ON, wait=ANY, timeout=ANY)
# new
get_mock_put(driver.capture).assert_called_once_with(Writing.ON, wait=ANY)
```
## `super().__init__` required for `Device` subclasses
```python
# old
class MyDevice(Device):
        self.signal, self.backend_put = soft_signal_r_and_setter(int)
# new
class MyDevice(Device):
        self.signal, self.backend_put = soft_signal_r_and_setter(int)
        super().__init__(name=name)
```
## Arbitrary `BaseModel`s not supported, pending use cases for them
The `Table` type has been suitable for everything we have seen so far, if you need an arbitrary `BaseModel` subclass then please make an issue
## Child `Device`s set parent on attach, and can't be public children of more than one parent
```python
class SourceDevice(Device):
        self.signal = soft_signal_rw(int)
        super().__init__(name=name)

# old
class ReferenceDevice(Device):
    def __init__(self, signal: SignalRW[int], name: str = ""):
        self.signal = signal
        super().__init__(name=name)

    def set(self, value) -> AsyncStatus:
        return self.signal.set(value + 1)
# new
from ophyd_async.core import Reference

class ReferenceDevice(Device):
    def __init__(self, signal: SignalRW[int], name: str = ""):
        self._signal_ref = Reference(signal)
        super().__init__(name=name)

    def set(self, value) -> AsyncStatus:
        return self._signal_ref().set(value + 1)
```


coretl:
Apologies, another unreviewable PR, but if you start from the tests you can see the changes needed

src/ophyd_async/core/_utils.py
```diff
@@ -119,19 +123,38 @@ async def wait_for_connection(**coros: Awaitable[None]):
...
-    if getattr(typ, "__origin__", None) == np.ndarray:
-        # datatype = numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]]
-        # so extract numpy.float64 from it
-        return np.dtype(typ.__args__[1].__args__[0])  # type: ignore
-    return None
+    if not get_origin(datatype) == np.ndarray:
+        raise TypeError(f"Expected np.ndarray, got {datatype}")
+    # datatype = numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]]
+    # so extract numpy.float64 from it
+    return np.dtype(get_args(get_args(datatype)[1])[0])
```
evalott100, line 138: Do we allow for np arrays without subtype? If so we need:

```suggestion
    if get_origin(datatype) == np.ndarray:
        # np.ndarray[typing.Any, np.dtype[np.float64]] returns np.float64
        return np.dtype(get_args(get_args(datatype)[1])[0])
    elif datatype == np.ndarray:
        return np.float64
    else:
        raise TypeError(f"Expected np.ndarray, got {datatype}")
```

If not then perhaps a modification to the error message that the numpy array must be explicitly typed.

src/ophyd_async/epics/signal/_signal.py
```diff
@@ -2,46 +2,81 @@
...
+    class TransportNotAvailable:
+        def __init__(*args, **kwargs):
+            raise NotImplementedError("Transport not available") from error
+
+    return TransportNotAvailable
+
+
+class EpicsProtocol(Enum):
+    ca = "ca"
+    pva = "pva"
```
evalott100, line 28:
```suggestion
    NONE_AVAILABLE = None # Or smth
    CA = "ca"
    PVA = "pva"
```

src/ophyd_async/epics/signal/_signal.py
```diff
@@ -2,46 +2,81 @@
...
+
+
+_default_epics_protocol = EpicsProtocol.ca
```
evalott100, line 31:
```suggestion
_default_epics_protocol = EpicsProtocol.NONE_AVAILABLE
```

evalott100 (changes requested):
Most of these are discussion points more than "requested changes".

pyproject.toml
```diff
@@ -13,12 +13,12 @@ classifiers = [
...
 dependencies = [
     "networkx>=2.0",
-    "numpy<2.0.0",
+    "numpy",
     "packaging",
     "pint",
     "bluesky>=1.13.0a3",
-    "event_model",
-    "p4p",
+    "event-model @ git+https://github.com/bluesky/event-model@main",
```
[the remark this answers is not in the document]
coretl: There will probably be a few other import level changes to event model before I'm done, then we can release event model too and go back to a release here

src/ophyd_async/core/_device.py
```diff
@@ -66,7 +92,7 @@ def set_name(self, name: str):
             del self.log
 
         self._name = name
-        for attr_name, child in self.children():
+        for attr_name, child in self.children().items():
```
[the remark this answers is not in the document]
coretl: I like the first of these

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -1,244 +1,218 @@
...
-        }
-
-    def make_initial_value(self, datatype: type[T] | None) -> T:
-        if datatype is None:
-            return cast(T, None)
-
-        if issubclass(datatype, Enum):
-            return cast(T, list(datatype.__members__.values())[0])  # type: ignore
-        return cast(T, self.choices[0])
+    def write_value(self, value: Any) -> EnumT:
```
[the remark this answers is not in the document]
coretl: In typing world then we will never `signal.put()` anything other than `None | SignalDatatypeT`. However in real life people could put anything to it, so I chose to make the `Converter` be true to this rather than the stricter typing sense. This could do with a comment...

src/ophyd_async/epics/signal/_signal.py
coretl: I've changed this again

src/ophyd_async/core/_device.py
```diff
@@ -1,39 +1,63 @@
...
+class DeviceConnector:
+    @abstractmethod
+    async def connect(
+        self, device: Device, mock: bool, timeout: float, force_reconnect: bool
+    ) -> None: ...
+
+
+class DeviceChildConnector(DeviceConnector):
+    async def connect(
+        self, device: Device, mock: bool, timeout: float, force_reconnect: bool
```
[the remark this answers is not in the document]
coretl: I'd rather not repeat the defaults in every `DeviceConnector` as they should only be used via a `Device` anyway...

src/ophyd_async/core/_device.py
```diff
@@ -47,10 +71,12 @@ def log(self):
...
-    def children(self) -> Iterator[tuple[str, "Device"]]:
-        for attr_name, attr in self.__dict__.items():
-            if attr_name != "parent" and isinstance(attr, Device):
-                yield attr_name, attr
+    def children(self) -> dict[str, Device]:
+        return {
+            attr_name: attr
+            for attr_name, attr in self.__dict__.items()
+            if attr_name != "parent" and isinstance(attr, Device)
+        }
```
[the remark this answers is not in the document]
evalott100: Tbh we don't have to define use an abstract class, instead we can:

``` python
class Device(HasName):
    @property
    def children(self) -> "DeviceVector":
        return DeviceVector(
            {
                attr_name: attr
                for attr_name, attr in self.__dict__.items()
                if attr_name != "parent" and isinstance(attr, Device)
            }
        )
    def connect(self, ...):
        # DO STUFF
        self.children.connect(...)

class DeviceVector(dict[int, VT], Device):
    """
    Defines device components with indices.

    In the below example, foos becomes a dictionary on the parent device
    at runtime, so parent.foos[2] returns a FooDevice. For example usage see
    :class:`~ophyd_async.epics.demo.DynamicSensorGroup`
    """

    def connect(self, *args, **kwargs):
        # Logic for connecting which can be used for by Device
        # For its children
        gather({
            device.connect(*args, **kwargs)
            for device_name, device in self.items():
        })
        

    def children(self) -> DeviceVector:
        return self
```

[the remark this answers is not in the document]
coretl: This reduces the code a bit for the current use case, but I think it will break what I had planned for PVI. Let me write that first then we can revisit this and see if this still applies...

src/ophyd_async/core/_table.py
```diff
@@ -64,90 +88,44 @@ def __add__(self, right: TableSubclass) -> TableSubclass:
...
+    def numpy_table(self, selection: slice | None = None) -> np.ndarray:
+        array = None
+        for k, v in self:
+            if selection:
+                v = v[selection]
+            if array is None:
+                array = np.empty(v.shape, dtype=self.numpy_dtype(self))
+            array[k] = v
+        assert array
+        return array
```
evalott100, line 114: I think we can do:

```suggestion
    def numpy_table(self, selection: slice = [::]) -> np.ndarray:
        return np.array(
            [
                (k, v[selection])
                for k, v in self
            ],
            dtype=self.numpy_dtype()
        )
```

Has the perk of vacuously working with empty tables.

evalott100: Ah I'm incorrect, ignore me

src/ophyd_async/core/_device.py
```diff
@@ -1,39 +1,107 @@
...
+    def __init__(
+        self,
+        name: str = "",
+        connector: DeviceConnectorType | None = None,
+    ) -> None:
+        if connector is None:
+            # TODO: this is ugly, maybe we remove the option to pass None as the
+            # connector so this goes away?
+            connector = cast(DeviceConnectorType, DeviceChildConnector(self))
+        self.connect = connector
```
evalott100, line 104: Apologies, I left a comment before which had an incorrect example - what I meant was

``` python
def foo(device: Device[PviDeviceConnector]):
    assert isinstance(device.connect, DeviceChildConnector)

class DeviceA(Device):
    ...
```
class DeviceB(Device[DeviceChildConnector]):
    ...

device_1: DeviceA = DeviceA()
foo(device_1) # passes type hinting

device_2: DeviceB = DeviceB()
foo(device_2) # correctly fails type hinting
```

in the above code we don't have nice default performance because of this cast.

Comment (coretl) [reply src/ophyd_async/core/_device.py:104]: In `class Device`, we need a `connect: DeviceConnector`. In `class Signal(Device)`, we need a `connect: SignalConnector`. To make this compatible we make `class Device(Generic[DeviceConnectorT])`, and make `class Signal(Device[SignalConnector])`. However, `Device.__init__` takes `DeviceConnectorT | None`. If we supply a `DeviceConnectorT` then typing is happy, but if we supply `None` then typing cannot infer its type, so when we create it in the code above, it is of the wrong type (`DeviceChildConnector`, not `DeviceConnectorT`). This is a valid concern, typing does not know that `Signal` will always pass `SignalConnector` in the constructor rather than `None`. I'm not sure what to do about this concern, and I'm not sure if there's anything in your comment that will fix this particular concern? Also, at runtime `typing.cast` [does nothing](https://github.com/python/cpython/blob/cceb25c5033bce5d19311d63f6943215a1a27283/Lib/typing.py#L2182-L2190) so there is no performance hit.

File: src/ophyd_async/core/_signal_backend.py [hunk 11]
```diff
@@ -161,11 +164,12 @@ def make_datakey(
     source: str,
     metadata: SignalMetadata,
 ) -> DataKey:
+    dtn = _datakey_dtype_numpy(datatype, value)
     return DataKey(
         dtype=_datakey_dtype(datatype),
         shape=_datakey_shape(value),
         # Ignore until https://github.com/bluesky/event-model/issues/308
-        dtype_numpy=_datakey_dtype_numpy(datatype, value).str,  # type: ignore
+        dtype_numpy=dtn.descr if len(dtn.descr) > 1 else dtn.str,  # type: ignore
```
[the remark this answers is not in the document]
coretl: > I should maybe head back to event model and allow this.

Yes please

src/ophyd_async/core/_device.py
evalott100: Here's one solution if we want to forbid subclasses of types without a Connector in the generic e.g `class A(Device):`

``` python
from typing import Generic, TypeVar, get_args, cast
import pytest

class SomeClass:
    def __init__(self, base: "Base"): ...

class SomeClass1(SomeClass): ...

class SomeClass2(SomeClass): ...

T = TypeVar("T", bound=SomeClass)

class Base(Generic[T]):
    value: T

    def __init__(self, value: T | None = None):
        # In 3.12 you can use types.get_original_bases
        annotation = type(self).__orig_bases__[0] # pyright: ignore
        print(type(self).__name__, "ANNOTATED", annotation)
        arg = get_args(annotation)
        print(type(self).__name__, "ARGS", annotation)
        type_in_generic = arg[0]
        if type_in_generic is T:
            # This means they didn't provide a type
            raise RuntimeError

        print(type(self).__name__, "TYPE IN GENERIC", type_in_generic)

        # Type hinting would complain if this wasn't true as well
        assert issubclass(type_in_generic, SomeClass)

        value = value or cast(T, type_in_generic(self))

class Sub1(Base[SomeClass1]): ...

class Sub2(Base[SomeClass2]): ...

class Sub3(Base): ...

def foo(x: Base[SomeClass1]) -> None: ...


sub1 = Sub1()
foo(sub1) # This correctly passes type checking

sub2 = Sub2()
foo(sub2) # This correctly fails type checking

with pytest.raises(RuntimeError):
    sub3 = Sub3() # This can't even be initialised
```

With output:
```
Sub1 ANNOTATED __main__.Base[__main__.SomeClass1]
Sub1 ARGS __main__.Base[__main__.SomeClass1]
Sub1 TYPE IN GENERIC <class '__main__.SomeClass1'>
Sub2 ANNOTATED __main__.Base[__main__.SomeClass2]
Sub2 ARGS __main__.Base[__main__.SomeClass2]
Sub2 TYPE IN GENERIC <class '__main__.SomeClass2'>
Sub3 ANNOTATED typing.Generic[~T]
Sub3 ARGS typing.Generic[~T]
```

This is very hacky (until python 3.12), but it means that Devices require the connector.

src/ophyd_async/core/_signal_backend.py
```diff
@@ -1,97 +1,175 @@
...
+    | Array1D[np.int64]
+    | Array1D[np.uint64]
+    | Array1D[np.float32]
+    | Array1D[np.float64]
+    | np.ndarray
+    | SubsetEnum
+    | Sequence[str]
+    | Sequence[SubsetEnum]
+    | Table
+)
```
coretl, line 35: Need to put these in an ADR as per #601

coretl:
@evalott100 if you get a chance please could you review `core._device.py` and `core._signal.py`

src/ophyd_async/tango/demo/_counter.py
```diff
@@ -19,7 +19,7 @@ class TangoCounter(TangoReadable):
     counts: SignalR[int]
     sample_time: SignalRW[float]
     start: SignalX
-    _reset: SignalX
+    reset_: SignalX
```
coretl, line 22: @burkeds I've made it a trailing underscore rather than leading underscore to fit better with https://peps.python.org/pep-0008/#function-and-method-arguments

src/ophyd_async/tango/base_devices/_base_device.py
```diff
@@ -150,76 +93,62 @@ def decorator(cls):
...
+                .union(self.proxy.get_attribute_list())
+                .union(self.proxy.get_command_list())
+            )
+            for name in children:
+                # TODO: strip attribute name
+                full_trl = f"{self.trl}/{name}"
+                signal_type = await infer_signal_type(full_trl, self.proxy)
+                if signal_type:
+                    backend = self._filler.make_child_signal(name, signal_type)
+                    backend.datatype = await infer_python_type(full_trl, self.proxy)
```
coretl, line 140: @burkeds if I remove the `infer_python_type` it still works with `TangoSignalBackend.datatype = None`. Is there any reason to keep `infer_python_type`?

tests/tango/test_tango_signals.py
```diff
@@ -27,6 +25,11 @@
 from tango.test_context import MultiDeviceTestContext
 from tango.test_utils import assert_close
 
+
+def __tango_signal_auto(*args, **kwargs):
+    raise RuntimeError("Fix this later")
```
coretl, line 30: @burkeds I've removed tango_signal_auto as it was only used in `TangoDevice` creation of Signals that are not type hinted in the class. Is it used anywhere else? Can I delete the tests that go with them? Or replace them with `tango_signal_rw(None,...)` etc?

burkeds: tango_signal_auto is only used when automatically creating signals. Why would you want to delete it? Will you implement another method that does the same thing?

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: They will initialize but the datatype of the signal will be None. This should throw up an error if you try to say, write a float to the signal. infer_python_type makes sure automatically generated signals are typed according to the signal's Tango type.

src/ophyd_async/tango/demo/_counter.py
burkeds: That's fine with me. When signals are automatically created, we just have to make sure that this underscore placement is treated correctly.

src/ophyd_async/tango/base_devices/_base_device.py
coretl: What should throw up an error? Signal or the Tango proxy?

src/ophyd_async/core/_device.py
```diff
@@ -1,128 +1,155 @@
...
+        return instance
+
+    def __init__(self, name: str = "", backend: DeviceBackend | None = None) -> None:
+        if backend:
+            # Copy into the new backend the children that have already been added
+            backend.children.update(self._backend.children)
+            self._backend = backend
+        super().__init__(name)
+
+    def _set_child_name(self, child: DeviceBase, child_name: str):
```
evalott100, line 104: Are the annotations in here `DeviceBase` instead of `Device` because we expect others to implement their own `Device`?

evalott100 (changes requested):

src/ophyd_async/core/_device.py
```diff
@@ -1,128 +1,155 @@
...
+    task: asyncio.Task | None = None
+
+    async def need_connect(self, mock: bool, force_reconnect: bool) -> bool:
+        can_use_previous_connect = (
+            mock is self.mock_arg
+            and self.task
+            and not (self.task.done() and self.task.exception())
+        )
+        if can_use_previous_connect and not force_reconnect:
+            assert self.task, "Connect caching not working"
```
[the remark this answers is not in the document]
coretl: yes, the logic above seems too much for pyright

src/ophyd_async/core/_device.py
```diff
@@ -1,128 +1,155 @@
...
     _name: str = ""
     #: The parent Device if it exists
-    parent: Optional["Device"] = None
-    # None if connect hasn't started, a Task if it has
-    _connect_task: asyncio.Task | None = None
-
-    # Used to check if the previous connect was mocked,
-    # if the next mock value differs then we fail
-    _previous_connect_was_mock = None
+    parent: DeviceBase | None = None
```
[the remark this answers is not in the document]
coretl: [Problem is that parent is set from the outside. Agree this needs more thought](https://github.com/bluesky/ophyd-async/issues/615)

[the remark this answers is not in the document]
evalott100: What about a getter and a setter? I'm alright with it being set from the outside, I just want it to be impossible (or as close as you can get in python) for it to be overwritten by subclass definitions.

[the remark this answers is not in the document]
coretl: that sounds reasonable, let's handle it as part of that ticket

src/ophyd_async/core/_utils.py
coretl: Updated the error message

src/ophyd_async/tango/base_devices/_base_device.py
burkeds: I would say that Signal should throw up the error. The proxy knows what type it should be, so it's the signal configuration which is the issue. I think that currently, the type mismatch is checked in `get_trl_descriptor`.

burkeds:
```python
import numpy as np
# old
import numpy.typing as npt
x = epics_signal_rw(npt.NDArray[np.int32], "PV")
# new
from ophyd_async.core import Array1D
x = epics_signal_rw(Array1D[np.int32], "PV")
```

I suspect this will break type enforcement on the tango backend. PyTango will read in non-scalar values as numpy arrays. This change will necessitate a rewrite of `get_trl_descriptor`. I really don't like this method. I think I want to refactor it and put its type enforcement elsewhere. What do you think?

src/ophyd_async/core/_device.py
```diff
@@ -60,23 +75,25 @@ def set_name(self, name: str):
...
 
-        self._name = name
-        for attr_name, child in self.children():
-            child_name = f"{name}-{attr_name.rstrip('_')}" if name else ""
-            child.set_name(child_name)
-            child.parent = self
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name != "parent" and isinstance(value, Device):
+            _setup_child(self, name, value)
+        return super().__setattr__(name, value)
```
evalott100, line 89: With this, a device can be an attribute in multiple parents, but it'll get it's name and `parent` field from whichever parent happens to `setattr` last? We should maybe outlaw this? (The below suggestion wouldn't work alongside the above review comment).

```suggestion
    def __setattr__(self, name: str, value: Any) -> None:
        if name == "parent" and getattr(self, "parent", None) is not None:
            raise TypeError("Can't overwrite a device's parent after it's been assigned.")
        elif isinstance(value, Device):
            _setup_child(self, name, value)
        return super().__setattr__(name, value)
```

The side effect would be that you couldn't

``` python
def __init__(self: Device, other_sub_device: Device):
    self.other_sub_device = other_sub_Device
```

where `other_sub_device` is a child in another tree.

src/ophyd_async/core/_readable.py
```diff
@@ -150,19 +150,19 @@ def add_children_as_readables(
         :meth:`HintedSignal.uncached`
         """
 
-        dict_copy = self.__dict__.copy()
+        dict_copy = dict(self.children())
```
evalott100, line 153: May as well only take the keys we need
```suggestion
        keys_before = set(self.children().keys())
```
...

src/ophyd_async/core/_readable.py
```diff
@@ -150,19 +150,19 @@ def add_children_as_readables(
...
 
         yield
 
         # Set symmetric difference operator gives all newly added keys
-        new_keys = dict_copy.keys() ^ self.__dict__.keys()
-        new_values = [self.__dict__[key] for key in new_keys]
+        new_dict = dict(self.children())
+        new_keys = dict_copy.keys() ^ new_dict.keys()
+        new_values = [new_dict[key] for key in new_keys]
```
evalott100, line 160: ...

```suggestion
        new_dict = dict(self.children())
        new_keys = keys_before ^ new_dict.keys()
        new_values = [new_dict[key] for key in new_keys]
```

src/ophyd_async/core/_signal.py
```diff
@@ -132,30 +109,29 @@ def close(self):
...
             f"Updated subscription: reading of source {self._signal.source} changed"
             f"from {self._reading} to {reading}"
         )
         self._reading = reading
-        self._value = value
         self._valid.set()
         for function, want_value in self._listeners.items():
             self._notify(function, want_value)
 
     def _notify(self, function: Callback, want_value: bool):
```
evalott100, line 131: We should maybe type hint `Callback` more precisely here

``` python
Callback[dict[str, Reading[SignalDatatypeT]] | SignalDatatypeT]
```

evalott100 (changes requested):
Will examine further tomorrow.

src/ophyd_async/core/_device.py
evalott100: Alternatively if we don't want to deal with this now, we could just change the above to only `_setup_child` if it hasn't been done already, so you could still do
``` python
def __init__(self: Device, other_sub_device: Device):
    self.other_sub_device = other_sub_Device
```
But `other_sub_device` will keep its original parent and name.

src/ophyd_async/core/_readable.py
coretl: `self.children()` is not a dict yet, so we need to turn it into one first...

src/ophyd_async/core/_readable.py
coretl: So I think it's clearer as it is

src/ophyd_async/tango/base_devices/_base_device.py
coretl: I think we need a followup PR to look at the Tango tests, let's get this one merged first though

rtuck99:
Seems like `assert_configuration()` (and presumably also `assert_reading()`) are broken for numpy array types. I guess they probably were already broken before these changes, but one of the tests in dodal uses assert_configuration and I encountered this as I'm currently in the process of going through dodal and mx-bluesky to test out this PR.
I will keep you posted if I find any other issues.

Tom-Willemsen:
Just tried this branch on our devices at ISIS, I'm getting a timeout when doing `trigger` on a device, despite explicitly specifying no timeout.

Problematic line is something like:
```
await dae.controls.end_run.trigger(wait=True, timeout=None)
```

where the relevant signal will give an EPICS completion callback eventually, but can be quite slow (~30s). My understanding was that specifying `timeout=None` should explicitly say "no timeout", and this seemed to work previously, but on this branch I'm seeing a timeout error. Full stack trace listed in [this PR](https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/41).

src/ophyd_async/core/_device.py
```diff
@@ -60,23 +75,31 @@ def set_name(self, name: str):
...
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "parent":
+            if self.parent not in (value, None):
+                raise TypeError(
+                    f"Cannot set the parent of {self} to be {value}: "
+                    f"it is already a child of {self.parent}"
+                )
+        elif isinstance(value, Device):
+            _setup_child(self, name, value)
```
rtuck99, line 94: This now breaks the use of the name parameter in signal initialiser because lines like
```
            self.test_signal = soft_signal_rw(int, 0, test_signal_name)
```
result in the desired signal name getting applied to the signal during construction, but then it is clobbered as soon as the assignment happens.

coretl: Good to know, will try and work out a way to restore the old behaviour and add a test for it.

coretl: Added in [be6d103](https://github.com/bluesky/ophyd-async/pull/594/commits/be6d10326f921d7b80fe7c51fcd6f4b7ea93dab7)

src/ophyd_async/core/_device.py
```diff
@@ -60,23 +69,32 @@ def set_name(self, name: str):
...
+        for child_name, child in self.children():
+            child_name = f"{self.name}-{child_name.rstrip('_')}" if self.name else ""
             child.set_name(child_name)
-            child.parent = self
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "parent":
+            if self.parent not in (value, None):
+                raise TypeError(
+                    f"Cannot set the parent of {self} to be {value}: "
```
rtuck99, line 85: This is now breaking the Thawer as the ThawingTimer has a reference to the control signal of the parent Thawer Device.

Can we please not make assumptions about what subclasses may be doing with their own attributes? Or at the very least, allow them to have unmolested underscore private attribs?

coretl: Done in e2df3248ea28619afc820080b12c755b5860f96c

src/ophyd_async/core/_table.py
```diff
@@ -17,34 +22,31 @@ def _concat(value1, value2):
...
-                )
-            default_array = field_value.default_factory()
-            if isinstance(default_array, np.ndarray):
-                arrayified_kwargs[field_name] = np.array(
-                    [value], dtype=default_array.dtype
-                )
-            elif issubclass(type(value), Enum) and isinstance(value, str):
-                arrayified_kwargs[field_name] = [value]
+    # Allow extra so we can use this as a generic Table
+    model_config = ConfigDict(extra="allow")
```
callumforrester, line 26: I'm not sure how this works, given the different config in `__init_subclass__`, this config only applies to fields of the `Table` base class, and there aren't any...?

src/ophyd_async/core/_table.py
```diff
@@ -64,83 +66,66 @@ def __add__(self, right: TableSubclass) -> TableSubclass:
...
+        for k, v in self:
+            if selection:
+                v = v[selection]
+            if array is None:
+                array = np.empty(v.shape, dtype=self.numpy_dtype())
+            array[k] = v
+        assert array is not None
+        return array
+
+    # TODO: would like to do this validation, but too strict for CA which returns
```
callumforrester, line 94: Should: Either do TODO or delete and make an issue

src/ophyd_async/core/_utils.py
```diff
@@ -2,25 +2,35 @@
...
-ReadingValueCallback = Callable[[Reading, T], None]
 DEFAULT_TIMEOUT = 10.0
 ErrorText = str | dict[str, Exception]
 
 
+class StrictEnum(str, Enum):
+    """All members should exist in the Backend, and there will be no extras"""
+
+
+class SubsetEnumType(EnumMeta):
```
callumforrester, line 23: Should: Rename to `SubsetEnumMeta` for consistency

docs/explanations/decisions/0008-signal-types.md
callumforrester, line 1: Should: I like this and the how-to, but I think there should also be an explanation page about devices, signals and backends are now structured and built. I had to read through most of `core` to figure it out and it would be useful to document to onboard new people.

callumforrester (changes requested):
This is looking really good! I like the simplified structure and not embedding the magic too deep inside the device tree. Net negative LOC is also never a bad thing!

I've left a few minor comments, I would also appreciate it if there is a plan (in writing!) for testing this on P45 soon before/after release. Maybe with a release candidate?

src/ophyd_async/fastcs/panda/_block.py
```diff
@@ -32,26 +35,28 @@ class PulseBlock(Device):
...
+class PcompDirection(StrictEnum):
     positive = "Positive"
     negative = "Negative"
     either = "Either"
 
 
-EnableDisableOptions = SubsetEnum["ZERO", "ONE"]
+class BitMux(SubsetEnum):
+    zero = "ZERO"
+    one = "ONE"
```
evalott100, line 46:
```suggestion
    ZERO = "ZERO"
    ONE = "ONE"
```

evalott100 (changes requested):

src/ophyd_async/core/_device_filler.py
```diff
@@ -0,0 +1,183 @@
...
+SignalBackendT = TypeVar("SignalBackendT", bound=SignalBackend)
+DeviceConnectorT = TypeVar("DeviceConnectorT", bound=DeviceConnector)
+
+
+class DeviceFiller(Generic[SignalBackendT, DeviceConnectorT]):
+    def __init__(
+        self,
+        device: Device,
+        signal_backend_type: type[SignalBackendT],
+        device_connector_type: type[DeviceConnectorT],
```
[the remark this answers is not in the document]
coretl: But then we lose the type inference for the return value of `make_child_device`...

src/ophyd_async/fastcs/panda/_hdf_panda.py
```diff
@@ -13,17 +13,15 @@
 class HDFPanda(CommonPandaBlocks, StandardDetector):
     def __init__(
         self,
-        prefix: str,
+        uri: str,
```
rtuck99, line 16: Is it possible to please retain prefix so that I don't have to modify `device_instantiation()` in dodal to explicitly not send it?

> Robert Tuck
  Today at 10:17 AM
Thanks. After some further investigation, it looks like it will lead to some ugliness in dodal's device_instantiation because the panda doesn't accept a prefix parameter, whereas every single other dodal device does, so I have to add code to explicitly not send prefix. Can it be made an optional keyword parameter with backwards compatibility?
New


> Tom Cobb
>   [4 minutes ago](https://diamondlightsource.slack.com/archives/C04RRAM68LD/p1729681756673319?thread_ts=1729600610.523609&cid=C04RRAM68LD)
> I can put it back to prefix , and change it back to uri when device_instantiation has gone and device_factory is used instead

src/ophyd_async/core/_device.py
rtuck99: I may be having buyer's remorse on this but would like to solicit feedback from users of ophyd_async before asking to revert this change:

Previously, we had control over whether something was parented or not by deciding when to call the `super()._init__()` method, if the signal was present it got parented otherwise not. From conversation with @coretl apparently this was unintended behaviour but it seems we have been using it in dodal anyway.

With the change to parent on setting the attribute we lost control and all devices are parented - which resulted in the above change as it broke the Thawer as it has a device which has a private reference to a signal. However we also have some devices which are parented but had the signals private (underscore prefixed) because they aren't interesting to users of the device, and are only implementation, like the attenuator and the mirror voltages.

To some extent though private signals are still very visible anyway as they are exposed via events etc. so I don't know whether making them public is that big of a deal or not.

I think if this change were to be reverted though, we would ideally need some other explicit way of being able to control whether a signal is a child or not.

stan-dot: > . From conversation with @coretl apparently this was unintended behaviour but it seems we have been using it in dodal anyway.

https://www.hyrumslaw.com/ strikes again

Tom-Willemsen: Only just catching up on some of this, but as a simple "user" of ophyd-async, I agree it's surprising if `self.foo = some_signal(...)` and `self._foo = some_signal(...)` have different behaviour. That feels very magical & wrong.

Is there a more explicit API we could use?

```python
with self.dont_overwrite_parent():
    self.foo = some_signal_reference
```
?

callumforrester: The other option is if we can find a way to make all this work with immutable attributes

src/ophyd_async/core/_device.py
```diff
@@ -1,55 +1,68 @@
...
 from bluesky.protocols import HasName
 from bluesky.run_engine import call_in_bluesky_event_loop
 
+from ._protocol import Connectable
 from ._utils import DEFAULT_TIMEOUT, NotConnected, wait_for_connection
 
 
-class Device(HasName):
-    """Common base class for all Ophyd Async Devices.
+class DeviceConnector:
```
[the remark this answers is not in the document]
stan-dot: but shouldn't then *all* devices have a 'Connector' suffix? all of ophyd is `Connector` abstraction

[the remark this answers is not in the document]
callumforrester: Not sure I understand, `Device` and `DeviceConnector` are two different things. A `DeviceConnector` is a thing that connects a `Device`, it's separation of concerns.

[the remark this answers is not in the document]
stan-dot: nvm I misread the git diff

src/ophyd_async/core/_device.py
coretl: Discussed this with @rtuck99 and @callumforrester and decided that:
- Having different parent-child-naming behaviour for `self._foo` and `self.foo` is indeed surprising behaviour, so will revert the change so they behave the same
- To solve the `ThawerTimer` case we will provide a new box-like class, that is simply a container for a Device. This means it can be a public or private member of the class, but not take part in the parent-child-naming behaviour of Device, like we do with `DetectorWriter` and other logic classes:
```python
class Box(Generic[T]):
    def __init__(self, obj: T): ...
    def get(self) -> T: ...

class DeviceWithRefToSignal(Device):
    def __init__(self, signal: SignalRW[int[):
        self.signal_box = Box(signal)

    def set(self, value) -> AsyncStatus:
        signal = self.signal_box.get()  # typing knows that signal is a SignalRW[int]
        return signal.set(value + 1)
```

The questions:
- If `parent.name == "foo"`, and `parent._bar` is a `Device`, should `parent._bar.name` be "parent-bar" or "parent-_bar"?
- What should `Box` be called?

coretl: I could model it after a [`weakref.proxy`](https://docs.python.org/3/library/weakref.html#weakref.proxy):
```python
class Reference(Generic[T]):
    def __init__(self, obj: T):
        self._obj = obj
    def __call__(self) -> T:
        return self._obj

class DeviceWithRefToSignal(Device):
    def __init__(self, signal: SignalRW[int[):
        self.signal_ref = Reference(signal)

    def set(self, value) -> AsyncStatus:
        return self.signal_ref().set(value + 1)
```
@Tom-Willemsen any better?

Tom-Willemsen: That seeems fine to me at first glance - assuming the docstring / docs in general will explain why one might need to use it. Thanks!

src/ophyd_async/fastcs/panda/_hdf_panda.py
callumforrester: FWIW we're ultimately going to replace `device_instantiation` with `DeviceInitialisationFactory`.

docs/explanations/decisions/0008-signal-types.md
coretl: Docs is my next month's work...

https://github.com/bluesky/ophyd-async/issues/622

src/ophyd_async/core/_device_filler.py
```diff
@@ -0,0 +1,183 @@
...
+            if name == "parent" or name.startswith("_") or not origin:
+                # Ignore
+                pass
+            elif issubclass(origin, Signal):
+                # SignalX doesn't need datatype, all others need one
+                datatype = self.get_datatype(name)
+                if origin != SignalX and datatype is None:
+                    self._raise(
+                        name,
+                        f"Expected SignalX or SignalR/W/RW[type], got {annotation}",
```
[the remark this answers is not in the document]
coretl: I found myself falling down this hole a couple of times, so that's what `get_origin_class` above does

src/ophyd_async/core/_mock_signal_backend.py
```diff
@@ -5,46 +5,37 @@
...
-
-        self.datatype = datatype
-
-        if not isinstance(self.initial_backend, SoftSignalBackend):
-            # If the backend is a hard signal backend, or not provided,
-            # then we create a soft signal to mimic it
-
-            self.soft_backend = SoftSignalBackend(datatype=datatype)
-        else:
+        if isinstance(self.initial_backend, SoftSignalBackend):
```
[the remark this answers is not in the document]
coretl: Done above

src/ophyd_async/core/_signal.py
```diff
@@ -2,128 +2,105 @@
...
+        self,
+        device: Device,
+        mock: bool,
+        timeout: float,
+        force_reconnect: bool,
+    ):
+        if mock:
+            self.backend = MockSignalBackend(self._init_backend)
+        else:
+            self.backend = self._init_backend
```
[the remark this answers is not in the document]
coretl: I think we need this as we can switch from mock back to real

src/ophyd_async/core/_signal.py
```diff
@@ -132,30 +109,33 @@ def close(self):
...
         assert self._reading is not None, "Monitor not working"
         return self._reading
 
-    async def get_value(self) -> T:
-        await self._valid.wait()
-        assert self._value is not None, "Monitor not working"
-        return self._value
+    async def get_value(self) -> SignalDatatypeT:
+        reading = await self.get_reading()
+        return reading["value"]
```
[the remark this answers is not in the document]
coretl: We either extract from a dict in `_callback`, or here, so the difference in performance is negligible and dependant on whether the signal updates more, or if you call get_value more...

src/ophyd_async/core/_signal_backend.py
```diff
@@ -1,97 +1,163 @@
...
-from ._utils import DEFAULT_TIMEOUT, ReadingValueCallback, T
+from event_model import DataKey, Dtype, Limits
+
+from ._table import Table
+from ._utils import Callback, StrictEnum, T
+
+DTypeScalar_co = TypeVar("DTypeScalar_co", covariant=True, bound=np.generic)
+Array1D = np.ndarray[tuple[int], np.dtype[DTypeScalar_co]]
+Primitive = bool | int | float | str
+SignalDatatype = (
```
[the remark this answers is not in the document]
coretl: Not worth the automation, but added a note to clarify connection

src/ophyd_async/core/_signal_backend.py
```diff
@@ -1,97 +1,163 @@
...
+def make_datakey(
+    datatype: type[SignalDatatypeT],
+    value: SignalDatatypeT,
+    source: str,
+    metadata: SignalMetadata,
+) -> DataKey:
+    dtn = _datakey_dtype_numpy(datatype, value)
+    return DataKey(
+        dtype=_datakey_dtype(datatype),
+        shape=_datakey_shape(value),
```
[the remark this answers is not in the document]
coretl: We are saying that `dtype_numpy` expresses a single row, so shape should be `(num_rows,)`. We did this because the columns can all be of different types, so can't be expressed as a 2D structure.

src/ophyd_async/core/_table.py
coretl: You can use `Table` in 2 ways:
1. `Table(**whatever_pva_gives_us)` when pvi adds a Signal to a Device that is not type hinted
2. `MyTable(**whatever_pva_gives_us)` where the Signal is type hinted

For 1 we want extra="allow" so it is passed through as is. There are no base class fields, only "extra" fields, so they must be allowed. For 2 we want extra="forbid" so it is strictly checked against the BaseModel we are supplied.

src/ophyd_async/core/_table.py
```diff
@@ -17,34 +22,31 @@ def _concat(value1, value2):
...
+        for k, anno in cls.__annotations__.items():
+            if get_origin(anno) is np.ndarray:
+                dtype = get_dtype(anno)
+                new_anno = Annotated[
+                    anno,
+                    NpArrayPydanticAnnotation.factory(
+                        data_type=dtype.type, dimensions=1, strict_data_typing=False
+                    ),
+                    Field(
+                        default_factory=lambda dtype=dtype: np.array([], dtype=dtype)
```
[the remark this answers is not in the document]
coretl: Because it is created in a for loop, so needs to be bound to the function or it will disappear:
https://docs.astral.sh/ruff/rules/function-uses-loop-variable

src/ophyd_async/fastcs/panda/_block.py
coretl: Will do this in https://github.com/bluesky/ophyd-async/issues/605

coretl:
@evalott100 just waiting on an event-model release (which I think we should put https://github.com/bluesky/event-model/issues/327 into) then this is ready to merge I think...

src/ophyd_async/core/_device.py
DominicOram: Sorry, fractionally too late to the party but only minor. Can we add a mention of `Reference` to this error message so that people know it is possible, they just have to do it in a different way?

PR #598, opened 2024-10-01
Declarative EPICS and StandardReadable Devices

coretl:
Add optional Declarative Device support

Includes:
- An ADR for optional Declarative Devices
- Support for `StandardReadable` Declarative Devices via `StandardReadableFormat` annotations
- Support for EPICS Declarative Devices via an `EpicsDevice` baseclass  and `PvSuffic` annotations
- Updates to the EPICS and Tango demo devices to use them

# Changes

## pvi structure changes
Structure read from `.value` now includes `DeviceVector` support. Requires at least PandABlocks-ioc 0.11.2

## Epics `signal` module moves
`ophyd_async.epics.signal` moves to `ophyd_async.epics.core` with a backwards compat module that emits deprecation warning.
```python
# old
from ophyd_async.epics.signal import epics_signal_rw
# new
from ophyd_async.epics.core import epics_signal_rw
```

## `StandardReadable` wrappers change to `StandardReadableFormat`
`StandardReadable` wrappers change to enum members of `StandardReadableFormat` (normally imported as `Format`)
```python
# old
from ophyd_async.core import ConfigSignal, HintedSignal
    def __init__(self):
        self.add_readables([sig1], ConfigSignal)
        self.add_readables([sig2], HintedSignal)
        self.add_readables([sig3], HintedSignal.uncached)
# new
from ophyd_async.core import StandardReadableFormat as Format
    def __init__(self):
        self.add_readables([sig1], Format.CONFIG_SIGNAL)
        self.add_readables([sig2], Format.HINTED_SIGNAL)
        self.add_readables([sig3], Format.HINTED_UNCACHED_SIGNAL
```

## Declarative Devices are now available
```python
# old
from ophyd_async.core import ConfigSignal, HintedSignal
from ophyd_async.epics.signal import epics_signal_r, epics_signal_rw

class Sensor(StandardReadable):
    def __init__(self, prefix: str, name="") -> None:
        with self.add_children_as_readables(HintedSignal):
            self.value = epics_signal_r(float, prefix + "Value")
        with self.add_children_as_readables(ConfigSignal):
            self.mode = epics_signal_rw(EnergyMode, prefix + "Mode")
        super().__init__(name=name)
# new
from typing import Annotated as A
from ophyd_async.core import StandardReadableFormat as Format
from ophyd_async.epics.core import EpicsDevice, PvSuffix, epics_signal_r, epics_signal_rw

class Sensor(StandardReadable, EpicsDevice):
    value: A[SignalR[float], PvSuffix("Value"), Format.HINTED_SIGNAL]
    mode: A[SignalRW[EnergyMode], PvSuffix("Mode"), Format.CONFIG_SIGNAL]
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,13 @@ addopts = """
     --doctest-glob="*.rst" --doctest-glob="*.md" --ignore=docs/examples
     """
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
-filterwarnings = "error"
+filterwarnings = [
+    "error",
+    "ignore:Use `ophyd_async.epics.core` instead of `ophyd_async.epics.signal` and `pvi`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.CONFIG_SIGNAL` instead of `ConfigSignal`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.HINTED_SIGNAL` instead of `HintedSignal`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.HINTED_UNCACHED_SIGNAL` instead of `HintedSignal.uncached`:DeprecationWarning",
+]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
 log_format = "%(asctime)s,%(msecs)03d %(levelname)s (%(threadName)s) %(message)s"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,16 +94,11 @@ reportMissingImports = false  # Ignore missing stubs in imported modules
 # Run pytest with all our checkers, and don't spam us with massive tracebacks on error
 addopts = """
     --tb=native -vv --strict-markers --doctest-modules
-    --doctest-glob="*.rst" --doctest-glob="*.md" --ignore=docs/examples
+    --doctest-glob="*.rst" --doctest-glob="*.md"
+    --ignore=docs/examples --ignore=src/ophyd_async/epics/signal.py
     """
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
-filterwarnings = [
-    "error",
-    "ignore:Use `ophyd_async.epics.core` instead of `ophyd_async.epics.signal` and `pvi`:DeprecationWarning",
-    "ignore:Use `StandardReadableFormat.CONFIG_SIGNAL` instead of `ConfigSignal`:DeprecationWarning",
-    "ignore:Use `StandardReadableFormat.HINTED_SIGNAL` instead of `HintedSignal`:DeprecationWarning",
-    "ignore:Use `StandardReadableFormat.HINTED_UNCACHED_SIGNAL` instead of `HintedSignal.uncached`:DeprecationWarning",
-]
+filterwarnings = "error"
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
 log_format = "%(asctime)s,%(msecs)03d %(levelname)s (%(threadName)s) %(message)s"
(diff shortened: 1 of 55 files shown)


Tom-Willemsen:
Some thoughts from office conversations at ISIS (@rerpha):
- We feel the procedural option is nicer / more supportable / less magical (basically in agreement with ADR 6)
- In most of our cases we probably _could_ write our devices using the declarative approach, but in a few places it would probably get a bit messy and much harder to read than the declarative approach
  * Slightly worried about loss of flexibility / explicitness of the declarative approach
- How would a type-generic device be written using the declarative approach? If I have a `Generic[T]` device would the `T` typevar be a valid datatype to pass to the declarative signal datatype? I'm a bit worried about getting tied in `typing` knots here and it's all very not-obvious I feel, whereas the explicit approach makes it quite obvious...

Would appreciate being involved in the discussion if it looks like we might be going towards declarative route.

coretl:
One key point that came out of the discussion is that we should always be free to drop to procedural. We should always be able to have an `__init__` method that does `epics_signal_rw` calls. This should be used for any of the `DeviceVector` cases, and any tricky generics (although I think they could mostly be made to work in the declarative approach). This means we should never be writing a formatted string in the declarative approach, if we need it we should drop to an `__init__` method.

The question here is should we *support* the declarative approach for a subset of the EPICS devices. I think the large number of areaDetector drivers and plugins we are writing are good use case for this, they have no logic, only interface, and are more naturally written in the declarative style. This would allow us to teach the declarative style first, then introduce the procedural style when we need the more complicated examples.

Happy to discuss this in more detail on a zoom call

Tom-Willemsen:
@coretl and I discussed on zoom but will add the result of the discussion here:

> One key point that came out of the discussion is that we should always be free to drop to procedural.

This is fine from our perspective then - I was worried that the annotation style would be enforced, as long as procedural is still a supported option this is fine for us.

DominicOram:
I echo @Tom-Willemsen's thoughts. Prefer the more explicit procedural approach but happy if `ophyd_async` supports both. Could I suggest we add a line in the ADR that explicitly says something like "This is not a step towards only supporting the declarative approach and there are no plans to drop the procedural approach"

evalott100 (approved):
I'll give a tentative approval, but I think we should have a review from someone from a dodal type layer.

pyproject.toml
```diff
@@ -97,7 +97,13 @@ addopts = """
...
     """
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
-filterwarnings = "error"
+filterwarnings = [
+    "error",
+    "ignore:Use `ophyd_async.epics.core` instead of `ophyd_async.epics.signal` and `pvi`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.CONFIG_SIGNAL` instead of `ConfigSignal`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.HINTED_SIGNAL` instead of `HintedSignal`:DeprecationWarning",
+    "ignore:Use `StandardReadableFormat.HINTED_UNCACHED_SIGNAL` instead of `HintedSignal.uncached`:DeprecationWarning",
+]
```
rtuck99, line 106: I think ophyd-async should adopt stricter same standards for ignoring warnings in tests, since tests in dodal and bluesky currently run with strict warnings-are-errors.
If ophyd-async code raises deprecation warnings then it means we have to add similar ignores to our own tests which negates the purpose of adding the deprecation warnings in the first place.

src/ophyd_async/epics/motor.py
```diff
@@ -21,7 +21,7 @@
     WatcherUpdate,
     observe_value,
 )
-from ophyd_async.epics.signal import epics_signal_r, epics_signal_rw, epics_signal_x
+from ophyd_async.epics.core import epics_signal_r, epics_signal_rw, epics_signal_x
```
rtuck99, line 24: Please fix deprecated use of ConfigSignal in motor.py

pyproject.toml
coretl: Let me see if we can localise those checks to a single test...

rtuck99 (approved):
Deprecation warnings fixed, approved

Issue #336, opened 2024-05-24
Add a shared parent mock for all signals in once device


DominicOram:
As a developer I occasionally want to make sure that one signal is changed before another. For example, see `test_retrieve_mock_and_assert` in https://github.com/bluesky/ophyd-async/pull/335/files or for a more real world example see numerous tests in https://github.com/DiamondLightSource/dodal/blob/main/tests/devices/unit_tests/test_aperture_scatterguard.py. To do this you need to create a parent mock that all the signals are children of.

If `ophyd_async` by default made all the mock's in one device have the same parent this would remove a bit of boilerplate from downstream.

coretl:
A suggestion for implementation:
- Change `Device.connect` to take `mock: Union[bool, Mock] = False`
- If `isinstance(mock, Mock)` then use it as the mock for the `Device`, otherwise create a new one
- When calling `child.connect`, then call `mock.attach_mock(child_name, child_mock)` and pass `mock=child_mock` to the child

This will mean that if you have a tree of devices where connect it called at the top level then you can call:
```python
get_mock_put(mydevice).assert_has_calls(
        [
            call.child1.grandchild1(100, wait=True, timeout=ANY),
            call.child2.grandchild2(67, wait=True, timeout=ANY),
        ]
    )
```
I'm a little concerned that we might decide we want to check the ordering of puts and other methods like `get_descriptor`, so maybe we should consider this before making a mock heirarchy we rely upon?

For instance, would we ever want something like this?
```python
get_mock(mydevice).assert_has_calls(
        [
            call.child1.grandchild1.put(100, wait=True, timeout=ANY),
            call.child1.grandchild1.get_descriptor(),
            call.child2.grandchild2.put(67, wait=True, timeout=ANY),
        ]
    )
```

@DominicOram thoughts?

coretl:
@DominicOram any thoughts on this?

coretl:
@dperl-dls any thoughts on this?

DominicOram:
>A suggestion for implementation:

Yes, I like this idea.

> I'm a little concerned that we might

This feels like feature creep at this point. I can't see a reason we would want to do this yet and it doesn't seem like the way we structure this hierarchy will have that much of an affect on how we want to do it in the future.

coretl:
> This feels like feature creep at this point. I can't see a reason we would want to do this yet and it doesn't seem like the way we structure this hierarchy will have that much of an affect on how we want to do it in the future.

[...]

PR #599, opened 2024-10-04
Allow shared parent mock to be passed to Device.connect

jsouter:
Addresses https://github.com/bluesky/ophyd-async/issues/336

Allows a Device to accept a unittest.Mock or a boolean.
If True passed, creates a new Mock()
if Mock() passed, finds any child mocks set with attach_mock if they exist, else passes down mock=True to children.

Will push some tests for this

edit: now realising that I misunderstood the requirements, will fix quickly!

### DIFF
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -13,6 +13,8 @@ from bluesky.run_engine import call_in_bluesky_event_loop, in_bluesky_event_loop
 from ._protocol import Connectable
 from ._utils import DEFAULT_TIMEOUT, NotConnected, wait_for_connection
 
+_device_mocks: dict[Device, Mock] = {}
+
 
 class DeviceConnector:
     """Defines how a `Device` should be connected and type hints processed."""
@@ -73,7 +75,6 @@ class Device(HasName, Connectable):
         self, name: str = "", connector: DeviceConnector | None = None
     ) -> None:
         self._connector = connector or DeviceConnector()
-        self.mock = None
         self.set_name(name)
 
     @property
@@ -131,16 +132,18 @@ class Device(HasName, Connectable):
         timeout:
             Time to wait before failing with a TimeoutError.
         """
+        uses_mock = bool(mock)
         can_use_previous_connect = (
-            bool(mock) is self._connect_mock_arg
+            uses_mock is self._connect_mock_arg
             and self._connect_task
             and not (self._connect_task.done() and self._connect_task.exception())
         )
         if mock is True:
             mock = Mock()  # create a new Mock if one not provided
-        self.mock = mock
         if force_reconnect or not can_use_previous_connect:
-            self._connect_mock_arg = bool(mock)
+            self._connect_mock_arg = uses_mock
+            if self._connect_mock_arg:
+                _device_mocks[self] = mock
             coro = self._connector.connect(
                 device=self, mock=mock, timeout=timeout, force_reconnect=force_reconnect
             )
@@ -202,6 +205,9 @@ class DeviceVector(MutableMapping[int, DeviceT], Device):
     def children(self) -> Iterator[tuple[str, Device]]:
         for key, child in self._children.items():
             yield str(key), child
+
+    def __hash__(self):  # to allow DeviceVector to be used as dict keys and in sets
+        return hash(id(self))
 
 
 class DeviceCollector:
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -5,6 +5,7 @@ import sys
 from collections.abc import Coroutine, Iterator, Mapping, MutableMapping
 from logging import LoggerAdapter, getLogger
 from typing import Any, TypeVar
+from unittest.mock import Mock
 
 from bluesky.protocols import HasName
 from bluesky.run_engine import call_in_bluesky_event_loop, in_bluesky_event_loop
@@ -37,7 +38,7 @@ class DeviceConnector:
     async def connect(
         self,
         device: Device,
-        mock: bool,
+        mock: bool | Mock,
         timeout: float,
         force_reconnect: bool,
     ):
@@ -47,12 +48,12 @@ class DeviceConnector:
         done in a different mock more. It should connect the Device and all its
         children.
         """
-        coros = {
-            name: child_device.connect(
-                mock=mock, timeout=timeout, force_reconnect=force_reconnect
+        coros = {}
+        for name, child_device in device.children():
+            child_mock = getattr(mock, name) if mock else mock  # Mock() or False
+            coros[name] = child_device.connect(
+                mock=child_mock, timeout=timeout, force_reconnect=force_reconnect
             )
-            for name, child_device in device.children()
-        }
         await wait_for_connection(**coros)
 
 
@@ -72,6 +73,7 @@ class Device(HasName, Connectable):
         self, name: str = "", connector: DeviceConnector | None = None
     ) -> None:
         self._connector = connector or DeviceConnector()
+        self.mock = None
         self.set_name(name)
 
     @property
@@ -114,7 +116,7 @@ class Device(HasName, Connectable):
 
     async def connect(
         self,
-        mock: bool = False,
+        mock: bool | Mock = False,
         timeout: float = DEFAULT_TIMEOUT,
         force_reconnect: bool = False,
     ) -> None:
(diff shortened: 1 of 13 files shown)


DominicOram (approved):
I think this looks about right. As you say a bunch of tests confirming all the use cases would be great

jsouter:
Would have to call attach_mock after each child device is connected as that is when the signal backend gets set to MockSignalBackend and the _backend.put_mock can be accessed. Tom has said he may touch how that is handled soon so I will wait for that to be done first.

coretl:
#594 is sufficiently advanced that you can work on top of it to see if you can apply this on top of it

src/ophyd_async/core/_device.py
```diff
@@ -89,29 +90,38 @@ async def connect(
...
             # Kick off a connection
-            coros = {
-                name: child_device.connect(
-                    mock, timeout=timeout, force_reconnect=force_reconnect
+            coros = {}
+
+            for child_name, child_device in self.children():
+                # if mock is Mock instance, get child Mock if exists
+                # else pass down boolean
+                child_mock = getattr(mock, child_name, True) if mock else mock
```
coretl, line 120: mock should always be a `Mock` here by now
```suggestion
                child_mock = getattr(mock, child_name) if mock else mock
```

jsouter:
I've rebased on top of https://github.com/bluesky/ophyd-async/pull/594, still a WIP but it appears to work with the existing tests. I will work on porting over the new tests I added in the force-pushed-over commits here

src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -1,8 +1,9 @@
 from collections.abc import Awaitable, Callable, Iterable
 from contextlib import asynccontextmanager, contextmanager
 from typing import Any
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, Mock
 
+from ._device import Device
```
jsouter, line 6: I think this is mostly in a sensible state now, depending on if we're happy with get_mock in _mock_signal_utils accepting Device | Signal

jsouter:
There's a few places where the signature for connect is still `mock: bool`: in the Connectable Protocol, PviDeviceConnector and TangoDeviceConnector etc, I assume we want to make it uniform?

src/ophyd_async/core/_device.py
```diff
@@ -106,12 +108,15 @@ async def connect(
...
         """
         can_use_previous_connect = (
-            mock is self._connect_mock_arg
+            bool(mock) is self._connect_mock_arg
             and self._connect_task
             and not (self._connect_task.done() and self._connect_task.exception())
         )
+        if mock is True:
+            mock = Mock()  # create a new Mock if one not provided
+        self.mock = mock
```
coretl, line 117: Hmm, having seen this I'm inclined to go back to having a global to store all the mocks and MockSignalBackends in, like we had ages ago:
https://github.com/bluesky/ophyd-async/blob/54deda30cae717472d1020bd5a67489380960ad8/src/ophyd_async/core/signal.py#L24

src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -47,6 +48,12 @@ def get_mock_put(signal: Signal) -> AsyncMock:
     return _get_mock_signal_backend(signal).put_mock
 
 
+def get_mock(device: Device | Signal) -> Mock:
+    if isinstance(device, Signal):
+        return _get_mock_signal_backend(device).mock
+    return device.mock
```
coretl, line 54: Then this would be:
```python
    return _device_mocks[device]
```

src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -47,6 +48,12 @@ def get_mock_put(signal: Signal) -> AsyncMock:
...
+
+
 def reset_mock_put_calls(signal: Signal):
     backend = _get_mock_signal_backend(signal)
```
coretl, line 58: And this would be:
```python
    return _mock_signal_backends[signal]
```

src/ophyd_async/core/_mock_signal_backend.py
```diff
@@ -13,7 +13,11 @@
 class MockSignalBackend(SignalBackend[SignalDatatypeT]):
     """Signal backend for testing, created by ``Device.connect(mock=True)``."""
 
-    def __init__(self, initial_backend: SignalBackend[SignalDatatypeT]) -> None:
+    def __init__(
+        self,
+        initial_backend: SignalBackend[SignalDatatypeT],
+        mock: bool | Mock = True,
```
coretl, line 19: I think we should create the Mock before it gets passed in:
```suggestion
        mock: Mock,
```

src/ophyd_async/core/_device.py
jsouter: Okay, do we want the mapping to be

- dict[Signal, MockSignalBackend]
- dict[MockSignalBackend, Mock]
- dict[Signal, Mock]

coretl:
```python
_mock_signal_backends: dict[Signal, MockSignalBackend]  # used for `set_mock_value`
_device_mocks: dict[Device, Mock]  # used for `get_mock`
```

src/ophyd_async/core/_device.py
```diff
@@ -121,6 +129,10 @@ async def connect(
         # Wait for it to complete
         await self._connect_task
 
+    def __hash__(self):
+        # to allow Devices to be used as dict keys
+        return hash(id(self))
+
```
coretl, line 135: I think Devices are already hashable...

src/ophyd_async/core/_device.py
```diff
@@ -175,6 +187,9 @@ def children(self) -> Iterator[tuple[str, Device]]:
         for key, child in self._children.items():
             yield str(key), child
 
+    def __hash__(self):  # to allow DeviceVector to be used as dict keys
+        return hash(id(self))
+
```
coretl, line 192: But you're right that we need this to make DeviceVector hashable

src/ophyd_async/core/_device.py
jsouter: You're right! Will remove that now

tests/epics/demo/test_demo.py
```diff
@@ -199,6 +199,28 @@ async def test_retrieve_mock_and_assert(mock_mover: demo.Mover):
     )
 
 
+async def test_mock_with_Mock():
```
DominicOram, line 202: Nit: This name is a bit too implementation focussed IMO, something like `test_mocks_in_device_share_parent` would be more descriptive of the usecase

src/ophyd_async/core/_mock_signal_backend.py
```diff
@@ -27,6 +31,11 @@ def __init__(self, initial_backend: SignalBackend[SignalDatatypeT]) -> None:
             self.soft_backend = SoftSignalBackend(
                 datatype=self.initial_backend.datatype
             )
+
+        # use existing Mock if provided
+        self.mock = mock
+        self.mock.attach_mock(AsyncMock(name="put", spec=Callable), "put")
```
coretl, line 37: Could:
```suggestion
        self.mock = mock
        self.put_mock = AsyncMock(name="put", spec=Callable)
        self.mock.attach_mock(put_mock, "put")
```

src/ophyd_async/core/_mock_signal_backend.py
```diff
@@ -40,7 +49,7 @@ async def connect(self, timeout: float) -> None:
 
     @cached_property
     def put_mock(self) -> AsyncMock:
-        return AsyncMock(name="put", spec=Callable)
+        return self.mock.put
```
coretl, line 52: Then you could delete this cached property...

src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -1,20 +1,19 @@
...
 def _get_mock_signal_backend(signal: Signal) -> MockSignalBackend:
-    backend = signal._connector.backend  # noqa:SLF001
-    assert isinstance(backend, MockSignalBackend), (
-        "Expected to receive a `MockSignalBackend`, instead "
-        f" received {type(backend)}. "
-    )
-    return backend
+    assert (
+        signal in _mock_signal_backends
+    ), "No `MockSignalBackend` registered for signal."
```
coretl, line 15:
```suggestion
    ), f"Signal {signal} not connected in mock mode"
```

tests/epics/demo/test_demo.py
```diff
@@ -199,6 +199,28 @@ async def test_retrieve_mock_and_assert(mock_mover: demo.Mover):
...
+
+    await mock_mover.velocity.set(100)
+    await mock_mover.setpoint.set(67)
+
+    mock.assert_has_calls(
+        [
+            call.velocity.put(100, wait=True),
+            call.setpoint.put(67, wait=True),
+        ]
+    )
```
coretl, line 221: would prefer to be explicit about the previous calls (assert_has_calls does a subset check):
```suggestion
    mock.reset_mock()
    await mock_mover.velocity.set(100)
    await mock_mover.setpoint.set(67)

    assert mock.call_args_list == [
        call.velocity.put(100, wait=True),
        call.setpoint.put(67, wait=True),
    ]
```

coretl (approved):

coretl:
Shall I merge?

PR #600, opened 2024-10-04
Add test for context race condition

DominicOram:
Adds test to cover https://github.com/DiamondLightSource/dodal/issues/791.

To test:
* On latest `aioca` confirm this test times out
* Run `pip install git+https://github.com/DiamondLightSource/aioca.git@contexts` and confirm it passes

### DIFF
--- a/tests/epics/signal/test_signals.py
+++ b/tests/epics/signal/test_signals.py
@@ -13,11 +13,14 @@ from types import GenericAlias
 from typing import Any, Dict, Literal, Optional, Sequence, Tuple, Type
 from unittest.mock import ANY
 
+import bluesky.plan_stubs as bps
 import numpy as np
 import numpy.typing as npt
 import pytest
 from aioca import CANothing, purge_channel_caches
 from bluesky.protocols import DataKey, Reading
+from bluesky.run_engine import RunEngine
+from ophyd.signal import EpicsSignal
 from typing_extensions import TypedDict
 
 from ophyd_async.core import (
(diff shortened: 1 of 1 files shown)


tests/epics/signal/test_signals.py
```diff
@@ -901,3 +904,21 @@ async def test_signals_created_for_not_prec_0_float_cannot_use_int(ioc: IOC):
...
+
+
+async def test_can_read_using_ophyd_async_then_ophyd(ioc: IOC):
+    oa_read = f"{ioc.protocol}://{PV_PREFIX}:{ioc.protocol}:float_prec_1"
+    ophyd_read = f"{PV_PREFIX}:{ioc.protocol}:float_prec_0"
+
+    ophyd_async_sig = epics_signal_rw(float, oa_read)
+    await ophyd_async_sig.connect()
+    ophyd_signal = EpicsSignal(ophyd_read)
+    ophyd_signal.wait_for_connection()
```
coretl, line 916: Can we put any timeouts in here, or does it just hang anyway?

DominicOram: It hangs either way but I've added it anyway

DominicOram:
Linting seems broken on main so I don't think it's related to this change specifically

coretl:
> Linting seems broken on main so I don't think it's related to this change specifically

Yes, mainly https://github.com/bluesky/bluesky/issues/1809

coretl (approved):

Issue #471, opened 2024-07-19
Change capture mode for PandA HDF writer to be "forever"


coretl:
We want to capture multiple PCAPs into a single HDF file, which we do with `capture_mode = FOREVER`.
Something like:

```
--- a/src/ophyd_async/panda/_common_blocks.py
+++ b/src/ophyd_async/panda/_common_blocks.py
@@ -6,6 +6,10 @@ from ophyd_async.core import Device, DeviceVector, SignalR, SignalRW
 from ophyd_async.core.signal_backend import SubsetEnum
 from ophyd_async.panda._table import DatasetTable, SeqTable
 
+class CaptureMode(str, Enum):
+    first_n = "FIRST_N"
+    last_n = "LAST_N"
+    forever = "FOREVER"
 
 class DataBlock(Device):
     # In future we may decide to make hdf_* optional
@@ -13,6 +17,7 @@ class DataBlock(Device):
     hdf_file_name: SignalRW[str]
     num_capture: SignalRW[int]
     num_captured: SignalR[int]
+    capture_mode: SignalRW[CaptureMode]
     capture: SignalRW[bool]
     flush_period: SignalRW[float]
     datasets: SignalR[DatasetTable]
diff --git a/src/ophyd_async/panda/writers/_hdf_writer.py b/src/ophyd_async/panda/writers/_hdf_writer.py
index bde53c17..7fdfe40e 100644
--- a/src/ophyd_async/panda/writers/_hdf_writer.py
+++ b/src/ophyd_async/panda/writers/_hdf_writer.py
@@ -14,7 +14,7 @@ from ophyd_async.core import (
 from ophyd_async.core.signal import observe_value
 from ophyd_async.epics.areadetector.writers.general_hdffile import _HDFDataset, _HDFFile
 
-from .._common_blocks import CommonPandaBlocks
+from .._common_blocks import CommonPandaBlocks, CaptureMode
 
 
 class PandaHDFWriter(DetectorWriter):
@@ -48,7 +48,7 @@ class PandaHDFWriter(DetectorWriter):
             self.panda_device.data.hdf_file_name.set(
                 f"{info.prefix}{self.panda_device.name}{info.suffix}.h5",
             ),
-            self.panda_device.data.num_capture.set(0),
+            self.panda_device.data.capture_mode.set(CaptureMode.forever),
         )
 
         # Wait for it to start, stashing the status that tells us when it finishes
```

PR #604, opened 2024-10-08
add Capture mode signal to Panda DataBlock

jsouter:
Addresses https://github.com/bluesky/ophyd-async/issues/471

Implemented CaptureMode as an int enum to match what's in PandABlocks-ioc

### DIFF
--- a/src/ophyd_async/fastcs/panda/_block.py
+++ b/src/ophyd_async/fastcs/panda/_block.py
@@ -7,6 +7,12 @@ from ophyd_async.core import Device, DeviceVector, SignalR, SignalRW, SubsetEnum
 from ._table import DatasetTable, SeqTable
 
 
+class CaptureMode(str, Enum):
+    FIRST_N = "FIRST_N"
+    LAST_N = "LAST_N"
+    FOREVER = "FOREVER"
+
+
 class DataBlock(Device):
     # In future we may decide to make hdf_* optional
     hdf_directory: SignalRW[str]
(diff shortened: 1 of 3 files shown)


jwlodek:
Should we make the enum use the string names for each value, instead of int? Like with https://github.com/bluesky/ophyd-async/blob/1559aa0e451ffc93da46f5a0fdc7cf282def7fab/src/ophyd_async/epics/adcore/_core_io.py#L71

jsouter:
> Should we make the enum use the string names for each value, instead of int? Like with
> 
> https://github.com/bluesky/ophyd-async/blob/1559aa0e451ffc93da46f5a0fdc7cf282def7fab/src/ophyd_async/epics/adcore/_core_io.py#L71

Would require a small change to the IOC but that would probably be nicer

src/ophyd_async/fastcs/panda/_block.py
```diff
@@ -7,6 +7,12 @@
 from ._table import DatasetTable, SeqTable
 
 
+class CaptureMode(int, Enum):
+    first_n = 0
+    last_n = 1
+    forever = 2
```
evalott100, line 13:
```suggestion
class CaptureMode(str, Enum):
    FIRST_N = "FIRST_N"
    LAST_N = "LAST_N"
    FOREVER = "FOREVER"
```

evalott100 (dismissed):

evalott100:
> > Should we make the enum use the string names for each value, instead of int? Like with
> > https://github.com/bluesky/ophyd-async/blob/1559aa0e451ffc93da46f5a0fdc7cf282def7fab/src/ophyd_async/epics/adcore/_core_io.py#L71
> 
> Would require a small change to the IOC but that would probably be nicer

I everythings already sorted on the pandablocks-ioc and ophyd async epics backend for this.

PR #612, opened 2024-10-15
with DeviceCollector doesn't work in plans, so error rather than hang

### LINKED ISSUE #548: Cannot connect devices from both ophyd and ophyd-async while a plan is running
Issue encountered on I24

If a plan is running and I attempt to connect ophyd and ophyd-async devices within it, the connection freezes, code to reproduce below:

```python
from bluesky.run_engine import RunEngine
from ophyd import Device, Component
from ophyd.signal import EpicsSignalRO
from ophyd_async.core import StandardReadable, DeviceCollector
from ophyd_async.epics.signal import epics_signal_r
import bluesky.plans as bp
from softioc import softioc, builder, asyncio_dispatcher
from multiprocessing.pool import Pool
import signal

def make_ioc() -> None:
    builder.SetDeviceName("FAKEIOC")
    builder.aIn("FOO", initial_value=5.0)
    builder.LoadDatabase()
    softioc.iocInit(asyncio_dispatcher.AsyncioDispatcher())


def run_ioc_in_subprocess() -> None:
    subprocess = Pool(
        initializer=lambda: signal.signal(signal.SIGINT, signal.SIG_IGN), processes=1
    )
    subprocess.apply(make_ioc)


class OphydAsyncDevice(StandardReadable):
        self.foo = epics_signal_r(float, prefix + "FOO")
        super().__init__(name)


class OphydDevice(Device):
    foo = Component(EpicsSignalRO, "FOO")

    def __init__(self, prefix: str = "", *, name: str, **kwargs):
        super().__init__(prefix, name=name, **kwargs)


RE = RunEngine()


def ophyd_device() -> OphydDevice:
    od = OphydDevice(prefix="FAKEIOC:", name="od")
    od.wait_for_connection()
    print("connected ophyd device")
    return od


def ophyd_async_device() -> OphydAsyncDevice:
    with DeviceCollector():
        oad = OphydAsyncDevice(prefix="FAKEIOC:")
    print("connected ophyd-async device")
    return oad


def my_plan():
    od = ophyd_device()
    oad = ophyd_async_device()
    yield from bp.count([od, oad])
    print("counted devices")


if __name__ == "__main__":
    make_ioc()
    # od = ophyd_device()
    # oad = ophyd_async_device()
    RE(my_plan())
```

If I comment out these lines in the plan 

```python
    od = ophyd_device()
    oad = ophyd_async_device()
```

...and uncomment them in main, outside where the plan is running, everything works.

I suspect this is more interplay issues between pyepics and aioca. See https://github.com/bluesky/ophyd-async/issues/89 and https://github.com/bluesky/ophyd-async/issues/291. Also I don't think we should be making blocking calls to `connect` inside a plan in the first place, but this is what dodal potentially does if you have, say, `i24.pmac()` inside your plan. Ideally we should defer to the `RunEngine`, maybe with an `ensure_connected()` plan?

coretl:
Fixes #548

### DIFF
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -12,7 +12,7 @@ from typing import (
 )
 
 from bluesky.protocols import HasName
-from bluesky.run_engine import call_in_bluesky_event_loop
+from bluesky.run_engine import call_in_bluesky_event_loop, in_bluesky_event_loop
 
 from ._utils import DEFAULT_TIMEOUT, NotConnected, wait_for_connection
 
@@ -224,6 +224,11 @@ class DeviceCollector:
         await self._on_exit()
 
     def __exit__(self, type_, value, traceback):
+        if in_bluesky_event_loop():
+            raise RuntimeError(
+                "Cannot use DeviceConnector inside a plan, instead use "
+                "`yield from ophyd_async.plan_stubs.ensure_connected(device)`"
+            )
         self._objects_on_exit = self._caller_locals()
         try:
             fut = call_in_bluesky_event_loop(self._on_exit())
--- a/tests/core/test_device_collector.py
+++ b/tests/core/test_device_collector.py
@@ -67,6 +67,16 @@ def test_sync_device_connector_run_engine_created_connects(RE):
     assert working_device.connected
 
 
+def test_connecting_in_plan_raises(RE):
+    def bad_plan():
+        yield from bps.null()
+        with DeviceCollector():
+            working_device = WorkingDevice("somename")  # noqa: F841
+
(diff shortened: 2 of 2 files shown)


src/ophyd_async/core/_device.py
```diff
@@ -224,6 +224,11 @@ async def __aexit__(self, type, value, traceback):
         await self._on_exit()
 
     def __exit__(self, type_, value, traceback):
+        if in_bluesky_event_loop():
+            raise RuntimeError(
+                "Cannot use DeviceConnector inside a plan, instead use "
```
callumforrester, line 229: I like this error message, it is very clear, is there any circumstance under which `in_bluesky_event_loop()` could return `True` _without_ our being inside a plan?

callumforrester (approved):

PR #625, opened 2024-10-25
Make pvi aware of DeviceVector

### LINKED ISSUE #624: Make PVI detail DeviceVectors
Could: Is this there because PVI has a standard that says something like "`thing_n` is part of a list of 'thing's on the device"? Or is it more of a loose convention?

Either way, since we get to decide what PVI says, could we make it return a structure rather than a string to make it clear this is the case? Out of scope for this PR obviously, would be a separate issue.

_Originally posted by @callumforrester in https://github.com/bluesky/ophyd-async/pull/594#discussion_r1812340170_
            

Can do this by using QSRV structures with `[n]` in them:
```
    info(Q:group, {
        "$(IOC_NAME=PANDAQSRV):PVI": {
            "value.pulse[1].d": {
                "+channel": "VAL",
                "+type": "plain"
            }
        }
    })
```
We can make this back compatible with https://github.com/PandABlocks/PandABlocks-ioc/issues/138

coretl:
Fixes #624

### DIFF
--- a/tests/fastcs/panda/db/panda.db
+++ b/tests/fastcs/panda/db/panda.db
@@ -122,7 +122,7 @@ record(stringin, "$(IOC_NAME=PANDAQSRV):PULSE1:_PVI")
     field(VAL,  "$(IOC_NAME=PANDAQSRV):PULSE1:PVI")
     info(Q:group, {
         "$(IOC_NAME=PANDAQSRV):PVI": {
-            "value.pulse1.d": {
+            "value.pulse[1].d": {
                 "+channel": "VAL",
                 "+type": "plain"
             }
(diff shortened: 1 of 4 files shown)


evalott100 (commented):
Nothing to add, I don't mind doing another review once you de-draft it.

callumforrester (approved):
Provisionally approving if this works how I think it does. I can't actually see how you've changed the underlying structure but I am assuming that is because I don't know how PVI works.

tests/fastcs/panda/db/panda.db
```diff
@@ -441,7 +441,7 @@ record(stringin, "$(IOC_NAME=PANDAQSRV):SEQ1:_PVI")
     field(VAL,  "$(IOC_NAME=PANDAQSRV):SEQ1:PVI")
     info(Q:group, {
         "$(IOC_NAME=PANDAQSRV):PVI": {
-            "value.seq1.d": {
+            "value.seq[1].d": {
```
coretl, line 444: @callumforrester this diff is the change to the PVI structure as published over PVA.

Once inside the IOC, instead of publishing:
```yaml
value:
  seq1:
    d: PANDA:SEQ1:PVI
  seq2:
    d: PANDA:SEQ2:PVI
```
we now publish
```yaml
value:
  seq:
    - null
    - d: PANDA:SEQ1:PVI
    - d: PANDA:SEQ2:PVI
```

Horrific eh?

callumforrester: Yuck, but also more structure so yay

evalott100: https://github.com/PandABlocks/PandABlocks-ioc/pull/139

Issue #385, opened 2024-06-13
Restructure `epics/test_signals.py`


coretl:
At the moment we have:
- A parameterized fixture that provides an IOC, with records either prefixed CA or PVA
- A few tests that use this and are parameterized to run on both
- A lot of tests that skip the CA parts

I suggest we refactor so:
- There is a single IOC that loads `test_records` twice with both CA and PVA prefixes
- The tests that run with both have are further parameterized with CA and PVA param
- The others just run using the IOC with PVA prefix

This will mean breaking the massive `test_backend_get_put_monitor` test matrix into two: PVA only tests, and CA and PVA tests

coretl:
Also we want to have an `ophyd_async.epics.testing` Device with a signal of every type to be used in this test and the device load saving tests

PR #630, opened 2024-10-30
Restructure epics/test_signals.py

jsouter:
closes https://github.com/bluesky/ophyd-async/issues/385

Now has a single IOC containing all the CA and PVA records, and a Device group which provides access to them.


tests/epics/signal/test_signals.py
```diff
@@ -43,41 +43,98 @@
...
+
+    def __create_signal(
+        self, protocol: Protocol, dtype: type[SignalDatatypeT], name: str
+    ):
+        signal = epics_signal_rw(dtype, f"{protocol}://{self._name}:{protocol}:{name}")
+        setattr(self, f"{protocol}_{name}".replace(":", "_"), signal)
+        self._all_signals[protocol][name] = signal
+
+    def get_signal(self, protocol: Protocol, name: str):
+        return self._all_signals[protocol][name]
```
coretl, line 92: Actually, if we base this on https://github.com/bluesky/ophyd-async/pull/598 then we could write this like:
```python
from typing import Annotated as A
from ophyd_async.epics.core import EpicsDevice, PvSuffix

class CaAndPvaTestDevice(EpicsDevice):
    int: A[SignalRW[int], PvSuffix("int")]
    ...
    stra: A[Sequence[str], PvSuffix("stra")]
```

Then we can add the PVA only ones in a second device:
```python
class PvaOnlyTestDevice(EpicsDevice):
    int8a: A[Array1D[np.int8], PvSuffix("int8a")]
    ...
    ntndarray: A[np.ndarray, PvSuffix("ntndarray")]
```

tests/epics/signal/test_signals.py
```diff
@@ -43,41 +43,98 @@
...
+PVA_RECORDS = str(Path(__file__).parent / "test_records_pva.db")
 PV_PREFIX = "".join(random.choice(string.ascii_lowercase) for _ in range(12))
 
+Protocol = Literal["ca", "pva"]
+
+
+PARAMETERISE_PROTOCOLS = pytest.mark.parametrize("protocol", get_args(Protocol))
+
+
+class DeviceTestRecordsGroup(Device):
```
coretl, line 56: Please can we put this in a new `ophyd_async.epics.testing` package

tests/epics/signal/test_signals.py
```diff
@@ -43,41 +43,98 @@
...
+@pytest.fixture(scope="module")
+def ioc(request: pytest.FixtureRequest, tmpdir_factory) -> subprocess.Popen:
+    with open(CA_PVA_RECORDS) as f:
+        ca_pva_template = f.read()
+    with open(PVA_RECORDS) as f:
+        pva_template = f.read()
+
+    template = ca_pva_template.replace("$(R)", "ca:")
+    template += ca_pva_template.replace("$(R)", "pva:")
+    template += pva_template.replace("$(R)", "pva:")
```
coretl, line 122: We can pass many `-m` and `-d` flags to build up the database, like in https://github.com/bluesky/ophyd-async/blob/7d6070b30dc44e77bfd3c7024642460a2dffd27a/src/ophyd_async/epics/demo/__init__.py#L36-L38, rather than having to create a new `ALL_RECORDS` file...

coretl: Come to think of it, we should try and work out a nice way of starting up the IOC subprocess for the `epics.demo` and `epics.testing` packages as a module level fixture here, and in the docs. Will discuss this with you in person...

tests/epics/signal/test_signals.py
```diff
@@ -43,41 +43,98 @@
...
-        backend = _epics_signal_backend(typ, pv, pv)
-        await backend.connect(timeout=timeout)
-        return backend
+
+async def _make_backend(typ: type | None, protocol: str, suff: str, timeout=10.0):
+    pv = f"{protocol}://{PV_PREFIX}:{protocol}:{suff}"
+    # Make and connect the backend
+    backend = _epics_signal_backend(typ, pv, pv)
+    await backend.connect(timeout=timeout)
+    return backend
```
coretl, line 107: Can we replace this with `getattr(device, suffix)` in some way?

jsouter:
New changes are WIP, not working as decorators yet, if that is what we want.
Trying to figure out the best way to instantiate devices and iocs with fixtures - right now the test that use device_test_records need also use the ioc fixture so that they work when run alone - when run as a group it can be a race condition if any of the other tests use the ioc fixture - then it becomes available for all other tests. I don't know if we should make the devices dependent on the ioc fixture - I'm not sure the best way to do this generally with how the fixture names are handled in the module scope - it's easy to do if we have lots of boilerplate in the test modules themselves or possibly in conftest somewhere

jsouter:
Will update test_device_save_loader to use the the example PvaDevice class

stan-dot:
the only codecov complaint is `testing._utils` 

```python
[src/ophyd_async/epics/testing/_utils.py](https://app.codecov.io/gh/bluesky/ophyd-async/pull/630/blob/src/ophyd_async/epics/testing/_utils.py)
12
73.33%
73.33%
-
```

@coretl petition to exclude testing._utils from testing coverage

coretl:
@jsouter should I review this now or are there more changes to come?

jsouter:
> @jsouter should I review this now or are there more changes to come?

I've got some more changes to push, will try and get those done soon!

coretl (approved):

PR #633, opened 2024-10-31
Windows: fix unit tests & enable CI

Tom-Willemsen:
## Description

- Fixes a number of unit tests which were either failing or inconsistent  under windows: mostly race conditions, paths, numpy type differences, and asyncio scheduling differences.
- Enable `windows-latest` as part of the testing CI matrix.

## Motivation and Context

At isis we're running on windows, we'd like to be able to run the unit tests.

## How Has This Been Tested?

See the CI runs.


tests/epics/test_motor.py
```diff
@@ -90,7 +90,7 @@ async def test_motor_moving_well_2(sim_motor: motor.Motor) -> None:
     s.watch(watcher)
     done = Mock()
     s.add_callback(done)
-    await asyncio.sleep(A_BIT)
```
coretl, line 93: I'm concerned about just bumping up arbitrary sleeps in the test code, so I had a go at writing something that actually waits for all pending task wakeups to happen (which is what we really want):
```python
async def wait_for_wakeups(max_yields=10):
    loop = asyncio.get_event_loop()
    # If anything has called loop.call_soon or is scheduled a wakeup
    # then let it run
    for _ in range(max_yields):
        if loop._scheduled:
            await asyncio.sleep(0)
        else:
            return
    raise RuntimeError(f"Tasks still scheduling wakeups after {max_yields} yields")
```

Replacing all `await asyncio.sleep(A_BIT)` with `await wait_for_wakeups()` seems to work on Linux, but it's entirely event loop specific (and so OS specific). Does this work on windows?

tests/fastcs/panda/test_hdf_panda.py
```diff
@@ -149,7 +149,7 @@ def flying_plan():
             "uid": ANY,
             "data_key": data_key_name,
             "mimetype": "application/x-hdf5",
-            "uri": "file://localhost" + str(tmp_path / "test-panda.h5"),
+            "uri": "file://localhost/" + str(tmp_path / "test-panda.h5").lstrip("/"),
```
coretl, line 152: What does one of these windows paths actually look like? Not sure what the right `file://` semantics are for drive letter etc. What does a web browser show if you point to `file://something_on_c_drive` on windows?

coretl (commented):
Thanks for the additions! A couple fo comments...

tests/fastcs/panda/test_hdf_panda.py
Tom-Willemsen: Web browsers (at least firefox, chrome and edge) do correctly understand a path like `file:///C:\Instrument\Var\logs\ioc\ARINST-20241030.log`. Forward slashes rather than backslashes would also work - but the drive letter part is what it wants as far as i understand it.

tests/epics/test_motor.py
Tom-Willemsen: Just to check - [this](https://github.com/bluesky/ophyd-async/compare/tomc_wakeup_test?expand=1#diff-a20f02ff3b0d811a166f5a1c5d752f08f0e79637adcd53349a64534cb91f0e58) is what you were suggesting? If so that's not working for me on any of the modified tests, locally on windows, nor in a linux container, nor [on github CI linux or windows](https://github.com/bluesky/ophyd-async/actions/runs/11627138318/job/32379976029) - am I missing something? 

If you had something different working and want to push to a branch I'm happy to test locally...

Tom-Willemsen: Ah sorry just seen your edits... will retest...

coretl (approved):

PR #638, opened 2024-11-04
Resolves issue with getting table value correctly on panda writer.open

jwlodek:
Currently, doing a `writer.open()` on a PandA writer will fail with an error instantiating the `DatasetTable` object. You can actually reproduce the issue by just doing:

```
await panda1.data.datasets.get_value()
```

This is because the pvget of the corresponding table PV will return something like:

```
{'name': {'r': ['PhotoDiode]}, 'type': {'r': [float64]}}
```

which does not unpack into something that matches the init signature of `DatasetTable`.

This change resolves the immediate issue in that it lets the writer open, and the scan can execute. However, I am not sure if this is the correct solution across the board.


coretl:
There shouldn't be an access level in the structures in the first place, so this needs to be fixed on the IOC side...

Issue #640, opened 2024-11-06
 mx-bluesky unit tests taking ~3x longer since signal-typing changes


rtuck99:
Since updating mx-bluesky to be compatible with #594, the unit tests are taking significantly longer to complete.
On my desktop machine the difference is ~97s to complete compared to ~27s before

It seems that most of this slowdown is due to waiting in `call_in_bluesky_event_loop()` waiting for `wait_for_connection(coros=device.connect(mock=True))` to complete.

Can this be made faster?

coretl:
Can you post a typical test please?

olliesilvester:
See https://github.com/DiamondLightSource/mx-bluesky/blob/3845a51a2f083c578b92476702950f4125813b4b/tests/unit_tests/beamlines/i04/test_thawing.py#L118 for a typical test

In our tests, devices are created through fixtures which look like 
`smargon = i03.smargon(fake_with_ophyd_sim=True)`

which in turn calls
`call_in_bluesky_event_loop(...)` and the ophyd-async `wait_for_connection(..)`

DominicOram:
I'm happy this is fixed by https://github.com/bluesky/ophyd-async/pull/641

PR #641, opened 2024-11-08
Speed up device creation and connection in mock mode

coretl:


coretl:
@rtuck99 or @DominicOram does this improve your test speeds?

tests/core/test_device.py
```diff
@@ -174,43 +175,45 @@ def __init__(self, name: str) -> None:
         super().__init__(name)
 
 
+async def test_many_individual_device_connects_not_slow():
```
DominicOram, line 178: Could: Might be worth the equivalent test for connecting in parallel?

DominicOram (approved):
Confirmed that this speeds our tests back up to ~30s and so fixes https://github.com/bluesky/ophyd-async/issues/640, thank you! Comments in code are all optional, otherwise looks good.

PR #642, opened 2024-11-08
An idea...

coretl:
In https://github.com/bluesky/ophyd-async/pull/641 we make connect in mock mode sync rather than async, but can't complete the job because `Device.connect` is async so child connects must also be async. This PR make `connect` return an optional `Task` which means we can make the mock path fully sync.

Too magic?

Also, tests need fixing...

### DIFF
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -35,10 +35,10 @@ class DeviceConnector:
         during ``__init__``.
         """
 
-    async def connect_mock(self, device: Device, mock: LazyMock):
+    def connect_mock(self, device: Device, mock: LazyMock):
         # Connect serially, no errors to gather up as in mock mode
         for name, child_device in device.children():
-            await child_device.connect(mock=mock.child(name))
+            assert not child_device.connect(mock=mock.child(name))
 
     async def connect_real(self, device: Device, timeout: float, force_reconnect: bool):
         """Used during ``Device.connect``.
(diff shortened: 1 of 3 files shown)


callumforrester:
Mocking is always magic anyway so the magic does not bother me

DominicOram (commented):
I feel like `connect` is a semi-user facing interface. For example, we use it quite a lot in `dodal`/`mx_bluesky` (which may be wrong I guess). This complicates the job for the caller quite a bit so I'm not a big fan. We could do something instead like:

```python

class DeviceConnector:
    def connect_mock(self, device: Device, mock: LazyMock):
        # Connect serially, no errors to gather up as in mock mode
        for name, child_device in device.children():
            assert not child_device._mock_connect(mock=mock.child(name)) #Bit ugly this calls something private but I wouldn't want to confuse external callers of `Device`

class Device:
    ...

    def _mock_connect(self, mock):
        # Always connect in mock mode serially
        if isinstance(mock, LazyMock):
            # Use the provided mock
            self._mock = mock
        elif not self._mock:
            # Make one
            self._mock = LazyMock()
        self._connector.connect_mock(self, self._mock)

    async def connect(
        self,
        mock: bool | LazyMock = False,
    ) -> None:
        """Connect self and all child Devices.

        Contains a timeout that gets propagated to child.connect methods.

        Parameters
        ----------
        mock:
            If True then use ``MockSignalBackend`` for all Signals
        timeout:
            Time to wait before failing with a TimeoutError.
        """
        if mock:
            self._mock_connect(mock)
        else:
            ...

coretl:
I think the gains are marginal, so let's abandon this idea

PR #646, opened 2024-11-12
Fix some small issues discovered in testing

coretl:
- Raise `NotConnected` errors in mock mode too when failing on creation of mock signal because of invalid datatype
- Support bare `np.ndarray` in SoftSignalBackend
- Support tagged and variant unions for NTNDArray
- Check that Devices have unique names in `ensure_connected`

### DIFF
--- a/src/ophyd_async/plan_stubs/_ensure_connected.py
+++ b/src/ophyd_async/plan_stubs/_ensure_connected.py
@@ -1,3 +1,5 @@
+from collections.abc import Awaitable
+
 import bluesky.plan_stubs as bps
 
 from ophyd_async.core import DEFAULT_TIMEOUT, Device, LazyMock, wait_for_connection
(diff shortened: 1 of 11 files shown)


DominicOram (approved):
Great, thanks. Looks good and only involved some minor changes to `dodal` (https://github.com/DiamondLightSource/dodal/pull/899)

src/ophyd_async/plan_stubs/_ensure_connected.py
```diff
@@ -9,18 +11,19 @@ def ensure_connected(
...
-        [
-            lambda: wait_for_connection(
-                **{
-                    device.name: device.connect(
-                        mock=mock, timeout=timeout, force_reconnect=force_reconnect
-                    )
-                    for device in devices
-                }
+    if len(devices) != len({device.name for device in devices}):
+        raise RuntimeError(f"Devices do not have unique names {devices}")
```
callumforrester, line 15: Could: Use something more specific than `RuntimeError`, maybe `ValueError` fits here?

Nit: This also fails if the user just passes the same device twice, I'm not too opinionated about whether we want to support that, but the error message makes less sense in that case.

callumforrester: Additionally, having played with this, some sort of diff or other useful information for finding the conflict would be useful. If I have 30 devices and 1 duplicate name it may be hard to spot.

callumforrester: @coretl as requested, can't think of a one-liner but this is close

```python
device_names = [device.name for device in devices]
duplicates = {name for name in device_names if device_names.count(name) > 1}
```

PR #648, opened 2024-11-13
Yield in each loop of observe_value

coretl:
This helps in the very specific case of an observe_value directly or indirectly modifying the signal that is being updated. This creates a busy loop which will not be interrupted by wrapping in asyncio.wait_for.  To demonstrate, added
test_observe_value_times_out_with_no_external_task

Originally discussed in https://github.com/DiamondLightSource/dodal/pull/897

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -449,12 +449,6 @@ async def observe_value(
     """
 
     q: asyncio.Queue[SignalDatatypeT | Status] = asyncio.Queue()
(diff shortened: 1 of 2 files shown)


DominicOram (approved):
I agree that I'm still not 100% convinced that this will fix our issue but:
* It does no harm
* The tests are great because they match our actual usecases

I would be happy to have this merged instead of https://github.com/DiamondLightSource/dodal/pull/897 and then revisit if we see it again on the beamline

PR #650, opened 2024-11-14
Add new timeout for observe_value

coretl:
@DominicOram @olliesilvester this is the epics test I mentioned in https://github.com/DiamondLightSource/dodal/pull/897#issuecomment-2474190961

It looks like I need to yield after the `q.get` rather than before to let the timeout break without an extra yield


coretl:
Spoke too soon, doesn't work on CI...

coretl:
Ok, looks like there are actually numerous python bugs in `asyncio.wait_for`, which is fixed by its reimplementation on top of `asyncio.timeout` in 3.12: https://github.com/python/cpython/issues/96764

I think this fixes your particular use case, but you can't pass a `timeout` to `observe_value` and also wrap it in `wait_for` with a timeout on python 3.11 and below.

I would also suggest that you move the opencv calls to another thread and consume all the values from the Queue between the two, dropping old updates.

olliesilvester:
Great, thank you!

coretl:
Ok, still some issues will this, will investigate next week

coretl:
@DominicOram @olliesilvester @rtuck99 I think this is now working, but would be good to talk through the code with one of you to check I understand the problem correctly

olliesilvester:
> @DominicOram @olliesilvester @rtuck99 I think this is now working, but would be good to talk through the code with one of you to check I understand the problem correctly

I'm about today, I'm just reading through the changes now

olliesilvester:
Discussed with @coretl, I think this change fixes our issue but I will investigate further to see if there are any safer solutions

olliesilvester:
I tried using `asyncio.timeout` instead, and we seem to get the same issue. I guess this explains why python 3.12 still had this issue.

I'm leaning towards just not using `asyncio.wait_for` for this scenario. We could just do
```
async def observe_value(...):
  if time.time() - start_time > overall_timeout:
                  raise TimeoutError
...
```
instead, and warn against using `asyncio.wait_for` (or timeout) with `observe_value`

coretl:
> I'm leaning towards just not using `asyncio.wait_for` for this scenario

It's messy, but it's probably the most pragmatic solution here, then we relegate `wait_for_pending_wakeups` to `ophyd_async.testing` and only use it in tests

DominicOram:
I'm not a big fan of this, having to roll you're own timeout code when there's provision for it in `asyncio` looks on the face of it like a code-smell if you're not familiar with the bug. How would we feel about adding the functionality into `observe_value` itself e.g. optionally pass in a `overall_timeout` value? This would then make the path of least resistance be to just use that

olliesilvester:
Yeah, the timeout will be in `observe_value` as an optional parameter

olliesilvester:
I've given this a go but there are some extra things to think about

1. The regular timeout defaults to `None`, which means that if we do our `overall_timeout`, as above, then the code can get stuck forever doing `get_value`.

Solution here is either to either give `timeout` a default value or make a clever decorator which automatically supplies our `overall_timeout` parameter with the same value which was used in our timeout in `asyncio.wait_for(timeout=...)`

2. If we don't go with the latter solution above, then we could add a warning (with a decorator?) if `observe_value` has been used with `asyncio.timeout`

Thoughts @coretl?

olliesilvester:
Although people will probably still do `asyncio.wait_for(observe_value...)`, this issue (if anyone even sees it again) should be easier to spot with the extra docstring

src/ophyd_async/core/_signal.py
```diff
@@ -504,13 +528,19 @@ def queue_value(value: SignalDatatypeT, signal=signal):
...
-            await asyncio.sleep(0)
-            item = await get_value()
+            if overall_deadline and time.monotonic() >= overall_deadline:
+                raise asyncio.TimeoutError(
+                    f"observe_value was still observing signals "
+                    f"{[signal.source for signal in signals]} after "
+                    f"timeout {except_after_time}s"
+                )
+
+            item = await get_value(_get_iteration_timeout(timeout, overall_deadline))
```
coretl, line 543: `asyncio.wait_for` accepts timeout=None without scheduling a task, so I would just use that...
```suggestion
            iteration_timeout = _get_iteration_timeout(timeout, overall_deadline)
            item = await asyncio.wait_for(q.get(), iteration_timeout)
```

src/ophyd_async/core/_signal.py
```diff
@@ -486,12 +511,11 @@ async def observe_signals_value(
...
-    if timeout is None:
-        get_value = q.get
-    else:
 
-        async def get_value():
-            return await asyncio.wait_for(q.get(), timeout)
+    async def get_value(timeout: float | None = None):
+        if timeout is None:
+            return await q.get()
+        return await asyncio.wait_for(q.get(), timeout)
```
coretl, line 518: ... then this can go
```suggestion
```

src/ophyd_async/core/_signal.py
```diff
@@ -425,6 +426,7 @@ async def observe_value(
     signal: SignalR[SignalDatatypeT],
     timeout: float | None = None,
     done_status: Status | None = None,
+    except_after_time: float | None = None,
```
coretl, line 429: slight preference for `done_timeout` instead of `except_after_time`

src/ophyd_async/core/_utils.py
```diff
@@ -311,3 +311,24 @@ def __call__(self) -> Mock:
...
+    """
+    loop = asyncio.get_event_loop()
+    # If anything has called loop.call_soon or is scheduled a wakeup
+    # then let it run
+    for _ in range(max_yields):
+        await asyncio.sleep(0)
+        if not loop._ready:  # type: ignore # noqa: SLF001
+            return
+    if raise_if_exceeded:
+        raise RuntimeError(f"Tasks still scheduling wakeups after {max_yields} yields")
```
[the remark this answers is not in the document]
coretl: Prefer to ship them in `ophyd_async.testing` as they can be used in things like `dodal`, following the pattern of `numpy.testing`

Issue #639, opened 2024-11-04
 PandA writer fails to open, caused by issue getting dataset table


jwlodek:
Currently, doing a `writer.open()` on a PandA writer will fail with an error instantiating the `DatasetTable` object. You can actually reproduce the issue by just doing:

```
await panda1.data.datasets.get_value()
```

This is because the pvget of the corresponding table PV will return something like:

```
{'name': {'r': ['PhotoDiode]}, 'type': {'r': [float64]}}
```

which does not unpack into something that matches the init signature of `DatasetTable`

See #638 for a potential solution.

## Steps To Reproduce
1. Instantiate an HDFPanda device
2. Run await `panda1.writer.open()`

## Acceptance Criteria
- Able to open writer without crashing

coretl:
@evalott100 this is a bug in the IOC side. All the recent changes should only have happened on the PVI structures, but it looks like some changes crept in on the Table structures. I can see how this might have happened as they probably use the same QSRV helpers.

evalott100:
Will look into this this morning.

evalott100:
Before and after of `pvget -v`

```
PANDA1:DATA:DATASETS epics:nt/NTTable:1.0 
    structure record
        structure _options
            boolean atomic true
    string[] labels ["Name", "Type"]
    structure value
        string[] name []
        string[] type []
```
```
PANDA1:DATA:DATASETS epics:nt/NTTable:1.0 
    structure record
        structure _options
            boolean atomic true
    string[] labels ["Name", "Type"]
    structure pvi
        structure name
            string[] r []
        structure type
            string[] r []
    structure value
        structure name
            string[] r []
        structure type
            string[] r []
```

The cause of which is

https://github.com/PandABlocks/PandABlocks-ioc/blob/355a6ab5a584005661492590e607b95eb602c3cb/src/pandablocks_ioc/_tables.py#L106-L115

evalott100:
The cause was passing in access here:

https://github.com/PandABlocks/PandABlocks-ioc/blob/355a6ab5a584005661492590e607b95eb602c3cb/src/pandablocks_ioc/_tables.py#L106-L115

evalott100:
@jwlodek Just released the fix on the ioc side, let me know if you run into any more problems.

jwlodek:
The `DATASETS` PV now does not have the access level: 
```
In [9]: !pvget -v XF:31ID1-ES{PANDA:1}:DATA:DATASETS
[...]

PR #651, opened 2024-11-14
Rename hdf5_type to data_type in DatasetTable

jwlodek:
Decided to go with `dtype` rather than `data_type` - so the table name on the ioc side doesn't need an underscore.


jwlodek:
As discussed, this + https://github.com/PandABlocks/PandABlocks-ioc/pull/151 will solve https://github.com/bluesky/ophyd-async/issues/639

evalott100 (approved):

PR #655, opened 2024-11-19
Add Python 3.12 to CI now that p4p is updated

OCopping:
I noticed Python 3.12 never got added to the CI after support for it in `p4p` got released.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -114,7 +114,12 @@ def fail_test_on_unclosed_tasks(request: FixtureRequest):
     """
 
     fail_count = request.session.testsfailed
-    loop = asyncio.get_event_loop()
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+
+    asyncio.set_event_loop(loop)
     loop.set_debug(True)
 
     request.addfinalizer(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -113,20 +113,21 @@ def fail_test_on_unclosed_tasks(request: FixtureRequest):
     by the end of the test.
     """
 
-    fail_count = request.session.testsfailed
     try:
+        fail_count = request.session.testsfailed
         loop = asyncio.get_running_loop()
-    except RuntimeError:
-        loop = asyncio.new_event_loop()
 
-    asyncio.set_event_loop(loop)
-    loop.set_debug(True)
+        asyncio.set_event_loop(loop)
+        loop.set_debug(True)
 
-    request.addfinalizer(
-        lambda: _error_and_kill_pending_tasks(
-            loop, request.node.name, request.session.testsfailed == fail_count
+        request.addfinalizer(
+            lambda: _error_and_kill_pending_tasks(
+                loop, request.node.name, request.session.testsfailed == fail_count
+            )
         )
-    )
+    except RuntimeError as error:
+        if str(error) != "no running event loop":
+            raise error
 
 
 @pytest.fixture(scope="function")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -89,7 +89,9 @@ def _error_and_kill_pending_tasks(
     unfinished_tasks = {
         task
         for task in asyncio.all_tasks(loop)
-        if task.get_coro().__name__ not in _ALLOWED_PYTEST_TASKS and not task.done()
+        if (coro := task.get_coro()) is not None
+        and coro.__name__ not in _ALLOWED_PYTEST_TASKS
+        and not task.done()
     }
     for task in unfinished_tasks:
         task.cancel()
@@ -117,7 +119,6 @@ def fail_test_on_unclosed_tasks(request: FixtureRequest):
         fail_count = request.session.testsfailed
         loop = asyncio.get_running_loop()
 
-        asyncio.set_event_loop(loop)
         loop.set_debug(True)
 
         request.addfinalizer(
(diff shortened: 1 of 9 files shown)


DiamondJoseph (changes requested):
If we're supporting 3.12 add it to the pyproject.toml too
Good work getting this over the line :partying_face:

OCopping:
The tests are failing because of a `get_event_loop()` deprecation warning/error.
From the [docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop):
```Deprecated since version 3.12: Deprecation warning is emitted if there is no current event loop. In some future Python release this will become an error.```

**Comment by coretl** [comment]
> The tests are failing because of a `get_event_loop()` deprecation warning/error. From the [docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop): `Deprecated since version 3.12: Deprecation warning is emitted if there is no current event loop. In some future Python release this will become an error.`

@evalott100 I think this is coming from the "detect pending tasks" fixture, would you mind taking a look please?

**Comment by evalott100** [comment]
Taking this over.

File: tests/epics/adcore/test_drivers.py [hunk 1]
```diff
@@ -90,10 +90,12 @@ async def wait_then_fail():
...
+
     acquiring = await adcore.start_acquiring_driver_and_ensure_status(
         driver, timeout=0.1
     )
-    await wait_then_fail()
-
-    with pytest.raises(ValueError):
+    with pytest.raises(
+        ValueError, match="Final detector state DetectorState.DISCONNECTED"
+    ):
```
evalott100, line 100: We should be starting setting up `wait_then_fail` before we schedule the task which should fail.

panda_connect.py
```diff
@@ -0,0 +1,29 @@
...
+        "PANDA1:", StaticPathProvider(StaticFilenameProvider("test-panda"), Path("."))
+    )
+    await panda.connect()
+    assert isinstance(panda, HDFPanda)
+    table = await panda.seq[1].table.get_value()
+    assert isinstance(table, SeqTable)
+    print("OUTA1", table.outa1)
+    outa2 = table.outa2
+    position = table.position
+    assert isinstance(outa2, np.ndarray) and outa2.dtype == np.bool_, outa2.dtype
```
DiamondJoseph:
```suggestion
    assert isinstance(outa2, np.ndarray) and outa2.dtype == np.bool_, f"{outa2.dtype} is of wrong type!"
```

Not exactly the wording I want, but I was confused until I realised the arg after the comma was for the AssertionError, not part of the comparison

panda_connect.py
```diff
@@ -0,0 +1,29 @@
...
+    assert (
+        isinstance(position, np.ndarray) and position.dtype == np.int32
+    ), position.dtype
```
DiamondJoseph: as above

panda_connect.py
DiamondJoseph: But is this for local testing only and not part of the code to be pushed?

tests/conftest.py
```diff
@@ -113,15 +115,21 @@ def fail_test_on_unclosed_tasks(request: FixtureRequest):
     by the end of the test.
     """
 
-    fail_count = request.session.testsfailed
-    loop = asyncio.get_event_loop()
-    loop.set_debug(True)
+    try:
+        fail_count = request.session.testsfailed
+        loop = asyncio.get_running_loop()
```
evalott100, line 120: `get_event_loop` gives a deprecation warning that future python versions will raise an error if there is no event loop, `get_running_loop` does this already. It doesn't seem possible to check for an event loop without failing/warning if there isn't one, so we just catch the exception for the tests which don't use one.

tests/conftest.py
```diff
@@ -113,15 +115,21 @@ def fail_test_on_unclosed_tasks(request: FixtureRequest):
...
-            loop, request.node.name, request.session.testsfailed == fail_count
+        request.addfinalizer(
+            lambda: _error_and_kill_pending_tasks(
+                loop, request.node.name, request.session.testsfailed == fail_count
+            )
         )
-    )
+    except RuntimeError as error:
+        if str(error) != "no running event loop":
+            raise error
```
DiamondJoseph, line 132: nit: Is it worth having a `pass` explaining that if `str(error) == "no running event loop"` then the test either does not require a loop or will fail when attempting to use the loop?

DiamondJoseph (approved):
Looks good from my side, but as I'm not as involved in ophyd-async consider this advisory only

tests/conftest.py
evalott100: > Is it worth having a `pass`

What do you mean by this?
``` python
except RuntimeError as error:
    if str(error) == "no running event loop":
        pass
    else:
        raise error
```
?

DiamondJoseph: Yes exactly that, just a bit more obvious what is intended

evalott100: I would prefer the current `!=` comprehension but I'll add a comment explaining!

tests/core/test_observe.py
```diff
@@ -86,7 +86,7 @@ async def tick():
 
     async def watch():
         async for val in observe_value(sig, done_timeout=0.2):
-            time.sleep(0.15)
+            await asyncio.sleep(0.1)
```
coretl, line 89: Ok, my fault for not putting a comment in here, but this really was meant to be a `time.sleep`! This is a test to prove a subtle timing bug where the inner loop of observe_value was blocking the event loop. Please can you restore this and put a comment to this effect in?

evalott100: Made a [new issue](https://github.com/bluesky/ophyd-async/pull/692), easier than a revert.

PR #660, opened 2024-11-21
FIX: Handle NonSupportedFeature exception for pipes

burkeds:
Old tango servers running tango version less than 9 don't support the pipe API or get_pipe_list(). Since we don't support pipes yet anyway, we should handle this exception and move on.

### DIFF
--- a/src/ophyd_async/tango/signal/_signal.py
+++ b/src/ophyd_async/tango/signal/_signal.py
@@ -16,7 +16,7 @@ from ophyd_async.core import (
     SignalW,
     SignalX,
 )
-from tango import AttrDataFormat, AttrWriteType, CmdArgType, DeviceProxy, DevState
+from tango import AttrDataFormat, AttrWriteType, CmdArgType, DeviceProxy, DevState, NonSupportedFeature
 from tango.asyncio import DeviceProxy as AsyncDeviceProxy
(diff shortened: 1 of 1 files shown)


burkeds:
@coretl It looks like CI is failing here due to type checking error in: 

/home/runner/work/ophyd-async/ophyd-async/src/ophyd_async/core/_table.py:49:80

coretl:
> @coretl It looks like CI is failing here due to type checking error in:
> 
> /home/runner/work/ophyd-async/ophyd-async/src/ophyd_async/core/_table.py:49:80

Just fixed in #665

coretl (approved):

burkeds:
> > @coretl It looks like CI is failing here due to type checking error in:
> > /home/runner/work/ophyd-async/ophyd-async/src/ophyd_async/core/_table.py:49:80
> 
> Just fixed in #665

Thanks Tom. It looks like this brings down the coverage though. Since this relies on interacting with an old version of Tango, I am not sure how I could write a test for it.

coretl:
> It looks like this brings down the coverage though. Since this relies on interacting with an old version of Tango, I am not sure how I could write a test for it.

I don't think I care about this, it's such a small change, I'll merge as is...

Issue #525, opened 2024-08-16
Uses cases for signals that derive from other signals


canismarko:
Following up from a conversation on mattermost: this issue can hold use cases for a DerivedSignal class. I'll include ours as comments.

> is there an ophyd-async way to derive signals from other signals, in the spirit of OG ophyd's DerivedSignal and pcdsdevices' MultiDerivedSignal?

canismarko:
# Calculated signals

We have several situations where we use one or more epics signals to calculate new values. For example, an ion chamber's scaler measures counts from the voltage-to-frequency convertor. We use a derived signal for calculating what the voltage was at the input to the V-to-F, and another one that uses this voltage to calculate the current at the input to the pre-amplifier. Eventually we will add another that calculates the photon counts based on this current.

https://github.com/spc-group/haven/blob/c31960b010ce22c6d4d7bd909458acdfd800b122/src/haven/instrument/ion_chamber.py#L39

canismarko:
# Standardizing Interfaces

We have a set of 4 filters arranged in a single XIA PFCU4 filter bank, and two of them are fitted with solid tungsten to work as a shutter. To operate the shutter I need to read the existing state of all four filters as bits and specifically set the two bits for the W filters to the correct position for the shutter to be open (``XX10``) or closed (``XX01``) where X is the existing state of the other two non-shutter filters.

I used a DerivedSignal with some bitmasking to create a shutter device that can operate like any other shutter using e.g. ``set(0)`` or ``set(1)`` to open and close the shutter respectively.

https://github.com/spc-group/haven/blob/c31960b010ce22c6d4d7bd909458acdfd800b122/src/haven/instrument/xia_pfcu.py#L58

canismarko:
# Sharing Start/Stop Signals

To operate our undulators, there are PVs for setting the desired energy, gap, energy taper and gap taper. Then a single "start" PV exists for moving the undulator to all these targets. I wrote a device that has positioners for these 4 values so that I can do ``undulator.energy_taper.set(100)`` and it will move the undulator to that taper. To do this, each of those four positioners uses derived signals linking back to the undulator's start and stop signals so that they can all operate the undulator properly.

https://github.com/spc-group/haven/blob/c31960b010ce22c6d4d7bd909458acdfd800b122/src/haven/instrument/xray_source.py#L25

canismarko:
# Encoding Data in a PV as a string

[...]

PR #661, opened 2024-11-21
Initial stab at DerivedSignal #525

coretl:


src/ophyd_async/core/_protocol.py
```diff
@@ -96,6 +96,14 @@
         """
 
 
+@runtime_checkable
+class AsyncMovable(Protocol):
+    @abstractmethod
+    def set(self, value) -> AsyncStatus:
+        """Return a ``Status`` that is marked done when the device is done moving."""
+        ...
```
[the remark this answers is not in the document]
evalott100: Wow, AI is an even more pedantic reviewer than I am

src/ophyd_async/core/_signal.py
```diff
@@ -56,7 +56,15 @@ def __init__(self, backend: SignalBackend):
...
 
     async def connect_mock(self, device: Device, mock: LazyMock):
-        self.backend = MockSignalBackend(self._init_backend, mock)
+        # TODO: ugly circular import
+        from ._derived_signal import DerivedSignalBackend
+
+        if isinstance(self._init_backend, DerivedSignalBackend):
+            # It makes no sense to mock a derived signal backend, so
+            # use the real thing which will be backed by mocked signals
+            self.backend = self._init_backend
```
evalott100, line 65: Doesn't it? We still want mock utilities...

Is there a way to do it [here](https://github.com/bluesky/ophyd-async/blob/bdffa0111d4e7c432ce0ce8f3214b09e81f526ac/src/ophyd_async/core/_mock_signal_backend.py#L27-L34) like we do with soft signal backend?

coretl: The problem is what does it mean to call `set_mock_value` on a `DerivedSignal`? That makes it more of a "half mocked" signal, which might be confusing...

evalott100: Yea, I was thinking of maybe allowing it and putting an assert here

https://github.com/bluesky/ophyd-async/blob/93ac49901f6b32e5a8243a278119c9ced53dc7b4/src/ophyd_async/testing/_mock_signal_utils.py#L35-L38

but we probably don't have a use case really.

src/ophyd_async/core/_signal.py
```diff
@@ -56,7 +56,15 @@
         self.backend = self._init_backend = backend
 
     async def connect_mock(self, device: Device, mock: LazyMock):
-        self.backend = MockSignalBackend(self._init_backend, mock)
+        # TODO: ugly circular import
+        from ._derived_signal import DerivedSignalBackend
```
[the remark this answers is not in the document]
coretl: Yes I know...

src/ophyd_async/epics/motor.py
```diff
@@ -228,11 +238,18 @@ async def _prepare_velocity(
...
+        )
+        return Location(setpoint=setpoint, readback=readback)
+
+    def subscribe(self, function: Callback[dict[str, Reading[float]]]) -> None:
+        """Subscribe."""
+        self.user_readback.subscribe(function)
+
+    def clear_sub(self, function: Callback[dict[str, Reading[float]]]) -> None:
+        """Unsubscribe."""
+        self.user_readback.clear_sub(function)
```
DominicOram, line 252: Could: Might want a unit test covering just these motor changes?

tests/core/test_multi_derived_signal.py
```diff
@@ -0,0 +1,167 @@
...
+    jack1: F
+    jack2: F
+
+
+class TwoJackDerived(TypedDict, Generic[F]):
+    height: F
+    angle: F
+
+
+class TwoJackTransform(Transform):
```
DominicOram, line 48: Could: Couldn't we just reduce this all down to floats rather than get numpy involved:
```python
from math import tan, atan


class TwoJackRaw(TypedDict):
    jack1: float
    jack2: float


class TwoJackDerived(TypedDict):
    height: float
    angle: float


class TwoJackTransform(Transform):
    distance: float

    def raw_to_derived(self, *, jack1: float, jack2: float) -> TwoJackDerived:
        diff = jack2 - jack1
        return TwoJackDerived(
            height=jack1 + diff / 2,
            # need the case as returns numpy float rather than float64, but this
            # is ok at runtime
            angle=atan(diff / self.distance),
        )

    def derived_to_raw(self, *, height: float, angle: float) -> TwoJackRaw:
        diff = tan(angle) * self.distance
        return TwoJackRaw(
            jack1=height - diff / 2,
            jack2=height + diff / 2,
        )
```
Seems more readable?

tests/core/test_multi_derived_signal.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class Mirror(Device):
+    def __init__(self, name=""):
+        # Raw signals
+        self.x1 = SimMotor()
+        self.x2 = SimMotor()
+        # Parameter
+        self.x1_x2_distance = soft_signal_rw(float, initial_value=1)
+        # Derived signals
```
DominicOram, line 80: Should: Obviously we need docs in general on this but specifically defining raw, parameter and derived

tests/core/test_single_derived_signal.py
```diff
@@ -0,0 +1,161 @@
...
+        super().__init__(name=name)
+
+    def _get_position(self, x: float, y: float) -> BeamstopPosition:
+        if abs(x) < 1 and abs(y) < 2:
+            return BeamstopPosition.IN_POSITION
+        else:
+            return BeamstopPosition.OUT_OF_POSITION
+
+
+class Positioner(Device):
```
DominicOram, line 46: Could: At first glance it wasn't obvious how this related to the above. I would call it `MoveableBeamstop` and the one above `ReadOnlyBeamStop` or something like that

tests/core/test_single_derived_signal.py
```diff
@@ -0,0 +1,161 @@
...
+        super().__init__(name=name)
+
+    def _get_position(self, x: float, y: float) -> BeamstopPosition:
+        if abs(x) < 1 and abs(y) < 2:
+            return BeamstopPosition.IN_POSITION
+        else:
+            return BeamstopPosition.OUT_OF_POSITION
+
+    @AsyncStatus.wrap
+    async def set(self, value: BeamstopPosition) -> None:
```
DominicOram, line 64: Should: Naming this set is ambiguous as it means I can now do either `yield from bps.abs_set(device, IN)` or  `yield from bps.abs_set(device.position, IN)`? I would rename this to something like `_set_from_position` to be clear that the device is not designed to be used like `yield from bps.abs_set(device, IN)`.

tests/core/test_single_derived_signal.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        super().__init__(name=name)
+
+    def _get_position(self, x: float, y: float) -> BeamstopPosition:
+        if abs(x) < 1 and abs(y) < 2:
+            return BeamstopPosition.IN_POSITION
+        else:
+            return BeamstopPosition.OUT_OF_POSITION
+
+    @AsyncStatus.wrap
```
DominicOram, line 63: Nit: Do we have to do the wrapping here? Is there anyway we could just define an async func and `derived_signal_rw` do the wrapping?

tests/core/test_multi_derived_signal.py
```diff
@@ -0,0 +1,167 @@
...
+            jack1=self.x1,
+            jack2=self.x2,
+            distance=self.x1_x2_distance,
+        )
+        self.x = self._creator.derived_signal_rw(float, "height", "x")
+        self.roll = self._creator.derived_signal_rw(float, "angle", "roll")
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, derived: MirrorDerived) -> None:
```
DominicOram, line 93: Should: Is this designed to be externally callable? In which case the device should be `Movable`? Or if it's not meant to be externally callable rename from `set` for clarity

tests/core/test_single_derived_signal.py
```diff
@@ -0,0 +1,161 @@
...
+        # Raw signals
+        self.x = soft_signal_rw(float)
+        self.y = soft_signal_rw(float)
+        # Derived signals
+        self.position = derived_signal_rw(
+            self._get_position, self.set, x=self.x, y=self.y
+        )
+        super().__init__(name=name)
+
+    def _get_position(self, x: float, y: float) -> BeamstopPosition:
```
DominicOram, line 57: Nit: This is probably something we discussed already in person but it's a little annoying the type system won't catch this having incorrect arguments. I assume you've thought about it and there's no clean way round it. We can live with it

DominicOram (changes requested):
Thanks! I have just reviewed the tests as an example of how it would be used downstream and don't have time to review the implementation, sorry. Some of my comments may be a bit harsh/nitpicky but I was trying to review it from the perspective of "these are documented example cases that are the gold standard for how to use it".

Most of the comments are optional but I would like docs on this somewhere before I would approve

tests/core/test_multi_derived_signal.py
coretl: I will need this feature in 6 months as I need to transform a whole scanspec of points from derived to raw space, so I needed to make sure typing worked with that approach, however now I know it works I can remove it from the test and put it in an issue for the future...

tests/core/test_single_derived_signal.py
coretl: 3 days of head banging and I can't see a way of convincing the type system to check this...

I will make another ticket to add runtime checking of these at init time though

tests/core/test_single_derived_signal.py
coretl: Can you think of a usecase where we would want the `set` method publicly accessible on the device? Or is this never the case for a single derived signal?

src/ophyd_async/epics/motor.py
coretl: This is temporary until https://github.com/bluesky/ophyd-async/issues/828 so I'll cover tests as part of that ticket

src/ophyd_async/core/_derived_signal.py
```diff
@@ -0,0 +1,184 @@
...
+    Transform,
+    TransformT,
+)
+from ._device import Device
+from ._signal import SignalR, SignalRW, SignalT, SignalW
+from ._signal_backend import SignalDatatypeT
+from ._status import AsyncStatus
+
+
+class DerivedSignalCreator(Generic[TransformT]):
```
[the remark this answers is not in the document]
coretl: That's a much better name!

tests/core/test_multi_derived_signal.py
coretl: https://github.com/bluesky/ophyd-async/issues/830

tests/core/test_single_derived_signal.py
coretl: Nope, we can remove that

DominicOram (approved):
Great, thanks! I still haven't looked at the implementation in detail but the interface now looks good, just some minor suggestions, take them or leave them

src/ophyd_async/core/_derived_signal.py
```diff
@@ -10,27 +10,57 @@
...
     def __init__(
         self,
         transform_cls: type[TransformT],
-        set_derived: Callable[..., AsyncStatus] | None = None,
+        set_derived: Callable[..., Awaitable[None]] | None = None,
         **raw_and_transform_devices,
     ):
         self._set_derived = set_derived
+        # Check the raw and transform devices match the input arguments of the Transform
+        if transform_cls is not Transform:
```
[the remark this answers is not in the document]
DominicOram: https://github.com/bluesky/ophyd-async/issues/836

src/ophyd_async/testing/_single_derived.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, num_signals: int, name=""):
+        with self.add_children_as_readables():
+            self.signals = DeviceVector(
+                {i: soft_signal_rw(int, units="cts") for i in range(1, num_signals + 1)}
+            )
+        self.set_all = derived_signal_w(self._set_all, derived_units="cts")
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
```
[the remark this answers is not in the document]
coretl: correct

tests/core/test_multi_derived_signal.py
```diff
@@ -70,12 +72,33 @@ async def test_monitoring_position():
     assert results.empty()
 
 
-async def test_setting_position():
-    inst = Mirror("mirror")
+async def test_setting_position_stright_through():
```
DominicOram, line 75: Nit:
```suggestion
async def test_setting_position_straight_through():
```

DominicOram (approved):
One nit, otherwise looks good. Thanks! If you get this into a release I will try and get and convert some of our usecases in `dodal`

PR #668, opened 2024-11-22
Add a hint to PandA if we can't find a block

### LINKED ISSUE #574: Output custom error message when `fill_pvi_entries` fails on PAndA
As per https://github.com/DiamondLightSource/mx-bluesky/issues/463#issuecomment-2348452523 instead of erroring with `"create_directory" was not provided by pvi`, we should report `"create_directory" was not provided by pvi: is PandABlocks-ioc at least version x.x?`

We should do this by passing the string "is PandABlocks-ioc at least version x.x" to `fill_pvi_entries` in some way

coretl:
Fixes #574

### DIFF
--- a/src/ophyd_async/fastcs/panda/_hdf_panda.py
+++ b/src/ophyd_async/fastcs/panda/_hdf_panda.py
@@ -9,6 +9,8 @@ from ._block import CommonPandaBlocks
 from ._control import PandaPcapController
 from ._writer import PandaHDFWriter
 
+MINIMUM_PANDA_IOC = "0.11.4"
+
 
 class HDFPanda(CommonPandaBlocks, StandardDetector):
     def __init__(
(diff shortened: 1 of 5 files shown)


src/ophyd_async/fastcs/panda/_hdf_panda.py
```diff
@@ -19,7 +19,9 @@ def __init__(
         name: str = "",
     ):
         # This has to be first so we make self.pcap
-        connector = fastcs_connector(self, prefix)
+        connector = fastcs_connector(
+            self, prefix, error_hint="Is PandABlocks-ioc at least version 0.11.4?"
```
DominicOram, line 23: This is very likely to be forgotten about and not updated. Some suggestions:
* Should: Pull it out into a `MINIMUM_PANDA_IOC` variable - this should at least make it stand out more
* Do you have any where in you're PR/release process you could add a reminder to update it if necessary?
* We could just change the text to say "the latest panda IOC version" - I don't really like this but if it's forgotten about and suggests an old version it's actually worse

DominicOram (changes requested):
Thank you, one suggestion. @olliesilvester, could you add some docs somewhere on our side of how to actually test what panda IOC you're running and how to update it if required?

DominicOram (approved):
Much better, thanks!

PR #669, opened 2024-11-22
Add the python package and the config and make the contracts kept

### LINKED ISSUE #611: Add import-linter
To avoid making circular imports it would be good to make some rules and add a linter as per https://github.com/DiamondLightSource/mx-bluesky/pull/568 but with the config in a TOML file.

The rules:
- epics depends only on core
- tango depends only on core
- sim depends only on core
- fastcs depends only on core, epics, tango
- plan_stubs can depend on anything

## Acceptance Criteria
- Anything that breaks the above rules errors pre-commit

stan-dot:
implemented as discussed in the issue description

Fixes #611

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -49,6 +49,10 @@ jobs:
         with:
           python-version: ${{ inputs.python-version }}
           pip-install: ".[dev]"
+
+      - name: Run import linter
+        run: lint-imports
+
       - name: Run tests
         run: tox -e tests
 
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -50,9 +50,6 @@ jobs:
           python-version: ${{ inputs.python-version }}
           pip-install: ".[dev]"
 
-      - name: Run import linter
-        run: lint-imports
-
       - name: Run tests
         run: tox -e tests
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,14 +14,14 @@ jobs:
     uses: ./.github/workflows/_tox.yml
     with:
       tox: pre-commit,type-checking
-  
+
   test:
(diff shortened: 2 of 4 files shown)


.github/workflows/_test.yml
```diff
@@ -49,6 +49,10 @@ jobs:
         with:
           python-version: ${{ inputs.python-version }}
           pip-install: ".[dev]"
+
+      - name: Run import linter
+        run: lint-imports
+
```
coretl, line 55: This isn't needed as pre-commit runs the test on all files anyway

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     uses: ./.github/workflows/_tox.yml
     with:
       tox: pre-commit,type-checking
-
+
```
coretl, line 17: Please remove spurious whitespace

pyproject.toml
```diff
@@ -164,3 +165,41 @@ lint.preview = true # so that preview mode PLC2701 is enabled
...
+type = "forbidden"
+source_modules = "ophyd_async.tango"
+forbidden_modules = [
+    "ophyd_async.fastcs",
+    "ophyd_async.plan_stubs",
+    "ophyd_async.sim",
+    "ophyd_async.epics",
+]
+
+
```
coretl, line 200: Please can you also add the rules for sim?

coretl (changes requested):
Thanks! This is almost there...

pyproject.toml
stan-dot: ah yes I missed that one

.github/workflows/_test.yml
stan-dot: ah right

coretl (approved):

PR #673, opened 2024-11-25
Updates all Enums to UPPER_CASE across the code base

### LINKED ISSUE #605: Make all Enums `UPPER_CASE` across the code base
To match [conventions](https://docs.python.org/3.11/library/enum.html)

So:
```python
class MyEnum(str, Enum):
    FOO = "foo"
    BAR = "bar"
```

jennmald:
Resolves #605 
Fixes all enums to have the uppercase standard and updates all calls + tests to uppercase as well.

### DIFF
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -30,13 +30,13 @@ class DetectorTrigger(StrictEnum):
     """Type of mechanism for triggering a detector to take frames"""
 
     #: Detector generates internal trigger for given rate
-    internal = "internal"
+    INTERNAL = "internal"
     #: Expect a series of arbitrary length trigger signals
-    edge_trigger = "edge_trigger"
+    EDGE_TRIGGER = "edge_trigger"
     #: Expect a series of constant width external gate signals
-    constant_gate = "constant_gate"
+    CONSTANT_GATE = "constant_gate"
     #: Expect a series of variable width external gate signals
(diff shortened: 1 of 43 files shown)


coretl:
Excellent! Thanks for this. As it looks like we can't check with ruff, we should put checking in `StrictEnum` and `SubsetEnum`. Please can you replace the definition of `StrictEnum` and `SubsetEnum` in `ophyd_async/core/_utils.py` with the following:
```python
class StrictEnumMeta(EnumMeta):
    def __new__(metacls, *args, **kwargs):
        ret = super().__new__(metacls, *args, **kwargs)
        bad_names = [x.name for x in ret if not x.name.isupper()]  # type: ignore
        if bad_names:
            raise TypeError(f"Names {bad_names} should be uppercase")
        return ret


class StrictEnum(str, Enum, metaclass=StrictEnumMeta):
    """All members should exist in the Backend, and there will be no extras"""


class SubsetEnumMeta(StrictEnumMeta):
    def __call__(self, value, *args, **kwargs):  # type: ignore
        if isinstance(value, str) and not isinstance(value, self):
            return value
        return super().__call__(value, *args, **kwargs)


class SubsetEnum(StrictEnum, metaclass=SubsetEnumMeta):
    """All members should exist in the Backend, but there may be extras"""
```

Hopefully the tests will still pass after this and we can merge...

coretl (approved):

PR #675, opened 2024-11-25
Make name separator configurable

### LINKED ISSUE #666: Naming of signals in ophyd-async is different to ophyd.v1
In ophyd v1 the names of signals of components of a device are created by joining names with a `_`

so a device like:

```python

class Stage(Device):
    x = Cpt(EpicsMotor, 'X')
    y = Cpt(EpicsMotor, 'Y')

class StageOfStage(Device):
    a = Cpt(Stage, 'A')
    b = Cpt(Stage, 'B')

stage_of_stage = StageOfStage('TEST:', name='stage_of_stage'

```

would have component names like: 

```
stage_a_x
stage_a_y
stage_b_x
stage_b_y
```

In ophyd-async component names are joined with a `-` character instead and any `_` characters are removed from the child name. 
```
            child_name = f"{self.name}-{child_name.strip('_')}" if self.name else ""

```
 https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/_device.py#L113

If I include `ophyd.v1` and `ophyd-async` devices in a plan together then the signals in event documents from `ophyd-async` and `ophyd.v1` devices look different. 

## Why is this useful?

@coretl 

> The reason it was changed it because given a dash separated name you can split on the dash and walk the tree of devices from the root. With an underscore separated name you cannot. 

@DiamondJoseph suggests it's possible to iterate through the descriptor document to walk the tree. (I've not tried this)

## Why is this a problem?

1. Inconsistency in the signal names in documents is confusing when using ophyd v1 and ophyd-async in the same place
2. The `-` can be interpreted as a -Ve sign in some cases e.g if the signal is monitored in a plan and then you try to access the stream from the run using the databroker client. 

## Proposal

Standardise the signal naming of signals to be the same as in ophyd.v1

coretl:
Remove the stripping of underscore while we're at it, this can be done by overriding set_name if required.

Fixes #666

### DIFF
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -71,13 +71,16 @@ class Device(HasName, Connectable):
     _connect_task: asyncio.Task | None = None
     # The mock if we have connected in mock mode
     _mock: LazyMock | None = None
+    # The separator to use when making child names
+    _child_name_separator: str = "-"
 
     def __init__(
         self, name: str = "", connector: DeviceConnector | None = None
     ) -> None:
         self._connector = connector or DeviceConnector()
         self._connector.create_children_from_annotations(self)
-        self.set_name(name)
+        if name:
+            self.set_name(name)
 
     @property
     def name(self) -> str:
@@ -97,21 +100,30 @@ class Device(HasName, Connectable):
             getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
         )
 
-    def set_name(self, name: str, separator: str = "-"):
+    def set_name(self, name: str, *, child_name_separator: str | None = None) -> None:
         """Set ``self.name=name`` and each ``self.child.name=name+"-child"``.
 
         Parameters
         ----------
         name:
             New name to set
+        child_name_separator:
+            Use this as a separator instead of "-". Use "_" instead to make the same
+            names as the equivalent ophyd sync device.
         """
         self._name = name
+        if child_name_separator:
+            self._child_name_separator = child_name_separator
         # Ensure logger is recreated after a name change
         if "log" in self.__dict__:
             del self.log
-        for child_name, child in self.children():
-            child_name = f"{self.name}{separator}{child_name}" if self.name else ""
-            child.set_name(child_name, separator)
+        for attr_name, child in self.children():
+            child_name = (
+                f"{self.name}{self._child_name_separator}{attr_name}"
+                if self.name
+                else ""
+            )
+            child.set_name(child_name, child_name_separator=self._child_name_separator)
 
     def __setattr__(self, name: str, value: Any) -> None:
         # Bear in mind that this function is called *a lot*, so
@@ -147,6 +159,10 @@ class Device(HasName, Connectable):
         timeout:
             Time to wait before failing with a TimeoutError.
         """
+        assert hasattr(self, "_connector"), (
+            f"{self}: doesn't have attribute `_connector`,"
+            " did you call `super().__init__` in your `__init__` method?"
+        )
         if mock:
             # Always connect in mock mode serially
             if isinstance(mock, LazyMock):
@@ -247,6 +263,8 @@ class DeviceCollector:
     set_name:
         If True, call ``device.set_name(variable_name)`` on all collected
         Devices
+    child_name_separator:
+        Use this as a separator if we call ``set_name``.
     connect:
         If True, call ``device.connect(mock)`` in parallel on all
         collected Devices
@@ -271,11 +289,13 @@ class DeviceCollector:
     def __init__(
         self,
         set_name=True,
+        child_name_separator: str = "-",
         connect=True,
         mock=False,
         timeout: float = 10.0,
     ):
         self._set_name = set_name
+        self._child_name_separator = child_name_separator
         self._connect = connect
         self._mock = mock
         self._timeout = timeout
@@ -311,7 +331,7 @@ class DeviceCollector:
         for name, obj in self._objects_on_exit.items():
             if name not in self._names_on_enter and isinstance(obj, Device):
                 if self._set_name and not obj.name:
-                    obj.set_name(name)
+                    obj.set_name(name, child_name_separator=self._child_name_separator)
                 if self._connect:
                     connect_coroutines[name] = obj.connect(
                         self._mock, timeout=self._timeout
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -97,7 +97,7 @@ class Device(HasName, Connectable):
             getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
         )
 
-    def set_name(self, name: str):
+    def set_name(self, name: str, separator: str = "-"):
         """Set ``self.name=name`` and each ``self.child.name=name+"-child"``.
 
         Parameters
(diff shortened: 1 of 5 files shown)


tacaswell:
- positional only for the new kwarg?
- should it be threaded through `__init__` in at least some of the examples?

Tom-Willemsen:
This is fine for us at ISIS, and corresponds with my understanding of the dev meeting yesterday. If I was going to use a custom separator it would be useful to have it mentioned briefly in the docs somewhere.

src/ophyd_async/core/_device.py
```diff
@@ -97,7 +97,7 @@ def log(self) -> LoggerAdapter:
             getLogger("ophyd_async.devices"), {"ophyd_async_device_name": self.name}
         )
 
-    def set_name(self, name: str):
+    def set_name(self, name: str, separator: str = "-"):
```
DominicOram, line 100: Should: Would be good to have this in the docstring

DominicOram (changes requested):
Generally looks good but I think we need something in the docs about the tradeoffs of setting it and the reasoning behind making it different by default

coretl:
> * should it be threaded through `__init__` in at least some of the examples?

Thinking a bit more about this, I'm wondering if we should remove the ability to set name from `__init__` in the first place. It's annoying having to thread it through everywhere, and we generally call `set_name` on the device after it is created anyway. To ease the usage we could return `self` from `set_name`, possibly renaming it to `named` while we're at it, so instead of:
```python
device = MyDevice("PV_PRE", name="foo")
```
we would do
```python
device = MyDevice("PV_PRE").named("foo")
```
This would make it obvious where to put the separator if we wanted to override it without having to thread it through init:
```python
device = MyDevice("PV_PRE").named("foo", separator="_")
```

> * positional only for the new kwarg?

I'm not convinced, I think the name adds something, maybe it should be `child_name_separator` like you've done in https://github.com/bluesky/ophyd/pull/1222 though?


@tacaswell @DominicOram @Tom-Willemsen thoughts?

Tom-Willemsen:
> we would do
>
> `device = MyDevice("PV_PRE").named("foo")`

That's a bigger change than originally proposed - that'll need every `ophyd-async` device and corresponding instantion functions updating, right? At ISIS that's not a huge deal, we've only deployed to one beamline and a handful of test machines so far, but I can imagine it's going to be a bigger churn for `dodal` etc at Diamond?

---

Under that model our instantiation wrappers for lots of devices at ISIS would basically look like:

```python
# When bluesky name is related to PV name (which is common for us at ISIS)
def my_device_1(some, args, name):
    return MyDevice1(some, args, name).named(name)

# Or for when we want to have a default name for the device,
# only overridden in special cases (99% of the time "THING" is correct).
def my_device_2(some, args, name="THING"):
    return MyDevice2(some, args).named(name)
```

I don't necessarily _mind_ doing that, the above is quite simple if a bit repetitive, but especially in the second case it's an extra layer of indirection that we need to set up as a consumer of `ophyd-async`, and an extra thing downstream consumers need to remember to do if they ever want to create `MyDevice2` directly.

---

We could live with either case, but I'd say I slightly prefer the existing approach.

tacaswell:
I like keyword only as comparing

```python
obj.set_name('foo', '-')
obj.set_name('foo', separator='-')
obj.set_name('foo', child_name_separator='-')
```

the first is pretty unclear.  I like the longer name, but at NSLS-II we have the rule "Don't let Tom name things".

----

There is a gentle path to adding `obj.named()` as you don't have to on day 1 make `Obj(..., name='bob')` fail as the two methods (set in `__init__` and "fix later") are already co-existing.

coretl:
> That's a bigger change than originally proposed - that'll need every `ophyd-async` device and corresponding instantion functions updating, right? At ISIS that's not a huge deal, we've only deployed to one beamline and a handful of test machines so far, but I can imagine it's going to be a bigger churn for `dodal` etc at Diamond?

True, but we are about to move to a model where the name of the instantiation function is also used for the returned device, so shouldn't be used in more than a handful of cases...


> There is a gentle path to adding `obj.named()` as you don't have to on day 1 make `Obj(..., name='bob')` fail as the two methods (set in `__init__` and "fix later") are already co-existing.

Ok, I hear the message about a gentle path, but I would also like to get 1.0 out soon and passing `name` in init is a making it a little messy. There is also `DeviceCollector` which names things and will continue to work without change.

I suggest:
- Add `Device.named(name: str, *, child_name_separator="-") -> Self`
- Make `Device.set_name` a deprecated wrapper to the above that emits a warning
- Make `Device.__init__(name="something-non-blank")` a deprecated argument that emits a warning
- Remove `set_name` and the init arg for 1.0

What do you reckon?

coretl:
@DiamondJoseph what's happening with the device factory wrapper in dodal? Am I right in thinking that this wouldn't cause us too much pain?

DiamondJoseph:
> what's happening with the device factory wrapper in dodal? Am I right in thinking that this wouldn't cause us too much pain?

The device_factory decorator is back in, and [can name the device if it has not been named prior to this when called](https://github.com/DiamondLightSource/dodal/blob/34d2b0d0435f89d6628fd99280032b02bd673c7c/src/dodal/utils.py#L201) - which is almost certainly done prior to connect? We can modify this call however we want to.

It's only in use on i22, training_rigs, with PRs for non-MX beamlines open while I track down the state of those beamlines.

coretl:
Ok, a slight modification to the suggestion:
- Keep `Device.set_name(name: str, *, child_name_separator="-") -> None`
- Add a utility function `ophyd_async.core.named_device(device: T, name: str, *, child_name_separator="-") -> T`
- Make Device.__init__(name="something-non-blank") a deprecated argument that emits a warning
- Remove the init arg for 1.0

Then we can still write our one-liner if we need to:
```python
device = named_device(MyDevice("PV_PRE"), "foo", separator="_")
```

whs92:
@coretl does the utility function allow us to set the child separator for all ophyd-async devices in a particular environment? 

Or is the intention we have to do it per instance or per class definition? 

Sorry I've not had enough time to follow this discussion in depth after the call in Monday.

whs92:
> No, but we can make DeviceCollector do that

Then for me,  this works

coretl:
I baulked at the sheer number of tests that would have to change if we deprecated passing `name` to `Device.__init__`, so you'll be pleased to hear this is now back to the original scope of the PR

src/ophyd_async/core/_device.py
```diff
@@ -71,13 +71,16 @@ class Device(HasName, Connectable):
...
+    _child_name_separator: str = "-"
 
     def __init__(
         self, name: str = "", connector: DeviceConnector | None = None
     ) -> None:
         self._connector = connector or DeviceConnector()
         self._connector.create_children_from_annotations(self)
-        self.set_name(name)
+        if name:
+            self.set_name(name)
```
DominicOram, line 83: Nit: Without the refactor this change is a bit redundant? Unless we don't trust the the type system? But it does no harm

DominicOram (approved):
Minor comment, take it or leave it, otherwise all good from my perspective.

src/ophyd_async/core/_device.py
coretl: It is a tiny performance improvement to not set the child names if the parent name isn't set

PR #606, opened 2024-10-08
Split out common AD file plugin logic into core writer class, create ADTiffWriter

jwlodek:
This now seems to mostly work, at least from the collection side. ~~The data cannot be read back via tiled at the moment, however, so for now marking as draft, since I may need to edit the docs generated somewhat to account for this~~. PRs into `bluesky` and `tiled` have been merged to allow reading back collected TIFF data with this branch.

Addresses #595 .

TODO:

- [x] Add tests for tiff writer
- [x] Make sure we can read the data back through tiled
- [x] Make sure HDF writer still works as it did.

PR #679, opened 2024-11-27
Port andor to ophyd-async

DiamondJoseph:
Moves work by @Relm-Arrowny to ophyd-async from https://github.com/DiamondLightSource/dodal/pull/807

Adds Andor2 AreaDetector implementation in standard ophyd-async form: ophyd-async aims to unify all StandardDetector implementations prior to 1.0, so moving implementation here to ensure requirements remain met and to ensure device updates smoothly.

### DIFF
--- a/src/ophyd_async/epics/adandor/_andor_controller.py
+++ b/src/ophyd_async/epics/adandor/_andor_controller.py
@@ -1,0 +1,65 @@
+import asyncio
+
+from ophyd_async.core import (
+    AsyncStatus,
+    DetectorController,
+    DetectorTrigger,
+    TriggerInfo,
+    set_and_wait_for_value,
+)
+from ophyd_async.epics.adcore import (
+    set_exposure_time_and_acquire_period_if_supplied,
+    stop_busy_record,
+)
+
+from ._andor_io import Andor2DriverIO, Andor2TriggerMode, ImageMode
+
+_MIN_DEAD_TIME = 0.1
+_MAX_NUM_IMAGE = 999_999
+
+
+class Andor2Controller(DetectorController):
+    def __init__(
+        self,
+        driver: Andor2DriverIO,
+    ) -> None:
+        self._drv = driver
+        self._arm_status: AsyncStatus | None = None
+
+    def get_deadtime(self, exposure: float | None) -> float:
+        return _MIN_DEAD_TIME + (exposure or 0)
+
+    async def prepare(self, trigger_info: TriggerInfo):
+        await set_exposure_time_and_acquire_period_if_supplied(
+            self, self._drv, trigger_info.livetime
+        )
+        await asyncio.gather(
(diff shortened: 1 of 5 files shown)


DiamondJoseph:
I have not touched the ophyd-async StandardDetectors in a while, so I'd appreciate some careful eyes. I'm going to try and connect it to p99 this week.

coretl:
@jwlodek is making a much nicer AreaDetector baseclass in #606 so bear in mind that this will need changing after this, whether we merge this before or after

DiamondJoseph:
Yep, I just wanted the responsibility to update this here rather than in dodal :D

jwlodek:
I've made a new PR that slightly modifies these to work with the new base classes. Tests are all passing, but would be good to test with the real thing if we can.

Relm-Arrowny:
> I've made a new PR that slightly modifies these to work with the new base classes. Tests are all passing, but would be good to test with the real thing if we can.
I can try fire it up, I guess we just want to take a few imagie and make sure it all still work?

src/ophyd_async/epics/adandor/_andor_controller.py
```diff
@@ -0,0 +1,65 @@
...
+        self,
+        driver: Andor2DriverIO,
+    ) -> None:
+        self._drv = driver
+        self._arm_status: AsyncStatus | None = None
+
+    def get_deadtime(self, exposure: float | None) -> float:
+        return _MIN_DEAD_TIME + (exposure or 0)
+
+    async def prepare(self, trigger_info: TriggerInfo):
```
evalott100, line 32: Could we not subclass `TriggerInfo` like eiger does, making an `Andor2TriggerInfo` which contains `trigger: Andor2TriggerMode` instead of `trigger: DetectorTrigger`?

src/ophyd_async/epics/adandor/_andor_controller.py
```diff
@@ -0,0 +1,65 @@
...
+            DetectorTrigger.INTERNAL: Andor2TriggerMode.INTERNAL,
+            DetectorTrigger.EDGE_TRIGGER: Andor2TriggerMode.EXT_TRIGGER,
+        }
+        if trigger not in supported_trigger_types:
+            raise ValueError(
+                f"{self.__class__.__name__} only supports the following trigger "
+                f"types: {supported_trigger_types} but was asked to "
+                f"use {trigger}"
+            )
+        return supported_trigger_types[trigger]
```
evalott100, line 62: **Nit:** If we can't make a new `TriggerInfo` for Andor, then it might nicer to replace this code with a `from_detector_trigger(cls, trigger: DetectorTrigger) -> Andor2TriggerMode` method on `Andor2TriggerMode`.

coretl:
Closing in favour of #724. @evalott100 please could you review that one instead?

Issue #667, opened 2024-11-22
Make the Tango tests faster


coretl:
EPICS tests take 15s to run. Tango tests take 85s to run. This makes waiting for CI painful. I think a target of 20s for Tango tests is reasonable.

Some ideas:
- Some tests take a long time because of long timeout or many repeats of the same code. See if these can be reduced.
- All tango tests are currently forked, because this is apparently necessary, but when there is a module level fixture then we maybe we could get away with forking then running multiple tests? Test this theory by taking one of the parameterized tests and turning the parameters into a for loop inside the tests. If this is a good strategy then we could work out if it possible to extend `pytest-forked` to support forked fixtures, where every test that uses the fixture runs in the same fork.
- Some tests may be duplicates and could be deleted

## Acceptance Criteria
- Tango tests take 20s to run

PR #684, opened 2024-11-29
Faster tango tests

jsouter:
Closing https://github.com/bluesky/ophyd-async/issues/667

Primarily reduces the number of tests by moving parametrized parameters inside the test -> reduces the total number of forks. Am investigating if it's possible to fork after the fixtures are loaded in case we want to retain the nice logging that prints pass/fail per parametrized test.

Also some small QOL changes, using a dataclass to make looping over ATTRIBUTES_SET nicer

### DIFF
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -189,7 +189,7 @@ class AttributeProxy(TangoProxy):
                 raise TimeoutError(f"{self._name} attr put failed: Timeout") from te
             except DevFailed as de:
                 raise RuntimeError(
-                    f"{self._name} device" f" failure: {de.args[0].desc}"
+                    f"{self._name} device failure: {de.args[0].desc}"
                 ) from de
 
         else:
--- a/tests/tango/test_tango_signals.py
+++ b/tests/tango/test_tango_signals.py
@@ -402,155 +402,154 @@ async def test_backend_get_put_monitor_cmd(
 
 # --------------------------------------------------------------------
 @pytest.mark.asyncio
-@pytest.mark.parametrize("use_proxy", [True, False])
 async def test_tango_signal_r(
     echo_device: str,
-    use_proxy: bool,
 ):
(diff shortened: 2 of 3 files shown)


evalott100:
As part of this, can we change from using

https://github.com/bluesky/ophyd-async/blob/629e764b2c1e5cc7418b1c9c81edae0e93355297/tests/conftest.py#L240-L245

to putting an explicit `@pytest.mark.forked` on the tests (we can remove the `@pytest.mark.asyncio`).

jsouter:
> As part of this, can we change from using
> 
> https://github.com/bluesky/ophyd-async/blob/629e764b2c1e5cc7418b1c9c81edae0e93355297/tests/conftest.py#L240-L245
> 
> to putting an explicit `@pytest.mark.forked` on the tests (we can remove the `@pytest.mark.asyncio`).

Yep I like this, potentially makes it easier to unfork specific tests (though currently don't think there are any that can be).

evalott100:
I'm having much difficulty trying to adjust tango tests to work in 3.12.

I'll wait for this to be merged before I continue trying.

jsouter:
> I'm having much difficulty trying to adjust tango tests to work in 3.12.
> 
> I'll wait for this to be merged before I continue trying.

Looks like tests are broken on main for me right now, are you seeing that?

jsouter:
@evalott100 looks like pytest-forked is a bit hacked together, and marking the individual tests gives us this error, so may be simplest for me to roll back that last change (see https://github.com/pytest-dev/pytest/issues/9621)

```
AssertionError: previous item was not torn down properly
===========short test summary info =============
ERROR tests/test_cli.py::test_cli_version - AssertionError: previous item was not torn down properly
````

evalott100:
> @evalott100 looks like pytest-forked is a bit hacked together, and marking the individual tests gives us this error, so may be simplest for me to roll back that last change

Ah no worries, strange problem.

coretl:
@jsouter should I review this now?

jsouter:
> @jsouter should I review this now?

Yep, I think it's in an okay state

coretl (approved):

PR #686, opened 2024-11-29
use ophyd async logger for logging.debug statements

rerpha:
in an interactive shell we found we got a global logging handler added after a PV failed to connect. 

We found these logging.debug statements which use the "simple"/global logging API which is not recommended as per: https://docs.python.org/3/library/logging.html#logging.debug

where it mentions this: 
> The only difference is that if the root logger has no handlers, then [basicConfig()](https://docs.python.org/3/library/logging.html#logging.basicConfig) is called, prior to calling debug on the root logger.
> For very short scripts or quick demonstrations of logging facilities, debug and the other module-level functions may be convenient. However, most programs will want to carefully and explicitly control the logging configuration, and should therefore prefer creating a module-level logger and calling [Logger.debug()](https://docs.python.org/3/library/logging.html#logging.Logger.debug) (or other level-specific methods) on it, as described at the beginnning of this documentation.

This PR gets the ophyd-async logger  and uses it instead of the global functions.

### DIFF
--- a/.github/pages/make_switcher.py
+++ b/.github/pages/make_switcher.py
@@ -33,7 +33,7 @@ def get_versions(ref: str, add: str | None) -> list[str]:
         builds = set(get_branch_contents(ref))
     except CalledProcessError:
         builds = set()
-        logging.warning(f"Cannot get {ref} contents")
+        logging.warning(f"Cannot get {ref} contents")  # noqa: LOG015
 
     # Add and remove from the list of builds
     if add:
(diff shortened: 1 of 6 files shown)


Tom-Willemsen:
- Looks like we missed one in https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/_utils.py#L117 (detected via ruff rule `LOG015`)
- Maybe we could consider selecting ruff rule `LOG015` to make this a lint error?

coretl (approved):

Issue #687, opened 2024-11-29
 Motors don't stop


Tom-Willemsen:
`ophyd-async` currently uses an `epics_signal_x` to write to the `.STOP` field of a motor to stop it.

This doesn't work, as the [motor record documentation](https://epics.anl.gov/bcda/synApps/motor/motorRecord.html#Fields_command) specifies, `1` must explicitly be written:

> When this field is set to 1, the record will immediately reset it to 0, and the motor will decelerate to a stop. When the motor has stopped, VAL will be set equal to RBV, and DVAL will be set equal to DRBV. (This ensures that the motor will not start moving the next time the record is processed, unless a drive field is explicitly changed. If you want the motor to pause, use the SPMG field.)

## Steps To Reproduce
1. Spin up a motor IOC
2. Set off a long-running move
3. Run:
```python
from ophyd_async.epics.motor import Motor
from ophyd_async.plan_stubs import ensure_connected
import bluesky.plan_stubs as bps

def stop_plan():
    mot = Motor("some_motor_pv")
    yield from ensure_connected(mot)
    yield from bps.stop(mot)

RE(stop_plan())
```
4. Motor does not stop

## Acceptance Criteria
- [ ] Calling `stop` on a motor (e.g. via `bps.stop(motor)`) stops a motor.

PR #688, opened 2024-11-29
Ensure motors can be stopped

Tom-Willemsen:
Fixes https://github.com/bluesky/ophyd-async/issues/687

As noted on the issue, the problem is that `1` needs to explicitly be written to the motor record for it to actually stop.


coretl (approved):
Thanks!

PR #689, opened 2024-12-03
adjusted tests to avoid `pytest.approx` `bool` comparison

evalott100:
pytango fails in it's internal test utils, so we will have to wait for that to be fixed before we add `"boolean`" back to the tango signal tests.


coretl (approved):

burkeds (commented):
Looks good to me

PR #695, opened 2024-12-05
Move testing utils from core to testing

coretl:
The following have moved to `ophyd_async.testing`:
    - `assert_configuration`
    - `assert_emitted`
    - `assert_reading`
    - `assert_value`
    - `callback_on_mock_put`
    - `get_mock`
    - `get_mock_put`
    - `mock_puts_blocked`
    - `reset_mock_put_calls`
    - `set_mock_put_proceeds`
    - `set_mock_value`
    - `set_mock_values`
    - `wait_for_pending_wakeups`

This will make it clear that they are only to be used for testing and not clutter the `core` namespace

## Changes
Change import paths
```python
# old
from ophyd_async.core import set_mock_value
# new
from ophyd_async.testing import set_mock_value
```


dperl-dls (approved):
Looks good to me, I like this much better! But what happened to `Connectable`?

coretl:
> Looks good to me, I like this much better! But what happened to `Connectable`?

It wasn't used apart from in `Device`, and it wasn't quite right, so I deleted it rather than fixing it...

PR #700, opened 2024-12-09
Add 0.10 docs

### LINKED ISSUE #193: Clearer documentation on epics demo
Currently, the epics demo module includes facility to start a Python IOC with all of the PVs that the demo devices point at. It can be tested and played with via `ipython -i docs/user/examples/epics_demo.py`. There is an expectation that a developer makes sure the Python soft IOC actually contains all the needed PVs when they add new demo objects. 

This process needs to be documented somewhere and possibly made easier to test. Some suggestions:

- Add a tutorial page to the docs on how to use the demo module
- Create an entry point for running up the demo (e.g. `> ophyd-async demo`)
- Add a test that proves the demo IOC starts
- Add a test that proves the demo module starts

The one thing we can't automate is instantiation of the demo devices, since that's manually embedded inside `with DeviceCollector` in `docs/user/examples/epics_demo.py`, unless we can also think of a way to do that. It would be very nice to have the CI automatically flag up if someone has broken the demo or left it incomplete.

### LINKED ISSUE #227: Enable Sphinx nitpick
Sphinx nitpick mode has always been disabled since the project was initially created as part of the ophyd repository.
This means the docs may be full of links that are intended to be linking to internal and external docstrings.
This should be re-enabled and the references that are not working should be fixed.
The copier template attempted to set this to true but this blocked other issues from progressing.

n.b. this is in docs/conf.py

### LINKED ISSUE #246: Remove unwanted Python dependencies
As part of [this commit](https://github.com/bluesky/ophyd-async/commit/a9f13d6c6892fb59cbf1237ce7f1c78cf778ab52#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) a large number of dependencies were added to pyproject.toml. This includes "ophyd", as well as "pyepics", which can cause issues (crashes) due to conflicts with the existing epicscorelibs installation.

The list of dependencies should be reviewed and cleaned up.

### LINKED ISSUE #303: Write docs for new test util functions
We should add all the new [mock utility functions](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/mock_signal_utils.py) to the docs. This will involve making new tests within [test demo](https://github.com/bluesky/ophyd-async/blob/main/tests/epics/demo/test_demo.py).

https://github.com/bluesky/ophyd-async/blob/1e8423347249b2ab34f2abbe04f9790818cac3b6/docs/how-to/write-tests-for-devices.rst?plain=1#L33-L50

### LINKED ISSUE #349: Clearer documentation on how to use the reading context manager in a `StandardReadable`
See https://github.com/DiamondLightSource/dodal/pull/575#discussion_r1622002749
[
```I did initially consider the using the the context manager as shown in the tutorial but it wasn't clear to me from the name of HintedSignal what it was and I couldn't find any documentation clarifying this. read and config made more sense and there were more existing examples using this format.```]

### LINKED ISSUE #386: Document strategies of waiting for a signal to match a value, pushing it down to EPICS or checking in the device or plan
As a developer I would expect `set` on a `SignalRW` to only finish when the readback of the signal matches the setpoint. This would, I think match the behaviour of `ophyd`.

### LINKED ISSUE #443: fix pydantic models in API
In https://github.com/bluesky/ophyd-async/pull/430

We include `autodoc-pydantic` in order for the pydantic models to be included in the API. `autodoc-pydantic` seems to allow for warningless generation of docs, but doesn't generate entries for the models in the API.

### LINKED ISSUE #459: Fix docs build
It is currently reporting:
```
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.areadetector' with 'areadetector'.
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.demo' with 'demo'.
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.motion' with 'motion'.
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.pvi' with 'pvi'.
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.signal' with 'signal'.
WARNING: Summarised items should not include the current module. Replace 'ophyd_async.epics.demo.demo_ad_sim_detector' with 'demo_ad_sim_detector'.
```

### LINKED ISSUE #479: from ophyd v1 to ophyd-async (was v2?)
Hi !

Prior to try ophyd-async, and prior to embarking into some heavy development, I am wondering if any
guide / tutorial / documentation / notes exist to tell how to do a transition from Ophyd v1 to the new version.

What are the differences ? (apart from the asynchronous execution model)
What is compatible ?
What is incompatible ?
What has to be changed ?

Thanks in advance for some hints.

### LINKED ISSUE #480: Investigate if T can be imported per file that uses it instead of via utils
T was moved into utils and imported from there to prevent sphinx from importing it for every single file for the docs. It would be preferable to have `T = TypeVar("T")` defined in each file that uses it as this is a common use case. We need to investigate with sphinx and see if we can resolve it.

### LINKED ISSUE #484: Update docs to reflect new structure for making epics devices
We need a how to in the doc section for how to add support for a new epics device. This needs to tell people how to structure it to follow the defined conventions.

### LINKED ISSUE #622: There should also be an explanation page about devices, signals and backends are now structured and built
Should: I like this and the how-to, but I think there should also be an explanation page about devices, signals and backends are now structured and built. I had to read through most of `core` to figure it out and it would be useful to document to onboard new people.

_Originally posted by @callumforrester in https://github.com/bluesky/ophyd-async/pull/594#discussion_r1812405921_

### LINKED ISSUE #671: Add docs on waiting for signals
Include this use case:

> I want to use ophyd_async.core.set_and_wait_for_other_value() to write to a setpoint of a temperature controller and wait until the temperature reaches that value. I am worried that because they're both floats and because temperature is analog that they will never EXACTLY equal to each other and this function may not return when I want it to. How should I handle float precision here?
> 2 replies
> Tom Cobb (DLS)
> [15 minutes ago](https://mattermost.hzdr.de/bluesky/pl/6xozspka6pnniqqrbzkmcak35e)
> 
> In main, you can pass a callable as match_value to the function, so you could pass lambda v: np.isclose(v, 32.567) to that
> [12m ago](https://mattermost.hzdr.de/bluesky/pl/cucrigsfxtnptnn1qc6uhxkzny)
> 
> Or you could pass pytest.approx(32.567) to it

## Acceptance Criteria
- Docs tell you how to wait for a signal to be at a value to within a tolerance

### LINKED ISSUE #677: Unclear Annotation in Documentation Example
In the [Readable section of the Make a Simple Device how-to guide](https://blueskyproject.io/ophyd-async/main/how-to/make-a-simple-device.html#readable), the type annotation in the example code reference `A`:
```
class Sensor(StandardReadable, EpicsDevice):
    """A demo sensor that produces a scalar value based on X and Y Movers"""

    value: A[SignalR[float], PvSuffix("Value"), Format.HINTED_SIGNAL]
    mode: A[SignalRW[EnergyMode], PvSuffix("Mode"), Format.CONFIG_SIGNAL]
```

I am not sure it is clear that `A` refers to `typing.Annotated`, as the import alias is not shown.

Acceptance Criteria:
- The documentation should be readable by a programmer unfamiliar with the codebase.

### PR DESCRIPTION
**Pull request by coretl**
Rewrite of the documentation and its structure including:
- Making sim and epics.demo devices do the same
- Placeholder for the same for tango.demo and fastcs.demo devices
- New tutorials, how-tos and explanations
- Reduce installed dependencies

The first half of #721 
Fixes #677 
Fixes #671
Fixes #484 
Fixes #480
Fixes #479
Fixes #459 
Fixes #443 (well enough anyway)
Fixes #386
Fixes #349 
Fixes #303 
Fixes #246 
Fixes #227 
Fixes #193 
Fixes #622

### CODE REVIEW

**Comment by coretl** [comment]
@evalott100 please can you make a new branch off this one, then PR back to the docs branch fixing some of the warnings in https://github.com/bluesky/ophyd-async/actions/runs/12911955696/job/36005625932?

Some of them are just missing docstrings, but there are a bunch that say something like:
```
docstring of ophyd_async.core._signal.set_and_wait_for_value:1: WARNING: py:class reference target not found: ophyd_async.core._signal_backend.SignalDatatypeT
``` 
I think this is because importing `ophyd_async.core._signal_backend.SignalDatatypeT` into `ophyd_async.core` loses its doc-comment (doc-comment because it is a TypeVar not a class). I have no idea how to get sphinx to recognise that `SignalDatatypeT` exists and is of a particular type.

I have tried:
- Adding the following to `ophyd_async.core.__init__`:
```python
"""
.. data:: SignalDatatypeT

    The supported `Signal` datatypes
"""
```
- Doing the same with `.. class:: SignalDatatypeT`
- Doing the same but in `ophyd_async.core._signal_backend`
- Peppering the codebase with `from __future__ import annotations`

My next steps were going to be:
- Add `from __future__ import annotations` for everything (don't know if you can do this for the whole module)
- Use [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases) to map `ophyd_async.core._signal_backend.SignalDatatypeT` to `ophyd_async.core.SignalDatatypeT` as defined by one of the methods above
- Do some googling

coretl:
@evalott100 please could you give this another review?

I propose to merge this in approximately this state with further work in other tickets:
- Docstring format not decided yet: #755 
- AreaDetector modules need to be more uniform: https://github.com/bluesky/ophyd-async/issues/754

I will write up the last remaining tutorial (detectors) as part of another PR

jwlodek (dismissed):
Found a few typos

docs/explanations/design-goals.md
```diff
@@ -0,0 +1,67 @@
+# Design goals and differences with ophyd sync
+
+Ophyd-async was designed to be a library for asynchronously interfacing with hardware. As such it fulfils the same role in the bluesky ecosystem as [ophyd sync](https://github.com/bluesky/ophyd): an abstraction layer that enables experiment orchestration and data acquisition code to operate above the specifics of particular devices and control systems. This document details the design goals and the differences with ophyd sync.
```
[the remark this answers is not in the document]
coretl: Ha, US vs UK spelling... I think I'll leave it at the UK one...

src/ophyd_async/epics/demo/_point_detector_channel.py
```diff
@@ -0,0 +1,21 @@
...
+    """Energy mode for `DemoPointDetectorChannel`"""
+
+    #: Low energy mode
+    LOW = "Low Energy"
+    #: High energy mode
+    HIGH = "High Energy"
+
+
+class DemoPointDetectorChannel(StandardReadable, EpicsDevice):
+    """A channel for `DemoPointDetector` with int value based on X and Y Motors"""
```
evalott100, line 18:
```suggestion
    """A channel for `DemoPointDetector` with int value based on X and Y Motors."""
```

src/ophyd_async/epics/demo/_stage.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StandardReadable
+
+from ._motor import DemoMotor
+
+
+class DemoStage(StandardReadable):
+    """A simulated sample stage with X and Y movables"""
```
evalott100, line 7:
```suggestion
    """A simulated sample stage with X and Y movables."""
```

src/ophyd_async/sim/_point_detector.py
```diff
@@ -0,0 +1,87 @@
...
+    soft_signal_r_and_setter,
+    soft_signal_rw,
+)
+from ophyd_async.core import StandardReadableFormat as Format
+
+from ._pattern_generator import PatternGenerator
+
+
+class EnergyMode(StrictEnum):
+    """Energy mode for `SimPointDetector`"""
```
evalott100, line 22:
```suggestion
    """Energy mode for `SimPointDetector`."""
```

src/ophyd_async/sim/_point_detector.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, value_signal: SignalR[int], name=""):
+        with self.add_children_as_readables(Format.HINTED_SIGNAL):
+            self.value = value_signal
+        with self.add_children_as_readables(Format.CONFIG_SIGNAL):
+            self.mode = soft_signal_rw(EnergyMode)
+        super().__init__(name)
+
+
+class SimPointDetector(StandardReadable):
+    """Simalutes a point detector with multiple channels"""
```
evalott100, line 40:
```suggestion
    """Simalutes a point detector with multiple channels."""
```

src/ophyd_async/sim/_stage.py
```diff
@@ -0,0 +1,19 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.sim._pattern_generator import PatternGenerator
+
+from ._motor import SimMotor
+
+
+class SimStage(StandardReadable):
+    """A simulated sample stage with X and Y movables"""
```
evalott100, line 8:
```suggestion
    """A simulated sample stage with X and Y movables."""
```

src/ophyd_async/testing/_mock_signal_utils.py
```diff
@@ -173,3 +143,28 @@ def callback_on_mock_put(
     backend = _get_mock_signal_backend(signal)
     backend.put_mock.side_effect = callback
     return _unset_side_effect_cm(backend.put_mock)
+
+
+def set_mock_put_proceeds(signal: Signal, proceeds: bool):
+    """Allow or block a put with wait=True from proceeding"""
```
evalott100, line 149:
```suggestion
    """Allow or block a put with wait=True from proceeding."""
```

evalott100 (commented):
Lots here, changes to the docs are great!

All the following should be regarded as **NIT**s.

evalott100 (approved):

PR #702, opened 2024-12-10
618 assert statements broken in tests

### LINKED ISSUE #618: Assert statements broken in tests
Seems like `assert_configuration()` (and presumably also `assert_reading()`) are broken for numpy array types. I guess they probably were already broken before these changes, but one of the tests in dodal uses assert_configuration and I encountered this as I'm currently in the process of going through dodal and mx-bluesky to test out this PR.
I will keep you posted if I find any other issues.

_Originally posted by @rtuck99 in https://github.com/bluesky/ophyd-async/issues/594#issuecomment-2428651732_
            
I suggest that we move the `assert` functions into a new `ophyd_async.testing` package, make `pytest` a dependency of it, then use `pytest.approx` on the value

Relm-Arrowny:
Fix #618

Added _approx_readable_value which insert pytest.approx to expected_reading/configuration  in assert_reading and assert _configuration

### DIFF
--- a/src/ophyd_async/testing/_assert.py
+++ b/src/ophyd_async/testing/_assert.py
@@ -1,6 +1,7 @@
 from collections.abc import Mapping
 from typing import Any
 
+import pytest
 from bluesky.protocols import Reading
 
 from ophyd_async.core import AsyncConfigurable, AsyncReadable, SignalDatatypeT, SignalR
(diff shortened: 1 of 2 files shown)


coretl (approved):

coretl:
Thanks!

Issue #508, opened 2024-08-05
Implement support for bluesky suspenders


burkeds:
There is currently no support for bluesky suspenders. SignalR.subscribe() may need a refactor to comply with the expected structure.

```python
from bluesky.suspenders import SuspendFloor
suspender = SuspendFloor(signal=i1.counts, suspend_thresh=5.0, resume_thresh=10)
RE.install_suspender(suspender)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[11], line 3
      1 from bluesky.suspenders import SuspendFloor
      2 suspender = SuspendFloor(signal=i1.counts, suspend_thresh=5.0, resume_thresh=10)
----> 3 RE.install_suspender(suspender)

File ~/projects/BLUESKY/venv/lib/python3.11/site-packages/bluesky/run_engine.py:1083, in RunEngine.install_suspender(self, suspender)
   1070 """
   1071 Install a 'suspender', which can suspend and resume execution.
   1072 
   (...)
   1080 :meth:`RunEngine.clear_suspenders`
   1081 """
   1082 self._suspenders.add(suspender)
-> 1083 suspender.install(self)

File ~/projects/BLUESKY/venv/lib/python3.11/site-packages/bluesky/suspenders.py:76, in SuspenderBase.install(self, RE, event_type)
     74 with self._lock:
     75     self.RE = RE
---> 76 self._sig.subscribe(self, event_type=event_type, run=True)

TypeError: SignalR.subscribe() got an unexpected keyword argument 'event_type'
```

coretl:
Need to add support for the `run` argument to the `subscribe` method, as per https://github.com/bluesky/bluesky/issues/1789

coretl:
This means adding the `run` argument to both of these:
https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_signal.py#L215-L221

and if it is False then after https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_signal.py#L158 add it to a `set` of listeners that want to drop the next notification

In https://github.com/bluesky/ophyd-async/blob/406ac0885c119e21439918d08ff10325c415dd19/src/ophyd_async/core/_signal.py#L151 then if `function` is in the `set` of listeners that want to drop the next notification, then remove from the set and return without notifying, otherwise do the current behaviour

[...]

PR #703, opened 2024-12-10
Add tests for bluesky suspenders

jsouter:
Depends on https://github.com/bluesky/bluesky/pull/1806

Closing https://github.com/bluesky/ophyd-async/issues/508

Setting as draft for now, still having some issues with the RunEngine not entering a suspended state when the signal is set to a suspending value from a task outside of a plan. See the `test_suspension_from_async_task` test which passes (when running against the bluesky branch) for the `task_in_plan=True` case but not `task_in_plan=False`


jsouter:
Closing, current plan is to add ophyd-async as an optional dev dependency to bluesky and put the tests in there

Issue #707, opened 2024-12-13
Use the `OneOfEverythingDevice` to check the assert functions 


coretl:
So that every type is checked.

The Device will need extending so that all Signals will appear in both Read and Configure. Possibly choosing which via an init arg.

## Acceptance Criteria
- All types of signal and value are testing when testing the `assert_*` functions

jsouter:
Should I change assert_value to work with arrays?

coretl:
Yes please

Issue #710, opened 2024-12-17
 use of `assert` in non-test methods


kivel:
It appears that `assert` is used in core methods. 

## Why is this problematic?
From the [documentation](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) it becomes clear that this code is only executed when debugging is enabled.
When executed with the `-O` argument, `assert` statements are typically completely disabled.

## fix
replace `assert` with proper check + exceptions.

coretl:
Personally I have always used `assert` for things that really shouldn't ever fail, like:
https://github.com/bluesky/ophyd-async/blob/4f4458c916071859c44b03b2f39a6a9af05d4aac/src/ophyd_async/core/_device.py#L185

In this case the logic will always set `self._connect_task`, but the conditionals are too complicated for the type checker to infer this. The `assert` statement is then really for the benefit of the type checker. I seem to have picked up this habit from the (mypy cheat sheet)[https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#useful-built-in-types]. Would you rewrite this statement with a check and raise?

kivel:
Yes, I would rewrite those as check and raise.
I'd enable [S101](https://docs.astral.sh/ruff/rules/assert/), to ensure `assert` isn't used.

coretl:
Fixed by https://github.com/bluesky/ophyd-async/pull/728

PR #711, opened 2024-12-18
Expand assert_* functions to work with all signal types

### LINKED ISSUE #717: Remove `_generate_assert_error_msg` and its usage now we have pytest assert rewriting
After #714, this function is no longer necessary:
https://github.com/bluesky/ophyd-async/blob/c781e27691efde1dd7b4d185252840566b01d041/src/ophyd_async/testing/_assert.py#L10-L18

Remove it, and change all `assert x, _generate_assert_error_msg(...)` to bare `assert x`

jsouter:
Fixes https://github.com/bluesky/ophyd-async/issues/707
Fixes #717

Expands OneOfEveryThingDevice into a StandardDetector so we can call read_configuration, and adds tests to tests/core/test_signal.py.

pytest.approx does not work on a whole dict when Tables are present, so now having to manually loop over each key in a dict[str, Reading], but now assert_reading, assert_value and assert_configuration work with arrays and table types.


src/ophyd_async/testing/_assert.py
```diff
@@ -18,6 +24,20 @@ def _generate_assert_error_msg(name: str, expected_result, actual_result) -> str
...
+
+
+def _table_equal(table1: Table, table2: Table) -> bool:
+    # keys and values of all fields should be the same
+    if not table1.model_fields == table2.model_fields:
+        return False
+    for field, value1 in table1:
+        if not _value_equal(value1, getattr(table2, field)):
+            return False
+    return True
```
coretl, line 38: I've got another strategy for you:
```python
class EqualSignalValue:
    def __init__(self, value: SignalDatatype):
        self.value = value
    def __ne__(self, other):
        return _is_different(other, self.value)  # this function should be extracted from ophyd_async.plan_stubs._settings::apply_settings, renamed, and put in core

async def assert_reading(readable: AsyncReadable, expected_reading: dict[str, Reading]):
    actual_reading = await readable.read()
    expected_reading = {k: dict(**v, value=EqualSignalValue(v["value"])) for k, v in expected_reading.items()}
    assert actual_reading == expected_reading
```
Then you don't need to do the waking in each function, just change each value into an `EqualSignalValue`. You might need to read up on which of `__eq__`, `__ne__` you need to override to make this work properly

src/ophyd_async/testing/_assert.py
```diff
@@ -79,18 +61,12 @@ async def assert_reading(
...
     """
     actual_reading = await readable.read()
-    assert (
-        _approx_readable_value(expected_reading) == actual_reading
-    ), _generate_assert_error_msg(
-        name=readable.name,
-        expected_result=expected_reading,
-        actual_result=actual_reading,
-    )
+    _compare_readings_dicts(actual_reading, expected_reading)
```
coretl, line 64: Could this be:
```python
approx_expected_reading = {k: v.copy() for k, v in expected_reading.items()}
for reading in approx_expected_reading.values():
    reading["value"] = approx_value(reading["value"])
assert approx_expected_reading == actual_reading
```

jsouter:
```
approx_expected = {k: dict(v, value=approx_value(expected[k]["value"]))
                   for k, v in expected.items()}
assert actual == approx_expected
```
is slightly more concise though maybe less readable?

coretl: Concise is good, let's go with yours

src/ophyd_async/testing/_one_of_everything.py
```diff
@@ -53,29 +59,58 @@ def float_array_signal(
...
-        self.int64a = int_array_signal(np.int64)
-        self.uint64a = int_array_signal(np.uint64)
-        self.float32a = float_array_signal(np.float32)
-        self.float64a = float_array_signal(np.float64)
-        self.stra = soft_signal_rw(Sequence[str], ["one", "two", "three"])
+        self.int = soft_signal_rw(int, 1, name="int")
+        self.float = soft_signal_rw(float, 1.234, name="float")
+        self.str = soft_signal_rw(str, "test_string", name="str")
+        self.bool = soft_signal_rw(bool, True, name="bool")
+        self.enum = soft_signal_rw(ExampleEnum, ExampleEnum.B, name="enum")
```
coretl, line 100: Why did you need to add all these names? Naming the parent should name all the children too...

src/ophyd_async/testing/_one_of_everything.py
```diff
@@ -53,29 +59,58 @@ def float_array_signal(
...
+    async def get_indices_written(self) -> int:
+        return 0
+
+    def collect_stream_docs(self, indices_written: int):
+        raise NotImplementedError()
+
+    async def close(self) -> None: ...
+
+
+class OneOfEverythingDevice(StandardDetector):
```
coretl, line 93: Rather than making it a `StandardDetector`, would it be better to make it a `StandardReadable`? Then you wouldn't need the DummyController and DummyWriter...

kivel:
This might link to #710

src/ophyd_async/testing/_one_of_everything.py
jsouter: Yep, that works, I didn't have the device named in test_signal.py

coretl:
> This might link to #710

The `ophyd_async.testing` module (and `ophyd_async.*.testing` modules) are the only places that should use assert, as they are meant for tests, and we want pytest to do its assert rewriting, so must actually use an assert

src/ophyd_async/testing/_one_of_everything.py
```diff
@@ -53,10 +57,11 @@ def float_array_signal(
...
-    return soft_signal_rw(Array1D[dtype], value)
+    return soft_signal_rw(Array1D[dtype], value, name)
 
 
-class OneOfEverythingDevice(Device):
+class OneOfEverythingDevice(StandardReadable):
+    # make a detector to test assert_configuration
     def __init__(self, name=""):
         self.int = soft_signal_rw(int, 1)
         self.float = soft_signal_rw(float, 1.234)
```
coretl, line 67: Could you do this in the native StandardReadable way...
```suggestion
        with self.add_children_as_readables(Format.CONFIG_SIGNAL):
            self.int = soft_signal_rw(int, 1)
            self.float = soft_signal_rw(float, 1.234)
            ...
```

src/ophyd_async/testing/_one_of_everything.py
```diff
@@ -88,7 +97,11 @@ def __init__(self, name=""):
             ),
         )
         self.ndarray = soft_signal_rw(np.ndarray, np.array(([1, 2, 3], [4, 5, 6])))
-        super().__init__(name=name)
+        # add all signals to configuration
+        self._read_config_funcs = tuple(  # type: ignore
+            sig.read for sig in self.__dict__.values() if isinstance(sig, SignalR)
+        )
```
coretl, line 103: ...rather than this
```suggestion
```

coretl (commented):
One minor comment then we're good to go!

jsouter:
sorry, ignore the review request until I fix the conflicts

coretl (approved):

coretl:
Feel free to merge this when you're ready

PR #723, opened 2025-01-08
Update core and tango tests to match structure of epics tests

### LINKED ISSUE #715: Update core and tango tests to match structure of epics tests
#714 rewrote the epics tests, improving the test matrix approach and adding some utils like `MonitorQueue` and `ApproxTable` to `ophyd_async.testing`. This ticket is to update the core and tango tests to use the same approach and utils.

## Acceptance Criteria
- Core and tango tests use `ophyd_async.testing.MonitorQueue` and a similar style matrix approach.

### LINKED ISSUE #751: Testing assert functions give confusing message on failure
`assert_reading` and `assert_configuration` give a confusing `full diff` when an equality fails, making it hard to see the actual issue.

As an example, if we change `test_assert_reading` in `tests/core/test_signal.py` so that the `dummy_reading` has `timestamp: 5`, this is the test output

```
AssertionError: assert {'mock_readable-float_array': {'value': array([  1.1231,  -2.3   , 451.15  ,   6.6233], dtype=float32), 'timestamp': 366659.164213077, 'alarm_severity': 0}, 'mock_readable-int_array': {'value': array([1, 2, 4, 7], dtype=int8), 'timestamp': 366659.164199919, 'alarm_severity': 0}, 'mock_readable-int_value': {'value': 188, 'timestamp': 366659.164185914, 'alarm_severity': 0}} == {'mock_readable-int_value': {'alarm_severity': 0, 'timestamp': 5, 'value': 188 ± 1.9e-04}, 'mock_readable-int_array': {'alarm_severity': 0, 'timestamp': <ANY>, 'value': approx([1 ± 1.0e-06, 2 ± 2.0e-06, 4 ± 4.0e-06, 7 ± 7.0e-06])}, 'mock_readable-float_array': {'alarm_severity': 0, 'timestamp': <ANY>, 'value': approx([1.1231 ± 1.1e-06, -2.3 ± 2.3e-06, 451.15 ± 4.5e-04, 6.6233 ± 6.6e-06])}}
  
  Common items:
  {'mock_readable-float_array': {'alarm_severity': 0,
                                 'timestamp': 366659.164213077,
                                 'value': array([  1.1231,  -2.3   , 451.15  ,   6.6233], dtype=float32)},
   'mock_readable-int_array': {'alarm_severity': 0,
                               'timestamp': 366659.164199919,
                               'value': array([1, 2, 4, 7], dtype=int8)}}
  Differing items:
  {'mock_readable-int_value': {'alarm_severity': 0, 'timestamp': 366659.164185914, 'value': 188}} != {'mock_readable-int_value': {'alarm_severity': 0, 'timestamp': 5, 'value': 188 ± 1.9e-04}}
  
  Full diff:
    {
        'mock_readable-float_array': {
            'alarm_severity': 0,
  -         'timestamp': <ANY>,
  -         'value': approx([1.1231 ± 1.1e-06, -2.3 ± 2.3e-06, 451.15 ± 4.5e-04, 6.6233 ± 6.6e-06]),
  +         'timestamp': 366659.164213077,
  +         'value': array([  1.1231,  -2.3   , 451.15  ,   6.6233], dtype=float32),
        },
        'mock_readable-int_array': {
            'alarm_severity': 0,
  -         'timestamp': <ANY>,
  -         'value': approx([1 ± 1.0e-06, 2 ± 2.0e-06, 4 ± 4.0e-06, 7 ± 7.0e-06]),
  +         'timestamp': 366659.164199919,
  +         'value': array([1, 2, 4, 7], dtype=int8),
        },
        'mock_readable-int_value': {
            'alarm_severity': 0,
  -         'timestamp': 5,
  +         'timestamp': 366659.164185914,
  ?                      ++++ +++++++++++
  -         'value': 188 ± 1.9e-04,
  ?                     ----------
  +         'value': 188,
        },
    }
```
Really, there is only one value causing an issue. It would be easier to write and fix tests if the output here only contained the breaking differences.

jsouter:
Fixes #715
Fixes #751
very WIP, I should rework the tango version of OneOfEverythingDevice to create signals for each type as scalar, spectrum and image, so I can combine test_tango_signals and test_tango_one_of_everything.

Also trying to rework MonitorQueue to use the existing assert_value/assert_reading methods


src/ophyd_async/core/_signal_backend.py
```diff
@@ -30,6 +31,7 @@
     | Array1D[np.float32]
     | Array1D[np.float64]
     | np.ndarray
+    | IntEnum
```
coretl, line 34: From my brief skim of the docs, PyTango enumerations look like:
```python
class MyEnum(IntEnum):
    ENUM_STRING1 = 0
    ENUM_STRING2 = 1
```
I guess this means that Tango enum string values can't contain spaces. The value is superfluous to us as we can set value by string like in EPICS.

I would prefer that we don't support this as a signal type, but instead may our own Enum to request, using the tango name as the value:
```python
class MyEnum(StrictEnum):  # or SubsetEnum
    WHATEVER1 = "ENUM_STRING1"
    WHATEVER2 = "ENUM_STRING2"
```

We should then convert in the signal converter between the Tango IntEnum and the requested ophyd StrictEnum

src/ophyd_async/testing/_assert.py
```diff
@@ -60,11 +61,19 @@ async def assert_reading(
...
+
+
+def _assert_readings_approx_equal(expected, actual):
     approx_expected_reading = {
-        k: dict(v, value=approx_value(expected_reading[k]["value"]))
-        for k, v in expected_reading.items()
+        k: dict(
+            v,
+            value=approx_value(expected[k]["value"]),
+            timestamp=approx_value(expected[k]["timestamp"]),
```
coretl, line 72: Would be nice to support not having to put timestamps in the dictionary
```suggestion
            timestamp=pytest.approx(expected[k].get("timestamp", ANY)),
```
Maybe alarm severity too?

src/ophyd_async/testing/_assert.py
```diff
@@ -144,29 +149,27 @@ def __eq__(self, value):
...
-        assert value == expected_value
-        assert type(value) is expected_type
-        expected_reading = {
+        # reading = await self.signal.read()
+        await assert_value(self.signal, expected_value)
+        expected_reading: dict[str, Reading] = {
             self.signal.name: {
                 "value": expected_value,
-                "timestamp": pytest.approx(time.time(), rel=0.1),
+                "timestamp": self._timestamp_provider(),  # type: ignore
```
coretl, line 167: Is this just so we can use `time.monotonic`? If so, I think it's fine to change that just in that function without introducing a timestamp provider...

src/ophyd_async/testing/_one_of_everything.py
```diff
@@ -57,48 +54,64 @@ def float_array_signal(
...
+            ExampleTable(
+                bool=np.array([False, False, True, True], np.bool_),
+                int=np.array([1, 8, -9, 32], np.int32),
+                float=np.array([1.8, 8.2, -6, 32.9887], np.float64),
+                str=["Hello", "World", "Foo", "Bar"],
+                enum=[ExampleEnum.A, ExampleEnum.B, ExampleEnum.A, ExampleEnum.C],
+            ),
+        ),
+        EverythingSignal("ndarray", np.ndarray, np.array(([1, 2, 3], [4, 5, 6]))),
+    ]
```
coretl, line 105: Although this saves typing, it's actually pretty useful to see how you would make the Tango Device and it's associated ophyd device (and the same for EPICS and sim). Rather than try to be clever here I think I'd prefer to just see the test data written out longhand, with this file not being imported from Tango...

tests/tango/test_tango_one_of_everything.py
```diff
@@ -0,0 +1,180 @@
...
+            1.79769313e308,
+            2.22507386e-308,
+            4.94065646e-324,
+            0.00000000e000,
+            1.23400000e000,
+            2.34000000e005,
+            3.45000000e-006,
+        ],
+        dtype=np.float64,
+    ),
```
coretl, line 149: Although it would be fine to put these in a fixture and share these

jsouter:
Still very WIP, sketching out how to handle tango enums, also probably some debug comments left in.
test_tango_signals.py is a little broken, will probably merge with test_tango_one_of_everything.py when I figure out how everything should work.

src/ophyd_async/core/_signal_backend.py
jsouter: Okay, have figured out one way of doing converters for scalar/spectrum/image enum attributes. For the sake of working with assert_value etc it may be easiest if the converters return strs (or numpy arrays of strs), can create an Enum class in place in the converter but not sure if that provides much benefit. 

The shift away from using IntEnums means we can currently use a StrictEnum with assert_value etc but we can't test those same values with the tango provided assert_close.

jsouter:
Seems like running the test context in a subprocess does allow us to run tango tests on windows without forking. Still need to fix some race conditions in the tango one_of_everything tests. Possibly I should add a fixture that adds a short sleep between the tango tests too as there is still a small chance of the tests freezing without this.

tests/tango/test_tango_signals.py
```diff
@@ -1,30 +1,30 @@
...
+    OneOfEverythingTangoDevice,
+    everything_signal_info,
+)
+from ophyd_async.testing import MonitorQueue, assert_reading, assert_value
+from tango import AttrDataFormat, DevState
 from tango.asyncio import DeviceProxy
-from tango.asyncio_executor import set_global_executor
-from tango.server import Device, attribute, command
-from tango.test_context import MultiDeviceTestContext
 from tango.test_utils import assert_close
```
coretl, line 28: Ideally we would have no `tango.*` imports here at all, is that achievable?

tests/tango/test_tango_signals.py
```diff
@@ -631,7 +456,196 @@ async def test_tango_signal_auto_badtrl(tango_test_device: str):
...
+        dtype=np.float64,
+    ),
+    "strenum": np.array(
+        [ExampleStrEnum.A.value, ExampleStrEnum.B.value, ExampleStrEnum.C.value],
+        dtype=str,
+    ),
+    "str": ["one", "two", "three"],
+    "bool": np.array([False, True]),
+    "my_state": np.array(
+        [DevState.INIT, DevState.ON, DevState.MOVING]
```
coretl, line 526: I think we should reimplement `DevState` as a `SubsetEnum` or `StrictEnum`...

tests/tango/test_tango_signals.py
```diff
@@ -275,85 +108,53 @@ async def make_backend(
...
 ):
-    await prepare_device(echo_device, pv, initial_value)
-    source = echo_device + "/" + pv
-    backend = await make_backend(datatype, source, allow_events=True)
+    source = get_full_attr_trl(trl, pv)
+    signal = tango_signal_rw(datatype, source)
+    backend = signal._connector.backend
+    await signal.connect()
+    converter = signal._connector.backend.converter  # type: ignore
+    converted_put = converter.value(put_value)
```
coretl, line 125: Can we get these from the everything device rather than making the signals directly here?

tests/tango/test_tango_signals.py
jsouter: Yep, though may need to keep DeviceProxy unless we're happy to remove the ability to instantiate a TangoDevice from an existing DeviceProxy rather than a TRL. That would be my preference as it would remove some complexity in handling the TRLs of children when the server is remote.

coretl: Yes, please remove the ability to pass in a DeviceProxy

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -628,6 +632,60 @@ async def get_tango_trl(
...
+
+def make_converter(info: AttributeInfoEx | CommandInfo) -> TangoConverter:
+    if isinstance(info, AttributeInfoEx):
+        if info.enum_labels:  # enum_labels should be discarded for non enum types
+            if info.data_format == AttrDataFormat.SCALAR:
+                return TangoEnumConverter(list(info.enum_labels))
+            elif info.data_format == AttrDataFormat.SPECTRUM:
+                return TangoEnumSpectrumConverter(list(info.enum_labels))
+            elif info.data_format == AttrDataFormat.IMAGE:
+                return TangoEnumImageConverter(list(info.enum_labels))
```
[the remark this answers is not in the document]
jsouter: To be passed in as strings?

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -88,6 +100,7 @@ class TangoProxy:
     support_events: bool = True
     _proxy: DeviceProxy
     _name: str
+    _converter: TangoConverter = TangoConverter()
```
jsouter, line 103: I will try and make this nicer. It made sense for me to move the converter to the TangoProxy classes so that the Readings in the backend and cache would have values of the same type. But then I'm not sure if we should have a DisconnectedTangoConverter and require that connect be called on AttributeProxy/CommandProxy and set the converter based on the config info there before we call any get/get_reading/put on it

src/ophyd_async/tango/core/_tango_transport.py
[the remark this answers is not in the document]
jsouter: Is this for when we have a DevEnum on the server end or a DevString that we restrict to some set of values on the ophyd-async end? 

I think the one other thing I'm stuck on at the moment is commands that are backed by DevEnum, it seems like commands can't have specified enum_labels so we can't easily construct the converters from the labels. It would be possible to provide an enum class if the command signal is being explicitly declared with an override datatype, but in the case where ophyd-async is automatically creating the child command from the tango device trl we would still have to get and set using the underlying int value instead of the string name. This isn't a problem with attributes, just commands.

src/ophyd_async/tango/testing/_one_of_everything.py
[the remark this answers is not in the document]
jsouter: I've separated the dataclasses which provide the tests with the initial values to check etc, but otherwise the ophyd device is just a `TangoDevice(trl=...)` call, do you want each of the signals in the to be explicitly declared in a subclass of TangoDevice?

jsouter:
Note to self: I haven't touched the docs so I will have to go through and update those.

coretl (approved):

PR #724, opened 2025-01-08
Port andor new ad format

jwlodek:
Takes #679 and adds modifications to work with the changes to the new `AD` class structure. I don't have an Andor to test with, so would be good if we could test with the device.


src/ophyd_async/epics/adandor/_andor_controller.py
```diff
@@ -0,0 +1,49 @@
...
+        await self.set_exposure_time_and_acquire_period_if_supplied(
+            trigger_info.livetime
+        )
+        await asyncio.gather(
+            self.driver.trigger_mode.set(self._get_trigger_mode(trigger_info.trigger)),
+            self.driver.num_images.set(
+                trigger_info.total_number_of_triggers or _MAX_NUM_IMAGE
+            ),
+            self.driver.image_mode.set(ImageMode.MULTIPLE),
+        )
```
evalott100, line 36: Could we not subclass `TriggerInfo` like eiger does, making an `Andor2TriggerInfo` which contains `trigger: Andor2TriggerMode` instead of `trigger: DetectorTrigger`?

src/ophyd_async/epics/adandor/_andor_controller.py
```diff
@@ -0,0 +1,49 @@
...
+            DetectorTrigger.INTERNAL: Andor2TriggerMode.INTERNAL,
+            DetectorTrigger.EDGE_TRIGGER: Andor2TriggerMode.EXT_TRIGGER,
+        }
+        if trigger not in supported_trigger_types:
+            raise ValueError(
+                f"{self.__class__.__name__} only supports the following trigger "
+                f"types: {supported_trigger_types} but was asked to "
+                f"use {trigger}"
+            )
+        return supported_trigger_types[trigger]
```
evalott100, line 49: **Nit:** If we can't make a new `TriggerInfo` for Andor, then it might nicer to replace this code with a `from_detector_trigger(cls, trigger: DetectorTrigger) -> Andor2TriggerMode` method on `Andor2TriggerMode`.

src/ophyd_async/epics/adandor/_andor_controller.py
coretl: I'm not sure I understand, the only eiger code I can find is:
https://github.com/bluesky/ophyd-async/blob/4f4458c916071859c44b03b2f39a6a9af05d4aac/src/ophyd_async/epics/eiger/_eiger_controller.py#L39-L53

Plus, the point of the controller class it to translate from the generic `TrigerInfo` to the specific `Andor2TriggerMode`, so this translation should definitely happen here...

coretl: Ah, that was so that Eiger can add something extra, `energy`, that can't be determined from what's in `TriggerInfo`, while `Andor2TriggerMode` can already be determined from a standard `TriggerMode`. I'm not convinced we need `EigerTriggerMode` now that we have intelligent load either, so opened https://github.com/bluesky/ophyd-async/issues/727 to discuss

src/ophyd_async/epics/adandor/_andor_controller.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(
+        self,
+        driver: Andor2DriverIO,
+        good_states: frozenset[adcore.DetectorState] = adcore.DEFAULT_GOOD_STATES,
+    ) -> None:
+        super().__init__(driver, good_states=good_states)
+
+    def get_deadtime(self, exposure: float | None) -> float:
+        return _MIN_DEAD_TIME + (exposure or 0)
+
```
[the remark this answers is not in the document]
coretl: Except this is a sync method so we can't read PVs here...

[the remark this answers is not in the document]
Relm-Arrowny: Just out of interest, if I do want the deadtime to be from a readback channel, what is the best way to go about it?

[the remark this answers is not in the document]
coretl: The answer is either:
- We keep the ability to validate plans in advance, the we cannot touch hardware to calculate this number and have to duplicate the calculation of deadtime that's in the SDK into our code
- We abandon the ability to validate plans in advance, then we can make this an async method and use the value of that PV after exposure time has been set

@callumforrester do we want to open this box?

coretl:
@Relm-Arrowny will you have a chance to try this on a real Andor please?

Relm-Arrowny:
> @Relm-Arrowny will you have a chance to try this on a real Andor please?

sure, I will try this tomorrow or whenever it is free.

jwlodek:
@Relm-Arrowny I've pushed another change that moves the superclass init before the overrides to `image_mode` and `data_type`, and I've made the `data_type` a `SubsetEnum`. Could you pull this change and see if it allows you to connect?

Relm-Arrowny:
> @Relm-Arrowny I've pushed another change that moves the superclass init before the overrides to `image_mode` and `data_type`, and I've made the `data_type` a `SubsetEnum`. Could you pull this change and see if it allows you to connect?

Connected and did a few scans with it, as far as I can tell it working better than ever, thanks.

src/ophyd_async/epics/adandor/_andor_io.py
```diff
@@ -16,23 +16,30 @@ class Andor2TriggerMode(StrictEnum):
     SOFTWARE = "Software"
 
 
-class ImageMode(StrictEnum):
+class Andor2ImageMode(StrictEnum):
```
coretl, line 19: Could we make this one subclass the ADBase one and add `FAST_KINETICS` to it?

src/ophyd_async/epics/adandor/_andor_io.py
```diff
@@ -16,23 +16,30 @@ class Andor2TriggerMode(StrictEnum):
...
     def __init__(self, prefix: str, name: str = "") -> None:
+        super().__init__(prefix, name=name)
         self.trigger_mode = epics_signal_rw(Andor2TriggerMode, prefix + "TriggerMode")
         self.data_type = epics_signal_r(ADBaseDataType, prefix + "DataType_RBV")
         self.andor_accumulate_period = epics_signal_r(
             float, prefix + "AndorAccumulatePeriod_RBV"
         )
-        self.image_mode = epics_signal_rw_rbv(ImageMode, prefix + "ImageMode")
-        super().__init__(prefix, name=name)
+        self.image_mode = epics_signal_rw_rbv(Andor2ImageMode, prefix + "ImageMode")
```
coretl, line 45: This seems a relatively common pattern, overriding the image mode to be a subset of what the baseclass provides. We have `SubsetEnum` already, should we have `SupersetEnum` too? Or we could drop to string in the baseclass...

src/ophyd_async/epics/adandor/_andor_io.py
jwlodek: I just tried this and this is not allowed:

> subclassing an enumeration is allowed only if the enumeration does not define any members.

Source: https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing

In this case, I'd probably just leave this as-is, until we have `SupersetEnum` available.

coretl: I think `SupersetEnum` is the wrong way for this case, we need to make the baseclass do `SubsetEnum` to say that subclasses are allowed to add their own extra image modes. However if we never use the baseclass imagemode anywhere then what is here at the moment is fine.

coretl (approved):

PR #725, opened 2025-01-08
Add queue size signal to base AD plugin IO class

### LINKED ISSUE #621: Add `QueueSize` epics_signal_rw to `NDPluginBaseIO`
```diff
diff --git a/src/ophyd_async/epics/adcore/_core_io.py b/src/ophyd_async/epics/adcore/_core_io.py
index 796857911..15de202b1 100644
--- a/src/ophyd_async/epics/adcore/_core_io.py
+++ b/src/ophyd_async/epics/adcore/_core_io.py
@@ -38,6 +38,7 @@ class NDPluginBaseIO(NDArrayBaseIO):
         self.nd_array_address = epics_signal_rw_rbv(int, prefix + "NDArrayAddress")
         self.array_size0 = epics_signal_r(int, prefix + "ArraySize0_RBV")
         self.array_size1 = epics_signal_r(int, prefix + "ArraySize1_RBV")
+        self.queue_size = epics_signal_rw(int, prefix + "QueueSize")
         super().__init__(prefix, name)
```

jwlodek:
Resolves #621

### DIFF
--- a/src/ophyd_async/epics/adcore/_core_io.py
+++ b/src/ophyd_async/epics/adcore/_core_io.py
@@ -54,6 +54,7 @@ class NDPluginBaseIO(NDArrayBaseIO):
         self.nd_array_address = epics_signal_rw_rbv(int, prefix + "NDArrayAddress")
         self.array_size0 = epics_signal_r(int, prefix + "ArraySize0_RBV")
         self.array_size1 = epics_signal_r(int, prefix + "ArraySize1_RBV")
+        self.queue_size = epics_signal_rw(int, prefix + "QueueSize")
         super().__init__(prefix, name)
 
 


thomashopkins32 (approved):

coretl:
What do we need this signal for? It's not used in the code...

jwlodek:
At HEX we found that the file system could not keep up with the speed of the camera, so as a step in the plan we also set queue_size=num_frames for the writer plugin. I can add that to writer.open but wasn't sure if that would be desired behavior in all cases.

coretl:
> At HEX we found that the file system could not keep up with the speed of the camera, so as a step in the plan we also set queue_size=num_frames for the writer plugin. I can add that to writer.open but wasn't sure if that would be desired behavior in all cases.

You're probably right there. Let's just merge this as is.

PR #726, opened 2025-01-08
Rename multiplier to exposures_per_event and move to first dim of shape

### LINKED ISSUE #431: Shape and multiplier are wrong for PandA
https://github.com/bluesky/ophyd-async/blob/b153c5013f3ebb3f0d1697e96a33ed0746c95117/src/ophyd_async/panda/writers/_hdf_writer.py#L88

PandA always writes scalar datasets, so the shape should be `()` not `[1]`

https://github.com/bluesky/ophyd-async/blob/b153c5013f3ebb3f0d1697e96a33ed0746c95117/src/ophyd_async/panda/writers/_hdf_writer.py#L56-L59

Multipliers are valid here, and should be supported in the same way as the areaDetector HDF writer

### LINKED ISSUE #576: Remove multiplier from resource docs, add it instead as an extra first dimension in descriptor shape
Per discussion today, it was decided that in order to keep the dimensionality of `shape` in the event descriptor consistent, that the multiplier should be added there as the first dimension. So for example in the case of two detectors w/ x-y dimensions of 10x10, with one collecting a single frame per acq, and one collecting 5, we would have

```
shape: (1, 10, 10)
```
and
```
shape: (5, 10, 10)
```
respectively in the descriptor. Since this tells you the multiplier, and you can fetch this information given a resource doc by looking back at the associated descriptor, the multiplier field can be removed from the resource doc creation.

thomashopkins32:
This PR does the following:
- Renames `multiplier` -> `frames_per_event`
- Add the `frames_per_event` as the first dimension of `DataKey.shape`
- Ensure that the index provided by `DetectorWriter.get_indices_written()` and `DetectorWriter.observe_indices_written()` is divided by `frames_per_event` so that it actually captures the correct amount of exposures in each index (except for PandA which explicitly says it only has 1 "frame" per event)
- Add unit tests showing that describe() works as intended
- ~~Add unit tests showing that stream resources are actually batches of exposures~~
- Re-order `self._writer.open()` and `self._writer.get_indices_written()`. The writer needs to be opened in order to get the indices written. Otherwise, it has no idea what `frames_per_event` to use when returning the index last written.

I could not actually add tests using bluesky plans and inspecting the data afterword because `TriggerInfo` is hardcoded in `StandardDetector`. I think it is a separate issue that should be raised since it would enhance the scope of this PR. I will open an issue for this soon and mention it below.

Otherwise, I have a few open questions regarding my understanding of ophyd-async as well as the implementation which I will also leave as review comments. Please see below.

Closes #576
Closes #431


thomashopkins32 (commented):
Question about shapes when `frames_per_event` is 1: do we want to always "squeeze" the shape?
I.e. there are a couple of options:
- For 2d arrays:
    - `[1, h, w]` -> `[h, w]` when `frames_per_event = 1`
    - `[frames_per_event, h, w]` when `frames_per_event > 1`
- For scalar values:
    - `[1,]` -> `[]` when `frames_per_event = 1`
    - `[frames_per_event,]` when `frames_per_event > 1`

Currently, it is set up such that if the result would be a single scalar value, the shape would be replaced with `[]`. Otherwise, the shape always contains the extra dim.

src/ophyd_async/core/_detector.py
```diff
@@ -21,7 +21,7 @@
     Triggerable,
     WritesStreamAssets,
 )
-from event_model import DataKey
+from event_model import DataKey  # type: ignore
```
jwlodek, line 24: Do we need the `# type: ignore` here?

src/ophyd_async/core/_hdf_dataset.py
```diff
@@ -67,7 +66,7 @@ def __init__(
                 parameters={
                     "dataset": ds.dataset,
                     "swmr": ds.swmr,
-                    "multiplier": ds.multiplier,
+                    "shape": ds.shape,
```
jwlodek, line 69: I forgot what we decided here - did we want the shape in the resource doc, or just in the data key that you can get a reference to from the resource doc? Chunking info definitely makes sense in the resource doc since certain data formats won't really have chunking as an option.

src/ophyd_async/epics/adcore/_core_writer.py
```diff
@@ -181,10 +180,17 @@ async def collect_stream_docs(
                 self._emitted_resource = bundler_composer(
                     mimetype=self._mimetype,
                     uri=uri,
+                    # TODO: This is confusing, I expected this to be of type `DataKey`
+                    # but it is a string. Naming could be improved maybe?
                     data_key=self._name_provider(),
```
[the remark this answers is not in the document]
jwlodek: This is a good point. Maybe we should consider renaming to something like `data_key_name` or `data_key_reference`

src/ophyd_async/epics/adcore/_core_writer.py
```diff
@@ -181,10 +180,17 @@ async def collect_stream_docs(
...
+                    # Q: What are the parameters used for? Extra info?
                     parameters={
-                        # Assume that we always write 1 frame per file/chunk
-                        "chunk_shape": (1, *frame_shape),
+                        # Assume that we always write self._frames_per_event
+                        # frames per file/chunk
+                        # TODO: Validate this assumption and that it should
+                        # not be self._frames_per_event
+                        "chunk_shape": (self._frames_per_event, *frame_shape),
```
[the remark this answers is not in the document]
jwlodek: `shape` is the shape of a single event in our scan, so to make it interesting say each event we collect 4 color 1080p images. Our shape would be `[5, 3, 1920, 1080]`, since `frames_per_event` is 5, then 3 color channels, and then the resolution. `chunk_shape` on the other hand is the shape of the data as laid out in the file on the file system (the naming comes from `chunking` in an HDF file). They do not necessarily need to correspond to each other (although logically it would make sense).

For now, I would leave the `chunk_shape` parameters as-is, and we can revisit later to potentially make the default behavior be 1 chunk  =  1 event (which would optimize loading of individual events. This would involve using the `frames_per_event` value during `writer.open` to configure chunking settings on the corresponding IOC.

tests/core/test_flyer.py
```diff
@@ -95,7 +100,7 @@ async def collect_stream_docs(
                     parameters={
                         "path": "",
                         "dataset": "",
-                        "multiplier": False,
+                        "frames_per_event": self._frames_per_event,
```
jwlodek, line 103: If we are guaranteeing that the first dimension in `shape` is `frames_per_event` I think we won't need it in the parameters anymore.

src/ophyd_async/epics/adcore/_core_writer.py
```diff
@@ -181,10 +180,17 @@ async def collect_stream_docs(
...
+                        "shape": (self._frames_per_event, *frame_shape),
```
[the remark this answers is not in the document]
jwlodek: In the pre-SQL backed tiled days, parameters of resource docs and datum docs respectively would be the arguments passed to a handler's `__init__` and `__call__` respectively, see: 

https://github.com/bluesky/area-detector-handlers/blob/f58ca85a0b8a5c07a9e90979acc3086086a91f23/area_detector_handlers/eiger.py#L16

They would contain any metadata necessary to find the file on the filesystem and read it back.

In the new model, the parameters of resource/datum docs are used by consolidator objects to organize relational database entries for data retrieval later, see: 

https://github.com/bluesky/bluesky/blob/0356f8f77800cfdd455787aa38fa81de56177714/src/bluesky/consolidators.py#L87

To go along with this PR, we will also need to update the consolidator code to correctly interpret this new extra dimension.

tests/epics/adaravis/test_aravis.py
```diff
@@ -105,7 +107,7 @@ async def test_can_collect(
     assert stream_resource["parameters"] == {
         "dataset": "/entry/data/data",
         "swmr": False,
-        "multiplier": 1,
+        "shape": (one_shot_trigger_info.frames_per_event, 10, 10),
```
jwlodek, line 110: Like I mentioned before, I don't know if we need `shape` in the resource doc since we can retrieve it from the data key outputted by `describe`? @coretl thoughts?

jwlodek (changes requested):
I think this is close, just a few minor notes. We should try setting this up in the lab and running a test w/ collecting data from different devices w/ different `frames_per_event` to make sure it behaves as expected (and also to maybe work out the needed changes to the consolidators).

jwlodek:
> Question about shapes when `frames_per_event` is 1: do we want to always "squeeze" the shape? I.e. there are a couple of options:
> 
>     * For 2d arrays:
>       
>       * `[1, h, w]` -> `[h, w]` when `frames_per_event = 1`
>       * `[frames_per_event, h, w]` when `frames_per_event > 1`
> 
>     * For scalar values:
>       
>       * `[1,]` -> `[]` when `frames_per_event = 1`
>       * `[frames_per_event,]` when `frames_per_event > 1`
> 
> 
> Currently, it is set up such that if the result would be a single scalar value, the shape would be replaced with `[]`. Otherwise, the shape always contains the extra dim.

I think I'd be in favor of avoiding such squeezing, because then we'd need a separate parameter to let us know if it had been squeezed or not. Say we have a frames_per_event of 1 w/ a dataset that's 10 x 10. If we squeeze we get [10, 10] as the shape, but there's no way of telling if this is actually a 1D dataset of size 10 w/ 10 frames per event.

src/ophyd_async/core/_hdf_dataset.py
coretl: I think we decided just in the descriptor

tests/epics/adaravis/test_aravis.py
coretl: yeah, don't need it in the resource doc if it is in the descriptor

src/ophyd_async/epics/adcore/_core_writer.py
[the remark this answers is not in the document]
thomashopkins32: This is a potential change for event-model though, not ophyd-async if I understand correctly.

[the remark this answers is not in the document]
jwlodek: Yeah, that's right. Definitely outside the scope of this PR. It would need to change here: https://github.com/bluesky/event-model/blob/acd40a918db4453869e4945e09526b278a7ee20f/src/event_model/schemas/stream_resource.json#L6

and then in a bunch of dependent places.

thomashopkins32:
@jwlodek so the current squeezing behavior for the shape is

- For 2d arrays:
    - [frames_per_event, h, w] (even if frames_per_event is 1)
- For scalar values:
    - [1,] -> [] when frames_per_event = 1
    - [frames_per_event,] when frames_per_event > 1

The final change I would make based on your comment would be to remove the squeezing on scalar values from [1,] -> [].

thomashopkins32:
Should be ready to review once more. The new shape behavior is such that the `frames_per_event` is always the first dimension of shape. If the `len(shape) > 1`, then the `dtype` is an array, otherwise, its a number.

src/ophyd_async/core/_detector.py
coretl: Does `pyright` also complain? If it needs a `py.typed` then please could you make an event-model PR to add it?

thomashopkins32: pyright does not seem to complain, event-model is missing a py.typed marker so I can add it

thomashopkins32: PR here: https://github.com/bluesky/event-model/pull/346

thomashopkins32: @coretl It looks like adding this `py.typed` file to event-model breaks bluesky's tests since there are a lot of unresolved typing issues. I can work on these typing issues but I don't think this should be a blocker for this PR to be merged. Let me know if you think otherwise.

thomashopkins32:
@jwlodek, @jennmald , and I did some testing on actual devices and found a few more issues that need to be resolved. We didn't get through all of the testing we planned for so we will continue next week most likely.

For ophyd-async:
- PandA needs to be able to handle `frames_per_event > 1`. @coretl do you know why this was limited to only being 1 for PandA?
- The computed `total_number_of_triggers` needs to be multiplied by `frames_per_event`

For bluesky:
- `ConsolidatorBase` needs to be reworked based on the new assumption that `frames_per_event` is the first dim of `datum_shape`.

For tiled:
- TBD

That covers pretty much everything that we tested a debugged on the devices so far. We will see if changes to tiled are necessary in further testing.

coretl:
> * PandA needs to be able to handle `frames_per_event > 1`. @coretl do you know why this was limited to only being 1 for PandA?

I'm not sure, I don't think that's a real restriction. If you remove it, what breaks?

jwlodek:
> > * PandA needs to be able to handle `frames_per_event > 1`. @coretl do you know why this was limited to only being 1 for PandA?
> 
> I'm not sure, I don't think that's a real restriction. If you remove it, what breaks?

Nothing actually, we removed it and got everything to work as expected, just into separate streams. We're going to make sure they can fit into the same stream this week

jwlodek (approved):
We've now tested this and it is working for us, pending the `Consolidator` PR.

thomashopkins32:
Actually we decided that it does not work well with tiled just yet. We want tiled to have the `frames_per_event` explicitly in the shape which is causing issues with reading the data back from the files (due to how chunking works).

Basically, ophyd-async has the descriptor shape with the first dim as `frames_per_event`. bluesky's consolidators uses this to figure out the proper chunking of the data prior to writing it to the hdf5 file. Then tiled needs to also understand this chunking in order to read the data back from the file and unpack it properly.

The shape of the data from the user perspective should always be `(num_events, frames_per_event, ...)`

jwlodek:
We tested this again, and we think we are happy to merge this pending the sister PR into bluesky being ready / tested as well.

src/ophyd_async/core/_detector.py
```diff
@@ -79,20 +79,22 @@ class TriggerInfo(BaseModel):
...
+    10 frames, but publish 2 StreamDatum indices, and describe() will show a
+    shape of (5, h, w) for each.
     """
 
     @computed_field
     @cached_property
     def total_number_of_triggers(self) -> int:
         return (
-            sum(self.number_of_triggers)
+            sum(self.number_of_triggers) * self.frames_per_event
```
coretl, line 95: Either we rename `number_of_triggers` to `number_of_events` and this logic is valid, or we remove the `* self.frames_per_event`.
At the moment we have:
- number_of_triggers
- livetime and deadtime of each trigger
- multiplier (which is effectively triggers_per_event)

We could move to:
- number_of_events
- livetime and deadtime of each frame
- frames_per_event

@jwlodek do you have an opinion?

thomashopkins32: I think it should be:

`TriggerInfo.number_of_triggers` -> `TriggerInfo.number_of_events`
`TriggerInfo.frames_per_event` -> `TriggerInfo.triggers_per_event`

This would make the `total_number_of_triggers` property meaningful. The word "frame" is really only relevant to detectors that capture 2-D data.

Then we can add that the word "event" here relates to a single `StreamDatum` index being emitted.

coretl: I'm happy with that

thomashopkins32:
@coretl @jwlodek 

This might seem like overkill but please help me check my understanding of the terms. Here is what I understand:

| Level | Term | Meaning | Example |
|-------|-------|-------------------------------------|-----------------------------------|
| EPICS (or some other control system) | "capture" | One single data point collected from a detector | A singular frame is collected from an area detector |
| ophyd-async | "trigger" | One or many "captures" that returns a status | 5 frames are collected from an area detector |
| bluesky & beyond | "event" | Occurrence of a single "trigger" that finishes with successful status |  5 frames are collected and a document is emitted |

The `TriggerInfo` schema is actually pretty confusing when you try to do the following:
```python
trigger_info = TriggerInfo(number_of_triggers=4, frames_per_event=2)
RE(bps.prepare(det, trigger_info))
RE(bp.count([det], num=10))
```
_This example currently fails to run without error._

Is this something that should be possible? I can't even tell what the expected behavior here should be...

This may be an unintended consequence of doing a step scan versus a fly scan but I need to dig a bit further.

coretl:
> This might seem like overkill but please help me check my understanding of the terms. 

My take:

| Level | Term | Meaning | Example |
|-------|-------|-------------------------------------|-----------------------------------|
| EPICS (or some other control system) | "exposure" | One single data point collected from a detector | A singular frame is collected from an area detector |
| ophyd-async | "trigger" | One software or hardware trigger that is sent to the detector to trigger an exposure (with the given livetime) | A singular frame is collected from an area detector |
| bluesky & beyond | "event" | An atomic row of data in a stream within a scan, synchronous data points from multiple detectors |  5 frames are collected from one detector as a single stream datum index, 1 frame from another as a single stream datum index, and a document is emitted with the motor position at this scan point |

> The `TriggerInfo` schema is actually pretty confusing when you try to do the following:
> 
> ```python
> trigger_info = TriggerInfo(number_of_triggers=4, frames_per_event=2)
> RE(bps.prepare(det, trigger_info))
> RE(bp.count([det], num=10))
> ```

Definitely. There are 3 terms here:
- `number_of_triggers` -> `number_of_events` which is the number of events to produce for each `kickoff` in a flyscan, it should always be set to 1 for a step scan as each step is an event
- `frames_per_event` -> `triggers_per_event` which is the number of triggers and therefore the number of frames the detector will produce in each event
- `num` argument to `count` plan which is the number of events to produce

So `number_of_events` and `num` are both specifying the same thing, but in a step scan the number has to be controlled from the plan, so `number_of_events` must equal 1

> _This example currently fails to run without error._

This should error with something like "Can only trigger() to contribute to a single event, but detector was prepared with number_of_events=4"

jwlodek:
I agree with @coretl 's understanding. One thing I thought of however when mulling this over, is that currently we are missing one thing when operating in step scan mode - namely that if I set `num_images` to say 5 in my AD, and run `count` with `num=3`, I'd expect 3 events w/ 5 frames per event, but currently we don't actually have a mechanism for setting `frames_per_event` in a step scan with `trigger`.

I think the easiest solution to this would be to add a property to `StandardDetector` that basically just returns the currently configured `frames_per_event` for a step scan, with it just returning `1` by default, with the expectation that the implementation should override this. So then for `AD` based detectors this would get the value of `num_images` for example, and would set `frames_per_event=THE_VALUE_OF_THIS_PROPERTY` in the implicit `prepare` in the `trigger` method.

Thoughts?

coretl:
> I think the easiest solution to this would be to add a property to `StandardDetector` that basically just returns the currently configured `frames_per_event` for a step scan, with it just returning `1` by default, with the expectation that the implementation should override this. So then for `AD` based detectors this would get the value of `num_images` for example, and would set `frames_per_event=THE_VALUE_OF_THIS_PROPERTY` in the implicit `prepare` in the `trigger` method.

How about we make `frames_per_event` optional? If it isn't set then the detector should read `num_images` and use it as the value of `frames_per_event`, if it isn't then it should set `num_images` to be `frames_per_event`.

> I suggest we use `exposures_per_event` to replace `frames_per_event`.

I'm not sure whether `frames` or `exposures` is better, @jwlodek thoughts?

thomashopkins32:
> How about we make `frames_per_event` optional? If it isn't set then the detector should read `num_images` and use it as the value of `frames_per_event`, if it isn't then it should set `num_images` to be `frames_per_event`.

This ends up being a bit strange since the `DetectorController` and `DetectorWriter` are separate.

In `StandardDetector.prepare`, we would have to do
```python
await self._controller.prepare(value)
self._describe = await self._writer.open(value.frames_per_event or 1)
```
where the controller's prepare does:
```python
if trigger_info.frames_per_event is None:
    trigger_info.frames_per_event = await self.driver.num_images.get_value()
```

The writer needs to know this information to report the descriptor shape accurately. But the controller needs to be able to set this value. I'm fine with this approach as long as users are aware that their `TriggerInfo` could be modified.

Maybe `DetectorController.prepare` should return a new copy of `TriggerInfo` to keep the old one unmodified?

thomashopkins32:
> How about we make `frames_per_event` optional? If it isn't set then the detector should read `num_images` and use it as the value of `frames_per_event`, if it isn't then it should set `num_images` to be `frames_per_event`.

I think this should be handled in a different PR. There is some complexity in doing this. Something like this is already done for the `TriggerInfo.deadtime` but I'm not sure we want to do it the same way for `frames_per_event`.

We would basically be doing the following:
```
frames_per_event <- num_images
num_images <- number_of_events * frames_per_event
```
So if someone re-uses a `TriggerInfo` that has `frames_per_event=None`, then the `num_images` could increase over-time if `number_of_events > 1`. I think this user would have to call `prepare` twice in the same bluesky run though?

thomashopkins32:
> I think I've got some terminology confused. In my mind we're mapping:
> 
>     * `TriggerInfo.livetime` -> `ADAcquireTime`
> 
>     * `TriggerInfo.livetime + TriggerInfo.deadtime` -> `ADAcquirePeriod`
> 
>     * `TriggerInfo.number_of_events` -> `ADNumImages`
> 
>     * `TriggerInfo.frames_per_event` -> `ADNumExposures`
> 
> 
> As defined in https://areadetector.github.io/areaDetector/ADCore/ADDriver.html

My understanding of the `ADDriver` is not great, so I will defer to @jwlodek on this one. My understanding was that we want multiple images (or data points) per event. So maybe "exposure" is not the right word and we just use "point" instead.

> All of those fields in `TriggerInfo` should be optional apart from `number_of_events`. If there is an optional field, it should take its value from the currently set areaDetector value. `number_of_events` should always be `1` in a step scan.

Agreed! I think this would be great.

> I think I made a mistake in the current implementation, the passed in TriggerInfo should be immutable and we make a copy with a different deadtime. That would stop any surprises when a `TriggerInfo` was reused.

Yes currently it modifies the existing `TriggerInfo`

tests/epics/adcore/test_writers.py
```diff
@@ -99,13 +136,14 @@ async def test_hdf_writer_file_not_found(hdf_writer: adcore.ADHDFWriter):
 
 async def test_hdf_writer_collect_stream_docs(hdf_writer: adcore.ADHDFWriter):
     assert hdf_writer._composer is None
-
+    _ = await hdf_writer.open(exposures_per_event=1)
```
coretl, line 139: Don't need the return value here as it's None
```suggestion
    await hdf_writer.open(exposures_per_event=1)
```

tests/epics/adcore/test_writers.py
```diff
@@ -99,13 +136,14 @@ async def test_hdf_writer_file_not_found(hdf_writer: adcore.ADHDFWriter):
...
     [item async for item in hdf_writer.collect_stream_docs(1)]
     assert hdf_writer._composer
 
 
 async def test_tiff_writer_collect_stream_docs(tiff_writer: adcore.ADTIFFWriter):
     assert tiff_writer._emitted_resource is None
+    _ = await tiff_writer.open(exposures_per_event=1)
```
coretl, line 146: ditto

src/ophyd_async/epics/adsimdetector/_sim.py
```diff
@@ -21,8 +21,8 @@ def __init__(
         config_sigs: Sequence[SignalR] = (),
         plugins: dict[str, adcore.NDPluginBaseIO] | None = None,
     ):
-        driver = SimDriverIO(prefix + drv_suffix)
-        controller = SimController(driver)
+        driver: SimDriverIO = SimDriverIO(prefix + drv_suffix)
+        controller: SimController = SimController(driver)
```
coretl, line 25: I'm surprised these were needed, can't pyright infer their type here without them?

tests/sim/test_sim_blob_detector.py
```diff
@@ -50,14 +50,16 @@ async def test_sim_blob_detector_count(
...
             "dtype": "array",
             "object_name": "det",
             "external": "STREAM:",
         },
         "det-sum": {
             "source": "sim://pattern-generator-hdf-file",
-            "shape": [],
+            "shape": [
+                1,
+            ],
```
coretl, line 62: @danielballan @genematx @jwlodek if we follow the conclusion of making the descriptor shape be the same rank as the final dataset, then we end up with `[1]` for scalar datasets rather than `[]`. Is that right?

jwlodek: Yes, that's my understanding on what we settled on

src/ophyd_async/epics/adsimdetector/_sim.py
thomashopkins32: Yes, it infers them fine. I forget why I did this.

tests/sim/test_sim_blob_detector.py
coretl: In which case we need to update every scalar Signal to also produce `[1]` as shape rather than `[]`

genematx: In `TiledWriter` we currently support [both cases](https://github.com/bluesky/bluesky/blob/53f765138d81619acf10fb17101cec9386bbdce1/src/bluesky/consolidators.py#L92C14-L92C25); in fact, we convert `()` to `(1,)`, which we can change to make it more consistent too.

jwlodek (approved):
Tested w/ the sister bluesky PR, in a variety of cases, and they all seem to work.

thomashopkins32:
@jwlodek @coretl @genematx Please review the change of default behavior for single scalars, then we should be good to merge.

Although either `[1]` or `[]` are acceptable to use in the descriptor, the default will be `[]`.

One final question is whether to say whether the `dtype` is a "number" or "array". As of the most recent commit, only `[]` and `[1]` are considered a "number" while everything else is an "array".

genematx:
Thank you, @thomashopkins32. LGTM!
I think how you treat `dtype` is consistent with what we had before, although we don't even use it in Tiled (`dtype_numpy` is very important though).

coretl (approved):
Fine to merge from what I can see, but I'll leave final approval to @jwlodek

PR #728, opened 2025-01-10
Remove assert in production code

kivel:
The changes will remove `assert` statements and replace them with check/raise.
Logic should remain identical, except for velocity related assertions in the timeout calculation.

### new ruff rule
rule S101 was added to avoid future use of `assert`

### Tango
There are a few remaining uses of assert in the Tango code. I ask for a Tango developer to take a look at those. If fail understand why they are used and also struggle with seemingly unrelated tests failing when I touch them.

### assert in general
I understand that a simple `assert foo.bar` will make language support tools happy, as it is certain that `bar` exists for subsequent code. There are alternatives, admittedly, not as straight forward, though.
The general question for code hygiene should be, what comes first, a happy type checker or readable code. It happened a few time in my quest to remove the assert statements that it was unclear to me, why assert was used in the first place. I'd at least add a comment, if assert is used to make Pylance happy.

### velocity checks
Previously, the velocity needed to be positive. The error message was stating however that the velocity is zero. Since the EPICS motorRecord will work with both, the check was changed to specifically check for zero. Technically, the check isn't even needed, as the subsequent timeout calculation will throw a DivisionByZero exception. This is of course, in the case we allow negative velocities ... which we should.

### DIFF
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -267,6 +267,13 @@ class StandardDetector(
     async def describe(self) -> dict[str, DataKey]:
         return self._describe
 
+    @staticmethod
+    def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
+        # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
+        if trigger_info is None:
+            raise RuntimeError("Trigger info must be set before calling this method.")
+        return trigger_info
+
     @AsyncStatus.wrap
     async def trigger(self) -> None:
         if self._trigger_info is None:
@@ -280,15 +287,7 @@ class StandardDetector(
                 )
             )
 
-        # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
-        def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
-            if trigger_info is None:
-                raise RuntimeError(
-                    "Trigger info must be set before calling this method."
-                )
-            return trigger_info
-
-        self._trigger_info = ensure_trigger_info_exists(self._trigger_info)
+        self._trigger_info = self.ensure_trigger_info_exists(self._trigger_info)
         if self._trigger_info.trigger is not DetectorTrigger.INTERNAL:
             msg = "The trigger method can only be called with INTERNAL triggering"
             raise ValueError(msg)
@@ -363,7 +362,7 @@ class StandardDetector(
 
     @WatchableAsyncStatus.wrap
     async def complete(self):
-        assert self._trigger_info
+        self._trigger_info = self.ensure_trigger_info_exists(self._trigger_info)
         indices_written = self._writer.observe_indices_written(
             self._trigger_info.frame_timeout
             or (
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -313,16 +313,17 @@ class StandardDetector(
         Args:
             value: TriggerInfo describing how to trigger the detector
         """
-        if value.trigger != DetectorTrigger.INTERNAL:
-            assert (
-                value.deadtime
-            ), "Deadtime must be supplied when in externally triggered mode"
-        if value.deadtime:
-            required = self._controller.get_deadtime(value.livetime)
-            assert required <= value.deadtime, (
-                f"Detector {self._controller} needs at least {required}s deadtime, "
-                f"but trigger logic provides only {value.deadtime}s"
+        if value.trigger != DetectorTrigger.INTERNAL and not value.deadtime:
+            msg = "Deadtime must be supplied when in externally triggered mode"
+            raise ValueError(msg)
+        required_deadtime = self._controller.get_deadtime(value.livetime)
+        if value.deadtime and required_deadtime > value.deadtime:
+            msg = (
+                f"Detector {self._controller} needs at least {required_deadtime}s "
+                f"deadtime, but trigger logic provides only {value.deadtime}s"
             )
+            raise ValueError(msg)
+
         self._trigger_info = value
         self._number_of_triggers_iter = iter(
             self._trigger_info.number_of_triggers
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -274,13 +274,6 @@ class StandardDetector(
     async def describe(self) -> dict[str, DataKey]:
         return self._describe
 
-    # @staticmethod
-    # def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
-    #     # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
-    #     if trigger_info is None:
-    #         raise RuntimeError("Trigger info must be set before calling this method.")
-    #     return trigger_info
-
     @AsyncStatus.wrap
     async def trigger(self) -> None:
         if self._trigger_info is None:
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -173,6 +173,13 @@ DetectorControllerT = TypeVar("DetectorControllerT", bound=DetectorController)
 DetectorWriterT = TypeVar("DetectorWriterT", bound=DetectorWriter)
 
 
+def _ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
+    # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
+    if trigger_info is None:
+        raise RuntimeError("Trigger info must be set before calling this method.")
+    return trigger_info
+
+
 class StandardDetector(
     Device,
     Stageable,
@@ -267,12 +274,12 @@ class StandardDetector(
     async def describe(self) -> dict[str, DataKey]:
         return self._describe
 
-    @staticmethod
-    def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
-        # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
-        if trigger_info is None:
-            raise RuntimeError("Trigger info must be set before calling this method.")
-        return trigger_info
+    # @staticmethod
+    # def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
+    #     # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
+    #     if trigger_info is None:
+    #         raise RuntimeError("Trigger info must be set before calling this method.")
+    #     return trigger_info
 
     @AsyncStatus.wrap
     async def trigger(self) -> None:
@@ -287,7 +294,7 @@ class StandardDetector(
                 )
             )
 
-        self._trigger_info = self.ensure_trigger_info_exists(self._trigger_info)
+        self._trigger_info = _ensure_trigger_info_exists(self._trigger_info)
         if self._trigger_info.trigger is not DetectorTrigger.INTERNAL:
             msg = "The trigger method can only be called with INTERNAL triggering"
             raise ValueError(msg)
@@ -362,7 +369,7 @@ class StandardDetector(
 
     @WatchableAsyncStatus.wrap
     async def complete(self):
-        self._trigger_info = self.ensure_trigger_info_exists(self._trigger_info)
+        self._trigger_info = _ensure_trigger_info_exists(self._trigger_info)
         indices_written = self._writer.observe_indices_written(
             self._trigger_info.frame_timeout
             or (
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -279,10 +279,20 @@ class StandardDetector(
                     frame_timeout=None,
                 )
             )
-        # TODO: check with core devs if this is actually still needed
-        # I see no way that self._trigger_info remains None
-        assert self._trigger_info
-        assert self._trigger_info.trigger is DetectorTrigger.INTERNAL
+
+        # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
+        def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
+            if trigger_info is None:
+                raise RuntimeError(
+                    "Trigger info must be set before calling this method."
+                )
+            return trigger_info
+
+        self._trigger_info = ensure_trigger_info_exists(self._trigger_info)
+        if self._trigger_info.trigger is not DetectorTrigger.INTERNAL:
+            msg = "The trigger method can only be called with INTERNAL triggering"
+            raise ValueError(msg)
+
         # Arm the detector and wait for it to finish.
         indices_written = await self._writer.get_indices_written()
         await self._controller.arm()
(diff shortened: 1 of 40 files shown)


src/ophyd_async/core/_readable.py
```diff
@@ -123,30 +123,31 @@ def hints(self) -> Hints:
             # we want to combine them when they are Sequences, and ensure they are
             # identical when string values.
             for key, value in new_hint.hints.items():
+                # fail early for unkwon types
+                if not isinstance(value, str | Sequence):
+                    msg = (
+                        f"{new_hint.name}: Unknown type for value '{value}'"
+                        f" for key '{key}'"
+                    )
+                    raise TypeError(msg)
```
coretl, line 132: Any reason to move this to the start of the if block? It means you're doing two `isinstance` checks for each of the happy paths, so is slightly less performant for the happy path which is the one we care about...

src/ophyd_async/core/_readable.py
```diff
@@ -204,6 +205,14 @@ def add_readables(
             `StandardReadableFormat` documentation
         """
 
+        def is_signalr(device: Device) -> TypeGuard[SignalR]:
+            return isinstance(device, SignalR)
+
+        def assert_device_is_signalr(device: Device) -> SignalR:
+            if not is_signalr(device):
+                raise TypeError(f"{device} is not a SignalR")
+            return device
```
coretl, line 214: Could this be written more succinctly?
```suggestion
        def assert_device_is_signalr(device: Device) -> SignalR:
            if not isinstance(device, SignalR):
                raise TypeError(f"{device} is not a SignalR")
            return device
```

src/ophyd_async/core/_signal.py
```diff
@@ -113,8 +113,11 @@ def close(self):
 
     async def get_reading(self) -> Reading[SignalDatatypeT]:
         await self._valid.wait()
-        assert self._reading is not None, "Monitor not working"
-        return self._reading
+        if self._reading:
```
coretl, line 116: Just to make it crystal clear that we are checking for None rather than an empty dict...
```suggestion
        if self._reading is not None:
```

src/ophyd_async/core/_signal.py
```diff
@@ -135,7 +138,9 @@ def _notify(
         function: Callback[dict[str, Reading[SignalDatatypeT]] | SignalDatatypeT],
         want_value: bool,
     ):
-        assert self._reading, "Monitor not working"
+        if not self._reading:
+            msg = "Monitor not working"
+            raise RuntimeError(msg)
```
coretl, line 143: Same as above, is it worth making a utility function for this?

src/ophyd_async/core/_table.py
```diff
@@ -102,7 +102,10 @@ def numpy_table(self, selection: slice | None = None) -> np.ndarray:
             if array is None:
                 array = np.empty(v.shape, dtype=self.numpy_dtype())
             array[k] = v
-        assert array is not None
+        if array is None:
+            # TODO: discuss whether this code is actually reachable
```
coretl, line 106: It is reachable, when there are no columns in the table, so good idea giving it an error message!

src/ophyd_async/epics/core/_aioca.py
```diff
@@ -321,16 +321,18 @@ async def get_setpoint(self) -> SignalDatatypeT:
         return self.converter.value(value)
 
     def set_callback(self, callback: Callback[Reading[SignalDatatypeT]] | None) -> None:
+        if callback and self.subscription:
+            msg = "Cannot set a callback when one is already set"
+            raise RuntimeError(msg)
+
+        if self.subscription:
+            self.subscription.close()
+            self.subscription = None
```
coretl, line 330: Excellent, you caught an unclosed subscription!

src/ophyd_async/epics/motor.py
```diff
@@ -155,13 +157,15 @@ async def set(self, value: float, timeout: CalculatableTimeout = CALCULATE_TIMEO
...
-        if timeout is CALCULATE_TIMEOUT:
-            assert velocity > 0, "Motor has zero velocity"
+        if timeout is CALCULATE_TIMEOUT and velocity != 0:
             timeout = (
-                abs(new_position - old_position) / velocity
+                abs((new_position - old_position) / velocity)
                 + 2 * acceleration_time
                 + DEFAULT_TIMEOUT
             )
+        else:
```
coretl, line 166: This else clause will fire if `timeout=1` which is not what we want...

src/ophyd_async/epics/sim/_mover.py
```diff
@@ -52,9 +52,11 @@ async def set(self, value: float, timeout: CalculatableTimeout = CALCULATE_TIMEO
             self.precision.get_value(),
             self.velocity.get_value(),
         )
-        if timeout == CALCULATE_TIMEOUT:
-            assert velocity > 0, "Mover has zero velocity"
-            timeout = abs(new_position - old_position) / velocity + DEFAULT_TIMEOUT
+        if timeout is CALCULATE_TIMEOUT and velocity != 0:
+            timeout = abs((new_position - old_position) / velocity) + DEFAULT_TIMEOUT
+        else:
+            msg = "Mover has zero velocity"
```
coretl, line 58: ditto

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -735,7 +735,10 @@ def set_callback(self, callback: Callback | None) -> None:
 
         if callback:
             try:
-                assert not self.proxies[self.read_trl].has_subscription()  # type: ignore
+                # TODO: remove assert, I fail to get the test to work
+                # as soon as move the assert into a separate RunTime error
+                # the `test_tango_sim` fails
+                assert not self.proxies[self.read_trl].has_subscription()  # type: ignore  # noqa: S101
```
coretl, line 741: @burkeds any idea why this is the case?

src/ophyd_async/plan_stubs/_nd_attributes.py
```diff
@@ -51,10 +51,12 @@ def setup_ndattributes(
...
 def setup_ndstats_sum(detector: Device):
     hdf = getattr(detector, "fileio", None)
-    assert isinstance(hdf, NDFileHDFIO), (
-        f"Expected {detector.name} to have 'fileio' attribute that is an NDFilHDFIO, "
-        f"got {hdf}"
-    )
+    if not isinstance(hdf, NDFileHDFIO):
+        msg = (
+            f"Expected {detector.name} to have 'fileio' attribute that is an "
+            f"NDFilHDFIO, got {hdf}"
```
coretl, line 57:
```suggestion
            f"NDFileHDFIO, got {hdf}"
```

src/ophyd_async/core/_device_filler.py
```diff
@@ -137,6 +137,12 @@ def check_created(self):
                 f"{self._device.name}: {uncreated} have not been created yet"
             )
 
+    @staticmethod
+    def check_device_annotation(annotation: Any) -> None:
+        if not isinstance(annotation, DeviceAnnotation):
+            msg = f"Annotation {annotation} is not a DeviceAnnotation"
+            raise TypeError(msg)
```
coretl, line 144: As this is only used in this file, please could you make it a free private function in this file...
```
def _check_device_annotation(annotation: Any) -> DeviceAnnotation:
```

src/ophyd_async/core/_device_filler.py
```diff
@@ -150,7 +156,7 @@ def create_signals_from_annotations(
             yield backend, extras
             signal = child_type(backend)
             for anno in extras:
-                assert isinstance(anno, DeviceAnnotation), anno
+                self.check_device_annotation(annotation=anno)
                 anno(self._device, signal)
```
coretl, line 160: Then this would be:
```suggestion
                device_annotation = _check_device_annotation(anno)
                device_annotation(self._device, signal)
```

src/ophyd_async/core/_detector.py
```diff
@@ -267,6 +267,13 @@ async def read(self) -> dict[str, Reading]:
     async def describe(self) -> dict[str, DataKey]:
         return self._describe
 
+    @staticmethod
+    def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
```
coretl, line 271: Please could this either be a private method which checks `self._trigger_info`:
```python
    def _ensure_trigger_info_exists(self) -> TriggerInfo:
```
Or a private free function that is passed `trigger_info`:
```python
def _ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
```

coretl (commented):
Thanks, this looks almost complete! Just a few minor requests

coretl:
> ### velocity checks
> 
> Previously, the velocity needed to be positive. The error message was stating however that the velocity is zero. Since the EPICS motorRecord will work with both, the check was changed to specifically check for zero. Technically, the check isn't even needed, as the subsequent timeout calculation will throw a DivisionByZero exception. This is of course, in the case we allow negative velocities ... which we should.

I think in this case we are calculating a timeout, which should always be positive, so we should be dividing by `abs(velocity)`. The reason we don't just allow the `DivisionByZero` to bubble up is so we can make a slightly nicer error message...

src/ophyd_async/core/_readable.py
kivel: I'll revert it back to the original order.
I personally like to fail early. But that's mostly personal preference.

src/ophyd_async/core/_readable.py
kivel: fixed: ab110f6d125f7d6b6ef714f02b4e70d6809b89d4

src/ophyd_async/core/_readable.py
kivel: fixed: 9dfbb926dda864cdde1aeb3af76edde8816b9c92

src/ophyd_async/core/_signal.py
kivel: given that we can't re-use the `get_reading` because it's async, but the `_callback` and `_notify` are both synchronous, I don't really the the usefulness of a dedicated utility method. Unless I misunderstood you.

src/ophyd_async/core/_table.py
kivel: I removed the TODO

src/ophyd_async/epics/motor.py
kivel: thanks for the catch!
fixed in 101b64ef34dad5a79988c33f4343a98205d39693
also added test cases to prevent this again.

src/ophyd_async/epics/sim/_mover.py
kivel: fixed in 9d8828168188fff0aa43db025d17ac6da05a97f1
I'm not sure how I can test this, though.

src/ophyd_async/core/_device_filler.py
kivel: done 8af924f9a25c4a8673e2eeafcfe51786dddba9c2

src/ophyd_async/core/_device_filler.py
kivel: done de964ad5bb209ef14ed5c5452eb0645de7b7c61b

src/ophyd_async/core/_device_filler.py
kivel: done cc79c3a4598d4ae06df639906742627ff5cc4ef3

src/ophyd_async/core/_detector.py
kivel: fixed 6860060329357f6adb3b4d3301bb84c8930de1e3

kivel:
@coretl  I addressed your requests to my abilities.

src/ophyd_async/core/_signal.py
coretl: I wondered if it was worth doing this:
```python
def _ensure_reading(self):
        if not self._reading:
            msg = "Monitor not working"
            raise RuntimeError(msg)
        return self._reading
```
Then in the 2 places we use `self._reading` we instead do `reading = self._ensure_reading()`

kivel: Sure, that makes it much nicer to read!
added here 5098783e629a91675d6ee133d831f4e2493ad6a3
test have been added for both happy cases, the RuntimeError was already there.

also added missing type hints for the class
a769e576f40286be50acd786e1ce285b96d658d8

src/ophyd_async/core/_detector.py
```diff
@@ -267,6 +274,13 @@ async def read(self) -> dict[str, Reading]:
     async def describe(self) -> dict[str, DataKey]:
         return self._describe
 
+    # @staticmethod
+    # def ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo:
+    #     # make absolute sure we realy have a valid TriggerInfo ... mostly for pylance
+    #     if trigger_info is None:
+    #         raise RuntimeError("Trigger info must be set before calling this method.")
+    #     return trigger_info
+
```
coretl, line 283: Please can you remove the commented out code

coretl (approved):

src/ophyd_async/core/_detector.py
kivel: sorry, totally overlooked that
done a3fa08d8d7b84681d18139e321a8ba2e1e648019

kivel:
> Please could you resolve the conflicts and remove that commented out code and then I will merge
sure!

merged main into feature and fixed all conflicts 34e0b702ffea4e805d1aa157641eac75cdef4fcf

coretl:
Looks like it still has conflicts? Also, lint seems to be failing...

kivel:
I have absolutely no idea why the `ruff format` step in the workflow insists on reformatting the 4 files and make it fail.
It looks to me as if a different formatting rule is used in the workflow. All the tools work on my host in the provide dev-container
```bash
(venv) root ➜ /workspaces/ophyd-async (710_remove_assert) $ ruff format
178 files left unchanged
(venv) root ➜ /workspaces/ophyd-async (710_remove_assert) $ pre-commit run -a
check for added large files..............................................Passed
check yaml...............................................................Passed
check for merge conflicts................................................Passed
fix end of files.........................................................Passed
lint with ruff...........................................................Passed
format with ruff.........................................................Passed
Ensure import directionality.............................................Passed
(venv) root ➜ /workspaces/ophyd-async (710_remove_assert) $ git status
On branch 710_remove_assert
Your branch is up to date with 'origin/710_remove_assert'.

nothing to commit, working tree clean
```

coretl:
> I have absolutely no idea why the `ruff format` step in the workflow insists on reformatting the 4 files and make it fail.

I think that is because ruff changed its formatting rules (looks like it does that once a year): https://github.com/astral-sh/ruff/releases/tag/0.9.0

If you run `pip install ruff --upgrade` in your venv then `ruff format` then it should match CI

src/ophyd_async/tango/core/_tango_transport.py
kivel: Hmm, I gave it a another go, using the same logic as I use for `aioca` and `p4p`, and that worked. When I moved the AssertionError, that is re-raised as RuntimeError into a separate if-statement, a bunch of tests failed.
fixed in 8592cbd7aa8780e048922bad95ab9776ab19eac0

kivel:
@burkeds Thanks for looking into the Tango part.

@coretl Thanks for pointing out the need to update `ruff`. That made the CI work.
That should conclude this PR.

coretl (approved):

Issue #750, opened 2025-01-27
Improve error message on bad enum set


DominicOram:
When I set a value to an `mbbi` to a string value that isn't one of the expected values I get the following error message:

<details>

```
Run aborted
Traceback (most recent call last):
  File "/scratch/ffv81422/dodal/src/dodal/devices/oav/oav_detector.py", line 51, in set
    await self.level.set(level_to_set, wait=True)
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/core/_signal.py", line 240, in set
    await _wait_for(self._connector.backend.put(value, wait=wait), timeout, source)
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/core/_signal.py", line 44, in _wait_for
    return await asyncio.wait_for(coro, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/tasks.py", line 479, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/epics/core/_aioca.py", line 283, in put
    await caput(
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/aioca/_catools.py", line 127, in call_wrapper
    return await throw_wrapper(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/aioca/_catools.py", line 109, in throw_wrapper
    return await awaitable
           ^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/aioca/_catools.py", line 159, in ca_timeout
    result = await awaitable
             ^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/aioca/_catools.py", line 840, in caput
    await done.wait()
  File "/scratch/ffv81422/mx-bluesky/.venv/lib/python3.11/site-packages/aioca/_catools.py", line 69, in wait
    raise self.value
[...]

PR #752, opened 2025-01-27
Update to offer more helpful error message if str fallback is used wi…

jwlodek:
…th enum

Addresses #750

### DIFF
--- a/src/ophyd_async/epics/core/_aioca.py
+++ b/src/ophyd_async/epics/core/_aioca.py
@@ -312,6 +312,8 @@ class CaSignalBackend(EpicsSignalBackend[SignalDatatypeT]):
                         f"valid choices: {self.converter.metadata.get('choices')}"
                     )
                     raise ValueError(msg) from exc
+                else:
+                    raise
             else:
                 raise
 
--- a/src/ophyd_async/epics/core/_aioca.py
+++ b/src/ophyd_async/epics/core/_aioca.py
@@ -293,13 +293,27 @@ class CaSignalBackend(EpicsSignalBackend[SignalDatatypeT]):
             write_value = self.initial_values[self.write_pv]
         else:
             write_value = self.converter.write_value(value)
-        await caput(
-            self.write_pv,
-            write_value,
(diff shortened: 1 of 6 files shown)


src/ophyd_async/epics/core/_aioca.py
```diff
@@ -293,13 +293,29 @@ async def put(self, value: SignalDatatypeT | None, wait: bool):
...
+                write_value,
+                datatype=self.converter.write_dbr,
+                wait=wait,
+                timeout=None,
+            )
+        except CANothing as exc:
+            # If we ran into a write error, check to see if there is a list
+            # of valid choices, and if the value we tried to write is in that list.
+            valid_choices = self.converter.metadata.get("choices")
+            if valid_choices is not None and len(valid_choices) > 0:
```
DominicOram, line 308: Could: 
```suggestion
            if valid_choices:
```
is equivalent.

src/ophyd_async/epics/core/_aioca.py
```diff
@@ -293,13 +293,29 @@ async def put(self, value: SignalDatatypeT | None, wait: bool):
...
+                if value not in valid_choices:
+                    msg = (
+                        f"{value} is not a valid choice for {self.write_pv}, "
+                        f"valid choices: {self.converter.metadata.get('choices')}"
+                    )
+                    raise ValueError(msg) from exc
+                else:
+                    raise
+            else:
+                raise
```
DominicOram, line 318: Could: I don't think the else statements are needed:
```suggestion
            if valid_choices is not None and len(valid_choices) > 0:
                if value not in valid_choices:
                    msg = (
                        f"{value} is not a valid choice for {self.write_pv}, "
                        f"valid choices: {self.converter.metadata.get('choices')}"
                    )
                    raise ValueError(msg) from exc
            raise
```

DominicOram (approved):
Awesome, thank you for addressing this quickly! I've added a couple of nit-picks but they're just over style, take them or leave them

coretl (approved):

PR #759, opened 2025-02-11
Tweak wording of article

coretl:


abbiemery (approved):
This looks like it covered the points I mentioned. The description of the device class is now clearer I think.

PR #764, opened 2025-02-13
Rewrite docstrings in markdown

### LINKED ISSUE #755: Decide on docstring format
After spending 2 weeks with sphinx autodoc and autosummary I have failed to get it to document typevars properly, which means a bunch of unclickable links in the docs everytime you write `SignalDatatypeT`. This is annoying as it happens in hundreds of places, and the user generally would like to know at this point "which datatypes could I pass here?" so really needs that link.

To fix, we could:
1. Manually link to the datatypes documentation in every docstring
2. [Fix autodoc](https://github.com/sphinx-doc/sphinx/pull/13277)
3. Switch to [`sphinx-autodoc2`](https://github.com/sphinx-extensions2/sphinx-autodoc2)

I think 1. is too manual and error prone, I failed at 2. after 2 days of hacking, so I suggest 3.

It is almost working, but there is one decision to make, what format should the docstrings be?

Given that prose documents are written in markdown and look like this:
```md
# Connecting the Device
Rather than calling [](#Device.connect) yourself which would use the wrong event loop you can 
use [](#init_devices) at startup or the equivalent [plan stub](#ensure_connected). Remember to 
pass `mock=True` during testing.
```
We can either write the docstring in markdown or RST, and we can format the arguments using param list, google style or numpy style. I've included a sample docstring below that we can comment on:

## Numpy style with RST links
```python
    def create_devices_from_annotations(
        self,
        filled=True,
    ) -> Iterator[tuple[DeviceConnectorT, list[Any]]]:
        """Create all Signals from annotations

        Parameters
        ----------
        filled
            If ``True`` then the Devices created should be considered already filled
            with connection data. If ``False`` then `fill_child_device` needs
            calling at parent device connection time before the child `Device` can
            be connected.

        Yields
        ------
        (connector, extras):
            The `DeviceConnector` that has been created for this Signal, and the list of
            extra annotations that could be used to customize it.
        """
```
## Google style with RST links
```python
    def create_devices_from_annotations(
        self,
        filled=True,
    ) -> Iterator[tuple[DeviceConnectorT, list[Any]]]:
        """Create all Signals from annotations

        Args:
            filled: If ``True`` then the Devices created should be considered
                already filled with connection data. If ``False`` then
                `fill_child_device` needs calling at parent device connection
                time before the child `Device` can be connected.

        Yields:
            (connector, extras): The `DeviceConnector` that has been created for this
                Signal, and the list of extra annotations that could be used to
                customize it.
        """
```
## Param list with markdown links
```python
    def create_devices_from_annotations(
        self,
        filled=True,
    ) -> Iterator[tuple[DeviceConnectorT, list[Any]]]:
        """Create all Signals from annotations

        :param filled: If `True` then the Devices created should be considered
            already filled with connection data. If `False` then
            [](#fill_child_device) needs calling at parent device connection
            time before the child [](#Device) can be connected.

        :yields: `(connector, extras)`: the [](#DeviceConnector) that has been
            created for this Signal, and the list of extra annotations that
            could be used to customize it.
        """
```

Or any combination of the above.

## Acceptance Criteria
- Decision is made and documented in an ADR

coretl:
And add ADR to describe

Fixes #755

1. Run `tox -e docs autobuild`
2. Look at the docs and check for mistakes

### Checks for reviewer
- [ ] Would the PR title make sense to a user on a set of release notes
- [ ] If the change requires a bump in an IOC version, is that specified in a `##Changes` section in the body of the PR
- [ ] If the change requires a bump in the PandABlocks-ioc version, is the `ophyd_async.fastcs.panda._hdf_panda.MINIMUM_PANDA_IOC` variable updated to match


src/ophyd_async/epics/motor.py
```diff
@@ -42,7 +47,7 @@ class InvalidFlyMotorException(Exception):
 
 
 class FlyMotorInfo(BaseModel):
-    """Minimal set of information required to fly a motor:"""
+    """Minimal set of information required to fly a motor."""
```
evalott100, line 50: Are the `#:` comments here also used for docs? There are other places where they're explicitly converted to `"""`.

src/ophyd_async/plan_stubs/_settings.py
```diff
@@ -85,15 +87,12 @@ def apply_settings_if_different(
...
+    """Set every SignalRW in settings, only if it is different to the current value.
 
-    Parameters
-    ----------
-    apply_plan:
+    :param apply_plan:
         A device specific plan which takes the Settings to apply and applies them to
         the Device. Used to add device specific ordering to setting the signals.
-    current_settings:
+    :current_settings:
```
evalott100, line 95:
```suggestion
    :param current_settings:
```
?

src/ophyd_async/tango/core/_signal.py
```diff
@@ -48,7 +48,7 @@ def tango_signal_rw(
     timeout: float = DEFAULT_TIMEOUT,
     name: str = "",
 ) -> SignalRW[SignalDatatypeT]:
-    """Create a `SignalRW` backed by 1 or 2 Tango Attribute/Command
+    """Create a `SignalRW` backed by 1 or 2 Tango Attribute/Command.
 
     Parameters
```
evalott100, line 53: Docstring isn't converted

src/ophyd_async/tango/core/_signal.py
```diff
@@ -76,7 +77,7 @@ def tango_signal_r(
     timeout: float = DEFAULT_TIMEOUT,
     name: str = "",
 ) -> SignalR[SignalDatatypeT]:
-    """Create a `SignalR` backed by 1 Tango Attribute/Command
+    """Create a `SignalR` backed by 1 Tango Attribute/Command.
 
     Parameters
```
evalott100, line 82: .

src/ophyd_async/tango/core/_signal.py
```diff
@@ -102,7 +104,7 @@ def tango_signal_w(
     timeout: float = DEFAULT_TIMEOUT,
     name: str = "",
 ) -> SignalW[SignalDatatypeT]:
-    """Create a `SignalW` backed by 1 Tango Attribute/Command
+    """Create a `SignalW` backed by 1 Tango Attribute/Command.
 
     Parameters
```
evalott100, line 109: etc.

evalott100 (approved):

src/ophyd_async/epics/motor.py
coretl: No, `sphinx-autodoc2` doesn't support doc comments, so I needed to convert to `"""`. I may have missed some...

src/ophyd_async/tango/core/_signal.py
coretl: I was going to leave Tango for another day...

src/ophyd_async/core/_readable.py
```diff
@@ -17,25 +17,34 @@
...
     CHILD = "CHILD"
-    #: Contribute the `Signal` value to ``read_configuration()`` and
-    #: ``describe_configuration()``
+    """Detect which verbs the child supports and contribute to:
+
+    - `read()`, `describe()` if it is [](#bluesky.protocols.Readable)
+    - `read_configuration()`, `describe_configuration()` if it is
+      [](#bluesky.protocols.Configurable)
+    - `stage()`, `unstage()` if it is [](#bluesky.protocols.Stageable)
+    - `hints` if it [](#bluesky.protocols.HasHints)
```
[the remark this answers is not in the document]
coretl: I think as the link to the protocol is right next to it this is probably good enough

PR #766, opened 2025-02-14
Pass datatype between eiger and odin

### LINKED ISSUE #529: FastCS Eiger: Pass datatype from eiger to odin
Following https://github.com/bluesky/ophyd-async/pull/528 we should continue implementing the full functionality for the `eiger` that we need in MX. The datatype of data coming out of the eiger is dependant on the exposure time set on the detector. Odin needs to know the datatype up front so that it can correctly write data. During arm/open we need to get this out of the eiger and put it into odin. e.g. see https://github.com/DiamondLightSource/dodal/blob/46132f0e066debaf906c1a80dcefca175b3f5c01/src/dodal/devices/eiger.py#L320

Note that we need to do this after the exposure time has been accepted by the eiger as this is when it calculates the new bit depth

## Acceptance Criteria
* When the `fastcs` eiger is armed it correctly forwards the bit depth to odin

shihab-dls:
Fixes #529 

This PR simply retrieves the bit_depth from Eiger and puts it into Odin. This needs to be done after setting exposure, although that is beyond the scope of this PR.


shihab-dls:
Seems linting is failing for an unrelated reason to this PR, so will request a review anyways.

coretl (commented):
This seems about right to me, but I'm assuming #806 will be merged first, so I'll re-review when that is in

shihab-dls:
Given the order in which we are merging eiger-odin changes (see final #806 comments), I'm closing this PR in favour of #870

PR #769, opened 2025-02-14
Add to observe_value docstring

### LINKED ISSUE #741: Document observe_value better
When working with `observe_value` it's not always obvious on the edge cases of it e.g.:
* Do I get a value immediately on calling `observe_value`?
* When do I get a new value? Do I get one if the there's a callback but the value remains the same?

I think the answers to this are both yes but we should make sure and document it in the docstring

## Acceptance Criteria
- The docstring of `observe_value` answers the above questions

coretl:
Fixes #741

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -364,6 +364,10 @@ async def observe_value(
 ) -> AsyncGenerator[SignalDatatypeT, None]:
     """Subscribe to the value of a signal so it can be iterated from.
 
+    The first value yielded in the iterator will be the current value of the
+    Signal, and subsequent updates from the control system will result in that
+    value being yielded, even if it is the same as the previous value.
+
     :param signal:
         Call subscribe_value on this at the start, and clear_sub on it at the end.
     :param timeout:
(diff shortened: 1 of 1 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -412,6 +416,10 @@ async def observe_signals_value(
 ) -> AsyncGenerator[tuple[SignalR[SignalDatatypeT], SignalDatatypeT], None]:
     """Subscribe to the value of a signal so it can be iterated from.
```
DominicOram, line 417: Nit: Outside scope I guess but:
```suggestion
    """Subscribe to a set of signals so they can be iterated from.
```

src/ophyd_async/core/_signal.py
```diff
@@ -412,6 +416,10 @@ async def observe_signals_value(
...
 
+    The first value yielded in the iterator will be the current value of the
+    Signal, and subsequent updates from the control system will result in that
```
DominicOram, line 420:
```suggestion
    The first values yielded in the iterator will be the current values of the
    Signals, and subsequent updates from the control system will result in that
```

DominicOram (approved):
Great, thank you. Minor comment, take it or leave it.

PR #780, opened 2025-02-21
add use/set switch

rerpha:
As per the epics motor record's "calibration related fields"  - we would like a way to redefine motor positions. 

I think ophyd does have this: https://github.com/bluesky/ophyd/blob/main/ophyd/epics_motor.py#L50

Please let me know if this can be achieved in another way without manually calculating a new offset based on the current position ( `.OFF`)


src/ophyd_async/epics/motor.py
```diff
@@ -77,6 +83,7 @@ def __init__(self, prefix: str, name="") -> None:
         self.motor_done_move = epics_signal_r(int, prefix + ".DMOV")
         self.low_limit_travel = epics_signal_rw(float, prefix + ".LLM")
         self.high_limit_travel = epics_signal_rw(float, prefix + ".HLM")
+        self.use_set_toggle = epics_signal_rw(UseSetMode, prefix + ".SET")
```
coretl, line 86: Please could you use the same name as ophyd?

coretl (approved):

PR #781, opened 2025-02-21
First pass at stopping standard detector from failing to unstage afte…

### LINKED ISSUE #744: Unstaging a `StandardDetector` breaks if it was previously triggered unsuccessfully
In this even `StandardDetector` tries to await the previous coroutine, which will never succeed.

## Steps To Reproduce
Steps to reproduce the behaviour:
- `count` a detector, kill and restart the IOC midway through (recommend counting a lot of points so you have time to do this)
- Try to `count` it again

## Acceptance Criteria
- Unstaging a `StandardDetector` does not break if it was previously triggered unsuccessfully

jwlodek:
…r unsuccessful trigger

Resolves #744  - tested with a VimbaDetector and it appears to work. Before I work on tests, I'd like a once over to see if the approach is sound.

TODO:

- [ ] Fix tests
- [ ] Add `is_open` and `is_armed` to all writer/controller classes respectively
- [ ] Test with a more complex real hardware setup

### DIFF
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -122,6 +122,10 @@ class DetectorController(ABC):
     async def disarm(self):
         """Disarm the detector, return detector to an idle state."""
 
+    @abstractmethod
+    def is_armed(self) -> bool:
+        """Return True if the detector is armed, False otherwise."""
+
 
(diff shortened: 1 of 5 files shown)


jwlodek:
After some further consideration, I think this may be overkill. Working on a simpler fix...

PR #783, opened 2025-02-24
add offset mode switch and other missing motor fields

rerpha:
like #780 but for the `.FOFF` field.

it's in ophyd but not here: https://github.com/bluesky/ophyd/blob/main/ophyd/epics_motor.py#L49 

I've also added `LLS` and `HLS` - might be useful if checking you aren't on a limit before starting a scan

### DIFF
--- a/src/ophyd_async/epics/motor.py
+++ b/src/ophyd_async/epics/motor.py
@@ -83,22 +83,14 @@ class Motor(StandardReadable, Locatable, Stoppable, Flyable, Preparable):
         self.motor_done_move = epics_signal_r(int, prefix + ".DMOV")
         self.low_limit_travel = epics_signal_rw(float, prefix + ".LLM")
         self.high_limit_travel = epics_signal_rw(float, prefix + ".HLM")
-        self.offset_mode = epics_signal_rw(OffsetMode, prefix + ".FOFF")
+        self.offset_freeze_switch = epics_signal_rw(OffsetMode, prefix + ".FOFF")
         self.high_limit_switch = epics_signal_r(float, prefix + ".HLS")
         self.low_limit_switch = epics_signal_r(float, prefix + ".LLS")
-        self.direction_of_travel = epics_signal_r(float, prefix + ".TDIR")
 
-        # You probably don't want to use this in most cases - .DMOV is
-        # probably more definitive. You might wish to use it if you use SPMG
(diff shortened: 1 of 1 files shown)


src/ophyd_async/epics/motor.py
coretl, line 1: I'm happy with `FOFF`, `HLS`, `LLS`, but I'd rather not add `HOMF` and `HOMR` as we explicitly hide these at DLS (we use custom homing routines instead), and I'm dubious about confusing people with `TDIR` and `MOVN`

Please could you check that the names match the ophyd ones.

rerpha: fair enough - thanks for reviewing, have just updated.

src/ophyd_async/epics/motor.py
```diff
@@ -83,12 +88,16 @@ def __init__(self, prefix: str, name="") -> None:
         self.motor_done_move = epics_signal_r(int, prefix + ".DMOV")
         self.low_limit_travel = epics_signal_rw(float, prefix + ".LLM")
         self.high_limit_travel = epics_signal_rw(float, prefix + ".HLM")
+        self.offset_freeze_switch = epics_signal_rw(OffsetMode, prefix + ".FOFF")
+        self.high_limit_switch = epics_signal_r(float, prefix + ".HLS")
+        self.low_limit_switch = epics_signal_r(float, prefix + ".LLS")
```
coretl, line 93: Does `bool` work? If not it should at least be `int`.
```suggestion
        self.high_limit_switch = epics_signal_r(bool, prefix + ".HLS")
        self.low_limit_switch = epics_signal_r(bool, prefix + ".LLS")
```
I'm confused that this didn't complain when you try to connect to a real motor, it should have errored saying `DBR_SHORT can't be coerced to type float`

rerpha: ah - turns out i was testing against the wrong thing (our motor wrapper) - whoops! 

yes - i think it needs to be `int` as i get: 
```
    high_limit_switch: TypeError: TE:NDW2932:MOT:MTR0101.HLS with inferred datatype int cannot be coerced to bool
    low_limit_switch: TypeError: TE:NDW2932:MOT:MTR0101.LLS with inferred datatype int cannot be coerced to bool
```

coretl (approved):

Issue #596, opened 2024-09-25
Create a version of AD StandardDetector that allows for collecting during continuous acquisition


jwlodek:
At some of our beamlines, the modus operandi for specific detectors is to run continuously and use the file plugin's capture button or save button (in single mode) to capture individual frames. This is specifically useful for some slow flat panel detectors that need to warm up and stay in acquisition mode to get acceptable noise levels, and with some of our monitoring/diagnostic cameras, where single snapshots are collected for logs.

## Acceptance Criteria
- Class similar to `SingleTriggerDetector` is created that will use either single mode / save or capture mode / num capture = 1 / capture as the trigger
- Tests that check all functionality
- Test with real detector/camera.

coretl:
Is this for file-writing detectors?

jwlodek:
Yes, specifically we want to integrate a Perkin Elmer flat panel detector. To get acceptable performance, it must run continuously, and we use the "save" button in the AD file plugin as the trigger in the ophyd v1 object.

coretl:
I suggest that we do this with `StandardDetector`:
- Make a `PerkinElmerController` that controls both a PerkinElmer and an [NDPluginCircularBuff](https://areadetector.github.io/areaDetector/ADCore/NDPluginCircularBuff.html):
  - On `prepare` (or maybe `stage`?) in internal trigger mode starts the driver in continuous mode and sets the number of triggers on the circular buffer plugin with a buffer of 1 and post-buffer of N-1
  - On `arm` starts the circular buffer plugin going
  - On `disarm` stops the circular buffer
- Then the standard HDF writer plugin should do the right thing

PR #793, opened 2025-02-26
Add support for a continuous acquisition interface for AD dets

jwlodek:
Uses the Circular Buffer plugin.

I've tested that this works with the `ContAcqVimbaDetector`. Before I continue it would be good if someone could take this for a spin, or have a look-over.

TODO:

- [x] New tests for the cont acq interface
- [x] Modify existing AD factory fixture to toggle between cont acq/not-cont acq
- [x] Update docs to include information on configuring cont acq detector.

Addresses #596


src/ophyd_async/epics/adcore/_core_detector.py
```diff
@@ -39,3 +39,28 @@ def get_plugin(
                 f"Expected {self.name}.{name} to be a {plugin_type}, got {plugin}"
             )
         return plugin
+
+
+class ContAcqAreaDetector(StandardDetector[ADBaseContAcqControllerT, ADWriter]):
```
coretl, line 44: Do we need this class? ...

src/ophyd_async/epics/adsimdetector/_sim.py
```diff
@@ -56,3 +56,52 @@ def __init__(
...
+        cb_plugin_prefix: str,
+        driver: SimDriverIO,
+    ):
+        super().__init__(cb_plugin_prefix, driver)
+
+    def get_deadtime(self, exposure: float | None) -> float:
+        return 0.001
+
+
+class ContAcqSimDetector(adcore.ContAcqAreaDetector[ContAcqSimController]):
```
coretl, line 75: ... couldn't we just derive from `adcore.AreaDetector` here?

src/ophyd_async/epics/adcore/_core_logic.py
```diff
@@ -112,3 +121,62 @@ async def complete_acquisition() -> None:
                 )
 
         return AsyncStatus(complete_acquisition())
+
+
+class ADBaseContAcqController(DetectorController, Generic[ADBaseIOT]):
```
coretl, line 126: I think this logic looks fine

src/ophyd_async/epics/advimba/_vimba.py
```diff
@@ -39,3 +39,38 @@ def __init__(
             name=name,
             config_sigs=config_sigs,
         )
+
+
+class ContAcqVimbaDetector(adcore.ContAcqAreaDetector[ContAcqVimbaController]):
```
coretl, line 44: Is this something we would ever want to use with a Vimba detector? If not, then we probably shouldn't include this class in case we confuse people...

jwlodek: We actually do have cases where this would be desirable - specifically in cases where these are monitoring cameras that the beamline wants to just take a snapshot with of some current setup for their logbook.

coretl: Hmm, but maybe for that case we should be just controlling the file writer rather than the driver, can we discuss in the collab meeting today?

src/ophyd_async/epics/adsimdetector/_sim.py
jwlodek: Originally I was going to follow the same model we did for the default classes, since different drivers can have slightly different behavior w.r.t. setting exposure while running for example, but I decided to now just make a class that covers 99% of cases that just checks to make sure we are already in the right mode. If for whatever reason people need to be able to change the modes they can make another class, or use `load_settings`.

jwlodek:
The test failures appear to be all tango related, so nothing to do with the changes in the PR...

jwlodek:
Tested w/ real camera and most recent changes and it all works as expected.

src/ophyd_async/epics/adcore/_core_detector.py
```diff
@@ -39,3 +41,40 @@ def get_plugin(
...
+
+
+class ContAcqAreaDetector(AreaDetector[ADBaseContAcqController]):
+    """Ophyd-async implementation of a continuously acquiring AreaDetector."""
+
+    def __init__(
+        self,
+        prefix: str,
+        path_provider: PathProvider,
+        drv_suffix: str = "cam1:",
```
coretl, line 53: Should we add a `drv_cls = ADBaseIO` here in case we want to expose the extra PVs the class has? Most of the time I imagine that doesn't matter, but it would probably be useful to have the option...

tests/epics/adcore/test_cont_acq_detector.py
```diff
@@ -0,0 +1,175 @@
...
+                trigger=trigger_mode,
+                livetime=livetime,
+                deadtime=0.001,
+            )
+        else:
+            return TriggerInfo(
+                number_of_triggers=1, trigger=trigger_mode, livetime=livetime
+            )
+
+    return generate_one_shot_trigger_info
```
coretl, line 50: I'm not a fan of wrapping functions as fixtures, not sure it adds anything...
```suggestion
def generate_one_shot_trigger_info(
    trigger_mode=DetectorTrigger.INTERNAL, livetime=0.8
):
        if trigger_mode != DetectorTrigger.INTERNAL:
            return TriggerInfo(
                number_of_triggers=1,
                trigger=trigger_mode,
                livetime=livetime,
                deadtime=0.001,
            )
        else:
            return TriggerInfo(
                number_of_triggers=1, trigger=trigger_mode, livetime=livetime
            )
```

tests/epics/adcore/test_cont_acq_detector.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+async def test_cont_acq_controller_invalid_trigger_mode(
+    cont_acq_controller: adcore.ADBaseContAcqController, one_shot_trigger_info_factory
+):
+    trigger_info = one_shot_trigger_info_factory(
+        trigger_mode=DetectorTrigger.CONSTANT_GATE
+    )
```
coretl, line 58: ...
```suggestion
async def test_cont_acq_controller_invalid_trigger_mode(
    cont_acq_controller: adcore.ADBaseContAcqController,
):
    trigger_info = generate_one_shot_trigger_info(
        trigger_mode=DetectorTrigger.CONSTANT_GATE
    )
```

coretl (commented):
Looks good, a couple of small suggestions

src/ophyd_async/epics/adcore/_core_detector.py
coretl: at the moment there is only one circular buffer class in areaDetector, so I think we should hardcode that, while there are many detector drivers that we should probably support

jwlodek: Right, since the Circular Buffer plugin is a default plugin that will be included with every driver, and there is only one variant, I think we can leave as is. There are multiple different writers and drivers, hence passing in classes.

thomashopkins32: Good with me then

PR #798, opened 2025-02-28
Auto fill devices

burkeds:
Added `auto_fill_signals` as an kwarg for `TangoDevice` which defaults to True. If false, the device will not auto-fill non-annotated signals on connect().

### DIFF
--- a/src/ophyd_async/tango/core/_base_device.py
+++ b/src/ophyd_async/tango/core/_base_device.py
@@ -33,7 +33,9 @@ class TangoDevice(Device):
         device_proxy: DeviceProxy | None = None,
         support_events: bool = False,
         name: str = "",
+        auto_fill_signals: bool = True,
     ) -> None:
+        self._auto_fill_signals = auto_fill_signals
         connector = TangoDeviceConnector(
             trl=trl, device_proxy=device_proxy, support_events=support_events
         )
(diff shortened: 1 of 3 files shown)


src/ophyd_async/tango/core/_base_device.py
```diff
@@ -102,28 +104,29 @@ async def connect_mock(self, device: Device, mock: LazyMock):
...
-            signal_type = await infer_signal_type(full_trl, self.proxy)
-            if signal_type:
-                backend = self.filler.fill_child_signal(name, signal_type)
-                backend.datatype = await infer_python_type(full_trl, self.proxy)
-                backend.set_trl(full_trl)
-        # Check that all the requested children have been filled
-        self.filler.check_filled(f"{self.trl}: {children}")
-        # Set the name of the device to name all children
-        device.set_name(device.name)
+        if getattr(device, "_auto_fill_signals", False) is True:
```
coretl, line 107: I don't think this is quite right, this won't fill any signals, including those that are type hinted. I think we need a new `get_unfilled() -> list[str]` method on `DeviceFiller`, then if `not self.auto_fill_signals` then we set `children = self.filler.get_unfilled()`

burkeds: I saw this wasn't working properly when I added a test. It should work not. In connect_real I populate a list of unfilled children and only fill those signals when auto_fill_signals is False.

coretl (approved):

PR #802, opened 2025-02-28
Ignore annotation

burkeds:
I created a simple type `Ignore`. When a signal is typehinted as `Ignore` it signals to device fillers to ignore this signal during device introspection.

```python
class SimulatedMotor(Motor):
    EncoderPosition: Ignore
```

### DIFF
--- a/src/ophyd_async/core/_device_filler.py
+++ b/src/ophyd_async/core/_device_filler.py
@@ -76,11 +76,7 @@ class DeviceFiller(Generic[SignalBackendT, DeviceConnectorT]):
         self._extras: dict[UniqueName, Sequence[Any]] = {}
         self._signal_datatype: dict[LogicalName, type | None] = {}
         self._vector_device_type: dict[LogicalName, type[Device] | None] = {}
-        self.ignored_signals: set[str] = {
-            name
-            for name, type_annotation in device.__annotations__.items()
-            if type_annotation is Ignore
-        }
+        self.ignored_signals: set[str] = set()
         # Backends and Connectors stored ready for the connection phase
         self._unfilled_backends: dict[
             LogicalName, tuple[SignalBackendT, type[Signal]]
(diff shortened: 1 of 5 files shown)


burkeds:
> Do we need this if you've disable auto filling?

I believe that the two solutions are not mutually exclusive. Auto-filling is still useful so this provides a simple way to exclude a signal from being filled. This serves two functions:

1. Prevent broken attributes/commands from being present in the device.
2. Prevents exposing attributes/commands to the user that you don't want to expose while still having the benefits of auto-fill.

src/ophyd_async/tango/core/_base_device.py
```diff
@@ -121,6 +121,13 @@ async def connect_real(self, device: Device, timeout: float, force_reconnect: bo
             .union(self.proxy.get_command_list())
         )
 
+        ignored_signals = []
+        for name, type_annotation in device.__annotations__.items():
+            if type_annotation is Ignore:
+                ignored_signals.append(name)
+
+        children = [child for child in children if child not in ignored_signals]
```
coretl, line 129: Can we put this code in `DeviceFiller` please?

src/ophyd_async/tango/core/_base_device.py
```diff
@@ -121,6 +121,13 @@ async def connect_real(self, device: Device, timeout: float, force_reconnect: bo
             .union(self.proxy.get_command_list())
         )
 
+        ignored_signals = []
```
coretl, line 124: And can this be a `set()` please...

src/ophyd_async/core/_device_filler.py
```diff
@@ -76,6 +76,11 @@ def __init__(
         self._extras: dict[UniqueName, Sequence[Any]] = {}
         self._signal_datatype: dict[LogicalName, type | None] = {}
         self._vector_device_type: dict[LogicalName, type[Device] | None] = {}
+        self.ignored_signals: set[str] = {
+            name
+            for name, type_annotation in device.__annotations__.items()
+            if type_annotation is Ignore
+        }
```
coretl, line 83: `device.__annotations__` might only contain strings if we have `from __future__ import annotations`.

Please can we put this in `self._scan_for_annotations` which already has the `get_type_hints` call?

burkeds: Done

coretl (approved):
Just realised that we should add this to the epics bit too, please can you either add it to:
https://github.com/bluesky/ophyd-async/blob/e926a6048427ccce24a34931481277278bbad133/src/ophyd_async/epics/core/_pvi_connector.py#L92

Or make an issue to say this needs adding

PR #803, opened 2025-02-28
Adding ConfigSignal version of get_current_settings and retrieve_settings.

burkeds:
Added config signal only plans to handle settings including get_current_config_settings and retrieve_config_settings. Not sure if I handled the tests correctly. I had to create a new test_data file for OneOfEverythingDevice.


coretl (commented):
Rather than duplicate the plans, would it be better to add an `only_config=False` arg to each of the existing plans?

coretl (approved):
Much neater, thanks!

PR #805, opened 2025-03-06
Don't return completed_status() in set_and_wait_for_value

### LINKED ISSUE #694: set_and_wait_for_value should not return completed_status
https://github.com/bluesky/ophyd-async/blob/55eb812de03ec7c15ecc27bb80614b08390049ab/src/ophyd_async/core/_signal.py#L671

It should return the status of the set operation, possibly awaiting it based on the wait_for_set_completion argument

This needs a test

jsouter:
Closes #694

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -579,33 +579,32 @@ async def set_and_wait_for_other_value(
 
     status = set_signal.set(set_value, timeout=set_timeout)
 
+    if callable(match_value):
+        matcher: Callable[[SignalDatatypeV], bool] = match_value  # type: ignore
+    else:
+
+        def matcher(value):
+            return value == match_value
+
+        matcher.__name__ = f"equals_{match_value}"
+
     # If the value was the same as before no need to wait for it to change
-    if (
-        callable(match_value) and not match_value(current_value)
-    ) or match_value != current_value:
+    if not matcher(current_value):
 
         async def _wait_for_value():
-            if callable(match_value):
-                async for value in values_gen:
-                    if match_value(value):
-                        break
-            else:
-                async for value in values_gen:
-                    if value == match_value:
-                        break
+            async for value in values_gen:
+                if matcher(value):
+                    break
 
         try:
             await asyncio.wait_for(_wait_for_value(), timeout)
             if wait_for_set_completion:
                 await status
         except asyncio.TimeoutError as e:
-            msg = (
-                f"{match_signal.name} value didn't match value from {match_value}()"
-                f" in {timeout}s"
-                if callable(match_value)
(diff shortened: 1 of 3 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -588,16 +588,15 @@ async def _wait_for_value():
                     break
 
         try:
-            await asyncio.wait_for(_wait_for_value(), timeout)
             if wait_for_set_completion:
+                await asyncio.wait_for(_wait_for_value(), timeout)
```
coretl, line 592: I think this was correct as it was before, we always wait for `_wait_for_value` to complete, but we only `await status` if `wait_for_set_completion`

src/ophyd_async/core/_signal.py
```diff
@@ -588,16 +588,15 @@ async def _wait_for_value():
                     break
```
coretl, line 588: Just noticed that somewhere along the line we lost the ability to pass a callable as `match_value`, please could you add it back in?

jsouter:
This works but just noticed the _ValueChecker class so should probably be using that

jsouter:
Reraising TimeoutError and asyncio.TimeoutError as asyncio.TimeoutError so that the error is raised for python pre and post 3.11

coretl:
> Reraising TimeoutError and asyncio.TimeoutError as asyncio.TimeoutError so that the error is raised for python pre and post 3.11

If you use `asyncio.TimeoutError` everywhere that seems to work for all pythons, what was raising a builtin `TimeoutError`?

jsouter:
> > Reraising TimeoutError and asyncio.TimeoutError as asyncio.TimeoutError so that the error is raised for python pre and post 3.11
> 
> If you use `asyncio.TimeoutError` everywhere that seems to work for all pythons, what was raising a builtin `TimeoutError`?

Hmm okay, thought I needed to account for asyncio.wait_for raising a base TImeoutError in 3.10 but looks like catching asyncio.TimeoutError covers that regardless. I think the breaking thing is that in the tests checking for base TimeoutError is raised fails if asyncio.TimeoutError is raised fails but only pre 3.10, but if we check for asyncio.TimeoutError everywhere it's fine.

src/ophyd_async/core/_signal.py
coretl, line 1: Could:
```python
if callable(match_value):
    matcher = match_value
else:
    def matcher(value):
        return value == match_value
    matcher.__name__ = f"equals_{match_value}"
```
Then use it in all the cases below, and make the error message always look at `matcher.__name__`

coretl (approved):
Had a possible suggestion for an improvement, but happy to merge as is

coretl (approved):

PR #806, opened 2025-03-06
Use PVI information to fill in PVs on FastCS Eiger

### LINKED ISSUE #551: Eiger FastCS: Use PVI information to fill in PVs

shihab-dls:
Fixes #551 

This PR should create FastCS modules for Eiger and Odin, and write the Eiger in a declarative style with typed annotations. Currently annotations are limited to signals internatlly used by the device, and those accessed in the MX DAQ Eiger arming chain. These signals should now follow the expected format of PVI attributes.

**Notes for reviewer:**
These changes were made for FastCS Odin as well; `capture` signal, which reads `Writing` and writes to `ConfigHDFWrite` was split into two signals, which seems messy. Moreover, attribute names seem quite ambiguous.


src/ophyd_async/fastcs/eiger/_eiger_controller.py
```diff
@@ -47,20 +43,20 @@ async def prepare(self, trigger_info: TriggerInfo):
...
+                    self._drv.count_time.set(trigger_info.livetime),
+                    self._drv.frame_time.set(trigger_info.livetime),
                 ]
             )
         await asyncio.gather(*coros)
 
     async def arm(self):
         # TODO: Detector state should be an enum see https://github.com/DiamondLightSource/eiger-fastcs/issues/43
-        self._arm_status = await set_and_wait_for_other_value(
+        self._arm_status = set_and_wait_for_other_value(
```
coretl, line 59: We still need the await here:
https://blueskyproject.io/ophyd-async/main/how-to/interact-with-signals.html#wait-for-the-value-to-match-some-expected-value
```suggestion
        self._arm_status = await set_and_wait_for_other_value(
```

src/ophyd_async/fastcs/eiger/_eiger_io.py
```diff
@@ -0,0 +1,37 @@
...
+    disarm: SignalW[int]
+    photon_energy: SignalRW[float]
+    beam_center_x: SignalRW[float]
+    beam_center_y: SignalRW[float]
+    detector_distance: SignalRW[float]
+    omega_start: SignalRW[float]
+    omega_increment: SignalRW[float]
+
+    def __init__(self, uri: str, name: str = ""):
+        super().__init__(name=name, connector=fastcs_connector(self, uri))
```
coretl, line 37: Let's discuss this in the standup later, I'm not sure what is a top level controller in FastCS here...

GDYendell:
This API doesn't seem very clear now that I am seeing it on the ophyd side... Maybe we can improve the names in fastcs-odin.

shihab-dls:
Current | API  | Rec |
| ------------- | ------------- | ------------- |
| capture (read_pv) | writing | Content Cell  |
| capture (write_pv) | config_hdf_write  | Content Cell  |
| num_captured | frames_written  | Content Cell  |
| num_to_capture | config_hdf_frames  | Content Cell  |
| image_height | data_dims_0  | Content Cell  |
| image_width | data_dims_1  | Content Cell  |
| num_frames_chunks | data_chunks_0  | Content Cell  |
| num_row_chunks | data_chunks_1  | Content Cell  |
| num_col_chunks | data_chunks_2  | Content Cell  |
| file_path | config_hdf_file_path  | Content Cell  |
| file_name | config_hdf_file_prefix  | Content Cell  |
| data_type | data_datatype  | Content Cell  |

Above is a table of current signal names (i.e., ADOdin), API signal names (i.e., Fast-CS Odin), and recommended names (yet to be filled).

DominicOram:
Can I suggest spinning the name changes off into a separate issue? Doing just a find and replace later won't be too much effort and it decouples getting the issue complete from fastcs changes

coretl:
> Can I suggest spinning the name changes off into a separate issue? Doing just a find and replace later won't be too much effort and it decouples getting the issue complete from fastcs changes

The name changes in ophyd-async? Unfortunately we can't do this, the name in PVI is the name of the attribute in the ophyd-async Device. This ticket is to use PVI to make the Eiger and Odin Devices, so I think we need to decide on names as part of this ticket...

It's possible to make a different ticket that keeps the old attribute names, but fills in the PVs from the new FastCS Eiger, but I'm not sure it is worth the effort to do that intermediate step unless you need to keep switching between the two IOCs, optimizing both...

src/ophyd_async/core/_signal.py
coretl, line 1: As discussed offline I think we don't need `trigger_and_wait_for_other_value`

src/ophyd_async/epics/core/_p4p.py
```diff
@@ -383,7 +383,7 @@ def _make_reading(self, value: Any) -> Reading[SignalDatatypeT]:
 
     async def put(self, value: SignalDatatypeT | None, wait: bool):
         if value is None:
-            write_value = self.initial_values[self.write_pv]
+            write_value = self.initial_values[self.write_pv]["value"]
```
coretl, line 386: Good catch!

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -26,7 +26,7 @@ def _get_signal_details(entry: Entry) -> tuple[type[Signal], str, str]:
         case {"rw": read_write_pv}:
             return SignalRW, read_write_pv, read_write_pv
         case {"x": execute_pv}:
-            return SignalX, execute_pv, execute_pv
+            return SignalX, execute_pv + ".PROC", execute_pv + ".PROC"
```
coretl, line 29: If we need `.PROC` here that should be pushed down to the PVI structure in FastCS rather than added here

shihab-dls: Ah, thank you for spotting this; I added this whilst debugging, but I believe it is redundant.

shihab-dls:
With the merging of FastCS [#141](https://github.com/DiamondLightSource/FastCS/pull/141), this PR should be ready for re-review.

### **Current State of Changes**

- `eiger` and `odin` moved to `fastcs` module from `epics` module, and written declaratively.
- Signals explicitly used in `ophyd_async` and `MX` are annotated.
- `EigerController.arm()` now simply returns the `SignalX`'s `trigger()` async status, where `EigerController.wait_for_idle()` awaits the status, instead of returning the status of `set_and_wait_for_other_value()`

src/ophyd_async/fastcs/eiger/_eiger.py
```diff
@@ -1,10 +1,11 @@
 from pydantic import Field
 
 from ophyd_async.core import AsyncStatus, PathProvider, StandardDetector, TriggerInfo
+from ophyd_async.fastcs.core import fastcs_connector
```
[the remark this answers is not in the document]
shihab-dls: Schooled by a bot, thanks!

src/ophyd_async/fastcs/odin/_odin_io.py
```diff
@@ -0,0 +1,102 @@
...
+    StrictEnum,
+    observe_value,
+    set_and_wait_for_value,
+)
+from ophyd_async.fastcs.core import fastcs_connector
+
+
+class Writing(StrictEnum):
+    ON = "ON"
+    OFF = "OFF"
```
coretl, line 23:
```suggestion
class OdinWriting(StrictEnum):
    ON = "ON"
    OFF = "OFF"
```

src/ophyd_async/fastcs/eiger/_eiger.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         name="",
     ):
         self.drv = EigerDriverIO(prefix + drv_suffix)
-        self.odin = Odin(prefix + hdf_suffix + "FP:")
+        self.odin = OdinHdfIO(prefix + hdf_suffix + "FP:")
```
coretl, line 29: As discussed in person, I wonder if this should be structured to point to the top level controller of the FastCS-odin. The we would have `det.odin.fp.hdf.file_path` rather than `det.odin.file_path`, in case there are other things to go in `fr` or other plugins in `fp` we need to reference later. OdinWriter can still take an `OdinHdfIO` in its constructor if all the PVs it needs are in that Device

DominicOram:
Following https://github.com/bluesky/ophyd-async/pull/862 and https://github.com/bluesky/ophyd-async/pull/863 the changes in here are effectively to move us from ADOdin to fast-cs odin. I would suggets closing this PR and making a new one for that but leave it up to @shihab-dls

PR #823, opened 2025-03-17
DOCS: change mermaid configuration

jacopoabramo:
Using `:config: { "theme" : "neutral" }` and `:align: center` options in mermaid figures to make the mermaid figures more easily visible in both dark and light mode.

I tend to use dark mode with pydata documentations and the original settings of mermaid are not usually correctly adapted for it. With this change diagrams look something like this:

- dark mode
image

- light mode
image

### DIFF
--- a/docs/explanations/devices-signals-backends.md
+++ b/docs/explanations/devices-signals-backends.md
@@ -5,6 +5,8 @@ The [](#bluesky.run_engine.RunEngine) facing interface is defined by the [bluesk
 ## Device and DeviceConnector
 
 ```{mermaid}
+:config: { "theme": "neutral" }
+:align: center
 classDiagram
 Device *-- DeviceConnector
 Device : connect(mock)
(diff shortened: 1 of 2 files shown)


coretl:
Thanks for this! Please could you put the values in `mermaid_params` in `conf.py` as per https://github.com/mgaitan/sphinxcontrib-mermaid?tab=readme-ov-file#config-values instead of in each directive instance?

jacopoabramo:
The link you mentioned only works for non-`raw` output formats as per mention of this [issue](https://github.com/mgaitan/sphinxcontrib-mermaid/issues/39).

I've been trying both:

```
mermaid_params = ["--theme", "neutral"]
```

and

```
mermaid_init_js = "mermaid.initialize({startOnLoad: true, theme: 'neutral'});"
```

which si the alternative approach suggested in the issue, but it doesn't seem to work.

coretl (approved):
In which case this is fine

PR #831, opened 2025-03-21
Allow setting read timeout

Tom-Willemsen:
Allow setting the timeout for operations on signals (not the initial connect, but for example a read operation).

We ran into this at ISIS when doing a lot of concurrent reads through a gateway, to an IOC serving PVs via pcas. This is a bit of a worst-case scenario.

This provides a hook to change the timeout on signals that could potentially take a long time to read, for example during calls like `await signal.get_value()`.


coretl (commented):
Looks good, please can you add to the docstrings of each function too?

coretl (approved):

PR #478, opened 2024-07-23
Remove hard-dependency on p4p

Tom-Willemsen:
With this PR it should be possible to use `ophyd-async[ca]` to create basic epics (ca) devices, and use plan stubs such as `ensure_connected`, _without_ a `p4p` dependency.

We noticed this at ISIS when trying to run on Python 3.12 - `p4p` isn't yet available for 3.12 due to `numpy.distutils` removal, but that led me to wondering why we were pulling in `p4p` at all...

For reference, this is one path from `import ophyd_async.plan_stubs` to depending on `p4p`:
```
    from ophyd_async.plan_stubs import ensure_connected
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\plan_stubs\__init__.py", line 2, in <module>
    from ._fly import (
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\plan_stubs\_fly.py", line 13, in <module>
    from ophyd_async.fastcs.panda import (
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\fastcs\panda\__init__.py", line 12, in <module>
    from ._hdf_panda import HDFPanda
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\fastcs\panda\_hdf_panda.py", line 9, in <module>
    from ._hdf_writer import PandaHDFWriter
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\fastcs\panda\_hdf_writer.py", line 6, in <module>
    from p4p.client.thread import Context
```
and this is the dependency path from `epics_signal_r[w]` to `p4p`:
```
    from ophyd_async.epics.signal import epics_signal_r, epics_signal_rw
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\epics\signal\__init__.py", line 2, in <module>
    from ._p4p import PvaSignalBackend
  File "C:\Instrument\dev\ophyd-async\src\ophyd_async\epics\signal\_p4p.py", line 12, in <module>
    from p4p import Value
```

(Checks are failing on codecov which I'm not sure is very relevant for this PR...)

PR #833, opened 2025-03-22
Remove hard-dependency on p4p from ophyd_async.plan_stubs

Tom-Willemsen:
Another attempt at https://github.com/bluesky/ophyd-async/pull/478

Reopened this because we lost the unconditional `p4p` dependency in https://github.com/bluesky/ophyd-async/commit/663ad9df2228465ee200e95f9514e19ede897039#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L22 but without guarding the types in `plan_stubs`, which means that this now affects all platforms and users of `ophyd-async[ca]` for example can't import `plan_stubs`. If there's a better/different way you want to do this please let me know?

To reproduce problem:

```
tom@machine:~$ python3 -m venv .temp_venv
tom@machine:~$ . .temp_venv/bin/activate
(.temp_venv) tom@machine:~$ python -m pip install ophyd_async[ca]@git+https://github.com/bluesky/ophyd-async
[snip]
(.temp_venv) tom@machine:~$ python -c "from ophyd_async import plan_stubs"
  File "<string>", line 1, in <module>
  File "/home/tom/.temp_venv/lib/python3.12/site-packages/ophyd_async/plan_stubs/__init__.py", line 4, in <module>
    from ._fly import (
  File "/home/tom/.temp_venv/lib/python3.12/site-packages/ophyd_async/plan_stubs/_fly.py", line 11, in <module>
    from ophyd_async.fastcs.panda import (
  File "/home/tom/.temp_venv/lib/python3.12/site-packages/ophyd_async/fastcs/panda/__init__.py", line 14, in <module>
    from ._hdf_panda import HDFPanda
  File "/home/tom/.temp_venv/lib/python3.12/site-packages/ophyd_async/fastcs/panda/_hdf_panda.py", line 10, in <module>
    from ._writer import PandaHDFWriter
  File "/home/tom/.temp_venv/lib/python3.12/site-packages/ophyd_async/fastcs/panda/_writer.py", line 7, in <module>
    from p4p.client.thread import Context
ModuleNotFoundError: No module named 'p4p'
(.temp_venv) tom@machine:~$
```

The second commit adds a CI smoke test that should catch this problem in future - by checking that what I think are "core" apis are importable even without any extras installed. If this is contentious, happy to drop this second commit.

### DIFF
--- a/.github/workflows/_import_with_no_extras.yml
+++ b/.github/workflows/_import_with_no_extras.yml
@@ -1,0 +1,26 @@
+on:
+  workflow_call:
+
+jobs:
+  run:
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Need this to get version number from last tag
+          fetch-depth: 0
+
+      - name: Install (with no extras)
+        uses: ./.github/actions/install_requirements
+        with:
+          pip-install: "."
+
+      - name: Check APIs are importable
+        run: |
+          python -c "import ophyd_async"
+          python -c "import ophyd_async.core"
+          python -c "import ophyd_async.epics.core"
+          python -c "import ophyd_async.epics.motor"
+          python -c "import ophyd_async.plan_stubs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,11 @@ jobs:
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  import_with_no_extras:
+    needs: check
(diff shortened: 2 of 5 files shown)


coretl:
I like the CI, but it looks like the import in `ophyd_async/fastcs/panda/_writer.py` is not actually needed. Delete this line:
https://github.com/bluesky/ophyd-async/blob/b53064e1ecc0e3e77eaec2ad12c3ea7e7ef4731c/src/ophyd_async/fastcs/panda/_writer.py#L26
and the relevant import and I think that sorts it for you...

coretl (approved):

PR #834, opened 2025-03-22
Allow no-op moves for SimMotor

Tom-Willemsen:
Allow a `SimMotor` to be "moved" to it's current position (as a no-op).

Currently this causes a crash - we ran into this in one of our tests which explicitly set the initial position of a motor to be `0`.


coretl (approved):
Thanks!

Issue #837, opened 2025-03-25
Make PvaSignalBackend put use completion if `wait=True`


coretl:
As per https://github.com/epics-base/pvxs/issues/105#issuecomment-2750956968 we should set `block=True` in the put request to make it block

## Acceptance Criteria
- Put with callback works, and there is a test to exercise it (using something like the record in the linked issue)

coretl:
Was already fixed, check added in #846

PR #846, opened 2025-03-28
Add test that checks to make sure put completion works for PVA signals

jwlodek:
Adds test to confirm #837 

I had a look, and it seems that this is already supported, so just adding a test to confirm.

A `put` will call

https://github.com/bluesky/ophyd-async/blob/f37968eba275003ac229de5578c7a92971188714/src/ophyd_async/epics/core/_p4p.py#L389

which passes along the `wait` argument, which then leads to the following from `p4p`:

https://github.com/epics-base/p4p/blob/9040d4bb56d58b674de1475dbb62444a056f6729/src/p4p/client/asyncio.py#L203

which adds the `block=True` to the request.

### DIFF
--- a/src/ophyd_async/epics/testing/test_records.db
+++ b/src/ophyd_async/epics/testing/test_records.db
@@ -112,6 +112,11 @@ record(mbbo, "$(device)enum_str_fallback") {
   field(PINI, "YES")
 }
 
+record(seq, "$(device)slowseq") {
+  field(DLY1, "5")
+  field(LNK1, "$(device)slowseq.DESC")
+}
+
 record(waveform, "$(device)uint8a") {
   field(NELM, "3")
   field(FTVL, "UCHAR")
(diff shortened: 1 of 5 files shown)


src/ophyd_async/epics/testing/test_records.db
```diff
@@ -112,6 +112,11 @@ record(mbbo, "$(device)enum_str_fallback") {
   field(PINI, "YES")
 }
 
+record(seq, "$(device)slowseq") {
+  field(DLY1, "5")
```
coretl, line 116: Please could you make this 0.5 so the tests run faster?

tests/epics/signal/test_signals.py
```diff
@@ -766,3 +766,21 @@ def a_plan():
                 assert yaml.safe_load(actual_file) == yaml.safe_load(expected_file)
 
     RE(a_plan())
+
+
+async def test_pva_put_completion(RE, ioc_devices: EpicsTestIocAndDevices):
```
coretl, line 771: Please could you make this do both CA and PVA?

coretl (commented):
Excellent, thanks for doing this! I hadn't thought to check if p4p was already doing this under the hood...

A couple of suggestions for the tests...

PR #847, opened 2025-03-31
Fix tests from broken branch name

thomashopkins32:
Fixes tests from the broken branch name:
```
Add-tests-for-"unhappy-path"-of-DerivedSignal
```
Having the `"` characters in paths makes Windows unhappy: https://github.com/bluesky/ophyd-async/actions/runs/14177724812/job/39716759198?pr=726

While we should restrict branch names from allowing `"` characters, we should also not pull every branch in our CI workflows.

### DIFF
--- a/.github/workflows/_codeql.yml
+++ b/.github/workflows/_codeql.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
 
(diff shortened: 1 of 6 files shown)


thomashopkins32:
Never mind, this is not possible since it always includes all remote references on `origin`: https://github.com/actions/checkout/blob/main/src/ref-helper.ts#L69-L77

thomashopkins32:
Looks like a known issue: https://github.com/actions/checkout/issues/1781

PR #850, opened 2025-04-04
Add tests for derived signal

Villtord:
Added tests to improve coverage for `_derived_signal_backend` and `_derived_signal`.
It looks like I shouldn't have done a rebase locally as now there are files from main branch which I didn't change but they have been merged already, sorry about it.


tests/core/test_multi_derived_signal.py
```diff
@@ -10,6 +10,13 @@
     DerivedSignalFactory,
     soft_signal_rw,
 )
+from ophyd_async.core._derived_signal_backend import (
+    DerivedSignalBackend,  # noqa: PLC2701
+    SignalTransformer,  # noqa: PLC2701
+    Transform,  # noqa: PLC2701
+)
+from ophyd_async.core._signal import SignalRW  # noqa: PLC2701
+from ophyd_async.core._table import Table  # noqa: PLC2701
```
coretl, line 19: Sorry, I should have written a more detail explanation, please can we write these tests against the public interface rather than using the private ones? If you need `# noqa: PLC2701` then we need to consider if we can write the test in a different way. The reason for this is that writing against the private interface makes refactoring later painful, ideally a refactor would not change any tests, but if you are testing a private interface then it probably has to.

In this case, you should be able to change most of these imports to `from ophyd_async.core import Transform, SignalRW, ...`, then for the remainder test by passing invalid arguments to `derived_signal_rw` and `DerivedSignalFactory` rather than to the private `SignalTransformer`

Villtord: Sorry I am confused, I though "private/public" in python is justified purely by name, here Transform class does not seem to be private, neither any of other classes. 
If then private is defined by module name - then all of them are private. So why exactly SignalTransformer is a "private" class and Transformer is not? What am I missing?

Villtord: OK ignore me, I found __initi__.py file where "public" objects are defined.

Villtord:
@coretl I think i got rid of all private imports, however I still have to get to certain methods addressing _methods in tests, I am not sure how to avoid it tbh.

tests/core/test_multi_derived_signal.py
```diff
@@ -131,3 +136,60 @@ def test_mismatching_args():
...
+    signal_r = soft_signal_rw(int, initial_value=4)
+
+    def _get(ts: int) -> float:
+        return ts
+
+    async def _put(value: float) -> None:
+        pass
+
+    derived = derived_signal_rw(_get, _put, ts=signal_r)
+    return derived._get_cache().backend
```
coretl, line 152: Continuing with the theme of testing the public interface, can we make this return the whole signal instead of the backend?
...

tests/core/test_multi_derived_signal.py
```diff
@@ -131,3 +136,60 @@ def test_mismatching_args():
...
+
+
+async def test_derived_signal_backend_connect_pass(
+    derived_signal_backend: SignalBackend,
+):
+    result = await derived_signal_backend.connect(0.1)
+    assert result is None
```
coretl, line 159: Then this would be
```suggestion
async def test_derived_signal_backend_connect_pass(
    derived_signal: SignalRW,
):
    result = await derived_signal.connect()
    assert result is None
```

tests/core/test_multi_derived_signal.py
```diff
@@ -131,3 +136,60 @@ def test_mismatching_args():
...
+) -> None:
+    with pytest.raises(RuntimeError):
+        derived_signal_backend.set_value(1.0)  # type: ignore
+
+
+async def test_derived_signal_backend_put_wait_fails(
+    derived_signal_backend: SignalBackend,
+) -> None:
+    with pytest.raises(RuntimeError):
+        await derived_signal_backend.put(value=None, wait=False)
```
coretl, line 173: And this would be
```suggestion
async def test_derived_signal_backend_put_wait_fails(
    derived_signal: SignalRW,
) -> None:
    with pytest.raises(RuntimeError):
        await derived_signal.set(value=None, wait=False)
```

tests/core/test_multi_derived_signal.py
```diff
@@ -131,3 +136,60 @@ def test_mismatching_args():
...
+):
+    result = await derived_signal_backend.connect(0.1)
+    assert result is None
+
+
+def test_derived_signal_backend_set_value(
+    derived_signal_backend: SignalBackend,
+) -> None:
+    with pytest.raises(RuntimeError):
+        derived_signal_backend.set_value(1.0)  # type: ignore
```
coretl, line 166: This is a bit trickier to test, but still possible with the public interface
```suggestion
def test_derived_signal_backend_set_value(
    derived_signal: SignalRW,
) -> None:
    await derived_signal.connect(mock=True)
    with pytest.raises(RuntimeError):
        set_mock_value(derived_signal, 1.0)
```

tests/core/test_multi_derived_signal.py
```diff
@@ -131,3 +136,60 @@ def test_mismatching_args():
...
+        set_derived=None,
+        distance=soft_signal_rw(float),
+        jack1=soft_signal_rw(float),
+        jack2=soft_signal_rw(float),
+    )
+    with pytest.raises(
+        ValueError,
+        match=re.escape("Must define a set_derived method to support derived"),
+    ):
+        factory._make_signal(signal_cls=SignalRW, datatype=Table, name="")
```
coretl, line 195: Public interface is
```suggestion
        factory.derived_signal_rw(datatype=Table, name="")
```

tests/core/test_single_derived_signal.py
coretl, line 1: I think all of these are possible without using `patch` or `MagicMock` or referring to private members of the variable, but let me know if any of them are difficult and I'll have a think how...

coretl (changes requested):

tests/core/test_single_derived_signal.py
Villtord: have to say am really struggling trying to find the way around using public interface without mock for e.g. ?

```
async def test_set_derived_not_initialized(derived_signal_backend: SignalBackend):
    with patch.object(
        derived_signal_backend.transformer,  # type: ignore
        "_set_derived",
        None,
    ):
        with pytest.raises(RuntimeError):
            await derived_signal_backend.put("name", True)
```
any ideas on that one?

coretl: Yeah, I don't think you can test that one with the public interface. Best I can do is:
```python
async def test_set_derived_not_initialized():
    def _get(ts: int) -> float:
        return ts

    sig = derived_signal_r(_get, ts=soft_signal_rw(int, initial_value=4))
    with pytest.raises(
        RuntimeError,
        match="Cannot put as no set_derived method given",
    ):
        await sig._connector.backend.put(1.0, True)
```
If you've tried using the public interface and it's not possible then I'm happy with patches or private member variables

src/ophyd_async/core/_derived_signal_backend.py
```diff
@@ -91,21 +91,23 @@ def __init__(
...
+        return validate_by_type(
             self._raw_devices | self._transform_devices, AsyncReadable
         )
 
     @cached_property
     def raw_and_transform_subscribables(self) -> dict[str, Subscribable]:
-        return filter_by_type(self._raw_devices | self._transform_devices, Subscribable)
+        return validate_by_type(
+            self._raw_devices | self._transform_devices, Subscribable
+        )  # noqa: E501
```
coretl, line 110:
```suggestion
        )
```

coretl (approved):
One comment to remove then good to merge, thanks!

PR #852, opened 2025-04-07
Support signal_r for one-to-many derived_signal_rw

coretl:


Relm-Arrowny (approved):
thanks, looks good and it is tested on dodal devices.

PR #854, opened 2025-04-07
Add datatype to signal and check that derived params have correct type

### LINKED ISSUE #836: Check types in derived signals
As of https://github.com/bluesky/ophyd-async/pull/661 there are now derived signals that I can set up like:
```python
class ReadOnlyBeamstop(Device):
    """Reads from 2 motors to work out if the beamstop is in position.

    E.g. bps.rd(beamstop.position)
    """

    def __init__(self, name=""):
        # Raw signals
        self.x = soft_signal_rw(float)
        self.y = soft_signal_rw(float)
        # Derived signals
        self.position = derived_signal_r(self._get_position, x=self.x, y=self.y)
        super().__init__(name=name)

    def _get_position(self, x: float, y: float) -> BeamstopPosition:
        if abs(x) < 1 and abs(y) < 2:
            return BeamstopPosition.IN_POSITION
        else:
            return BeamstopPosition.OUT_OF_POSITION
```
There is some typechecking on this such that if I change `get_position` to `def _get_position(self, x: float, y: float, bad_param: float) -> BeamstopPosition:` an exception is raised on signal creation. However, this only checks number and name of arguments, not the types. We should also check the type here so that if I do `def _get_position(self, x: float, y: str) -> BeamstopPosition:` an exception is raised on signal creation

## Acceptance Criteria
- Argument types are checked on derived signal creation

shihab-dls:
Fixes #836 

This PR should streamline the way we check our params in a `derived_signal`, checking that a `dict` of expected vs received are equal in length, keys, and values (where values are now the arg types).

### DIFF
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -41,8 +41,7 @@ class DerivedSignalFactory(Generic[TransformT]):
                 if k not in {"self", "return"}
             }
 
-            # received = {k: v.datatype for k, v in raw_and_transform_devices.items()}
-            received = {k: v for k, v in raw_and_transform_devices.items()}
+            received = {k: v.datatype for k, v in raw_and_transform_devices.items()}
 
             if expected != received:
                 msg = (
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -10,7 +10,7 @@ from ._derived_signal_backend import (
     TransformT,
 )
 from ._device import Device
-from ._signal import SignalR, SignalRW, SignalT, SignalW
+from ._signal import Signal, SignalR, SignalRW, SignalT, SignalW
 from ._signal_backend import SignalDatatypeT
 
 
@@ -50,7 +50,7 @@ class DerivedSignalFactory(Generic[TransformT]):
             # Populate received parameters and types
             # Use Signal datatype, or Locatable datatype, or set type as None
             received = {
-                k: v.datatype if hasattr(v, "datatype") else get_locatable_type(v)
+                k: v.datatype if isinstance(v, Signal) else get_locatable_type(v)
                 for k, v in raw_and_transform_devices.items()
             }
 
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -35,17 +35,22 @@ class DerivedSignalFactory(Generic[TransformT]):
         self._set_derived = set_derived
         # Check the raw and transform devices match the input arguments of the Transform
         if transform_cls is not Transform:
-            expected = list(transform_cls.model_fields) + [
-                x
-                for x in get_type_hints(transform_cls.raw_to_derived)
-                if x not in ["self", "return"]
-            ]
-            if set(expected) != set(raw_and_transform_devices):
+            expected = {
+                k: v
+                for k, v in get_type_hints(transform_cls.raw_to_derived).items()
+                if k not in {"self", "return"}
+            }
+
+            # received = {k: v.datatype for k, v in raw_and_transform_devices.items()}
+            received = {k: v for k, v in raw_and_transform_devices.items()}
+
+            if expected != received:
                 msg = (
-                    f"Expected devices to be passed as keyword arguments {expected}, "
-                    f"got {list(raw_and_transform_devices)}"
+                    f"Expected devices to be passed as keyword arguments "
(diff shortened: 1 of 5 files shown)


shihab-dls:
Need to fix `SimMotor`, which is failing tests, as this does not have a `datatype` property like `Signal`.

coretl:
> Need to fix `SimMotor`, which is failing tests, as this does not have a `datatype` property like `Signal`.

Hmm, datatype is only a Signal thing, why does it want datatype on a motor?

shihab-dls:
> > Need to fix `SimMotor`, which is failing tests, as this does not have a `datatype` property like `Signal`.
> 
> Hmm, datatype is only a Signal thing, why does it want datatype on a motor?

This seems to be the section:
```python
class HorizontalMirror(Device, Movable):
    def __init__(self, name=""):
        # Raw signals
        self.x1 = SimMotor()
        self.x2 = SimMotor()
```
which then results in:
```python
mappingproxy({'jack1': <ophyd_async.sim._motor.SimMotor object at 0x7f258b43fe10>, 'jack2': <ophyd_async.sim._motor.SimMotor object at 0x7f258b188750>, 'distance': <ophyd_async.core._signal.SignalRW object at 0x7f258b44a810>})
````
from `raw_and_transform_devices.items()`, which then fails the operation `{k: v.datatype for k, v in raw_and_transform_devices.items()}`.

**Comment by shihab-dls** [comment]
> Maybe only check the datatype for `Signal` subclasses then?

I think that's reasonable, but I feel as though, given the expected and received args aren't `Signal`, we should still check if they are the same `type` (e.g., jack1 received and expected are both `SimMotor`). If we **do** want to do this, I'm not yet sure if this should allow subclasses (I assume **no**), and my next concern would be the failing tests, as it seems `jack1` and `jack2` are returned as `floats` from `get_type_hints()`. I'd appreciate your input on the matter; I'll make changes tomorrow morning before PMAC.

**Comment by coretl** [comment]
> > Maybe only check the datatype for `Signal` subclasses then?
> 
> I think that's reasonable, but I feel as though, given the expected and received args aren't `Signal`, we should still check if they are the same `type` (e.g., jack1 received and expected are both `SimMotor`). If we **do** want to do this, I'm not yet sure if this should allow subclasses (I assume **no**), and my next concern would be the failing tests, as it seems `jack1` and `jack2` are returned as `floats` from `get_type_hints`.

One possibility is to make `SimMotor` derive from `Locatable[float]`:
https://github.com/bluesky/bluesky/blob/2151c66341d36005def29996848a4d84a69a4f1a/src/bluesky/protocols.py#L365-L376

Then check that eveything is either a `Signal[T]` or has `Locatable[T]` somewhere in its class heirarchy.

File: src/ophyd_async/core/_derived_signal.py [hunk 1]
```diff
@@ -35,15 +37,27 @@ def __init__(
...
+                    for k, v in get_type_hints(transform_cls.raw_to_derived).items()
+                    if k not in {"self", "return"}
+                },
+            }
+
+            # Populate received parameters and types
+            # Use Signal datatype, or Locatable datatype, or set type as None
+            received = {
+                k: v.datatype if hasattr(v, "datatype") else get_locatable_type(v)
+                for k, v in raw_and_transform_devices.items()
```
shihab-dls, line 54: I've made the assumption here that returning `None` will cause `if expected != received:` to be falsy, and thus raising the exception; I think this is a fair assumption, but probably good to point out. Moreover, I'm delaying the exception until after `received` is populated, instead of using an iterator for the first occurrence (which I felt was less readable).

src/ophyd_async/core/_derived_signal.py
```diff
@@ -35,15 +37,27 @@ def __init__(
...
+                    k: v
+                    for k, v in get_type_hints(transform_cls.raw_to_derived).items()
+                    if k not in {"self", "return"}
+                },
+            }
+
+            # Populate received parameters and types
+            # Use Signal datatype, or Locatable datatype, or set type as None
+            received = {
+                k: v.datatype if hasattr(v, "datatype") else get_locatable_type(v)
```
coretl, line 53: Slightly prefer explicit checking on `Signal` type:
```suggestion
                k: v.datatype if isinstance(v, Signal) else get_locatable_type(v)
```

coretl (approved):
One minor comment, then good to merge

src/ophyd_async/core/_derived_signal.py
shihab-dls: Oh yeah, I do prefer this as well! Amended.

PR #856, opened 2025-04-09
Improve error message from observe value

Villtord:
Improve error message from signals queue timeout in observe_signals_value()
Add test for signals queue timeout in observe_signals_value()

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -481,7 +481,7 @@ async def observe_signals_value(
             iteration_timeout = _get_iteration_timeout(timeout, overall_deadline)
             try:
                 item = await asyncio.wait_for(q.get(), iteration_timeout)
-            except TimeoutError as exc:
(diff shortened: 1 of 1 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -476,14 +479,22 @@ def queue_value(value: SignalDatatypeT, signal=signal):
                     f"timeout {done_timeout}s"
                 )
             iteration_timeout = _get_iteration_timeout(timeout, overall_deadline)
-            item = await asyncio.wait_for(q.get(), iteration_timeout)
+            try:
+                item = await asyncio.wait_for(q.get(), iteration_timeout)
+            except TimeoutError as exc:
```
coretl, line 484:
```suggestion
            except asyncio.TimeoutError as exc:
```

Villtord: well spotted, thanks

coretl (approved):

PR #858, opened 2025-04-16
Simplify HDFDocumentComposer

### LINKED ISSUE #795: Simplify HDFDocumentComposer
At the moment we have code like:
https://github.com/bluesky/ophyd-async/blob/f9f985a59eb2a9a967e87b67f24cc343e9a5d8e0/src/ophyd_async/epics/adcore/_hdf_writer.py#L131-L151

This is duplicate within PandA HDF writer and sim writer. It would be better to make `HDFDocumentComposer` have:
```python
async def collect_stream_docs(self, indices_written: int) -> AsyncIterator[StreamAsset]:
```
rather than `stream_resources` and `stream_data`.

This would mean that we could make the composer in `open()` rather than having to make it in `collect_stream_docs`.

## Acceptance Criteria
- The changes are made, along with usage in the 3 DetectorWriters, and tests not changed

shihab-dls:
Fixes #795 

This PR should mostly remove duplicate code in writers that perform the same `collect_stream_docs` operation. As #795 requested tests to remain unchanged, writers still have their `collect_stream_docs` methods, but this simply calls the composers method. 

With this, composers can now be created in `open()`; however, as panda's writer creates its dataset in `_update_datasets()`, I create the composer here instead. 

EDIT: I've moved this back to `collect_stream_docs` (for panda writer), as I think it's clearer here.


shihab-dls:
Given I did not change any of the tests, we still pass in `name: str` to writers `collect_stream_docs()`. I don't see how this is ever accessed, even prior to the changes of this PR? If I'm not missing something, I'll remove this and amend the tests that call `collect_stream_docs()`.

src/ophyd_async/core/_hdf_dataset.py
```diff
@@ -79,17 +78,19 @@ def __init__(
             for ds in datasets
         ]
 
-    def stream_resources(self) -> Iterator[StreamResource]:
-        for bundle in self._bundles:
-            yield bundle.stream_resource_doc
+    async def collect_stream_docs(
+        self, indices_written: int
+    ) -> AsyncIterator[StreamAsset]:
```
coretl, line 83: Ok, having seen the code it does no I/O, so we can put it back to a plain iterator

```suggestion
    def collect_stream_docs(
        self, indices_written: int
    ) -> Iterator[StreamAsset]:
```

src/ophyd_async/epics/adcore/_hdf_writer.py
```diff
@@ -134,20 +138,7 @@ async def collect_stream_docs(
...
-                # you get one stream resource and many stream datums per scan
-
-                for doc in self._composer.stream_resources():
-                    yield "stream_resource", doc
-            for doc in self._composer.stream_data(indices_written):
-                yield "stream_datum", doc
+        if self._composer:
+            await self.fileio.flush_now.set(True)
+            async for doc in self._composer.collect_stream_docs(indices_written):
+                yield doc
```
coretl, line 144: Which simplifies to
```suggestion
            yield from self._composer.collect_stream_docs(indices_written)
```

src/ophyd_async/fastcs/panda/_writer.py
```diff
@@ -145,17 +145,14 @@ async def collect_stream_docs(
...
-                for doc in self._composer.stream_resources():
-                    yield "stream_resource", doc
-            for doc in self._composer.stream_data(indices_written):
-                yield "stream_datum", doc
+        if not self._composer:
+            self._composer = HDFDocumentComposer(
+                Path(await self.panda_data_block.hdf_directory.get_value())
+                / Path(await self.panda_data_block.hdf_file_name.get_value()),
+                self._datasets,
+            )
```
coretl, line 153: What was the reasoning to put it here rather than in `open()`?

src/ophyd_async/sim/_blob_detector_writer.py
```diff
@@ -85,17 +85,9 @@ async def collect_stream_docs(
...
-            # the file
-            if not self.composer:
-                if not self.path:
-                    raise RuntimeError(f"open() not called on {self}")
-                self.composer = HDFDocumentComposer(self.path, self.datasets)
-                for doc in self.composer.stream_resources():
-                    yield "stream_resource", doc
-            for doc in self.composer.stream_data(indices_written):
-                yield "stream_datum", doc
+        if self.composer:
```
coretl, line 88: I think we should raise if `self.composer is None`
```suggestion
        if self.composer is None:
            msg = "open() not called on {self}"
            raise RuntimeError(msg)
```

coretl (changes requested):
Thanks, looks almost there but I spotted a few simplifications

PR #859, opened 2025-04-17
Allow constants in kwargs of derived_signal

### LINKED ISSUE #857: Allow constants in kwargs of derived_signal
There are cases where I want to make two derived signals that call the same deriving function but with different fixed signals. e.g.

```python
from ophyd_async.core import StandardReadable, derived_signal_r
from ophyd_async.epics.core import epics_signal_r


    def __init__(self, name: str = "") -> None:
        self._underlying_signal = epics_signal_r(float, "TEST")
        self.signal_add_5 = derived_signal_r(
            self._add_const_to_value, value=self._underlying_signal, const=5
        )
        self.signal_add_24 = derived_signal_r(
            self._add_const_to_value, value=self._underlying_signal, const=24
        )

        super().__init__(name=name)

    def _add_const_to_value(self, value, const):
        return value + const
```

However, this causes a typing error as `kwargs` passed to `derived_signal_r` cannot be literals. We can get round this using soft signals:

```python
from ophyd_async.core import StandardReadable, derived_signal_r, soft_signal_rw
from ophyd_async.epics.core import epics_signal_r


    def __init__(self, name: str = "") -> None:
        self._underlying_signal = epics_signal_r(float, "TEST")
        self.signal_add_5 = derived_signal_r(
            self._add_const_to_value,
            value=self._underlying_signal,
            const=soft_signal_rw(datatype=int, initial_value=5),
        )
        self.signal_add_24 = derived_signal_r(
            self._add_const_to_value,
            value=self._underlying_signal,
            const=soft_signal_rw(datatype=int, initial_value=24),
        )

        super().__init__(name=name)

    def _add_const_to_value(self, value, const):
        return value + const
```

But this is unintuitive. Instead we should allow literals to be passed in.

## Acceptance Criteria
- `derived_signal` can take literals as `kwargs` to pass to deriving functions

shihab-dls:
Fixes #857

This PR should allow literals of type `int | bool | str | float` to be passed in as raw_devices. The approach that I've taken is filtering out these constants in `SignalTransformer`, then passing them back in when a derived reading is made. There are probably more streamlined ways of doing this, so this is an initial pass to check that this is the intended functionality.


src/ophyd_async/core/_derived_signal.py
```diff
@@ -163,7 +163,7 @@ def _get_first_arg_datatype(
 def _make_factory(
     raw_to_derived: Callable[..., SignalDatatypeT] | None = None,
     set_derived: Callable[[SignalDatatypeT], Awaitable[None]] | None = None,
-    raw_devices: dict[str, Device] | None = None,
+    raw_devices: dict[str, Device | int | bool | float | str] | None = None,
```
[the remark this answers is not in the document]
shihab-dls: I agree with the messy typing concern; I've pushed a commit to rename to `raw_devices_and_constants`, but we should query @coretl's input on this. I'll then update the docs accordingly.

src/ophyd_async/core/_derived_signal.py
```diff
@@ -200,7 +207,7 @@ def derived_signal_r(
     raw_to_derived: Callable[..., SignalDatatypeT],
     derived_units: str | None = None,
     derived_precision: int | None = None,
-    **raw_devices: Device,
+    **raw_devices_and_constants: Device | int | bool | float | str,
```
coretl, line 210: Could: there is a union of these primitive types already in `._signal_backend`

```suggestion
    **raw_devices_and_constants: Device | Primitive,
```

src/ophyd_async/core/_derived_signal.py
```diff
@@ -50,8 +50,12 @@ def __init__(
...
             # Use Signal datatype, or Locatable datatype, or set type as None
             received = {
-                k: v.datatype if isinstance(v, Signal) else get_locatable_type(v)
-                for k, v in raw_and_transform_devices.items()
+                k: v.datatype
+                if isinstance(v, Signal)
+                else type(v)
+                if isinstance(v, int | bool | float | str)
+                else get_locatable_type(v)
+                for k, v in raw_and_transform_devices_and_constants.items()
```
coretl, line 58: Could: If we are having to do the type checking here, we could also partition into `raw_and_transform_devices` and `constants` and pass these down to the `DerivedSignalBackend`...

src/ophyd_async/core/_derived_signal_backend.py
```diff
@@ -77,15 +78,29 @@ def __init__(
...
+                k: v
+                for k, v in raw_and_transform_devices_and_constants.items()
+                if isinstance(v, Device)
+            },
+            {
+                k: v
+                for k, v in raw_and_transform_devices_and_constants.items()
+                if isinstance(v, int | bool | str | float)
+            },
+        )
```
coretl, line 102: ... then we wouldn't need to do the split here

coretl (commented):
Neat! A couple of suggestions to tidy...

shihab-dls:
If we have a multi derived signals using transforms, we still need to pass a signal as a parameter, as this will be a `transform_device`. Splitting into `raw_constants`, `raw_devices`, `transform_constants`, and `transform_devices` lets us unpack `raw_constants` when reading raw_values, and unpack `transform_constants` when populating the transform with args. This should thus be general for passing constants in single and multi derived signals. So the order of splitting this up is now:

- Take `raw_and_transform_devices_and_constants` in `DerivedSignalFactory`
- Split into `_raw_and_transform_devices` and `_raw_and_transform_constants` for checking against `transform_cls`, as this lets us `type(v)` for constants and `v.datatype` or `get_locatable_type(v)` for devices.
- Split into `raw_constants`, `raw_devices`, `transform_constants`, and `transform_devices` before passing to `SignalTransformer`

Then when we collect our derived readings we can:
```python
raw_values = {
    **{
        k: raw_and_transform_readings[sig.name]["value"]
        for k, sig in self._raw_devices.items()
    },
    **self._raw_constants,
}
```
and 
```python
return self._transform_cls(**(transform_args | self._transform_constants))
```
in `_make_transform_from_readings`. So, given a single derived signal, a constant would be populated as a `raw_constant`, and no `transform_constant` is passed as no specific transform is given. And, given a multi derived signal, a constant would be populated as a `transform_constant` and no `raw_constant` is expected (i.e., parameters should be passed as type hints to the transform, and thus a `transform_constant`).

coretl (approved):

PR #862, opened 2025-04-23
Use fastCS-Eiger Without fastCS-Odin Changes

shihab-dls:
Following testing done with @DominicOram on i03, I will be opening up two PRs (based on #806) in order to have fastcs-eiger using adodin. This PR solely introduces fastcs-eiger.

Subsequent PR is #863.


src/ophyd_async/epics/core/_p4p.py
```diff
@@ -383,7 +383,7 @@ def _make_reading(self, value: Any) -> Reading[SignalDatatypeT]:
 
     async def put(self, value: SignalDatatypeT | None, wait: bool):
         if value is None:
-            write_value = self.initial_values[self.write_pv]
+            write_value = self.initial_values[self.write_pv]["value"]
```
DominicOram, line 386: Should: Can we get a unit test on this?

src/ophyd_async/fastcs/eiger/_eiger_io.py
```diff
@@ -0,0 +1,53 @@
...
+    beam_center_x: SignalRW[float]
+    beam_center_y: SignalRW[float]
+    detector_distance: SignalRW[float]
+    omega_start: SignalRW[float]
+    omega_increment: SignalRW[float]
+    arm: SignalX
+    disarm: SignalX
+
+
+class EigerDriverIO(Device):
```
DominicOram, line 44: Nit: The [docs](https://blueskyproject.io/ophyd-async/main/explanations/decisions/0009-procedural-vs-declarative-devices.html#decision) talk about using `EpicsDevice` for declarative devices. It looks like this has an option for `pvi` (https://github.com/bluesky/ophyd-async/blob/48236a919cec15700bbeeda2324ab2efff1c4558/src/ophyd_async/epics/core/_epics_device.py#L8). I guess using `fastcs_connector` is functionally equivalent so doesn't really matter but having two ways to do the same thing is odd.

DominicOram (changes requested):
Thanks, some minor points about adding more tests

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -43,10 +43,10 @@
 
 def _get_signal_details(entry: Entry) -> tuple[type[Signal], str, str]:
     match entry:
-        case {"r": read_pv}:
-            return SignalR, read_pv, read_pv
         case {"r": read_pv, "w": write_pv}:
             return SignalRW, read_pv, write_pv
+        case {"r": read_pv}:
+            return SignalR, read_pv, read_pv
```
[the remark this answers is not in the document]
DominicOram: In which case can we have a test that passes in `{"r": read_pv, "w": write_pv}` here and confirms we get a `SignalRW`? This would presumably fail on main and pass on this branch

src/ophyd_async/fastcs/eiger/_eiger_io.py
shihab-dls: Oh yeah, good point; I guess the only difference here is we call `create_children_from_annotations` on the connector, but we should probably streamline this into one utility.

DominicOram (approved):
Great, thank you!

src/ophyd_async/fastcs/eiger/_eiger_io.py
coretl: Remember that `FastCS` supports both EPICS and Tango, so `Eiger` could be potentially used for both. I was going to make a `FastCsDevice` but it messed with the superclass init order in Panda so went with `fastcs_connector`. This will be documented in the FastCS version of that tutorial when it's written...

Issue #742, opened 2025-01-21
Introduce `SupersetEnum` to go along with `SubsetEnum` and `StrictEnum`


jwlodek:
Add a `SupersetEnum` class that specifies that a corresponding enum signal will have permitted values that are all values of the superset, but not necessarily all of them.

One consideration should also be how to handle empty values, since for example ADAndor uses empty strings "" for enum values for data types it does not support rather than removing them.

## Acceptance Criteria
- `SupersetEnum` exits, and can be used in cases where a Enum record is expected to have a subset of its values. See #724 for an example.
- Make `StrictEnum` and `SubsetEnum` both extend from `Enum`. Currently one is a subclass of the other.

muddi900:
If I am getting correctly, you want an Enum subclass, that only accepts a set values for each item. It can accept any of the type, but not some of the types.

If that is correct, I think I can implement it.

coretl:
We want 3 Enum subclasses:
```python
class UppercaseNameEnumMeta(EnumMeta):
    def __new__(metacls, *args, **kwargs):
        ret = super().__new__(metacls, *args, **kwargs)
        lowercase_names = [x.name for x in ret if not x.name.isupper()]  # type: ignore
        if lowercase_names:
            raise TypeError(f"Names {lowercase_names} should be uppercase")
        return ret

class StrictEnum(str, Enum, metaclass=UppercaseEnumMeta):
    """All members should exist in the Backend, and there will be no extras"""

class AnyStringUppercaseNameEnumMeta(UppercaseNameEnumMeta):
    def __call__(self, value, *args, **kwargs):  # type: ignore
        if isinstance(value, str) and not isinstance(value, self):
            return value
        return super().__call__(value, *args, **kwargs)

class SubsetEnum(str, Enum, metaclass=AnyStringUppercaseNameEnumMeta):
    """All members should exist in the Backend, but there may be extras"""

class SupersetEnum(str, Enum, metaclass=UppercaseEnumMeta):
    """Some members should exist in the Backend, and there should be no extras"""
```

Anywhere that references `SubsetEnum` in the rest of the codebase should be updated to handle `StrictEnum | SubsetEnum | SupersetEnum` with the relevant checking as per the docstrings above

PR #864, opened 2025-04-23
Add SupersetEnum and use it in ADBaseDataType

oliwenmandiamond:
Add SupersetEnum, rework Enums to not inherit from another and adjust framework to updated pattern. 

Also updated ADBaseDataType to use SupersetEnum to prevent the following error for some detectors:
```
data_type: TypeError: TEST-SPECS-02:DataType_RBV has choices ('Int8', 'UInt8', 'Int16', 'UInt16', 'Int32', 'UInt32', 'Float32', 'Float64'), but <enum 'ADBaseDataType'> requested ['Int8', 'UInt8', 'Int16', 'UInt16', 'Int32', 'UInt32', 'Int64', 'UInt64', 'Float32', 'Float64'] to be strictly equal to them.
```

Fixes https://github.com/bluesky/ophyd-async/issues/742


jwlodek (approved):
This LGTM, I would maybe just add a test or two that creates a superset enum signal and checks the behavior. Can probably even use `ADBaseDataType`

src/ophyd_async/core/_signal_backend.py
```diff
@@ -38,17 +38,17 @@
...
     | Sequence[str]
-    | Sequence[StrictEnum]
+    | Sequence[EnumTypes]
     | Table
 )
 """The supported [](#Signal) datatypes:
 
 - A python primitive [](#bool), [](#int), [](#float), [](#str)
-- A [](#StrictEnum) or [](#SubsetEnum) subclass
+- A [](#EnumTypes) or subclass
```
DiamondJoseph, line 47:
```suggestion
- An [](#EnumTypes) subclass
```

nit: grammar

src/ophyd_async/core/_signal_backend.py
```diff
@@ -38,17 +38,17 @@
...
 - A fixed datatype [](#Array1D) of numpy bool, signed and unsigned integers or float
 - A [](#numpy.ndarray) which can change dimensions and datatype at runtime
 - A sequence of [](#str)
-- A sequence of [](#StrictEnum) or [](#SubsetEnum) subclass
+- A sequence of [](#EnumTypes) or subclass
```
DiamondJoseph, line 51:
```suggestion
- A sequence of [](#EnumTypes) subclasses
```

src/ophyd_async/core/_signal_backend.py
```diff
@@ -167,7 +167,7 @@ def _datakey_dtype_numpy(
 
 
 def _datakey_shape(value: SignalDatatype) -> list[int]:
-    if type(value) in _primitive_dtype or isinstance(value, StrictEnum):
+    if type(value) in _primitive_dtype or isinstance(value, EnumTypes):
```
DiamondJoseph, line 170: Does this work? isinstance expects a class or tuple of classes, here you're passing a Union of classes.

Same for the places above.

src/ophyd_async/epics/core/_util.py
```diff
@@ -34,9 +36,12 @@ def get_supported_values(
     if issubclass(enum_cls, SubsetEnum):
         if not set(choices).issubset(pv_choices):
             raise TypeError(error_msg + "to be a subset of them.")
-    else:
+    elif issubclass(enum_cls, StrictEnum):
         if set(choices) != set(pv_choices):
             raise TypeError(error_msg + "to be strictly equal to them.")
+    elif issubclass(enum_cls, SupersetEnum):
+        if len(set(choices).intersection(pv_choices)) == 0:
+            raise TypeError(error_msg + ". None of them match.")
```
DiamondJoseph, line 44: This is checking that at least one of the returned values is in the superset enum type, is that the expected behaviour, or should all of the returned choices be in choices, i.e. the opposite of SubsetEnum        `set(pv_choices).issubset(choices):`? I'll read the ticket

DiamondJoseph:
Does this close #742? Should be added to the body of the PR so github picks it up

src/ophyd_async/core/_signal_backend.py
DiamondJoseph:
```python
>>> class Thing():
...     pass
... 
>>> class Thing2():
...     pass
... 
>>> all_things = Thing | Thing2
>>> instance = Thing()
>>> isinstance(instance, all_things)
True

```

Looks like it does work for python 3.12 at least

jennmald (approved):
Looks good so far!

DiamondJoseph (approved):

src/ophyd_async/core/_signal_backend.py
```diff
@@ -38,17 +38,17 @@
     | Array1D[np.float64]
     | np.ndarray
     | Sequence[str]
-    | Sequence[StrictEnum]
+    | Sequence[EnumTypes]
```
coretl, line 41: `Sequence[StrictEnum | SubsetEnum | SupersetEnum]` would be "a sequence of different enum subclasses". I think this should be:
```suggestion
    | Sequence[StrictEnum]
    | Sequence[SubsetEnum]
    | Sequence[SupersetEnum]
```

src/ophyd_async/epics/adcore/_utils.py
```diff
@@ -7,11 +7,12 @@
     SignalRW,
     StrictEnum,
     SubsetEnum,
+    SupersetEnum,
     wait_for_value,
 )
 
 
-class ADBaseDataType(StrictEnum):
+class ADBaseDataType(SupersetEnum):
```
coretl, line 15: Should probably allow the empty case here as we tend to blank entries the driver doesn't support...
```suggestion
class ADBaseDataType(SupersetEnum):
    _ = ""  # Driver database override will blank the enum string if it doesn't support a datatype
```
We then need to handle this in the functions below

src/ophyd_async/epics/core/_util.py
coretl: I'm not sure if I was clear in the ticket, but I meant what you said @DiamondJoseph

coretl (commented):
Thanks, this looks almost there, a few comments and an added test and this should be done

src/ophyd_async/epics/adcore/_utils.py
```diff
@@ -30,7 +30,7 @@ class ADBaseDataType(SupersetEnum):
 
 def convert_ad_dtype_to_np(ad_dtype: ADBaseDataType) -> str:
     ad_dtype_to_np_dtype = {
-        ADBaseDataType.UNDEFINED:  "U1",
+        ADBaseDataType.UNDEFINED: "U1",
```
coretl, line 33: I think we should error if we get `UNDEFINED`

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -50,10 +49,10 @@ def write_value(self, value: Any) -> Sequence[str]:
 
 
 @dataclass
-class SequenceEnumSoftConverter(SoftConverter[Sequence[EnumT]]):
-    datatype: type[EnumT]
+class SequenceEnumSoftConverter(SoftConverter[Sequence[EnumTypes]]):
+    datatype: type[EnumTypes]
 
-    def write_value(self, value: Any) -> Sequence[EnumT]:
+    def write_value(self, value: Any) -> Sequence[EnumTypes]:
```
coretl, line 55: This was correct as it was, `Sequence[EnumT]` means "a sequence of instances of an enum type inferred from the datatype field, as long as the datatype is a subclass of EnumTypes", while `Sequence[EnumTypes]` means "sequence of anything that is an EnumType subclass" which is less strict

https://pyright-play.net/?code=GYJw9gtgBApgdgV2gSwgBzCALlAooiAKFEiiwE81k4BzKVDbKAFUpgDUBDEAGigGUYARwTwAxjEKExAG04BneXgIBBABT4kASgBchKAahoF8qbJPKkAIQ0Fd%2Bw8cVTNEVmhhKAvJYgqoAD6%2BVoSuzFA%2B7hzcagBEYbF8AEZgCHAAJhG%2BUfJaUukwwFAQ5AD6wGliapw6AsKicBIA2mEAulpQALQAfNl6hlAgMFgIIHBQnE0ADK35hcVlFY0ATNW1giLiMC0EOe1dvWFs8v2GQyNjE9OzZnKKLAAW1DS21vYD-j6x8lggz7FSAAeWRK5UqaiazCetAAdCp2oQhgA3GCcGSlCgeNSAvKEcggxaVVaQ6E0OEI5Go9GYmBqch5IA

src/ophyd_async/core/_utils.py
```diff
@@ -20,18 +28,34 @@
...
 
 
+class AnyStringUppercaseNameEnumMeta(UppercaseNameEnumMeta):
+    def __call__(self, value, *args, **kwargs):  # type: ignore
+        if isinstance(value, str) and not isinstance(value, self):
+            return value
+        return super().__call__(value, *args, **kwargs)
+
+
 class SubsetEnumMeta(StrictEnumMeta):
```
coretl, line 59: And also delete `SubsetEnumMeta`, but move its docstring to `AnyStringUppercaseNameEnumMeta`

tests/epics/signal/test_common.py
```diff
@@ -45,3 +45,41 @@ class MyEnum(StrictEnum):
...
     assert "test_1" in supported_vals
     assert "test_2" in supported_vals
+
+
+def test_given_supersetenum_that_partial_matches_are_valid():
+    class MyEnum(SupersetEnum):
+        TEST_1 = "test_1"
+        TEST_2 = "test_2"
+
+    supported_vals = get_supported_values("", MyEnum, (MyEnum.TEST_1,))
```
coretl, line 55: We should be testing the string values lookup here
```suggestion
    supported_vals = get_supported_values("", MyEnum, ("test_1",))
```

coretl (dismissed):

src/ophyd_async/epics/core/_util.py
```diff
@@ -31,18 +33,30 @@ def get_supported_values(
...
     # But override those that we specify via the datatype
     for v in enum_cls:
         supported_values[v.value] = v
+
+    if issubclass(enum_cls, SupersetEnum):
+        # Convert to sets and use symmetric difference
+        missing_choices = list(set(pv_choices) ^ set(choices))
+        # Remove the missing choices from pv from supported values
+        for m in missing_choices:
+            del supported_values[m]
```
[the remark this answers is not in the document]
oliwenmandiamond: This change doesn't work. 
It only works with 

`supported_values = {x: str(enum_cls(x)) for x in pv_choices}`

But then causes the followings tests to fail inside of epics.siganl.test_singals.py:

test_backend_put_enum_string

test_epics_get_put_monitor_for_override_types

test_retrieve_apply_store_settings

PR #865, opened 2025-04-24
Tango improved descriptor

burkeds:
This improves the descriptor returned by Tango devices. If specified in the device server, the DataKey will now include limits including the Tango RDS range, choices, dims, precision, units and object_name.

An important note here is that we are still treating Tango commands with input/output types as SignalRW instead of SignalX. Tango commands have limited configuration information so their DataKeys can only carry the keys object_name, source, dtype, and shape. Applications using the document stream must be aware of this and not try to, for example, read the units of a SignalRW backed by a Tango command.

### DIFF
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -623,11 +623,10 @@ def get_trl_descriptor(
                 }
                 for key in ["control", "display", "warning", "alarm"]
             }
-            _limits["hysteresis"] = try_to_cast_as_float(alarm_info.delta_val)  # type: ignore
-            _limits["rds"] = None  # type: ignore
-            # If the hysteresis key is None, delete it
-            if _limits["hysteresis"] is None:
-                del _limits["hysteresis"]
+            _limits["rds"] = {
+                "time_difference": try_to_cast_as_float(alarm_info.delta_t),
+                "value_difference": try_to_cast_as_float(alarm_info.delta_val),
+            }
 
             _choices = list(config.enum_labels) if config.enum_labels else []
             _dims = []
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -47,7 +47,7 @@ from ._converters import (
     TangoEnumArrayConverter,
     TangoEnumConverter,
 )
-from ._utils import DevStateEnum, get_device_trl_and_attr
+from ._utils import DevStateEnum, get_device_trl_and_attr, try_to_cast_as_float
 
 logger = logging.getLogger("ophyd_async")
 
@@ -81,7 +81,7 @@ def get_python_type(tango_type: CmdArgType) -> tuple[bool, object, str]:
     if is_float(tango_type, True):
         return array, float, "number"
     if is_bool(tango_type, True):
-        return array, bool, "integer"
+        return array, bool, "boolean"
     if is_str(tango_type, True):
         return array, str, "string"
     if is_binary(tango_type, True):
@@ -518,19 +518,6 @@ def get_dtype_extended(datatype) -> object | None:
     return dtype
 
 
-def try_to_cast_as_float(value: Any) -> float | None:
-    """Attempt to cast a value to float, returning None on failure."""
-    try:
-        return float(value)
-    except (ValueError, TypeError):
-        return None
-
-
-def _update_descriptor(descriptor, **kwargs):
-    """Update the descriptor with non-empty key-value pairs."""
-    descriptor.update({key: value for key, value in kwargs.items() if value})
-
-
 def get_simple_datakey(
     datatype: type | None,
     tango_resource: str,
@@ -620,7 +607,9 @@ def get_trl_descriptor(
     tango_resource: str,
     tr_configs: dict[str, AttributeInfoEx | CommandInfo],
 ) -> DataKey:
-
+    def _update_descriptor(desc, **kwargs):
+        """Update the descriptor with non-empty key-value pairs."""
+        desc.update({key: value for key, value in kwargs.items() if value})
     descriptor = get_simple_datakey(datatype, tango_resource, tr_configs)
 
     for _, config in tr_configs.items():
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -2,7 +2,6 @@ import asyncio
 import functools
 import logging
 import time
-import copy
 from abc import abstractmethod
 from collections.abc import Callable, Coroutine
 from enum import Enum
@@ -518,6 +517,7 @@ def get_dtype_extended(datatype) -> object | None:
             dtype = CmdArgType.DevState
     return dtype
 
+
 def try_to_cast_as_float(value: Any) -> float | None:
     """Attempt to cast a value to float, returning None on failure."""
     try:
@@ -525,9 +525,11 @@ def try_to_cast_as_float(value: Any) -> float | None:
     except (ValueError, TypeError):
         return None
 
+
 def _update_descriptor(descriptor, **kwargs):
     """Update the descriptor with non-empty key-value pairs."""
     descriptor.update({key: value for key, value in kwargs.items() if value})
+
 
 def get_simple_datakey(
     datatype: type | None,
@@ -612,6 +614,7 @@ def get_simple_datakey(
 
     raise RuntimeError(f"Error getting descriptor for {tango_resource}")
 
+
 def get_trl_descriptor(
     datatype: type | None,
     tango_resource: str,
@@ -640,7 +643,7 @@ def get_trl_descriptor(
             elif config.data_format == AttrDataFormat.IMAGE:
                 _dims = ["y", "x"]
 
-            arr, tr_dtype, json_type = get_python_type(config.data_type)
+            _, tr_dtype, _ = get_python_type(config.data_type)
 
             if tr_dtype == CmdArgType.DevState:
                 _choices = list(DevState.names.keys())
@@ -670,6 +673,7 @@ def get_trl_descriptor(
                 object_name=config.cmd_name,
             )
     return DataKey(**descriptor)
+
 
 async def get_tango_trl(
     full_trl: str, device_proxy: DeviceProxy | TangoProxy | None, timeout: float
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -9,7 +9,7 @@ from typing import Any, ParamSpec, TypeVar, cast
 
 import numpy as np
 from bluesky.protocols import Reading
-from event_model import DataKey
+from event_model import DataKey, LimitsRange
 
 from ophyd_async.core import (
     AsyncStatus,
@@ -522,7 +522,7 @@ def get_simple_datakey(
     datatype: type | None,
     tango_resource: str,
     tr_configs: dict[str, AttributeInfoEx | CommandInfo],
-) -> DataKey:
+) -> dict:
     """Create a descriptor from a tango resource locator."""
     tr_dtype = {}
     for tr_name, config in tr_configs.items():
@@ -579,9 +579,9 @@ def get_simple_datakey(
                 raise TypeError(f"{tango_resource} has type [{tr_dtype}] not [{dtype}]")
 
         if tr_format == AttrDataFormat.SPECTRUM:
-            return DataKey(source=tango_resource, dtype="array", shape=[max_x])
+            return {"source": tango_resource, "dtype": "array", "shape": [max_x]}
         elif tr_format == AttrDataFormat.IMAGE:
-            return DataKey(source=tango_resource, dtype="array", shape=[max_y, max_x])
+            return {"source": tango_resource, "dtype": "array", "shape": [max_y, max_x]}
 
     else:
         if tr_dtype in (Enum, CmdArgType.DevState):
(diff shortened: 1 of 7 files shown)


src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -581,40 +579,106 @@ def get_trl_descriptor(
...
-            return DataKey(source=tango_resource, dtype="string", shape=[])
+            return {"source": tango_resource, "dtype": "string", "shape": []}
         else:
             if datatype and not issubclass(tr_dtype, datatype):
                 raise TypeError(
                     f"{tango_resource} has type {tr_dtype.__name__} "
                     f"not {datatype.__name__}"
                 )
-            return DataKey(source=tango_resource, dtype=tr_dtype_desc, shape=[])
+            return {"source": tango_resource, "dtype": tr_dtype_desc, "shape": []}
```
coretl, line 600: Any reason for this particular change? At runtime they are identical, just gives more IDE help as to the fields at edit time...

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -581,40 +579,106 @@ def get_trl_descriptor(
...
+                limits_rds["value_difference"] = delta_val
+            if limits_rds:
+                _limits["rds"] = limits_rds
+
+            _choices = list(config.enum_labels) if config.enum_labels else []
+            _dims = []
+            if config.data_format == AttrDataFormat.SPECTRUM:
+                _dims = ["x"]
+            elif config.data_format == AttrDataFormat.IMAGE:
+                _dims = ["y", "x"]
```
coretl, line 643: Isn't this reflected in `shape`?

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -581,40 +579,106 @@ def get_trl_descriptor(
...
+            if _choices:
+                descriptor["choices"] = _choices
+            if _dims:
+                descriptor["dims"] = _dims
+            if _precision:
+                descriptor["precision"] = _precision
+            if config.unit:
+                descriptor["units"] = config.unit
+            if config.name:
+                descriptor["object_name"] = config.name
```
coretl, line 673: This should will be added here so will override this key:
https://github.com/bluesky/bluesky/blob/985b14a76ed01661e2d16e08fe7d405734bdcaeb/src/bluesky/bundlers.py#L219

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -581,40 +579,106 @@ def get_trl_descriptor(
...
+
+        elif isinstance(config, CommandInfo):
+            if config.cmd_name:
+                descriptor["object_name"] = config.cmd_name
```
coretl, line 677: ditto

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -581,40 +579,106 @@ def get_trl_descriptor(
...
+
+    return DataKey(**descriptor)
```
coretl, line 679: Just wondering if you could make any use of https://github.com/bluesky/ophyd-async/blob/90209e187e869653286720033e9387fca84177c2/src/ophyd_async/core/_signal_backend.py#L180-L195

Like we do for EPICS:
https://github.com/bluesky/ophyd-async/blob/90209e187e869653286720033e9387fca84177c2/src/ophyd_async/epics/core/_p4p.py#L391-L396

src/ophyd_async/tango/core/_tango_transport.py
burkeds: I made this change because if `get_simple_datakey` returns a DataKey, I get a pyright error in get_trl_descriptor because _limits isn't a Limits object. For it to be a Limits object, I need RdsRange which I can't seem to import from event_model. It was easier to keep the descriptor as a dictionary until it's returned by get_trl_descriptor. If I can import RdsRange then I can make it a DataKey again.

src/ophyd_async/tango/core/_tango_transport.py
burkeds: I am not sure what you mean. `dims` is the label of the axes described by shape. Tango does not carry these labels so I just call them 'x' and 'y'.

src/ophyd_async/tango/core/_tango_transport.py
burkeds: I removed lines 672 and 673

src/ophyd_async/tango/core/_tango_transport.py
burkeds: Removed

src/ophyd_async/tango/core/_tango_transport.py
evvaaaa: Yes you should, are you sure you're on `event_model>=1.22.2`?

evvaaaa: Here it'`>=1.22.1`

https://github.com/bluesky/ophyd-async/blob/90209e187e869653286720033e9387fca84177c2/pyproject.toml#L18

So it should install the latest `event_model` unless any of the other projects put an upper bound. :thinking:

burkeds: That's probably what happened. I pushed an update to the required event-model version since the 'rds' key is not in 1.22.1.

src/ophyd_async/tango/core/_tango_transport.py
burkeds: Does this make the datakey from the signal itself? Is this because an EPICS PV does not easily provide this information? It doesn't look like this enforces that the signal datatype match the PV. With Tango, the descriptor can simply return information from the device server and check that the type matches.

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -615,6 +603,80 @@ def get_trl_descriptor(
...
+            delta_t, delta_val = map(
+                try_to_cast_as_float, (alarm_info.delta_t, alarm_info.delta_val)
+            )
+            limits_rds = RdsRange(
+                time_difference=delta_t or 0.0,
+                value_difference=delta_val or 0.0,
+            )
+
+            if any((delta_t, delta_val)):
+                _limits["rds"] = limits_rds
```
evvaaaa, line 640: Could `delta_t`, `delta_val` ever intentionally be `(0.0, 0.0)` instead of `(None, None)`? If so this check would leave this data out of the datakey.

```suggestion
            if (alarm_info.delta_t, alarm_info.delta_val) != (None, None):
                delta_t = try_to_cast_as_float(alarm_info.delta_t) or 0.0
                delta_val = try_to_cast_as_float(alarm_info.delta_val) or 0.0
                _limits["rds"] = RdsRange(
                    time_difference=delta_t,
                    value_difference=delta_val
                )
```

tacaswell:
It is better to provide no dims rather than dummy place holder dims.

The current primary consumer of this is generating xarrays from the collected data which can leverage dims if they exist (which in turn helps make sure that array operations are correctly aligned rather than blindly zipping by dimension index).   If dims aren't provided xarray will auto-generate them at the end.  Providing dummy dims reduces the value of intentionally provided ones.

src/ophyd_async/tango/core/_tango_transport.py
coretl: It uses the signal value for 2 reasons:
1. To get the dtype, both the loose json form ('integer`) and the numpy datatype ('<i8')
2. To get the shape, needed especially for waveforms

The first could be got from the EPICS metadata, but the second requires the current value

coretl:
@burkeds I've lost track of where we got to with this, do you need me to re-review?

src/ophyd_async/tango/core/_tango_transport.py
burkeds: @coretl I think we just need to resolve this question

burkeds: I'm a little lost too. I think I'll look at it again. My issue was this "converter" that was added to the tango code

burkeds: @coretl How are command signals carrying the None datatype handled? If I try to do something like this:

```python
test_datakey = make_datakey(
    datatype=self.datatype,
    value=self.converter.value(value),
    source=self.source("", True),
    metadata={}
)
```

`_datakey_dtype_numpy` throws an exception if datatype is None which is not a class.

coretl: Is this for `Commands` that don't have a value? In which case I think it is fine to raise an error when making the descriptor, because it will be wrapped with a `SignalX` that doesn't allow you to call the function `backend.get_datakey`

coretl: Yes, we decided to use `Sequence[str]` rather than `npt.NDArray[np.str_]` as the numpy arrays of strings are all fixed size and actually quite difficult to work with. We might revisit this when numpy variable length strings are supported in structured arrays...

burkeds: Ok, I think I understand. I can handle a sequence of strings. What about an image or 2d array of strings. Would it be `
`List[Sequence[str], Sequence[str]]`?

coretl: Do you have such things in Tango? If so, are they natively variable length or fixed length?

coretl: Reading that page it looks like the `(DEV_STRING + IMAGE)` is exploded into a 1D `Sequence[str]` so I suggest you do the same. For the `DEV_LONGSTRINGARRAY` and `DEV_DOUBLESTRINGARRAY` this looks very much like table types to me:
```python
class TangoLongStringTable(Table):
    long: Array1D[np.int32]
    string: Sequence[str]

class MyDevice(TangoDevice):
    an_attr: SignalRW[TangoLongStringTable]
```

burkeds: Ok thanks for the suggestion. I think I made some significant improvements to how we handle Tango types now. It looks like there is an issue with approximating arrays with nested values at index 0. 

```
**  File "/home/burkedev/projects/BLUESKY/ophyd-async/venv/lib/python3.11/site-packages/_pytest/python_api.py", line 375, in _check_type
    raise TypeError(msg.format(x, index, pprint.pformat(self.expected)))
TypeError: pytest.approx() does not support nested data structures: ('one', 'two', 'three') at index 0
  full sequence: (('one', 'two', 'three'), ('one', 'two', 'three'))**
```

coretl: Is that for a 2D array of strings? What datatype does pytango give you? It looks from the docs like it will be a `Sequence[str]` which is 1D...

coretl (commented):
A skim review looks fine.

There are a few commented out tests, should they be uncommented or deleted?

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -73,28 +85,91 @@ async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+
+class TangoDoubleStringTable(Table):
+    double: Array1D[np.float64]
+    string: Sequence[str]
+
+
+class TestConfig:
+    data_type: CmdArgType
+    data_format: AttrDataFormat
+    enum_labels: list[str]
```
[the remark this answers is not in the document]
burkeds: Done.

tests/tango/test_tango_signals.py
```diff
@@ -327,37 +335,37 @@ async def test_set_with_converter(everything_device_trl):
...
+    #                 ExampleStrEnum.C,
+    #             ],
+    #             [
+    #                 ExampleStrEnum.B,
+    #                 ExampleStrEnum.C,
+    #             ],
+    #         ],
+    #         dtype=ExampleStrEnum,
+    #     )
+    # )
```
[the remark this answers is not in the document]
burkeds: It no longer works since we are not supporting string images in Tango for now. I have removed the commented tests.

tests/tango/test_base_device.py
```diff
@@ -390,3 +478,41 @@ async def test_signal_autofill(tango_test_device, auto_fill_signals):
     else:
         assert not hasattr(test_device, "ignored_attr")
         assert not hasattr(test_device, "readback")
+
+
+@pytest.mark.asyncio
```
coretl, line 483: Tango under windows is currently tested, but the runners are quite weedy, so need a bigger timeout. Does something like this work?
```suggestion
@pytest.mark.asyncio
@pytest.mark.timeout(10.0)
```

coretl (changes requested):

tests/tango/test_base_device.py
burkeds: It seems as if this error occurs at the end of the last test in test_tango_transport. I was seeing the error on the last test, so I tried splitting that into smaller tests and I still see it at the end of the last newly defined test.

burkeds: The last 4 tests in test_tango_transport: `test_type_mismatch_justvalue`, `test_type_mismatch_array`, `test_type_mismatch_sequence`, and `test_type_mismatch_longstringarray`, seem to be the culprits. If I comment out all 4, the tests pass on windows. If I comment out any subset of them, it times out on the last one. Whether I include justvalue or all four, it times out after the last test.

burkeds: In this commit 1f4a3d45b45bc5b3a7cb34900664b9671e37483a you can see the windows tests all passed when I commented out those tests,

coretl: Ahh, I see. I didn't look far enough back, I just check the last couple of logs. I have no idea why the process hangs, but I notice we're currently pinning Tango for some reason:
https://github.com/bluesky/ophyd-async/blob/5e6b1b932ac54c18f8aceb10111219e27fd0a33b/pyproject.toml#L34

If we remove the pin does it fix it?

burkeds: @coretl I am not sure what to do about the windows tests. They seem to fail in different ways each time I make a change. It doesn't just seem to be the tango tests that are failing. On main it looks like we've been having problems with Windows tests for a while now. Could we put this PR through before the windows issues are sorted?

PR #870, opened 2025-04-28
Pass datatype between FastCS-Eiger and ADOdin

### LINKED ISSUE #529: FastCS Eiger: Pass datatype from eiger to odin

shihab-dls:
Fixes #529

This PR retrieves `bit-depth` from FastCS-Eiger and passes it as `datatype` for ADOdin.

### DIFF
--- a/src/ophyd_async/epics/eiger/_odin_io.py
+++ b/src/ophyd_async/epics/eiger/_odin_io.py
@@ -10,6 +10,7 @@ from ophyd_async.core import (
     Device,
     DeviceVector,
     PathProvider,
+    SignalR,
     StrictEnum,
     observe_value,
     set_and_wait_for_other_value,
@@ -80,9 +81,11 @@ class OdinWriter(DetectorWriter):
         self,
         path_provider: PathProvider,
         odin_driver: Odin,
+        eiger_bit_depth: SignalR[int],
     ) -> None:
         self._drv = odin_driver
         self._path_provider = path_provider
+        self._eiger_bit_depth = eiger_bit_depth
         super().__init__()
 
     async def open(self, name: str, exposures_per_event: int = 1) -> dict[str, DataKey]:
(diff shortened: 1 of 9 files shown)


src/ophyd_async/fastcs/eiger/_eiger_controller.py
```diff
@@ -51,6 +52,9 @@ async def prepare(self, trigger_info: TriggerInfo):
                     self._drv.detector.frame_time.set(trigger_info.livetime),
                 ]
             )
+
+        await wait_for_value(self._drv.stale_parameters, False, timeout=DEFAULT_TIMEOUT)
```
DominicOram, line 56: Should: I believe with the changes in https://github.com/DiamondLightSource/fastcs-eiger/pull/61 this is not required. Lets remove for now and see how we go.

src/ophyd_async/epics/eiger/_odin_io.py
```diff
@@ -80,9 +81,11 @@ def __init__(
         self,
         path_provider: PathProvider,
         odin_driver: Odin,
+        eiger_bit_depth: SignalR[int],
```
DominicOram, line 84: Should: This needs to be a reference (https://blueskyproject.io/ophyd-async/main/_api/ophyd_async/ophyd_async.core.html#ophyd_async.core.Reference) so that it doesn't get renamed

DominicOram (changes requested):
Thanks. A couple of comments in code, also can you add a test that makes sure the datatype gets correctly passed on before the writer is opened?

src/ophyd_async/epics/eiger/_odin_io.py
DominicOram: FYI https://github.com/bluesky/ophyd-async/issues/871

DominicOram (changes requested):
Thanks but still missing a specific test on the bit depth being set at the correct time

shihab-dls:
> Thanks but still missing a specific test on the bit depth being set at the correct time

Sorry about that; Two tests added now, at slightly different scopes (one checking detector `bit_depth` is passed to the writer, another checking that `data_type` is only set after `Writer.open()` even though `writer._eiger_bit_depth` is correctly passed on instantiation)

PR #887, opened 2025-05-08
Pass name to backend of derived signal

### LINKED ISSUE #886: Derived signals name not set when added as readables
Setting derived signals as readable such as:
```python
    self.x = self._factory.derived_signal_rw(float, "x")
    self.y = self._factory.derived_signal_rw(float, "y")
    self.z = self._factory.derived_signal_rw(float, "z")
```
will result in a reading such as:
`{'': {'value': 0.0, 'timestamp': 97893.330239932, 'alarm_severity': 0}}`
because the signals name is not properly set.

This is because we return 
```python
return signal_cls(backend)
```
in `_make_signal` of `DerivedSignalFactory`

## Steps To Reproduce
1. Add derived signals as readable
2. Read device
3. See that signal names are empty strings

## Acceptance Criteria
- Return `signal_cls(backend, name=name)`

shihab-dls:
Fixes #886 

This PR passes `name` to the signal returned by the derived signal factory, such that multiple derived signals can be correctly read via `StandardReadable`'s `read()`.

### DIFF
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -94,7 +94,7 @@ class DerivedSignalFactory(Generic[TransformT]):
         backend = DerivedSignalBackend(
             datatype, name, self._transformer, units, precision
         )
-        return signal_cls(backend)
+        return signal_cls(backend, name=name)
 
     def derived_signal_r(
         self,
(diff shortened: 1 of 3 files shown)


shihab-dls:
After discussion with @coretl, setting the device's name will set the signals names, so this is not required. This is tested [here](https://github.com/bluesky/ophyd-async/blob/main/tests/core/test_device.py#L112), so closing this PR and issue.

Issue #713, opened 2024-12-19
Simulated AreaDetector still in Acquire state after collection when `yield from trigger_and_read(detector, *devices)`


DiamondJoseph:
Stacktrace:
[bar.txt](https://github.com/user-attachments/files/18196680/bar.txt)

When calling trigger_and_read rapidly on the SimulatedAreaDetector, backed by the containerised IOC

```sh
ophyd-async --version
0.9.0a1
```

## Steps To Reproduce
1. Create a plan that loops over trigger_and_read with a the ADsim
2. Observe that it occasionally (~0.5% of the time?) is still in the Acquire state

## Acceptance Criteria
- The ADSim can reliably be used on scans that collection many, many points without failure.

DiamondJoseph:
This was observed by @DiamondRC who can provide the plan if behaviour cannot be replicated

coretl:
This looks like a race condition we fixed in Malcolm, but didn't apply here.
Malcolm:
https://github.com/DiamondLightSource/pymalcolm/blob/264ec4a319092a545853f41f89ddf1d21da56504/malcolm/modules/ca/parts/caactionpart.py#L98-L122
Ophyd-async:
https://github.com/bluesky/ophyd-async/blob/c781e27691efde1dd7b4d185252840566b01d041/src/ophyd_async/epics/adcore/_core_logic.py#L101-L107
So if we changed to:
```python
await status
state = None
try:
    async for state in observe_value(driver.detector_state, done_timeout=STATUS_TIMEOUT):
        if state in good_states:
            return
except asyncio.TimeoutError:
    raise ValueError(
        f"Final detector state {state.value} not in valid end "
        f"states: {good_states}"
    )
```
or something similar that should fix it...

canismarko:
We've been running into this bug I think with our xspress3 family of detectors. I implemented the suggested solution [in a fork](https://github.com/bluesky/ophyd-async/compare/main...spc-group:ophyd-async:detector_races) and so far we have not triggered this race condition.

coretl:
Ok, so this issue becomes making a PR out of https://github.com/bluesky/ophyd-async/compare/main...spc-group:ophyd-async:detector_races with a test where the state updates before or after acquire returns

canismarko:
PR is ready. Tests are passing, though the CI is failing because of an unrelated test of sim_motor move profiles.

canismarko:
Should be fixed now by https://github.com/bluesky/ophyd-async/pull/889

[...]

PR #889, opened 2025-05-08
Fix for detector race condition when checking acquire status

canismarko:
Fixes:
- https://github.com/bluesky/ophyd-async/issues/713

We've been using this fix with a real Xspress3 pulse processor, and since implementing the fix, it has not triggered the race condition.

The unit test fails with the upstream version of `_core_logic.py`, and passes for me with the updated version.

### DIFF
--- a/src/ophyd_async/epics/adcore/_core_logic.py
+++ b/src/ophyd_async/epics/adcore/_core_logic.py
@@ -136,7 +136,10 @@ class ADBaseController(DetectorController, Generic[ADBaseIOT]):
                     ) from exc
                 else:
                     # No updates from the detector, something else is wrong
-                    raise exc
+                    raise asyncio.TimeoutError(
+                        "Could not monitor detector state: "
+                        + self.driver.detector_state.source
+                    ) from exc
 
         return AsyncStatus(complete_acquisition())
 
--- a/src/ophyd_async/epics/adcore/_core_logic.py
+++ b/src/ophyd_async/epics/adcore/_core_logic.py
@@ -7,7 +7,8 @@ from ophyd_async.core import (
     DetectorController,
     DetectorTrigger,
     TriggerInfo,
-    set_and_wait_for_value,
+    observe_value,
+    set_and_wait_for_value
 )
 
 from ._core_io import (
@@ -111,11 +112,17 @@ class ADBaseController(DetectorController, Generic[ADBaseIOT]):
             # NOTE: possible race condition here between the callback from
             # set_and_wait_for_value and the detector state updating.
             await status
-            state = await self.driver.detector_state.get_value()
-            if state not in self.good_states:
+            state = None
+            try:
+                async for state in observe_value(
+                    self.driver.detector_state, done_timeout=DEFAULT_TIMEOUT
+                ):
+                    if state in self.good_states:
+                        return
+            except asyncio.TimeoutError:
                 raise ValueError(
-                    f"Final detector state {state.value} not "
-                    "in valid end states: {self.good_states}"
+                    f"Final detector state {state.value} not in valid end "
+                    f"states: {self.good_states}"
                 )
 
         return AsyncStatus(complete_acquisition())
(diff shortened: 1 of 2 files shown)


canismarko:
CI failure looks to be coming from the sim_motor, unrelated to the code being changed.

tests/epics/adcore/test_drivers.py
```diff
@@ -93,3 +93,48 @@ async def wait_then_fail():
...
+async def test_start_acquiring_driver_and_ensure_status_disconnected(
+    controller: adsimdetector.SimController,
+):
+    """This test ensures the function behaves gracefully if no detector
+    states are available.
+
+    """
+    acquiring = await controller.start_acquiring_driver_and_ensure_status()
+
+    with pytest.raises(asyncio.TimeoutError):
```
coretl, line 139: Please could we check the error message here

tests/epics/adcore/test_drivers.py
```diff
@@ -93,3 +93,48 @@ async def wait_then_fail():
         ValueError, match="Final detector state Disconnected not in valid end states:"
     ):
         await acquiring
+
+
+@patch("ophyd_async.core._detector.DEFAULT_TIMEOUT", 0.2)
```
coretl, line 98: Please could we add `timeout=DEFAULT_TIMEOUT` as an argument to `start_acquiring_driver_and_ensure_status` with docstring `timeout used for waiting for the driver to start acquiring, and for waiting for the detector to detector to be in a good state after it stops acquiring` then use it here instead of patching.

canismarko: Added in https://github.com/bluesky/ophyd-async/pull/889/commits/8111d62fdbea4c5bb5be492e82e6dd5dccde63ab. 

The full timeout is used for the first call to set the detector acquire. Then any remaining timeout is used to wait for the detector to return to a good status.

tests/epics/adcore/test_drivers.py
canismarko: Sure. The original ``TimeoutError`` exception does not have a message. I put in a message when re-raising the exception that provides the source of the detector state signal.

Is it safe to assume that `driver.detector_state` will always have a `.source`?

coretl: Yes

canismarko:
Looking through the linting failures, none of them seem to come from the code changed in this PR. If you want them fixed in this PR, let me know. Otherwise, I think it's ready for another review.

src/ophyd_async/epics/adcore/_core_logic.py
```diff
@@ -89,34 +91,55 @@ async def set_exposure_time_and_acquire_period_if_supplied(
...
-            # set_and_wait_for_value and the detector state updating.
             await status
-            state = await self.driver.detector_state.get_value()
-            if state not in self.good_states:
-                raise ValueError(
-                    f"Final detector state {state.value} not "
-                    "in valid end states: {self.good_states}"
-                )
+            state = None
+            timeout_left = timeout - (time.monotonic() - t0)
```
coretl, line 124: Hmm, nice idea to use the total timeout for both actions, but given they are separated by some period of time it probably doesn't add much to do this. On reflection then maybe we should just pass the two timeouts into the function directly:
```python
        :param start_timeout:
            Timeout used for waiting for the driver to start acquiring
        :param state_timeout:
            Timeout used for waiting for the detector to be in a good state
            after it stops acquiring
```

coretl (commented):
Apologies for changing my mind, but I think two parameters to the function is clearer.

Don't worry about the linting changes, I'll fix in a different PR

src/ophyd_async/epics/adcore/_core_logic.py
canismarko: Done: https://github.com/bluesky/ophyd-async/pull/889/commits/d2e5aa838b74188ac7164cff9a7d367f90f77004

coretl (approved):

coretl:
Thanks!

PR #892, opened 2025-05-12
Fix prepare motor path

tomtrafford:
_prepare_motor_path does not properly account for travel in a negative direction. The calculations assumed that `fly_velocity ` would provide a direction, however `abs(fly_velocity)` gets passed in during prepare. This change therefore calculate the direction based on start_position and end_position.

I've also modified the existing tests to show the correct behaviour (not using negative velocity and extending the limits due to the corrected position).

### DIFF
--- a/src/ophyd_async/epics/motor.py
+++ b/src/ophyd_async/epics/motor.py
@@ -260,10 +260,11 @@ class Motor(
             (await self.acceleration_time.get_value()) * fly_velocity * 0.5
         )
 
-        self._fly_completed_position = end_position + run_up_distance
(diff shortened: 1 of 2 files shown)


src/ophyd_async/epics/motor.py
coretl, line 1: This looks fine for now.

In the future we might want to modify it to look more like the sim version:
https://github.com/bluesky/ophyd-async/blob/147fd9495b078fa0e9e9d94fd191221387ed68f0/src/ophyd_async/sim/_motor.py#L22-L50

coretl (approved):

PR #895, opened 2025-05-14
Allow None in Shapes

Tom-Willemsen:
After https://github.com/bluesky/event-model/pull/352, bluesky descriptors are allowed to contain `None` in the shape (meaning unknown or variable length for that dimension).

This is an initial pass at broadening the type hints of `shape` in ophyd-async to allow this to be used.

I've made the areadetector implementation error on this for now, as it's not clear to me whether areadetector can conceptually do this, or how to deal with chunking. Happy to defer to areadetector users on best behaviour there.


src/ophyd_async/epics/adcore/_hdf_writer.py
```diff
@@ -75,6 +76,19 @@ async def open(
...
+            # - Can AreaDetector support this?
+            # - How to deal with chunking?
+            # Don't support for now - leave option open to support it later
+            raise ValueError(
+                "Datasets with partially unknown dimensionality "
+                "are not currently supported by ADHDFWriter."
+            )
+
+        # Pyright doesn't understand the above check as a type-narrowing expression
+        detector_shape = cast(tuple[int, ...], detector_shape)
```
DiamondJoseph, line 90:
```suggestion
      def is_fully_described(val: Tuple[int | None, ...]) -> TypeGuard[tuple[int, ...]]:
            # Questions:
            # - Can AreaDetector support this?
            # - How to deal with chunking?
            # Don't support for now - leave option open to support it later
            return None not in val

        if not is_fully_described():
            raise ValueError(
                "Datasets with partially unknown dimensionality "
                "are not currently supported by ADHDFWriter."
            )
```


I haven't used it, but I think this is how to tell Pyright that you're narrowing the permitted types
https://docs.python.org/3/library/typing.html#typing.TypeGuard

coretl (approved):

PR #896, opened 2025-05-15
ci: Update to latest ruff expectations of

DiamondJoseph:
Ruff has released a new version and that version appears to sort imports from nested packages above rather than alphabetically. We should either adapt or configure it not to. This is one of those.


coretl:
Hah, you beat me to it, but I think the correct solution is https://github.com/bluesky/ophyd-async/pull/897

DiamondJoseph:
Replaced by #897

PR #897, opened 2025-05-15
Don't consider tests as first party imports

coretl:
Highlighted by https://github.com/astral-sh/ruff/pull/16565 causing ruff to sort imports differently in 0.11.9+

https://docs.astral.sh/ruff/settings/#src was being used wrong (something I added to the copier template a long time ago)

It should actually point to the first party imports, so the default of `[".", "src"]` works for us. Because it was set to include `tests` it considered `tango` a first party import, so this corrects that and fixes lint.

### DIFF
--- a/src/ophyd_async/tango/core/_tango_transport.py
+++ b/src/ophyd_async/tango/core/_tango_transport.py
@@ -10,18 +10,6 @@ from typing import Any, ParamSpec, TypeVar, cast
 import numpy as np
 from bluesky.protocols import Reading
 from event_model import DataKey
-
-from ophyd_async.core import (
-    AsyncStatus,
-    Callback,
-    NotConnected,
-    SignalBackend,
-    SignalDatatypeT,
-    StrictEnum,
-    get_dtype,
-    get_unique,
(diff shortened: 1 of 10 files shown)


src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -40,6 +28,18 @@
 )
 from tango.utils import is_array, is_binary, is_bool, is_float, is_int, is_str
 
+from ophyd_async.core import (
+    AsyncStatus,
+    Callback,
+    NotConnected,
+    SignalBackend,
```
DiamondJoseph, line 35:
```suggestion
```

nit: Code scanning says this isn't required but I don't 100% believe if it the linting didn't catch it

DiamondJoseph (approved):
Removing config overrides probably means this is moving in the right direction

src/ophyd_async/tango/core/_tango_transport.py
coretl: the AI is wrong...

PR #899, opened 2025-05-19
Set device name when child device attached

### LINKED ISSUE #894: Decide once and for all about how Devices are named
At the moment, Device naming happens via `Device.set_name`. This is called:
- in `Device.__init__(name)` at construction
- after construction by the `init_devices` or `dodal.common.beamlines.beamline_utils.device_factory`

This leads to some odd edge conditions:

```python
class Base(Device):
    def __init__(self, name=""):
        self.signal1 = soft_signal_rw()
        super().__init__(name=name)

class Derived(Base):
    def __init__(self, name=""):
        super().__init__(name=name)
        self.signal2 = soft_signal_rw(name="blah")

# Naming at init only names those that are present when the superclass init is called
derived1 = Derived(name="derived")
assert derived1.signal1.name == "derived-signal1"
assert derived1.signal2.name == "blah"

# Naming after init names everything
derived2 = Derived()
derived2.set_name("derived")
assert derived2.signal1.name == "derived-signal1"
assert derived2.signal2.name == "derived-signal2"
```
This is inconsistent, so we should fix it.

Possible solutions:
1. Remove `name` from `Device.__init__` and require `Device.set_name` to be called after creation
2. Set the name of child devices in `Device.__setattr__`

Both of those solutions will break one particular @rtuck99 had (that I can't remember details of) where `set_name` was never used after init, so the name "blah" in the above example was actually the desired name.

To support that use case we would have to override `Derived.set_name` to explicitly name `signal2` to be "blah", either by overriding the baseclass `set_name` (for 1):

```python
class Derived(Base):
    def set_name(self, name: str, *, child_name_separator: str | None = None) -> None:
        super().set_name(name, child_name_separator=child_name_separator)
        self.signal2.set_name("blah")
```

or by making an intermediate function that could be overridden instead (for 2):

```python
class Derived(Base):
    def set_child_name(self, child, name):
        super().set_child_name(child, "blah" if child is self.signal2 else name)
```

I think 2 would cause less breakages, so I would favour that over 1, but I'd welcome thoughts on the best API to override the child name.

@DominicOram @DiamondJoseph @callumforrester @jwlodek thoughts?
        
## Acceptance Criteria
- Code changed
- Only the single test that mimics @rtuck99's use case is changed

coretl:
## Release notes

Previously the naming of child signals was inconsistent between those named by passing `name` in `__init__`, and those named after the fact using `set_name` via `init_devices` or something similar. Now child devices are now named when they are attached to a parent, which means that the `super` order in `__init__` doesn't matter from the perspective of naming. For example:
```python
class Base(Device):
        self.sig1 = soft_signal_rw(float)
        super().__init__(name=name)

class Derived1(Base):
        self.sig2 = soft_signal_rw(float)
        super().__init__(name=name)

class Derived2(Base):
        super().__init__(name=name)
        self.sig2 = soft_signal_rw(float)
```

Previously:
-If `device = Derived1(name="mydevice")` then `device.sig2.name == "mydevice-sig2"`
-If `device = Derived2(name="mydevice")` then `device.sig2.name == ""`

Now:
- `device.sig2.name == "mydevice-sig2"` in both cases

The canonical way to rename a signal to something different to `{parent.name}-{signal_name}` is to override `set_name`, e.g. in motor:
```python
    def set_name(self, name: str, *, child_name_separator: str | None = None) -> None:
        """Set name of the motor and its children."""
        super().set_name(name, child_name_separator=child_name_separator)
        # Readback should be named the same as its parent in read()
        self.user_readback.set_name(name)
```

Fixes #894

### DIFF
--- a/docs/explanations/devices-signals-backends.md
+++ b/docs/explanations/devices-signals-backends.md
@@ -23,7 +23,7 @@ The `Device` class is the base of all ophyd-async objects that are published to 
 - a [](#Device.parent) read-write property to read it's parent Device if it exists
 - a [](#Device.children) to iterate through the Device attributes, yielding the `(name, child)` child Devices
 - a `setattr` override that detects whether the attribute is also a Device and sets its parent
-- a [](#Device.set_name) method to set its name and also set the names of its children using the parent name as a prefix
+- a [](#Device.set_name) method to set its name and also set the names of its children using the parent name as a prefix, called at init and also when a new child is attached to an already named Device
 - a [](#Device.connect) method that connects it and its children
 
 All the above methods are concrete, but `connect()` calls out to a [](#DeviceConnector) to actually do the connection, only handling caching itself. This enables plug-in behavior on connect (like the introspection of child Attributes in Tango or PVI, or the special case for `Signal` we will see later).
--- a/docs/tutorials/implementing-devices.md
+++ b/docs/tutorials/implementing-devices.md
@@ -231,6 +231,7 @@ Finally, we need to communicate to bluesky that it has to `trigger()` and acquis
 
 Although the Signals are declared via type hints, the DeviceVector requires explicit instantiation in an `__init__` method. This is because it requires the `num_channels` to be passed in to the constructor to know how many channels require creation. This means that we also need to do the PV concatenation ourselves, so if the PV prefix for the device as `PREFIX:` then the first channel would have prefix `PREFIX:CHAN1:`. We also register them with `StandardReadable` in a different way, adding them within a [](#StandardReadable.add_children_as_readables) context manager which adds all the children created within its body.
 
+Whilst it is not required for the call to `super().__init__` to be after all signals have been created it is more efficient to do so. However, there may be some edge cases where signals need to be created after this e.g. for [derived signals](../how-to/derive-one-signal-from-others.md) that depend on their parent.
 :::
 
 :::{tab-item} Tango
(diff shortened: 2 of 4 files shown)


coretl:
@DominicOram @Tom-Willemsen @callumforrester I've added some release notes at the top of the page. I added a short note to the Device explanation. The override `set_name` pattern hasn't changed and is documented in the second tutorial, so I left that as it was. Do you think any more docs should be added?

> I also think it's worth stopping to think about what is the canonical way of setting names now. If we no longer need to call super at the end it's probably worth removing comments like [this](https://github.com/bluesky/ophyd-async/blob/f652399e24c045b0f6c675d173c1ef588f909062/src/ophyd_async/sim/_stage.py#L18) and [this](https://github.com/bluesky/ophyd-async/blob/f652399e24c045b0f6c675d173c1ef588f909062/src/ophyd_async/epics/demo/_stage.py#L14).

The comment is still technically correct, we should still do the superclass init with the name, it's just that after the device is named you can carry on adding child signals and they will be named too...

> I wonder if we should go through and move all the `super` calls back to the start of the `__init__`? Thoughts @callumforrester?

I would leave super at the end, as each addition of a child after naming the device will rename all the children in the device, which adds a little bit of overhead. Not sure where is best to document this...

DominicOram:
>Not sure where is best to document this...

Maybe add to https://github.com/bluesky/ophyd-async/blob/b28a3efc47b6d6dd3eef94b06b633261a5a58362/docs/tutorials/implementing-devices.md?plain=1#L317C7-L317C24 something like `Whilst it is not required for the call to super().__init__ to be after all signals have been created it is more efficient to do so. However, there may be some edge cases where signals need to be created after this e.g. for derived signals[link] that depend on their parent`

DominicOram (approved):
Great, thanks!

PR #900, opened 2025-05-20
Fix ADOdin writer.open() for dodal plan

shihab-dls:
This PR amends ADOdin logic based on testing done on i03 with fastcs-eiger for running a standard arming sequence. The dodal plan this was tested against can be found [here](https://github.com/DiamondLightSource/dodal/pull/1187)


DominicOram (approved):
Great, thank you!

PR #904, opened 2025-05-28
Updated DeviceVector key to be any type, updated references for declared types

### LINKED ISSUE #903: Extend DeviceVector to support string and integer
In  the ophyd-async documentation, there is the DeviceVector class:
"
Sometimes, it makes sense to group devices by number, say an array of sensors:

```Python
class SensorGroup(StandardReadable):
    def __init__(self, prefix: str, name: str = "", sensor_count: int = 3) -> None:
            self.sensors = DeviceVector(
                {i: Sensor(f"{prefix}{i}:") for i in range(1, sensor_count + 1)}
            )

        super().__init__(name)
```

DeviceVector allows writing maintainable, arbitrary-length device groups instead of fixed classes for each possible grouping. A DeviceVector can be accessed via indices, for example: my_sensor_group.sensors[2]. Here sensors is a dictionary with integer indices rather than a list so that the most semantically sensible indices may be used, the sensor group above may be 1-indexed, for example, because the sensors’ datasheet calls them “sensor 1”, “sensor 2” etc. 
"

However, it would also be useful to support str as well as int as the key for DeviceVector e,g

```Python
class ExcitationEnergySources(StandardReadable):
    def __init__(self, name: str = "", alias_name_to_device: dict[str, Reference[Device]]) -> None:
            self.sources = DeviceVector(
                alias_to_device
            )

        super().__init__(name)
```
```Python
pgm = PGM(...)
dcm = DCM(...)
energy_sources = ExcitationEnergySources("energy_sources", {"source1" : Reference(DCM), "source2" : Reference(PGM)})

#Now have a way to dynamically get excitation energy source for a region from a plan

def plan(region: AbstractBaseRegion, energy_sources = inject("energy_sources")) -> Device:
     source_alias : str = region.excitation_energy_source
     return energy_sources.sources[source_alias]
```

oliwenmandiamond:
DeviceVector only accepts int as the key. This change allows a key to be anything, provided all keys are of the same type e.g `DeviceVector[int, Device]`, `DeviceVector[Device, Device]`, `DeviceVector[str, Device]` etc. Use cases are described in this issue here:

https://github.com/bluesky/ophyd-async/issues/903

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -144,6 +144,7 @@ obj_ignore = [
     "ophyd_async.core._detector.DetectorControllerT",
     "ophyd_async.core._detector.DetectorWriterT",
     "ophyd_async.core._device.DeviceT",
+    "ophyd_async.core._device.KeyT",
     "ophyd_async.core._device_filler.SignalBackendT",
     "ophyd_async.core._device_filler.DeviceConnectorT",
     "ophyd_async.core._derived_signal_backend.TransformT",
--- a/docs/explanations/decisions/0009-procedural-vs-declarative-devices.md
+++ b/docs/explanations/decisions/0009-procedural-vs-declarative-devices.md
@@ -28,7 +28,7 @@ class EpicsProceduralDevice(StandardReadable):
 and a Tango/FastCS procedural equivalent would be (if we add support to StandardReadable for Format.HINTED_SIGNAL and Format.CONFIG_SIGNAL annotations):
 ```python
 class TangoDeclarativeDevice(StandardReadable, TangoDevice):
-    value: Annotated[DeviceVector[SignalR[float]], Format.HINTED_SIGNAL]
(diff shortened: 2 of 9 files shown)


coretl:
In ophyd-async at the moment we have 2 ways to make a tree of Devices where their read output is a tree of the children:
- `StandardReadable` if the children have string names
- `DeviceVector` if the children have integer names

I'm loath to extend `DeviceVector` to support arbitrary typed keys without understanding the use case properly.

For instance, these are pretty equivalent:
```python

class ParentWithVector(Device):
    def __init__(self, names=("foo", "bar")):
        self.children = DeviceVector({name: ChildDevice() for name in names})

class ParentWithChildren(StandardReadable):
    def __init__(self, names=("foo", "bar")):
            for name in names:
                setattr(self, name, ChildDevice())
```

Can you post your use case please?

oliwenmandiamond:
PR too out of date

PR #911, opened 2025-06-05
Amend FastCS-Eiger and ADOdin logic

### LINKED ISSUE #866: Setting odin capture signal does not return a put callback.
It seems that setting `self._drv.capture` to `Writing.CAPTURE` does not return a put callback. i.e., awaiting this put blocks until timeout, even though the read back value confirms the new value. To work around this, we set `self._drv.capture` and wait for `self._drv.capture_rbv` explicitly with `set_and_wait_for_other_value` with `set_timeout=None`

## Acceptance Criteria
- Setting `self._drv.capture` returns when put is complete.

shihab-dls:
This PR adds MX specific signals required after testing on i03. This also closes #866, to use identical `_writer.close()` logic to `ADWriter`. The test that checks the order of setting and waiting for signals when we `_writer.open()` has been amended to remove the need to `patch`.

### DIFF
--- a/src/ophyd_async/epics/adcore/_core_writer.py
+++ b/src/ophyd_async/epics/adcore/_core_writer.py
@@ -19,7 +19,7 @@ from ophyd_async.core._signal import (
 )
 from ophyd_async.core._status import AsyncStatus
 from ophyd_async.core._utils import DEFAULT_TIMEOUT, error_if_none
-from ophyd_async.epics.core._util import stop_busy_record
+from ophyd_async.epics.core import stop_busy_record
 
 # from ophyd_async.epics.adcore._core_logic import ADBaseDatasetDescriber
 from ._core_io import (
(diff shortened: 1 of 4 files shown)


DominicOram (approved):
Great, thank you!

src/ophyd_async/epics/adcore/_core_writer.py
```diff
@@ -16,10 +16,10 @@
 from ophyd_async.core._signal import (
     observe_value,
     set_and_wait_for_value,
-    wait_for_value,
 )
 from ophyd_async.core._status import AsyncStatus
 from ophyd_async.core._utils import DEFAULT_TIMEOUT, error_if_none
+from ophyd_async.epics.core._util import stop_busy_record
```
shihab-dls, line 22: I've moved this util function to somewhere more generic for odin and core_writer to use; let me know if I should change this.

PR #913, opened 2025-06-10
Implement the NullPatternGenerator

rtuck99:
This PR implements a NullPatternGenerator.

See related PR
* DiamondLightSource/dodal#1270

The  motivation for this is that usage of `SimBlobDetector` in dodal unit tests was generating several GB of temporary files. This was traced back to the `PatternGenerator` which persists the output to a file. As the usage of the `SimBlobDetector` in the tests did not require examination of the detector output (only the ability to exercise it), a NullPatternGenerator was created and @DominicOram suggested it should be made part of ophyd-async.


DominicOram:
If @coretl doesn't think this is useful then I'm happy to just have it in dodal, apologies.

PR #916, opened 2025-06-11
729 make it easier to spot when slow tests are added

### LINKED ISSUE #729: Make it easier to spot when slow tests are added
We should make sure that all new tests are relatively fast, with a few smoke tests that are allowed to take a bit longer. I suggest a default per-test timeout of 0.5s, with an override for the few tests that take longer than this.

This would involve:
- Adding `pytest-timeout` as a dev dependency in `pyproject.toml`
- Setting in `timeout = 0.5` in `[tool.pytest.ini_options]` in `pyproject.toml`
- Adding `@pytest.mark.timeout(<a_longer_timeout>)` for those tests that now start to fail

As per https://pypi.org/project/pytest-timeout/

In a later PR we will make the slow tests faster

Relm-Arrowny:
fixes #729


coretl (approved):
Excellent, thanks!

PR #917, opened 2025-06-11
signal walking added for debugging

stan-dot:


tests/core/test_signal.py
```diff
@@ -1019,3 +1019,24 @@ def test_remove_non_existing_listener():
...
+    # Get all devices in the tree
+    devices = walk_devices(mock_readable)
+
+    # Should include all direct children and nested children
+    # Check for a few known paths
+    assert "int_value" in devices
+    assert "int_array" in devices
+    assert "float_array" in devices
+    assert "str_value" in devices
+    assert "strictEnum_value" in devices
```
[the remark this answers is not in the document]
stan-dot: updated

coretl (changes requested):
Can we have a test for walk_signal_sources too please? Coverage is complaining at the moment...

stan-dot:
sure!

tests/core/test_signal.py
```diff
@@ -1040,3 +1044,35 @@ async def test_walk_devices_returns_all_devices(mock_readable: DummyReadableArra
...
+        assert sources[path] == signal.source
+
+    # No extra keys
+    assert set(sources.keys()) == {
+        "int_value",
+        "int_array",
+        "float_array",
+        "str_value",
+        "strictEnum_value",
+    }
```
coretl, line 1079: The same comment as yesterday, please compare `sources` to an absolute dictionary of sources.

tests/core/test_signal.py
```diff
@@ -29,7 +29,11 @@
     wait_for_value,
 )
 from ophyd_async.core import StandardReadableFormat as Format
-from ophyd_async.core._signal import _SignalCache  # noqa: PLC2701
+from ophyd_async.core._signal import (
+    _SignalCache,  # noqa: PLC2701
+    walk_devices,  # noqa: PLC2701
+    walk_signal_sources,  # noqa: PLC2701
+)
```
coretl, line 36: Wait, I missed that these are not public. Please can you put `walk_devices` and `walk_signal_sources` in `src/ophyd_async/core/__init__.py` and then change the import here to get them from there.

coretl (changes requested):

tests/core/test_signal.py
stan-dot: what if I put them in `walk_devices`, its own file? I'd err on the side of not overpopulating the `__init__` file

coretl: The public API for ophyd-async is flat, everything in `ophyd_async.core`. This means we can move things around inside that without breaking people. I'd like to keep it that way.

stan-dot: done

coretl (approved):
Excellent, thanks!

PR #920, opened 2025-06-12
Modify SimMotor to use the same FlyMotorInfo as the EPICS motor, move…

jwlodek:
… FlyMotorInfo to core


jwlodek:
@coretl I think this is now what you had in mind?

coretl (approved):
Thanks, that's what I meant!

@oliwenmandiamond please could you do a quick review of the names of the fields of the FlyMotorInfo?

Issue #777, opened 2025-02-19
Make all `BaseModel`s have `extra="forbid"`


coretl:
Need this so people don't do `TriggerMode(live_time=0.2)` and be surprised that it didn't error but instead gave the default `livetime` of `None`...

A solution like https://github.com/epics-containers/ibek/blob/03817d9a963f5806ce795609acdd642b89d40e8e/src/ibek/globals.py#L160-L166 that everything derives from, but don't call it `BaseSettings` as that doesn't make sense for ophyd_async.

## Acceptance Criteria
- All BaseModels in the codebase derive from a common BaseModel with extra="forbid"

PR #925, opened 2025-06-16
Add `ConfinedModel` to explicitly defined fields in the model schema

hyperrealist:
This PR addresses #777. A "`ConfinedModel`" is derived from `BaseModel` that is configured to restrict fields to explicitly defined ones.


coretl:
Please can you have a look at the docs failure? If you run `tox -e docs` locally you should be able to reproduce it

Issue #923, opened 2025-06-16
Allow arbitrary number of OdinNodes in Odin


RJCD-Diamond:
Odin has the number of OdinNode hardcoded to 4. The Eiger on B21 only has 1 node, and this causes device to fail

```
        self.nodes = DeviceVector(
            {i: OdinNode(f"{prefix[:-1]}{i + 1}:") for i in range(4)}
        )
```

You should be able to specify the number of nodes ie.

```
self.odin = Odin(prefix + hdf_suffix, nodes=4)
```

Such that the EigerDetetctor can be passed the number of nodes of device creation

PR #926, opened 2025-06-16
Allow n nodes in Odin

RJCD-Diamond:
This sets the default number of nodes in Odin to 4, but allows an arbitrary number of nodes to be used. This will close out #923


RJCD-Diamond:
> Thanks @RJCD-Diamond. Some comments:
> 
> * We should instead be able to read the number of nodes out of `BL03I-EA-EIGER-01:OD:FAN:Consumers_RBV`, this will avoid us having to specify it
> * Can we have a unit test to confirm it will give different numbers of nodes when the PV changes?

We decided we were unable to do this without connecting to the PV first, which would cause additional issues. Given that this will be replaced again in the future  - we didn't do it

RJCD-Diamond:
> Sorry to add another comment, do we not need to add this to the constructor for `EigerDetector` too so that we can instantiate it in `b21.py` and pass it in at
> 
> https://github.com/bluesky/ophyd-async/blob/b28a3efc47b6d6dd3eef94b06b633261a5a58362/src/ophyd_async/fastcs/eiger/_eiger.py#L34
> 
> ?
> @coretl, I would be interested on how you would suggest creating a device vector with a length based on the value of another PV, given that the PV will have to connect before the vector is created? I think it's fine not to here given we're moving to `FastCS odin` but for future reference...

Yes, good spot.

PR #928, opened 2025-06-16
Remove EigerTriggerInfo

### LINKED ISSUE #727: Can we do `EigerTriggerInfo` via load/save instead?
It is defined here to be the standard `TriggerInfo`, but with `energy` too, so that it can be done all in one go.

https://github.com/bluesky/ophyd-async/blob/4f4458c916071859c44b03b2f39a6a9af05d4aac/src/ophyd_async/epics/eiger/_eiger.py#L10-L11

We should discuss with @DominicOram and @olliesilvester if we still need that approach after #697. This will depend on what exactly is slow in the Eiger.

shihab-dls:
Closes #727 

This PR waits for `stale_parameters` to go low after setting the number of triggers and exposure time. As we should set energy in a Dodal plan instead of in `EigerController.prepare()`, `EigerTriggerInfo` has been removed.

EDIT: This PR now only removes EigerTriggerInfo, as waiting for `stale_parameters` not required (see https://github.com/bluesky/ophyd-async/pull/928#discussion_r2154314614)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,8 @@ dev = [
     "import-linter",
     "myst-parser",
     "numpydoc",
-    "ophyd",
+    # Unpin once an ophyd release is made containing https://github.com/bluesky/ophyd/pull/1253
+    "ophyd @ git+https://github.com/tangkong/ophyd@10652e158b4f02172f4590f3f862ee815983c4db",
     "pickleshare",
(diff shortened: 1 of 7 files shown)


DominicOram (changes requested):
Thanks, happy with removing the energy but can we discuss the stale params again?

PR #940, opened 2025-06-23
Allow for overriding read/uri location for stream resource docs.

### LINKED ISSUE #966: HDF Document Composer is created before full file name PV updated in HDF writer, uses old path value.
Following the modifications to `HDFDocumentComposer`, when running against a real AD IOC, the resource doc URI is not generated correctly, since the full file name PV has not updated before it is used.

## Steps To Reproduce
1. Count an EPICS areaDetector driver w/ HDF writer
2. See that Resource URI generated is using old filepath

## Acceptance Criteria
- Correct filepath is used when generating resource docs.
- Modify AD tests to more closely match the behavior of IOCs to catch this issue in the future.

jwlodek:
fixes #966 
@coretl how is this for a starting point?


src/ophyd_async/core/_providers.py
```diff
@@ -21,6 +22,21 @@ class PathInfo:
...
+    directory_uri: str | None = None
+
+    def __post_init__(self):
+        # If directory uri is not set, set it using the directory path.
+        if self.directory_uri is None:
+            self.directory_uri = urlunparse(
+                (
+                    "file",
+                    "localhost",
+                    f"{self.directory_path}/",
```
coretl, line 34: do we need `.absolute()` on it?

src/ophyd_async/epics/adcore/_core_writer.py
```diff
@@ -152,30 +153,21 @@ async def get_indices_written(self) -> int:
     async def collect_stream_docs(
         self, name: str, indices_written: int
     ) -> AsyncIterator[StreamAsset]:
+        if self._path_info is None:
+            raise RuntimeError("Writer must be opened before collecting stream docs!")
```
coretl, line 157: https://github.com/bluesky/ophyd-async/pull/931 added a utility that would help here

coretl:
```suggestion
        path_info = error_if_none(self._path_info, "Writer must be opened before collecting stream docs!")
```

coretl (commented):
Looks like the right approach

thomashopkins32 (commented):
So if I understand right, we are not using `pathlib.Path` for the HDF writer anymore because AD IOCs always append posix ending, `/`, when we need `\\`?

I see how it was done in Ophyd here: https://github.com/bluesky/ophyd/blob/main/ophyd/areadetector/paths.py

Maybe the logic for checking the separator should be within a PathProvider rather than in the ADWriter class. That way when you get the path from the path provider it's always in the correct format.

src/ophyd_async/sim/_blob_detector_writer.py
```diff
@@ -26,14 +25,14 @@ def __init__(
...
-        self.path: Path | None = None
         self.composer: HDFDocumentComposer | None = None
         self.datasets: list[HDFDatasetDescription] = []
 
     async def open(self, name: str, exposures_per_event: int = 1) -> dict[str, DataKey]:
         path_info = self.path_provider(name)
-        self.path = path_info.directory_path / f"{path_info.filename}.h5"
-        self.pattern_generator.open_file(self.path, WIDTH, HEIGHT)
+        write_path = path_info.directory_path / f"{path_info.filename}.h5"
+        read_path_uri = f"{path_info.directory_uri}{path_info.filename}.h5"
```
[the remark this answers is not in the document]
thomashopkins32: Maybe `PathInfo` should have a `full_path` property that does this concat.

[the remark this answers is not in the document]
jwlodek: `directory_uri` includes the trailing `/` character, since for tiff/jpeg cases the uri doesn't include the filenames (since they use a pattern with sequentially incremented integer tail)

jwlodek:
I think you are right, some of this should be moved back into the providers. I'll work on that and request another review.

DiamondJoseph:
@coretl @jwlodek instead of an alpha can we have a 0.12.1 bug fix release? What other changes have gone in since 0.12.0?

PR #943, opened 2025-06-24
Add retries to signals

### LINKED ISSUE #884: Create the ability to retry signals
We have a number of signals (specifically on the zebra but we have also seen some on the smargon) that, when set, will timeout occasionally. Sending the set again works and they don't timeout. Originally this was solved at the plan level but led to further issues https://github.com/DiamondLightSource/mx-bluesky/issues/1031.

Instead we should have the ability to specify that a signal is expected to potentially timeout and that we should re-set it if it does so. For example we could do this with a particular signal type:

```python
class RetrySignal(SignalRW):
    def __init__(retry_times: int, ...):
        ...

    async def set(...):
        number_of_retries = 0
        while(number_of_retries < self.retry_times):
            try:
                await super().set(...)
            except Timeout:
                logger.warn(...)
                number_of_retries += 1
```

However, the disadvantage of this is that it will be cumbersome for a user of `ophyd_async` to make a `Motor` retry without making an entirely new `RetryMotor` class. Similarly with any other `Device`.

## Acceptance Criteria
- There is a way to specify signals so that when they are set they will retry N times if there is a timeout on setting them
- When there is a timeout this should be logged
- As a nice to have it would be good to be able to do this on the `Motor` or `Device` level too

shihab-dls:
Fixes #884 

This uses `stamina` (https://stamina.hynek.me/en/stable/) to retry setting to signals.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,6 @@ dev = [
     "tox-direct",
     "types-mock",
     "types-pyyaml",
-    "stamina>=23.1.0"
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ dev = [
     "tox-direct",
     "types-mock",
     "types-pyyaml",
+    "stamina>=23.1.0"
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "colorlog",
     "pydantic>=2.0",
     "pydantic-numpy",
+    "stamina>=23.1.0"
 ]
 dynamic = ["version"]
 license.file = "LICENSE"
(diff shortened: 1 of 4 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -89,9 +90,11 @@ def __init__(
         backend: SignalBackend[SignalDatatypeT],
         timeout: float | None = DEFAULT_TIMEOUT,
         name: str = "",
+        retries: int = 1,
```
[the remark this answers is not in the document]
DominicOram: I think `attempts` might be clearer

pyproject.toml
```diff
@@ -65,6 +66,7 @@ dev = [
     "tox-direct",
     "types-mock",
     "types-pyyaml",
+    "stamina>=23.1.0"
```
coretl, line 69: If it's in the standard deps then no need to add it to dev deps too

src/ophyd_async/core/_signal.py
```diff
@@ -287,7 +290,16 @@ async def set(
             timeout = self._timeout
         source = self._connector.backend.source(self.name, read=False)
         self.log.debug(f"Putting value {value} to backend at source {source}")
-        await _wait_for(self._connector.backend.put(value, wait=wait), timeout, source)
+        async for attempt in retry_context(
+            on=(TimeoutError, asyncio.exceptions.TimeoutError),
```
coretl, line 294:
```suggestion
            on=asyncio.TimeoutError,
```

coretl (approved):

PR #944, opened 2025-06-25
Add partial reading

oliwenmandiamond:
Added partial_reading function. Removes duplicate creation of dictionary with value as key e.g `{"value" : your_value}`
and conveniently does it in a function instead. Updated tests to use.

### DIFF
--- a/src/ophyd_async/testing/__init__.py
+++ b/src/ophyd_async/testing/__init__.py
@@ -11,6 +11,7 @@ from ._assert import (
     assert_emitted,
     assert_reading,
     assert_value,
+    partial_reading,
 )
 from ._mock_signal_utils import (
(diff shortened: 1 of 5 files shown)


DiamondJoseph:
I think when it was 

```python
await assert_reading(dev, {"dev-foo": value(1)})
```

It made slightly more sense as `value(x)` is shorter than `{"value": <x>}` but with the new name it isn't

coretl (commented):
Should we replace the imports in the test files with `import ophyd_async.testing.partial_reading as value` too?

oliwenmandiamond:
I personally think is fine to leave as is as I don't think your going to change many multi lines, if any, to fit on a signal line if change to `value`

coretl (approved):

PR #946, opened 2025-06-27
Introduce and use default reading testing function

Villtord:
Similar to partial_reading() in testing there are also quite a few tests that use following dictionary:
`{"alarm_severity": 0, "timestamp": ANY, "value": val}`

In this change new function introduced that will return that dictionary and can be used instead of typing in all key-values.


coretl:
I think the tests are wrong, anywhere you have `{"alarm_severity": 0, "timestamp": ANY, "value": val}` we can replace it with `partial_reading(val)`. The timestamp part is equivalent, as is the value, and I don't think the difference in alarm_severity checking warrants a new function.

PR #949, opened 2025-07-03
Convert to use uv

coretl:
As an experiment, the results will inform https://github.com/DiamondLightSource/python-copier-template/pull/248


coretl:
Closed in favour of #1058

PR #950, opened 2025-07-07
update to use new Subscribable protocol

jsouter:
Requires https://github.com/bluesky/bluesky/pull/1923, once that is approved we should pin bluesky to a release (or restrict it with <=) before merging that, then when that is released, update this pr to unpin bluesky and merge this.

We should also pin dodal to a release of ophyd-async before merging either of these, though I don't think it requires any changes in code content.


src/ophyd_async/epics/motor.py
```diff
@@ -228,9 +228,13 @@ async def locate(self) -> Location[float]:
...
-    def subscribe(self, function: Callback[dict[str, Reading[float]]]) -> None:
-        """Subscribe."""
-        self.user_readback.subscribe(function)
+    def subscribe_reading(self, function: Callback[dict[str, Reading[float]]]) -> None:
+        """Subscribe to reading."""
+        self.user_readback.subscribe_reading(function)
+
+    def subscribe_value(self, function: Callback[float]) -> None:
+        """Subscribe to value."""
+        self.user_readback.subscribe_value(function)
```
coretl, line 237: Don't think we want `subscribe_value` for anything other than a `Signal` do we? If suspenders and monitor both use `subscribe_reading` then there's not much point in adding this

coretl: You know, seeing that `Signal` always subscribes to the entire reading, then plucks out what the user wants:
https://github.com/bluesky/ophyd-async/blob/222e0f524630ff7dc7656822b7172763d8288ffe/src/ophyd_async/core/_signal.py#L154-L161

I'm wondering if we delete `subscribe_value` completely. Most usage is via `observe_value` and `wait_for_value` which can both be modified to work with `subscribe_reading`.

What do you think?

jsouter: I thought that we might as well use `subscribe_value` in suspenders if it exists in the Protocol as I don't think we're interested in any of the metadata, but `subscribe_reading` should work fine, initially my thought was that adding `subscribe_value` was a way to ensure that the protocol is broken for sync ophyd signals, but now that we're renaming `subscribe_reading` it's not strictly needed.

coretl: Re-reading https://github.com/bluesky/bluesky/issues/1915 I didn't make myself particularly clear. I think adding `subscribe_value` as well as `subscribe_reading` adds complexity for the implementer of the protocol, and as it is used in precisely one place at the moment (`ophyd_async.core.SignalR`) and in that place it is effectively taking the reading and throwing away everything that isn't the value, we might as well do that in the consumer of the protocol. Will discuss in the collab meeting.

coretl: Decided to delete `subscribe_value` form ophyd-async and make the code that depends on it call `subscribe_reading` and throw away everything that isn't value

tests/test_tutorials.py
```diff
@@ -44,5 +44,7 @@ def test_implementing_devices(module, capsys, expected_scan_output):
         RE(main.bp.grid_scan([main.pdet], main.stage.x, 1, 2, 3, main.stage.y, 2, 3, 3))
         assert time.monotonic() - start == pytest.approx(2.5, abs=2.0)
         captured = capsys.readouterr()
-        assert captured.err == ""
+        if captured.err:
+            # suppress warning
+            assert "Unsubscribe failed for signal" in captured.err
```
coretl, line 49: Why would unsubscribe fail?

coretl: Would you mind digging a bit more please? Maybe turn the warning into an error and see where it bubbles up? I don't think there should be any double unsubscribes in the test code so we should find out where it is...

jsouter: Hmm, somehow the "Remove want_value arg from SignalCache._notify` commit indirectly solved the issue, still, will try and see if I can find out the cause...

jsouter: Aha, okay, my "Remove want_value arg from SignalCache._notify" indirectly solved this issue.
`_listener = self._listeners.pop(function, None)` was evaluating to False or None (False if want_value=False, None if not in _listeners), both falsy, probably the line should have been `_listener = self._listeners.pop(function, None) is not None` and that error message always prints when subscribe/subscribe_reading is used.

src/ophyd_async/core/_signal.py
```diff
@@ -154,29 +154,28 @@ def _callback(self, reading: Reading[SignalDatatypeT]) -> None:
         )
         self._reading = reading
         self._valid.set()
-        items = self._listeners.copy().items()
-        for function, want_value in items:
-            self._notify(function, want_value)
+        callbacks = self._listeners.copy()
+        for callback in callbacks:
```
coretl, line 158:
```suggestion
        # Copy the listeners in case one of the callbacks removes the listener from the set
        for callback in list(self._listeners):
```

src/ophyd_async/core/_signal.py
```diff
@@ -154,29 +154,28 @@ def _callback(self, reading: Reading[SignalDatatypeT]) -> None:
...
+            self._notify(callback)
 
     def _notify(
         self,
         function: Callback[dict[str, Reading[SignalDatatypeT]] | SignalDatatypeT],
```
coretl, line 163:
```suggestion
        function: Callback[dict[str, Reading[SignalDatatypeT]]],
```

PR #965, opened 2025-07-16
Use real time for reading soft signals

### LINKED ISSUE #964: Use actual time for soft signal timestamp
We have a usecase where we want to compare the time a soft signal was set to the time an EPICS PV was changed. Currently, when reading a soft signal we just get a monotonic time, the value of which is meaningless (https://github.com/bluesky/ophyd-async/blame/864c2c2574750a9405fd695a1b217a25ded6557f/src/ophyd_async/core/_soft_signal_backend.py#L151). It would be nicer if we could get the system time (which obviously will have some inaccuracies but will be a better estimate)

## Acceptance Criteria
- Reading a soft signal gives a more sensible timestamp

DominicOram:
Fixes #964

To test:
* Confirm new test passes

### DIFF
--- a/src/ophyd_async/core/_soft_signal_backend.py
+++ b/src/ophyd_async/core/_soft_signal_backend.py
@@ -148,7 +148,7 @@ class SoftSignalBackend(SignalBackend[SignalDatatypeT]):
         """Set the current value, alarm and timestamp."""
         self.reading = Reading(
             value=self.converter.write_value(value),
-            timestamp=time.monotonic(),
+            timestamp=time.time(),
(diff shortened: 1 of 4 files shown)


coretl (approved):

Tom-Willemsen:
> Confirm new test passes

On windows it doesn't... https://github.com/bluesky/ophyd-async/actions/runs/16317882701/job/46090638432

This is because in python <= 3.12 `time.time()` on windows uses a low-resolution timer, this has resolution somewhere in the range 1-16ms, so a 10ms tolerance on the test is too low.

PR #968, opened 2025-07-17
Relaxed the time accuracy on tests for windows

### LINKED ISSUE #967: Fix windows test
> Confirm new test passes

On windows it doesn't... https://github.com/bluesky/ophyd-async/actions/runs/16317882701/job/46090638432

This is because in python <= 3.12 `time.time()` on windows uses a low-resolution timer, this has resolution somewhere in the range 1-16ms, so a 10ms tolerance on the test is too low.

_Originally posted by @Tom-Willemsen in https://github.com/bluesky/ophyd-async/issues/965#issuecomment-3083795781_

DominicOram:
Fixes #967

### DIFF
--- a/tests/epics/signal/test_signals.py
+++ b/tests/epics/signal/test_signals.py
@@ -861,4 +861,4 @@ async def test_signal_timestamp_is_same_format_as_soft_signal_timestamp(
     real_data = await real_sig.read()
     sim_data = await sim_sig.read()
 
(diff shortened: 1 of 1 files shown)


Tom-Willemsen (approved):
Thanks - I started looking at this this afternoon, but then got distracted with _other_ windows test failures as well (same ones that are showing up in CI). Let's get this in first though.

PR #972, opened 2025-07-22
fix!: Prevent timeout in stop_busy_record

DiamondJoseph:
Breaking change: removes the unused `status_timeout` field from `stop_busy_record` and adjusts uses of the method to fall through to the DEFAULT_TIMEOUT.

I am going to update the documentation in another change to formalise that DEFAULT_TIMEOUT should always be the default for a `timeout` argument, and should only be overwritten when passing `timeout=expected_time + DEFAULT+TIMEOUT` with expected_time > 1s. 

While `stop_busy_record` has been very fast in practice, when running in CI with a container that is running multiple composefiles worth of containers, it has been slow enough to time out and make the tests inconsistent.


DiamondJoseph:
I may leave `status_timeout` as an argument to speed up the adoption of this change and remove it again prior to 0.13.0?

PR #973, opened 2025-07-22
Change the way Axis IO is done

### LINKED ISSUE #953: Change the way Axis IO is done
- Make `PmacIO` take a `raw_motors: Sequence[motor.Motor]` instead of axes
- Rename `PmacAxisIO` to `PmacAxisAssigmentIO` and make it not derive from motor
- Instantiate one for each motor passed in and put it in a `.assignment` `DeviceVector` where the index is just an incrementing number, *not* the axis number
- Create a public lookup of Motor to axis assigment device vector index

## Acceptance Criteria
- Class created and tests updated

gkalua:
fixes #953 

Change the way Axis io is done by using raw_motrors instead of axes. 
Updates PmacAxisIO to PmacAxisAssignmentIO 
Creates a device vector containing PmacAxisAssignmentIO  instances and an incremental index
Creates a public look up of Motor to axis assignment device vector index
Tests the above mentioned changes

### DIFF
--- a/src/ophyd_async/epics/pmac/_pmac_io.py
+++ b/src/ophyd_async/epics/pmac/_pmac_io.py
@@ -14,39 +14,39 @@ class PmacTrajectoryIO(StandardReadable):
 
     def __init__(self, prefix: str, name: str = "") -> None:
         self.time_array = epics_signal_rw(
-            Array1D[np.float64], prefix + ":ProfileTimeArray"
+            Array1D[np.float64], prefix + "ProfileTimeArray"
         )
-        self.user_array = epics_signal_rw(Array1D[np.int32], prefix + ":UserArray")
+        self.user_array = epics_signal_rw(Array1D[np.int32], prefix + "UserArray")
         # 1 indexed CS axes so we can index into them from the compound motor input link
         self.positions = DeviceVector(
             {
                 i + 1: epics_signal_rw(
-                    Array1D[np.float64], f"{prefix}:{letter}:Positions"
+                    Array1D[np.float64], f"{prefix}{letter}:Positions"
                 )
                 for i, letter in enumerate(CS_LETTERS)
             }
         )
         self.use_axis = DeviceVector(
             {
-                i + 1: epics_signal_rw(bool, f"{prefix}:{letter}:UseAxis")
+                i + 1: epics_signal_rw(bool, f"{prefix}{letter}:UseAxis")
                 for i, letter in enumerate(CS_LETTERS)
             }
         )
         self.velocities = DeviceVector(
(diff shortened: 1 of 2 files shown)


LuisFSegalla (approved):
Looks pretty good to me!
I had a run through the logic and everything worked fine!

src/ophyd_async/epics/pmac/_pmac_io.py
```diff
@@ -84,13 +84,24 @@ class PmacIO(Device):
...
+            {
+                i: PmacAxisAssignmentIO(motor_prefix)
+                for i, motor_prefix in enumerate(motor_prefixes)
+            }
         )
+
+        # Public Look up for motor to axis assignment device vector index
+        self.motor_cs_index = {}
+        for i in range(len(self.assignment)):
+            self.motor_cs_index[raw_motors[i]] = i
```
coretl, line 103:
```suggestion
        self.motor_cs_index = {raw_motors[i]: i for i in range(len(self.assignment))}
```

src/ophyd_async/epics/pmac/_pmac_io.py
```diff
@@ -59,19 +60,18 @@ class PmacAxisIO(Device):
     def __init__(self, prefix: str, name: str = "") -> None:
         self.cs_axis_letter = epics_signal_r(str, f"{prefix}:CsAxis_RBV")
         self.cs_port = epics_signal_r(str, f"{prefix}:CsPort_RBV")
+        self.cs_number = epics_signal_r(int, f"{prefix}:CsRaw_RBV")
```
coretl, line 63: Just spotted this, the convention is that the prefix includes the trailing colon, so this (and all other signals in this file) should lose the colon:

```suggestion
        self.cs_number = epics_signal_r(int, f"{prefix}CsRaw_RBV")
```

tests/epics/pmac/test_pmac_io.py
```diff
@@ -8,16 +9,29 @@
...
     """Instantiate a PmacIO object that looks like the P47 training beamline"""
+    raw_motors = [
+        Motor("BL47P-MO-MAP-01:STAGE:X"),
+        Motor("BL47P-MO-MAP-01:STAGE:A"),
+    ]
 
     pmac = PmacIO(
-        axis_nums=[1, 2],
-        coord_nums=[1, 9],
         prefix="BL47P-MO-BRICK-01",
```
coretl, line 18: And this should be
```suggestion
        prefix="BL47P-MO-BRICK-01:",
```

coretl (changes requested):

PR #990, opened 2025-07-30
Add feature flag and ADR for `all_updates` change

callumforrester:
Add feature flag and ADR for changes in #962

### DIFF
--- a/src/ophyd_async/epics/core/_aioca.py
+++ b/src/ophyd_async/epics/core/_aioca.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import sys
 import typing
 from collections.abc import Mapping, Sequence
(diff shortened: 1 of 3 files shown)


src/ophyd_async/epics/core/_aioca.py
```diff
@@ -40,6 +41,10 @@
 logger = logging.getLogger("ophyd_async")
 
 
+def _default_all_updates() -> bool:
+    return os.environ.get("OPHYD_ASYNC_EPICS_CA_KEEP_ALL_UPDATES", "True") == "True"
```
callumforrester, line 45: I have no strong feelings about whether the environment variable should be ``DROP_UPDATES` or `KEEP_UPDATES`

coretl (approved):

Issue #992, opened 2025-07-31
Create common enums that devices can reuse


oliwenmandiamond:
Many devices use enums that would be common to a lot of devices, however many are redefining them locally e.g 

```Python
class AravisTriggerMode(StrictEnum):
    """GigEVision GenICAM standard TriggerMode."""

    ON = "On"
    """Use TriggerSource to trigger each frame"""

    OFF = "Off"
    """Just trigger as fast as you can"""
```

```Python
class VimbaOnOff(StrictEnum):
    """On/Off modes on the Vimba detector."""

    ON = "On"
    OFF = "Off"
```

This should be updated so there are common enums defined so any device can import to use them

## Acceptance Criteria
- File defining common enums

PR #993, opened 2025-07-31
Create common enums and update devices to use

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/992

### DIFF
--- a/src/ophyd_async/core/_enums.py
+++ b/src/ophyd_async/core/_enums.py
@@ -1,4 +1,4 @@
-from ophyd_async.core import StrictEnum
+from ophyd_async.core._utils import StrictEnum
 
 
 class OnState(StrictEnum):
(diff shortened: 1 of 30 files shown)


coretl (commented):
Polling @DiamondJoseph, @DominicOram and @jwlodek on naming questions

src/ophyd_async/core/_enums.py
callumforrester, line 1: Should: I don't think we should have the `Thing` and `ThingCapitalised` pattern, we should pick one of these and say it is the DLS standard, if your IOC does not meet this standard you should either:
1. Get Controls to change it or, if that's not possible
2. Define your own enum

Allowing 2. leads to more duplication but it also does not legitimise 2 near-identical standards, it makes a device following the "wrong" standard a piece of technical debt, to be fixed. The right "standard" could also have a docstring explaining this.

If picking one to be the standard, then go with the more commonly used (for example `EnabledStateCapitalised`).

coretl (approved):
Excellent, thanks

src/ophyd_async/core/_enums.py
DominicOram: See https://github.com/DiamondLightSource/dodal/issues/1416

PR #995, opened 2025-08-01
Check that there are no imports from core._*

coretl:


DiamondJoseph (approved):
SLF linting rules are enabled and these private imports should have been caught by `SLF001`, not sure why they weren't?

coretl:
> SLF linting rules are enabled and these private imports should have been caught by `SLF001`, not sure why they weren't?

`SLF` only checks for third party private imports. We're kinda treating `ophyd_async.core` as third party, but there doesn't seem to be a way to tell ruff about that...

Issue #989, opened 2025-07-30
BUG: observe_signals_value hanging until timeout


burkeds:
I have a mass spectrometer device with a `State` attribute that carries a `DevStateEnum`. State is a `SignalR` with no readable Format. The device is set with `support_events=True` so the backend processes events from the Tango device server, it does not poll.

I am running into an issue where `wait_for_value` fails to return but only for some values of `DevStateEnum`.

```python
await mass.trigger()
await wait_for_value(mass.State, "RUNNING", timeout=30)
await asyncio.sleep(1.0)
print("Waiting for idle")
await wait_for_value(mass.State, "ON", timeout=30)
await asyncio.sleep(1.0)
print("Done")
```

When I run the above test, it works as expected at first. `wait_for_value` returns when the state transitions to `Running`. However, when the state transitions to `ON`, I can see that the backend processes the event, pushes the new reading to the callback, but `observe_signals_value` does not try to match the new value. I added a bunch of debug messages to `ophyd-async.core._signal` to try and figure out what's going on. I also have some print statements on the backend of `tango_transport.AttributeProxy` to show the callback is being processed.

```
[DEBUG_SIGNAL] wait_for_value called for <ophyd_async.core._signal.SignalR object at 0x7f0df7690d50> match=RUNNING timeout=30
[DEBUG_SIGNAL] _ValueChecker: matcher=<function wait_for_value.<locals>.<lambda> at 0x7f0e1ce81a80> matcher_name='RUNNING'
[DEBUG_SIGNAL] wait_for_value called for <ophyd_async.core._signal.SignalR object at 0x7f0df7690d50> timeout=30
[DEBUG_SIGNAL] wait_for_value: waiting for 'RUNNING' in 30s
[DEBUG_SIGNAL] _wait_for_value called for <ophyd_async.core._signal.SignalR object at 0x7f0df7690d50>
[DEBUG_SIGNAL] [3] observe_signals_value: ENTER signals = (<ophyd_async.core._signal.SignalR object at 0x7f0df7690d50>,) timeout=None done_status=None done_timeout=None
[DEBUG_SIGNAL] [3] Subscribing to signal: <ophyd_async.core._signal.SignalR object at 0x7f0df7690d50>
BACKEND PROCESSING CHANGE EVENT:  {'value': 'RUNNING', 'timestamp': 1753875705.317937, 'alarm_severity': tango._tango.AttrQuality.ATTR_VALID}
[...]

PR #1001, opened 2025-08-07
Making tango event subscription asynchronous

burkeds:
Fixing issue with how pytango handled event loops when subscribing Tango events. Synchronous callbacks are deprecated in DeviceProxy.subscribe_event. Since subscribe_callback must by synchronous, a task is scheduled _subscribe_to_event which now handles the subscription asynchronously. This addresses issue #989. 

After making this change, all tests in test_tango_transport that used echo_device would fail when run as part of a file. These tests pass in isolation, the issue being a shared test context. ~~Running tests in forked mode as recommended by the tango documentation solves this. I am pretty sure that this will further break windows tests because forked mode is not supported on Windows~~
Moving all echo device tests to test_tango_signals solves the shared context issue without having to use forked tests.

This should also take care of the deprecation warnings seen in tango tests regarding _event_processor not being asynchronous.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,8 +7,8 @@
   "lockFileMaintenance": {
     "enabled": true
   },
-  // Disable github actions that are managed by python-copier-template
   "packageRules": [
+    // Disable github actions that are managed by python-copier-template
     {
       "matchPackageNames": [
         "actions/checkout",
@@ -21,6 +21,14 @@
         "github-actions"
       ],
       "enabled": false
+    },
+    // Disable python version as that is managed by python-copier-template
+    {
+      "matchManagers": [
+        "pyenv"
+      ],
+      "enabled": false
     }
+
   ]
 }


tests/tango/conftest.py
```diff
@@ -23,6 +23,11 @@
 T = TypeVar("T")
 
 
+def pytest_configure(config):
+    config.option.forked = True
+    config.option.timeout = 2
+
+
```
jsouter, line 30: Tests were previously forked, I believe the incentive to get around forking was because it's not supported in windows and it makes the tests run much slower. If there is a way to fix this without forking it would be great, but admittedly the tests have been quite flaky since that change (especially test_tango_signal_w iirc). 

If we do have to re-enable forking we should bypass subprocess_helper and run the MultiDeviceTestContext directly as the module scope of the tango_test_device fixture is effectively ignored when forking.

burkeds: @jsouter Ok so if I put all of "echo_device" tests from test_tango_transport into test_tango_signals that seems to resolve the failing tests. We don't need to use forked mode now.

Now it's just one epics sim test that's failing... and the windows tests failing and they just say "This job failed"

coretl:
> Synchronous callbacks are deprecated in DeviceProxy.subscribe_event

So does this mean that `DeviceProxy.subscribe_event` is an `async def`? The documentation doesn't make that clear...

If so, then I guess that this is the only way round it...

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -272,14 +284,9 @@ def subscribe_callback(self, callback: Callback | None):
...
-            if not self._eid:
-                self._eid = self._proxy.subscribe_event(
-                    self._name,
-                    EventType.CHANGE_EVENT,
-                    self._event_processor,
-                    green_mode=False,
-                )
+            asyncio.run_coroutine_threadsafe(
+                self._subscribe_to_event(), asyncio.get_running_loop()
+            )
```
coretl, line 289: I was a bit confused by this, as I've never seen this called from inside the event loop. Would this be clearer?
```suggestion
            # We are in a sync function, so have to run the async subscribe_to_event
            # in the background
            asyncio.create_task(self._subscribe_to_event())
```

burkeds:
It is unclear to me as well. If the callback is sync, subscribe_event doesn't need to be awaited but if the callback is async then it must be awaited.

src/ophyd_async/tango/core/_tango_transport.py
burkeds: I was using the run_coroutine_threadsafe version while I was debugging. Since we should be in the event loop here, create_task makes more sense. I've made the change as suggested.

burkeds:
> I'm not sure what the issue is here, but one of the tests references this:
> 
> https://github.com/bluesky/ophyd-async/blob/4edf2f10dec4c0f5d3279239a5ef5daf91e43d50/src/ophyd_async/tango/core/_tango_transport.py#L223-L248
> 
> If you never use a signal with `wait=False`, then it's fine to raise an exception to say this is not supported. The only reason we have it in EPICS is because some signals are used to both start a device (with `wait=True`), and stop a device (with `wait=False`). We cannot call stop with `wait=True` otherwise it will deadlock.

I am not sure what you mean here. Why do you say "If you never use a signal with `wait=False`". There could be wait=False signals at the developers discretion.

burkeds:
> > I am not sure what you mean here. Why do you say "If you never use a signal with `wait=False`". There could be wait=False signals at the developers discretion.
> 
> For EPICS there are 3 ways you can use a signal
> 
>     1. `await signal.set(wait=True)` will send the value and wait until the device has responded
> 
>     2. `status = signal.set(wait=True)` will send the value, and return a status that tells when the device has responded
> 
>     3. `status = signal.set(wait=False)` will send the value but tell the IOC to return as soon as it has got the value, not to wait for the device
> 
> 
> Most of the time, if the developer does not want to wait until the device has responded they will use 2. They only use 3 if they required two set operations to happen at the same time (not supported by the IOC).
> 
> If Tango developers only need 1 and 2 then you could say that 3 is unsupported for tango, to use 2 instead.

I think I understand. Tango devices can be asynchronous so concurrent operations are supported. Currently, it looks like Tango is using these a bit differently.

1. The first case waits until the device server operation returns.
2. The second case returns a status object marked done when the device server operation returns.
3. The third case returns a status object marked done which does not wait for the operation to return (fire and forget). We can then take advantage of concurrency in the device server.

In the first two cases, if there is an between the device and the device server, the signal raises a corba exception. We miss it in the third case.

Do you find this slight difference acceptable?

coretl:
From the Tango server side, does it see any difference between 2 and 3?

coretl:
Doesn't look like the tutorials are at fault, it looks like tango is trying to do some asyncio work after the teardown of the tango tests (during the tutorials test) and failing because the event loop is no longer active. Do you know if there are some connections in the Tango proxy that need to be explicitly closed before the event loop is closed?

PR #1003, opened 2025-08-07
If number of frames in a chunk is set to zero, set it to 1

### LINKED ISSUE #1002: First `StreamResource` of `SimDetector` has different shape which prevents insertion into Tiled
The first StreamResource created by a SimDetector when run in our system tests (i.e. right after starting up) has a different shape `(0, 1024, 1024)` than following ones `(1, 1024, 1024)` which causes issues when trying to insert data into Tiled:
```python
blueapi.client.event_bus.BlueskyStreamingError: Chunk size in all dimensions must be at least 1: chunk_shape=(0, 1024, 1024).
```

## Steps To Reproduce
1. $ git clone https://github.com/epics-containers/example-services
2. $ docker compose -f example-services/compose.yaml up bl01t-di-cam-01 bl01t-mo-sim-01 ca-gateway --detach
3. $ EPICS_CA_NAME_SERVERS=127.0.0.1:9064 EPICS_PVA_NAME_SERVERS=127.0.0.1:9075 ipython
4. Make a RunEngine, and a subscriber that prints the docs
5.  ```python
    with init_devices():
        adsim = SimDetector(
            "BL01T-DI-CAM-01:",
            path_provider=StaticPathProvider(lambda _: "adsim", Path("/tmp")),
            drv_suffix="DET:",
            fileio_suffix="HDF5:",
        )

    RE(count([adsim]))
```

6. Observe that the StreamResource has a shape of `[0, 1024, 1024]`
7. `RE(count([adsim]))`
8. Observe that the new StreamResource has a shape of `[1, 1024, 1024]`

## Acceptance Criteria
- The above gives a consistent shape of StreamResource

jwlodek:
Resolves #1002

### DIFF
--- a/src/ophyd_async/epics/adcore/_core_io.py
+++ b/src/ophyd_async/epics/adcore/_core_io.py
@@ -229,7 +229,7 @@ class NDFileHDFIO(NDFilePluginIO):
     swmr_mode: A[SignalRW[bool], PvSuffix.rbv("SWMRMode")]
     flush_now: A[SignalRW[bool], PvSuffix("FlushNow")]
     xml_file_name: A[SignalRW[str], PvSuffix.rbv("XMLFileName")]
-    num_frames_chunks: A[SignalR[int], PvSuffix("NumFramesChunks_RBV")]
+    num_frames_chunks: A[SignalRW[int], PvSuffix.rbv("NumFramesChunks")]
     chunk_size_auto: A[SignalRW[bool], PvSuffix.rbv("ChunkSizeAuto")]
     lazy_open: A[SignalRW[bool], PvSuffix.rbv("LazyOpen")]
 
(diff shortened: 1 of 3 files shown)


DiamondJoseph (approved):
I'll make a 0.13.1 after this is merged

jwlodek:
So basically with the HDF plugin, on a fresh bootup, the chunk shape parameters are all set to zero. If ChunkSizeAuto=Enabled, the plugin will chunk per-frame, w/ NumFramesChunks deciding how many frames per chunk. Since we read the frame size from the detector, not the plugin, we have that information before the plugin sees its first frame. As for the NumFramesChunks, if that hasn't been set (i.e. it is still 0), the driver will automatically set it to 1, but only after the first frame is received. The other potential solution would be to wait until after the first frame has been read to set this parameter (or, in the IOC, `dbpf` a reasonable value at IOC startup)

PR #1004, opened 2025-08-07
Create PmacTrajectoryTriggerLogic

### LINKED ISSUE #957: Make `PmacTrajectoryTriggerLogic
Something like:
```python
class PmacTrajectoryTriggerLogic(TriggerLogic):
    async def prepare():
        motor_info = PmacMotorInfo.from_motors(self.pmac, motors=slice.midpoints.keys())
        slice = Path(spec.calculate()).consume()
        run_up_positions, run_up_duration = calculate_run_up(slice, motor_info)
        trajectory = create_trajectory(slice, motor_info, run_up_duration)
        trajectory = Trajectory.from_slice(slice, motor_info, run_up_duration)
        await asyncio.gather(
            self._move_to_start(motor_info, run_up_positions),
            self._build_trajectory(trajectory)
        )

    async def _move_to_start(self, motor_info, run_up_positions):
        coord = self.pmac.coord[motor_info.cs_number]
        await coord.defer_moves.set(True)
        for motor, position in run_up_positions.items():
            cs_index = motor_info.motor_cs_index[motor]            
            await coord.cs_axis_setpoint[cs_index].set(position)
        await coord.defer_moves.set(False)        

    async def _build_trajectory(self, trajectory):
        # Take the trajectory structures, put to positions, velocities, user programs, and durations
        # then hit the build button
        pass

    async def kickoff(self):
        self.status = await self.pmac.trajectory.execute_profile.set(
            True, timeout=self.scantime + 1
        )

    async def stop(self):
        await self.pmac.trajectory.abort_profile.set(True)

    async def complete(self):
        await wait_for_value(
            self.pmac.trajectory.execute_profile, False, timeout=self.scantime + 11
        )
```

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

LuisFSegalla:
Fixes #957.

### DIFF
--- a/src/ophyd_async/epics/pmac/_pmac_io.py
+++ b/src/ophyd_async/epics/pmac/_pmac_io.py
@@ -72,7 +72,7 @@ class PmacCoordIO(Device):
         self.cs_port = epics_signal_r(str, f"{prefix}Port")
         self.cs_axis_setpoint = DeviceVector(
             {
-                i + 1: epics_signal_rw(np.float64, f"{prefix}M{i + 1}:DirectDemand")
+                i + 1: epics_signal_rw(float, f"{prefix}M{i + 1}:DirectDemand")
                 for i in range(len(CS_LETTERS))
             }
         )
@@ -93,7 +93,7 @@ class PmacIO(Device):
 
         self.assignment = DeviceVector(
             {
-                i: PmacAxisAssignmentIO(motor_prefix)
+                i: PmacAxisAssignmentIO(motor_prefix + ":")
                 for i, motor_prefix in enumerate(motor_prefixes)
             }
         )
--- a/src/ophyd_async/epics/pmac/_pmac_trajectory.py
+++ b/src/ophyd_async/epics/pmac/_pmac_trajectory.py
@@ -1,0 +1,96 @@
+import asyncio
+from dataclasses import dataclass
+
+import numpy as np
+from scanspec.core import Path
+from scanspec.specs import Spec
+
+from ophyd_async.core import FlyerController, wait_for_value
+from ophyd_async.epics.motor import Motor
+from ophyd_async.epics.pmac import PmacIO
+from ophyd_async.epics.pmac._utils import (
+    TICK_S,
+    _PmacMotorInfo,
+    _Trajectory,
+    calculate_ramp_position_and_duration,
+)
+
+
+@dataclass
+class PmacTriggerLogic:
+    spec: Spec[Motor]
+
+
+class PmacTrajectoryTriggerLogic(FlyerController[PmacTriggerLogic]):
+    def __init__(self, pmac: PmacIO) -> None:
+        self.pmac = pmac
+        self.scantime: float | None = None
+
+    async def prepare(self, value: PmacTriggerLogic):
+        _slice = Path(value.spec.calculate()).consume()
+        motor_info = await _PmacMotorInfo.from_motors(self.pmac, _slice.axes())
+        ramp_up_pos, ramp_up_time = calculate_ramp_position_and_duration(
+            _slice, motor_info, True
+        )
+        ramp_down_pos, ramp_down_time = calculate_ramp_position_and_duration(
+            _slice, motor_info, False
+        )
+        traj = _Trajectory.from_slice(_slice, ramp_up_time)
+        await asyncio.gather(
+            self._build_trajectory(traj, motor_info, ramp_down_pos, ramp_down_time),
+            self._move_to_start(motor_info, ramp_up_pos),
+        )
+        return await super().prepare(value)
+
+    async def kickoff(self):
+        if not self.scantime:
+            raise RuntimeError("Cannot kickoff. Must call prepare first.")
+        self.status = await self.pmac.trajectory.execute_profile.set(
+            True, timeout=self.scantime + 1
+        )
+
+    async def complete(self):
+        if not self.scantime:
+            raise RuntimeError("Cannot complete. Must call prepare first.")
+        await wait_for_value(
+            self.pmac.trajectory.execute_profile,
+            False,
+            timeout=self.scantime + 11,  # Why 11s?
+        )
+
+    async def stop(self):
+        await self.pmac.trajectory.abort_profile.set(True)
+
+    async def _build_trajectory(
+        self,
+        trajectory: _Trajectory,
+        motor_info: _PmacMotorInfo,
+        ramp_down_pos: dict[Motor, np.float64],
+        ramp_down_time: float,
+    ):
+        await self.pmac.trajectory.profile_cs_name.set(motor_info.cs_port)
+        await self.pmac.trajectory.points_to_build.set(len(trajectory.positions))
+        new_time = np.append(trajectory.durations, [int(ramp_down_time / TICK_S)])
+        self.scantime = np.sum(new_time)
+        await self.pmac.trajectory.time_array.set(new_time)
+        await self.pmac.trajectory.user_array.set(trajectory.user_programs)
+
+        for motor, number in motor_info.motor_cs_index.items():
+            new_pos = np.append(trajectory.positions[motor], [ramp_down_pos[motor]])
+            new_vel = np.append(trajectory.velocities[motor], [0])
+            await self.pmac.trajectory.use_axis[number + 1].set(True)
+            await self.pmac.trajectory.positions[number + 1].set(new_pos)
+            await self.pmac.trajectory.velocities[number + 1].set(new_vel)
+
+        await self.pmac.trajectory.calculate_velocities.set(False)
+        await self.pmac.trajectory.build_profile.set(True)
+
+    async def _move_to_start(
+        self, motor_info: _PmacMotorInfo, ramp_up_position: dict[Motor, np.float64]
+    ):
+        coord = self.pmac.coord[motor_info.cs_number]
+        await coord.defer_moves.set(True)
+        for motor, position in ramp_up_position.items():
+            cs_index = motor_info.motor_cs_index[motor]
+            await coord.cs_axis_setpoint[cs_index + 1].set(position)
+        await coord.defer_moves.set(False)
--- a/src/ophyd_async/epics/pmac/_pmac_trajectory.py
+++ b/src/ophyd_async/epics/pmac/_pmac_trajectory.py
@@ -8,6 +8,7 @@ from scanspec.specs import Spec
 from ophyd_async.core import FlyerController, wait_for_value
 from ophyd_async.epics.motor import Motor
 from ophyd_async.epics.pmac import PmacIO
+from ophyd_async.epics.pmac._pmac_io import CS_LETTERS
 from ophyd_async.epics.pmac._utils import (
     TICK_S,
     _PmacMotorInfo,
@@ -40,7 +41,6 @@ class PmacTrajectoryTriggerLogic(FlyerController[PmacTriggerLogic]):
             self._build_trajectory(traj, motor_info, ramp_down_pos, ramp_down_time),
             self._move_to_start(motor_info, ramp_up_pos),
         )
-        return await super().prepare(value)
 
     async def kickoff(self):
         if not self.scantime:
@@ -69,19 +69,24 @@ class PmacTrajectoryTriggerLogic(FlyerController[PmacTriggerLogic]):
         ramp_down_time: float,
     ):
         await self.pmac.trajectory.profile_cs_name.set(motor_info.cs_port)
-        await self.pmac.trajectory.points_to_build.set(len(trajectory.positions))
         new_time = np.append(trajectory.durations, [int(ramp_down_time / TICK_S)])
         self.scantime = np.sum(new_time)
         await self.pmac.trajectory.time_array.set(new_time)
         await self.pmac.trajectory.user_array.set(trajectory.user_programs)
 
+        # Unselect axes to later select only the ones that will be used in the scan
+        for axis in range(len(CS_LETTERS)):
+            await self.pmac.trajectory.use_axis[axis + 1].set(False)
+
+        size = 0
         for motor, number in motor_info.motor_cs_index.items():
             new_pos = np.append(trajectory.positions[motor], [ramp_down_pos[motor]])
             new_vel = np.append(trajectory.velocities[motor], [0])
             await self.pmac.trajectory.use_axis[number + 1].set(True)
             await self.pmac.trajectory.positions[number + 1].set(new_pos)
             await self.pmac.trajectory.velocities[number + 1].set(new_vel)
-
+            size += len(new_pos)
+        await self.pmac.trajectory.points_to_build.set(size)
         await self.pmac.trajectory.calculate_velocities.set(False)
         await self.pmac.trajectory.build_profile.set(True)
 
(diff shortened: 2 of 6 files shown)


src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,100 @@
...
+            size += len(new_pos)
+        await self.pmac.trajectory.points_to_build.set(size)
+        await self.pmac.trajectory.calculate_velocities.set(False)
+        await self.pmac.trajectory.build_profile.set(True)
+
+    async def _move_to_start(
+        self, motor_info: _PmacMotorInfo, ramp_up_position: dict[Motor, np.float64]
+    ):
+        coord = self.pmac.coord[motor_info.cs_number]
+        await coord.defer_moves.set(True)
```
LuisFSegalla, line 97: To add a bit of context of parts of the code I think we need to discuss:
I believe on `_move_to_start` we were supposed to use `coord.cs_axis_setpoint` instead of the MotorRecord motors. But I noticed while testing it in the test rigs that if we use the CS setpoint, when doing the set of the desired position it would wait forever and the value would never be passed down to the controller. Not sure if this is expected when using `DirectDemand` PVs as sometimes they need to wait until the motor reaches the desired position before returning.

src/ophyd_async/epics/pmac/_pmac_io.py
```diff
@@ -93,7 +93,7 @@ def __init__(
 
         self.assignment = DeviceVector(
             {
-                i: PmacAxisAssignmentIO(motor_prefix)
+                i: PmacAxisAssignmentIO(motor_prefix + ":")
```
LuisFSegalla, line 96: `motor_prefix` comes from the sequence of Motors that are passed to this class when creating it. Most of the time the PVs that are used to create the motor object will not have `:` at the end of it. So for this case I believe it's better if we add the `:` suffix at the time of creating of the `PmacAssigmentIO` objects.

src/ophyd_async/epics/pmac/_pmac_trajectory.py
coretl: Actually it looks like malcolm waits on completion of the move:
https://github.com/DiamondLightSource/pymalcolm/blob/264ec4a319092a545853f41f89ddf1d21da56504/malcolm/modules/pmac/parts/cspart.py#L68-L85

So probably need code like:
```python
# Deferred moves means that setpoints will not be acted on until we leave deferred mode
await coord.defer_moves.set(True)
# Wait for the value to be set, then keep the statuses that will tell move completion
coros = [set_and_wait_for_value(motor, pos, set_timeout=move_time, wait_for_set_completion=False) for motor, pos in ramp_up_position.items()]
statuses = await asyncio.gather(*coros)
# Actually start the move
await coord.defer_moves.set(False)
# Then wait for completion
await asyncio.gather(*statuses)
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+            True, timeout=self.scantime + 1
+        )
+
+    async def complete(self):
+        if not self.scantime:
+            raise RuntimeError("Cannot complete. Must call prepare first.")
+        await wait_for_value(
+            self.pmac.trajectory.execute_profile,
+            False,
+            timeout=self.scantime + 11,  # Why 11s?
```
coretl, line 58:
```suggestion
            timeout=self.scantime + DEFAULT_TIMEOUT,
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+    TICK_S,
+    _PmacMotorInfo,
+    _Trajectory,
+    calculate_ramp_position_and_duration,
+)
+
+
+@dataclass
+class PmacTriggerLogic:
+    spec: Spec[Motor]
```
coretl, line 22: As we don't have any other inputs I would delete this class...

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class PmacTrajectoryTriggerLogic(FlyerController[PmacTriggerLogic]):
+    def __init__(self, pmac: PmacIO) -> None:
+        self.pmac = pmac
+        self.scantime: float | None = None
+
+    async def prepare(self, value: PmacTriggerLogic):
```
coretl, line 30: ...and make prepare take a Spec:
```suggestion
    async def prepare(self, value: Spec[Motor]):
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+        _slice = Path(value.spec.calculate()).consume()
```
coretl, line 31: Nit: local variables are always private, so don't need the leading undescore
```suggestion
        slice = Path(value.spec.calculate()).consume()
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+    ):
+        await self.pmac.trajectory.profile_cs_name.set(motor_info.cs_port)
+        new_time = np.append(trajectory.durations, [int(ramp_down_time / TICK_S)])
+        self.scantime = np.sum(new_time)
+        await self.pmac.trajectory.time_array.set(new_time)
+        await self.pmac.trajectory.user_array.set(trajectory.user_programs)
+
+        # Unselect axes to later select only the ones that will be used in the scan
+        for axis in range(len(CS_LETTERS)):
+            await self.pmac.trajectory.use_axis[axis + 1].set(False)
```
coretl, line 79: This does unnecessary sets, instead it is better to:
```python
        use_axis = {axis + 1: False for axis in range(len(CS_LETTERS))}
        ...
        for motor, number in motor_info.motor_cs_index.items():
            ...
            use_axis[number + 1] = True
        coros = [self.pmac.trajectory.use_axis[number].set(use) for number, use in use_axis.items()]
        await asyncio.gather(*coros)       
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+
+    async def _build_trajectory(
+        self,
+        trajectory: _Trajectory,
+        motor_info: _PmacMotorInfo,
+        ramp_down_pos: dict[Motor, np.float64],
+        ramp_down_time: float,
+    ):
+        await self.pmac.trajectory.profile_cs_name.set(motor_info.cs_port)
+        new_time = np.append(trajectory.durations, [int(ramp_down_time / TICK_S)])
```
coretl, line 72: I think `TICK_S` is going to confuse us here, I suggest we keep everything in seconds until the last minute, so change `Trajectory` to be in seconds, then divide by `TICK_S` here.

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,111 @@
...
+        for motor, number in motor_info.motor_cs_index.items():
+            new_pos = np.append(trajectory.positions[motor], [ramp_down_pos[motor]])
+            new_vel = np.append(trajectory.velocities[motor], [0])
+            await self.pmac.trajectory.use_axis[number + 1].set(True)
+            await self.pmac.trajectory.positions[number + 1].set(new_pos)
+            await self.pmac.trajectory.velocities[number + 1].set(new_vel)
+            size += len(new_pos)
+        await self.pmac.trajectory.points_to_build.set(size)
+        await self.pmac.trajectory.calculate_velocities.set(False)
+        await self.pmac.trajectory.build_profile.set(True)
```
coretl, line 91: In fact, please could you consider what could be set in parallel, then gather them together and await them:
```python
# Setup the values for the build
coros = [
    self.pmac.trajectory.profile_cs_name.set(motor_info.cs_port),
    self.pmac.trajectory.points_to_build.set(size),
    self.pmac.trajectory.calculate_velocities.set(False),
    self.pmac.trajectory.time_array.set(new_time),
    self.pmac.trajectory.user_array.set(trajectory.user_programs)
] + [
    self.pmac.trajectory.use_axis[number].set(use) for number, use in use_axis.items()
] + ...
await asyncio.gather(*coros)
# Run the build
await self.pmac.trajectory.build_profile.set(True)

File: src/ophyd_async/epics/pmac/_pmac_trajectory.py [hunk 10]
```diff
@@ -0,0 +1,111 @@
...
+        await self.pmac.trajectory.user_array.set(trajectory.user_programs)
+
+        # Unselect axes to later select only the ones that will be used in the scan
+        for axis in range(len(CS_LETTERS)):
+            await self.pmac.trajectory.use_axis[axis + 1].set(False)
+
+        size = 0
+        for motor, number in motor_info.motor_cs_index.items():
+            new_pos = np.append(trajectory.positions[motor], [ramp_down_pos[motor]])
+            new_vel = np.append(trajectory.velocities[motor], [0])
```
coretl, line 84: Hmm, this seems a little messy, maybe we should add a `Trajectory.with_ramp_down(ramp_down_pos, ramp_down_time)` method that creates a `Trajectory` object with the ramp down appended

src/ophyd_async/epics/pmac/_utils.py
```diff
@@ -218,7 +218,7 @@ def calculate_ramp_position_and_duration(
         ]
         velocities[axis] = velocity
         ramp_times.append(abs(velocity) / motor_info.motor_acceleration_rate[axis])
-
+    ramp_times.append(0.002)  # Adding a 2ms ramp time as a min tournaround time
```
coretl, line 221: Can we make this a global please?

coretl (dismissed):

src/ophyd_async/epics/pmac/_pmac_trajectory.py
LuisFSegalla: I think this might create more confusion though. I think with the addition of the `with_ramp_down` method all the calculations of durations happen inside of the `_Trajectory` class and the `PmacTrajectoryTriggerLogic` class then handles all the times directly in us.

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,113 @@
...
+    async def prepare(self, value: Spec[Motor]):
+        slice = Path(value.calculate()).consume()
+        motor_info = await _PmacMotorInfo.from_motors(self.pmac, slice.axes())
+        ramp_up_pos, ramp_up_time = calculate_ramp_position_and_duration(
+            slice, motor_info, True
+        )
+        ramp_down_pos, ramp_down_time = calculate_ramp_position_and_duration(
+            slice, motor_info, False
+        )
+        traj = _Trajectory.from_slice(slice, ramp_up_time)
```
shihab-dls, line 33: Nit: I would rename this cleanliness.
```suggestion
        trajectory = _Trajectory.from_slice(slice, ramp_up_time)
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
shihab-dls: I think the instance method approach is fine. Since we can only create a `_Trajectory` instance by passing in positions etc.., or using the class method `from_slice`, and we get the ramp info separately, using an instance method with the parsed/set instance variables seems okay to me.

src/ophyd_async/epics/pmac/_utils.py
```diff
@@ -104,6 +105,28 @@ def from_slice(cls, slice: Slice[Motor], ramp_up_time: float) -> _Trajectory:
...
+        new_vel = {
+            motor: np.append(self.velocities[motor], [ramp_down_velocity])
+            for motor in ramp_down_pos.keys()
+        }
+        return _Trajectory(
+            positions=new_pos,
+            velocities=new_vel,
+            user_programs=self.user_programs,
+            durations=new_time,
+        )
```
shihab-dls, line 128: Should: Since this is an instance method, and we are mutating the instance variables with the new ramp down info, we could just return self:

```suggestion
        return self
```

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -0,0 +1,117 @@
...
+        await self.pmac.trajectory.abort_profile.set(True)
+
+    async def _build_trajectory(
+        self,
+        trajectory: _Trajectory,
+        motor_info: _PmacMotorInfo,
+        ramp_down_pos: dict[Motor, np.float64],
+        ramp_down_time: float,
+    ):
+        trajectory = trajectory.with_ramp_down(ramp_down_pos, ramp_down_time, 0)
```
shihab-dls, line 71: Should: Given you change this method to return self, the method name should probably be a verb, like:
```suggestion
        trajectory = trajectory.append_ramp_down(ramp_down_pos, ramp_down_time, 0)
```

LuisFSegalla: I think that's a good point and the name looks better as `append_ramp_down`!

src/ophyd_async/epics/pmac/_utils.py
LuisFSegalla: That makes sense! And the code looks nicer with your suggestion

tests/epics/pmac/test_pmac_trajectory.py
```diff
@@ -0,0 +1,158 @@
...
+    spec = Fly(2.0 @ Line(sim_x_motor, 1, 5, 2))
+    pmac_trajectory = PmacTrajectoryTriggerLogic(pmacIO)
+    await pmac_trajectory.prepare(spec)
+
+    assert await pmacIO.coord[1].cs_axis_setpoint[7].get_value() == -1.2
+
+    assert await pmacIO.trajectory.positions[7].get_value() == pytest.approx(
+        [-1.0, 1.0, 3.0, 5.0, 7.0, 7.2]
+    )
+    assert pmac_trajectory.scantime == 4400000
```
shihab-dls, line 59: Should: Given we don't use sim_y_motor here, but the fixture returns it, I would use an underscore for clarity:

```suggestion

async def test_pmac_prepare(sim_motors: tuple[PmacIO, Motor, Motor]):
    pmacIO, sim_x_motor, _ = sim_motors
    spec = Fly(2.0 @ Line(sim_x_motor, 1, 5, 2))
    pmac_trajectory = PmacTrajectoryTriggerLogic(pmacIO)
    await pmac_trajectory.prepare(spec)

    assert await pmacIO.coord[1].cs_axis_setpoint[7].get_value() == -1.2

    assert await pmacIO.trajectory.positions[7].get_value() == pytest.approx(
        [-1.0, 1.0, 3.0, 5.0, 7.0, 7.2]
    )
    assert pmac_trajectory.scantime == 4400000
```
Similarily with other trajectory tests

LuisFSegalla: Done!

shihab-dls:
Feel free to squash and merge when you're happy with this too @LuisFSegalla.

PR #1027, opened 2025-08-27
creates a PandA Flyer for time based triggers

### LINKED ISSUE #448: Create a PandA flyer that uses a SEQ table to do time based triggers within a position based gate
Want something similar to https://github.com/bluesky/ophyd-async/blob/5474cf4210c15994854fc29f77a2842212b1c1ea/src/ophyd_async/panda/_trigger.py#L15-L39 but for a SEQ table implementing the following scheme:

image

## prepare
Make a dataclass `MresOffset` that includes the Mres and Offset got from the relevant PandA input.
Take a `List[scanspec.Frames[MresOffset]]`, calculate the entire trajectory (this will later be only the first 4000 points), find the gaps, in each gap insert:
- Wait for BITA=1 for GPIO
- Wait for POSA to exceed the value of the first compare point
- Generate N triggers for the row

At the moment we should fail if there are more than 4096 rows generated

## kickoff
Set `ENABLE=ONE` and wait for `ACTIVE`

## complete
Wait for `ACTIVE` to drop

gkalua:
resolves #448


src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -22,6 +26,11 @@ class SeqTableInfo(ConfinedModel):
     prescale_as_us: float = Field(default=1, ge=0)  # microseconds
 
 
+class ScanSpecInfo(BaseModel):
```
LuisFSegalla, line 29: I believe in here we could make `ScanSpecInfo` inherit from `ConfinedModel`? I'm not entirely sure what's the difference between them, but other `Info` type classes inherit from the `ConfinedModel`.

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +60,93 @@ async def stop(self):
...
+        await self.seq.enable.set(PandaBitMux.ZERO)
+        slice = Path(value.spec.calculate()).consume()
+
+        gaps = self._calculate_gaps(slice)
+        if gaps[0] == 0:
+            gaps = np.delete(gaps, 0)
+        scan_size = len(slice)
+
+        gaps = np.append(gaps, scan_size)
+        fast_axis = slice.axes()[len(slice.axes()) - 2]
```
LuisFSegalla, line 78: I believe the code originally was using `len(slice.axes()) - 2` because at the time the last axis would always be the `DURATION`. As we're not using that anymore I believe it can be done like this:
```suggestion
        fast_axis = slice.axes()[len(slice.axes()) - 1]
```

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +60,93 @@ async def stop(self):
...
+                dir = False if resolution > 0 else True
+
+            else:
+                trig = SeqTrigger.POSA_LT
+                dir = True if resolution > 0 else False
+            rows += SeqTable.row(
+                trigger=trig,
+                position=int(
+                    slice.lower[fast_axis][start]
+                    / await fast_axis.encoder_res.get_value()
```
LuisFSegalla, line 104:
```suggestion
                position=int(
                    slice.lower[fast_axis][start]
                    / resolution
```

src/ophyd_async/fastcs/panda/_trigger.py
gkalua: Yes, you're right. ConfinedModel restricts the class I think. Done.

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +60,93 @@ async def stop(self):
...
+                ),
+            )
+
+            # Time based Triggers
+            rows += SeqTable.row(
+                repeats=gap - start,
+                trigger=SeqTrigger.IMMEDIATE,
+                time1=(slice.duration[0] - value.deadtime) * 10**6,  # type: ignore
```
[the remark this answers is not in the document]
gkalua: Done

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +60,93 @@ async def stop(self):
...
+    async def prepare(self, value: ScanSpecInfo):
+        await self.seq.enable.set(PandaBitMux.ZERO)
+        slice = Path(value.spec.calculate()).consume()
+
+        gaps = self._calculate_gaps(slice)
+        if gaps[0] == 0:
+            gaps = np.delete(gaps, 0)
+        scan_size = len(slice)
+
+        gaps = np.append(gaps, scan_size)
```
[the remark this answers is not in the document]
gkalua: Addressed this

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +60,93 @@ async def stop(self):
...
+    async def stop(self):
+        await self.seq.enable.set(PandaBitMux.ZERO)
+        await wait_for_value(self.seq.active, False, timeout=1)
+
+    def _calculate_gaps(self, slice: Slice[Motor]):
+        inds = np.argwhere(slice.gap)
+        if len(inds) == 0:
+            return [len(slice)]
+        else:
+            return inds
```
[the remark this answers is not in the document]
gkalua: Addressed this

tests/fastcs/panda/test_trigger.py
```diff
@@ -51,6 +55,53 @@ async def set_active(value: bool):
...
+        SeqTrigger.BITA_1,
+        SeqTrigger.POSA_LT,
+        SeqTrigger.IMMEDIATE,
+        SeqTrigger.BITA_0,
+        SeqTrigger.BITA_1,
+        SeqTrigger.POSA_LT,
+        SeqTrigger.IMMEDIATE,
+        SeqTrigger.BITA_0,
+    ]
+    assert (out.position == [0, 0, 5, 0, 0, 0, 7, 0, 0, 0, 10, 0, 0]).all()
```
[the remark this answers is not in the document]
gkalua: No, you were right. it was as a result of this line.
`fast_axis = slice.axes()[len(slice.axes()) - 2]`

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -68,14 +73,16 @@ def __init__(self, seq: SeqBlock, name="") -> None:
     async def prepare(self, value: ScanSpecInfo):
         await self.seq.enable.set(PandaBitMux.ZERO)
         slice = Path(value.spec.calculate()).consume()
+        slice_duration = error_if_none(slice.duration, "Slice must have duration")
 
-        gaps = self._calculate_gaps(slice)
+        # gaps = self._calculate_gaps(slice)
```
LuisFSegalla, line 78: We can remove this commented line
```suggestion
```

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -68,14 +73,16 @@ def __init__(self, seq: SeqBlock, name="") -> None:
     async def prepare(self, value: ScanSpecInfo):
         await self.seq.enable.set(PandaBitMux.ZERO)
         slice = Path(value.spec.calculate()).consume()
+        slice_duration = error_if_none(slice.duration, "Slice must have duration")
```
LuisFSegalla, line 76: Might be a bit silly, but can you write a test for the case where the calculated Path actually has no duration so we can confirm we're catching this error?

coretl:
Please can we make it so:
- If we pass in a motor mapping, then it checks the fastest axis is in that map, and inserts a compare point after waiting for the GPIO trigger
- If we don't pass in a motor mapping, then just put the waiting for GPIO trigger in, no compare point

LuisFSegalla (changes requested):
The code looks pretty good and I think it's ready to be tested against real hardware whenever is convenient to you.

src/ophyd_async/epics/motor.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, prefix: str, name="") -> None:
         # we must write 1 to stop the motor. Simply processing the record is not
         # sufficient.
         self.motor_stop = epics_signal_w(int, prefix + ".STOP")
+        self.encoder_res = epics_signal_rw(float, prefix + ".ERES")
```
coretl, line 91: Don't need this line now

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +70,110 @@ async def stop(self):
...
+    offset: SignalRW[float]
+
+    @classmethod
+    def from_inenc(cls, panda: CommonPandaBlocks, number: int) -> PosOutScaleOffset:
+        inenc = panda.inenc[number]  # type: ignore
+        return cls(
+            name=f"INENC{number}.VAL",
+            scale=inenc.val_scale,  # type: ignore
+            offset=inenc.val_offset,  # type: ignore
+        )
```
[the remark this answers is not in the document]
coretl: I think we want to defer the get as late as possible, and also we want to be able to construct these classes without being in an async function

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +70,110 @@ async def stop(self):
...
+        if not self.motor_pos_outs.get(fast_axis):
+            raise RuntimeError("Failed to fetch motor")
+        else:
+            motor_pos_outs = self.motor_pos_outs
+            _, scale, _ = (
+                motor_pos_outs[fast_axis].name,
+                await motor_pos_outs[fast_axis].scale.get_value(),
+                await motor_pos_outs[fast_axis].offset.get_value(),
+            )
+            resolution = scale
```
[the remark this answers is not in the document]
coretl: We need offset below, so I would do:
```suggestion
            pos_out = self.motor_pos_outs[fast_axis]
            scale, offset = await asyncio.gather(
                pos_out.scale.get_value(),
                pos_out.offset.get_value(),
            )
```

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +70,110 @@ async def stop(self):
...
+            if (
+                slice.midpoints[fast_axis][gap - 1] * resolution
+                > slice.midpoints[fast_axis][start] * resolution
+            ):
+                trig = SeqTrigger.POSA_GT
+                dir = False if resolution > 0 else True
+
+            else:
+                trig = SeqTrigger.POSA_LT
+                dir = True if resolution > 0 else False
```
coretl, line 136: I think we should use the lower->midpoint to determine the direction, and if they are equal we can't insert a trigger position:
```suggestion
            lower = (slice.lower[fast_axis][start] - offset) / scale
            midpoint = (slice.midpoint[fast_axis][start] - offset) / scale            
            if midpoint > lower:
                trig = SeqTrigger.POSA_GT
            elif midpoint < lower:
                trig = SeqTrigger.POSA_LT
            else:
                trig = None
```

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +70,110 @@ async def stop(self):
...
+            rows += SeqTable.row(
+                trigger=trig,
+                position=int(slice.lower[fast_axis][start] / resolution),
+            )
```
coretl, line 140: The guard this one:
```suggestion
            if trig is not None:
                rows += SeqTable.row(
                    trigger=trig,
                    position=int(lower),
            )
```

src/ophyd_async/fastcs/panda/_trigger.py
```diff
@@ -51,6 +70,110 @@ async def stop(self):
...
+                outa2=False,
+                outb2=dir,
+            )
+
+            # GPIO goes low
+            rows += SeqTable.row(trigger=SeqTrigger.BITA_0)
+
+            start = gap
+        await asyncio.gather(
+            self.seq.prescale.set(1.0),
```
coretl, line 159: Also add:
```suggestion
            self.seq.posa.set(pos_out.name),
            self.seq.prescale.set(1.0),
```

coretl (changes requested):

coretl (approved):
Please could you fix the tests and the docs build, then this is good to merge

For the tests, I think you need to add the "posA" field to the test structures

For the docs, you need to put `PandaPosMux` in `fastcs/panda/__init__.py` and also put a docstring on `TableSubclass`

PR #1028, opened 2025-08-27
Raise exception if attr name is a bluesky protocol

### LINKED ISSUE #1010: Raise warning when device includes a `stop` signal
Many IOCs have a `stop` PV that will halt it. Naively, people will implement these as a signal called `stop` e.g. https://github.com/DiamondLightSource/dodal/blob/295ab88a410d66e3cff9edd399600e494fc8dc7b/src/dodal/devices/pressure_jump_cell.py#L210

However, this can cause issues as devices should implement a `stop` method instead. This is not obvious until the RunEngine tries to call `stop()` on the device and errors. It would be more user friendly if there was a warning (exception?) when instantiating the device that this can cause issues and to change it. We may want this for other keywords like `set` too.

## Acceptance Criteria
- When I instantiate a device with a signal called `stop` I am warned against doing this

shihab-dls:
fixes #1010

This PR collects the abstract methods from [bluesky](https://github.com/bluesky/bluesky/blob/main/src/bluesky/protocols.py), and raises an exception if a `Device` attr name is set to any of these.


coretl (changes requested):

coretl:
The PVI and Tango device fillers also need to be updated to put a trailing underscore in if the name is one of `DEVICE_RESERVED_ATTRS`

shihab-dls:
> The PVI and Tango device fillers also need to be updated to put a trailing underscore in if the name is one of `DEVICE_RESERVED_ATTRS`

I've put this in base `DeviceFiller` so that it's common between Tango and PVI

src/ophyd_async/core/_device_filler.py
```diff
@@ -465,6 +469,14 @@ def fill_child_signal(
             # We need to add a new child to the top level Device
             backend = self._signal_backend_factory(None)
             expected_signal_type = signal_type
+            # Prevent shadowing bluesky protocol with new signal by
+            # adding a trailing underscore
+            if name in DEVICE_RESERVED_ATTRS:
+                logger.warning(
+                    f"Signal `{name}` is used in one of the bluesky protocols. "
+                    f"Using `{name}_` instead."
+                )
```
coretl: Can you turn this into an info instead please?

coretl (changes requested):

coretl (approved):

Issue #933, opened 2025-06-17
Make set on a Motor fail if the value is beyound the motor limits


noemifrisina:
As a user, I would like an error to be raised on `set` if the requested position is beyond the motor limits. Epics does not allow the motors to move in this case, but calling a plan such as:

```python

def move_detector_stage(motor: Motor, value: float):
     yield from bps.mv(motor, value)
```

will return as `plan succeed` even if the motor hasn't moved - with no indication that something went wrong.

Recent use case: users on I24 have requested a detector distance during collection lower than the low limit on the motor - which did not move but the plan still showed as passed allowing the collection to proceed with the wrong set up.

Proposed solution would be to check the limit violation PV (eg. `BL24I-EA-DET-01:Z.LVIO`) at the device level and raise a `MotorLimitsException` if the signal goes high.

## Acceptance Criteria
- There's a new signalR for the soft limit violation PV 
- A MotorLimitsException is raised when limit violation goes high

coretl:
We should also consider the following conditions that would cause a motor to not get to its final destination:
1. Following error (causes `.MSTA` problem bit to be 1)
2. Someone else presses stop (causes `.RBV` to be nowhere near our expected value)

Does condition 2 mean that we need to think about checking `.RBV` closeness to `.VAL` (within `.RDBD`) and raising an error if they are outside? Or is that too invasive a change?

@jwlodek what do you think?

kivel:
Using the `.RDBD` is _not_ a reliable means of deducing whether the positioner is close to the target position.
Certain motion controller drivers require `.RDBD` to be Zero.

kivel:
To us (CLS) the current implementation is sufficient for now. But I concur that a more advanced mechanism should be offered to deduce a successful motion.
@jwlodek: I particularly like the idea of having a callable passed in. Rather than returning a `boolean`, it should return a status word, as multiple conditions can be true. For example the drive can be on a limit switch, which is a warning from a motion control point of view, but the target was reached successfully. Even with an an error, the motion can still be considered a success, only subsequent steps will suffer from the failure.
On that note, would it be possible to read a status word from the motion controller (additional signal), rather than using the sometimes crippled `.MSTA` from the motor record? Optional of course.
I can see infinite possibilities how this can fail, though.

coretl:
[...]

PR #1032, opened 2025-09-03
Add limit checking to motor.set

teoching0705:
# Purpose
This MR want to address #933 by adding a check on if the movement have exceeded the low_limit_travel or the high_limit_travel.

## Changes
The function 'check_motor_limit' basically mimics the original checking logic in prepare, just move it out and make it a separate function for it to be reused in set.

This MR also adds the mock values for low_limit_travel and high_limit_travel in test_motor.py for testing


DominicOram (approved):
Looks good to me, I'm not sure if someone from controls should review though @coretl or @shihab-dls maybe?

coretl (approved):

Issue #1037, opened 2025-09-04
Discuss requirements for deadtime in StandardDetector


olliesilvester:
The way to specify deadtime for scans using `StandardDetector` needs a bit of discussion.

- How do we handle deadtime for multiple frames per trigger VS deadtime between triggers?
- Should you require deadtime for externally triggered scans with one frame per trigger? Currently ophyd-async enforces that you do. However, in MX, we often don't want to specify this because the minimum deadtime of a detector is negligable compared to the exposure time
- How should we handle the case where we only know the deadtime of a detector during run-time?

@coretl has suggested a discussion with people with specific requirements here. @RJCD-Diamond I think you may have some more specific requirements on this topic?

coretl:
Having just re-read https://github.com/bluesky/ophyd-async/issues/308#issuecomment-3132579189 I wonder if the `read_configuration` solution has some merit, I will give this some more thought...

PR #1038, opened 2025-09-04
Add deadtime to create jungfrau external trigger info util function

olliesilvester:
ophyd-async currently enforces that you specify a deadtime when doing externally triggered scans, so the utility function to create jungfrau trigger info must include this.

Note that for the current Jungfrau use-cases we always want to use the minimum deadtime, so long-term it may not be a required parameter. See https://github.com/bluesky/ophyd-async/issues/1037

### DIFF
--- a/src/ophyd_async/fastcs/jungfrau/_utils.py
+++ b/src/ophyd_async/fastcs/jungfrau/_utils.py
@@ -2,10 +2,13 @@ from pydantic import PositiveInt
 
 from ophyd_async.core import DetectorTrigger, TriggerInfo
 
+from ._controller import JUNGFRAU_DEADTIME_S
+
 
 def create_jungfrau_external_triggering_info(
     total_triggers: PositiveInt,
     exposure_time_s: float,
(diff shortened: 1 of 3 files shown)


src/ophyd_async/fastcs/jungfrau/_utils.py
```diff
@@ -16,6 +19,7 @@ def create_jungfrau_external_triggering_info(
     Args:
         total_triggers: Total external triggers expected before ending acquisition.
         exposure_time_s: How long to expose the detector for each of its frames.
+        deadtime_s: Time between exposures
```
DominicOram, line 22: Should: I feel like deadtime doesn't really make sense to a user in the case we're using it here. I get that we're discussing further but in the meantime I would either:
1. Not let the user specify it in `create_jungfrau_external_triggering_info` at all
2. Expand this docstring to be more detailed about why it's needed

DominicOram (changes requested):
Sorry, one comment and this was meant to be the easy jungfrau issue!

PR #1050, opened 2025-09-17
Fix pedestal mode in Jungfrau

olliesilvester:
Fixes bugs spotted when running the Jungfrau in pedestal mode, see https://github.com/DiamondLightSource/mx-bluesky/issues/1266
Specifically:
- When the Jungfrau is triggered in pedestal mode, the total number of events is actually pedestal loops * 2, since it runs the entire scan for two different gain modes. I updated trigger info and `prepare` to account for this
- In pedestal mode, the Jungfrau still needs to be in internal trigger mode, but the settings cannot be changed once pedestal mode has been turned on. I updated `prepare` to make sure these are set in the right order.

### DIFF
--- a/src/ophyd_async/fastcs/jungfrau/_controller.py
+++ b/src/ophyd_async/fastcs/jungfrau/_controller.py
@@ -113,19 +113,33 @@ class JungfrauController(DetectorController):
                     ]
                 )
             case AcquisitionType.PEDESTAL:
-                coros.extend(
-                    [
-                        self._driver.pedestal_mode_frames.set(
-                            trigger_info.exposures_per_event
-                        ),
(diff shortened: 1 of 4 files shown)


DominicOram (approved):
This looks good, thanks. However, I disagree that it fixes:
>Changing trigger mode in pedestal looked happy but it wasn't and didn't give an error on the ophyd-async side. Looking on the IOC the readback and setpoint for trigger mode didn't match but no error bubbled: 

I think that we need to fix the underlying issue of not realising that the trigger was being set in error. I'm concerned that there may be other cases the trigger doesn't get set and we're missing them or that it might be true for other PVs too. Happy for you to spin into a new issue though.

Issue #1019, opened 2025-08-20
Separate out tests from system tests


coretl:
At the moment there are 2 classes of change to ophyd-async:
1. Those that can be tested purely with mocks
2. Those that need some kind of external process to communicate with

In development we will typically run 1 repeatedly, and 2 rarely. We can also partition them so that different runners pick them up in CI, and potentially limit the number of OS and python versions we test for.

This ticket is to find all those tests that have a fixture that is an external process and move it to the equivalent directory in `system_tests`.

I would expect at least:
- `system_tests/test_tutorials.py`
- `system_tests/epics/core`
- `system_tests/tango/core`
- `system_tests/fastcs/panda`

then all tests that need a fixture that starts a subprocess to be moved there.

Then a new input added to the `_test.yml` workflow to specify the directory to test:
https://github.com/bluesky/ophyd-async/blob/228f25c8badf4d9012862a7a11985d9e85b9e4ae/.github/workflows/_test.yml#L3

And an extra entry added in the test matrix the following section added in the test matrix to either pass `src tests` or `system_tests` as the directories to run:
https://github.com/bluesky/ophyd-async/blob/228f25c8badf4d9012862a7a11985d9e85b9e4ae/.github/workflows/ci.yml#L18-L33

The eiger system tests won't run on CI so will need to be `@pytest.mark.skip`ped. 

## Acceptance Criteria
- Tests that require a subprocess moved to `system_tests` and CI updated to run them

coretl:
Ok, I suggest a middle ground then:
- For each directory in `system_tests`:
  - Have a shell script `external_dependencies.sh` that is meant to run as a subprocess in the foreground, and terminate gracefully on CTRL-C
  - This script should have comments in it to explain what it is doing, and the additional args that can be passed to do things like start the phoebus screens
  - Have a pytest module level fixture that runs this `external_dependencies.sh` in the presence of a `START_EXTERNAL_DEPS=YES` environment variable
- The CI is as above, but setting this environment variable
- Locally we have instructions that tells people to run the `external_dependencies.sh` of the system tests they are currently running (possibly outside the container...)

hyperrealist:
I'll move them test-by-test as discussed in person.

PR #1055, opened 2025-09-22
Restructure tests

hyperrealist:
This PR addresses #1019

Passing CI tests on the same commit hash (`9d711012a3b1e75e1a524cdef7e90ff649930845`) are here: https://github.com/hyperrealist/ophyd-async/actions/runs/17917291219

TODO:
- [x] retest without branching and verify if coverage is not lost from structuring
- [x] ~~improve coverage~~ opened a separate issue #1071

### DIFF
--- a/tests/unit_tests/test_branching.py
+++ b/tests/unit_tests/test_branching.py
@@ -1,0 +1,52 @@
+# This module adds tests designed to improve branching coverage.
+
+# TODO: Set branch = true in [tool.coverage.run] block in
+#       pyproject.toml to include these tests in coverage.
+
+# TODO: Move to appropriate modules if they already exist.
+
+# import pytest
+from autodoc2.config import Config
+from autodoc2.db import InMemoryDb
+from bluesky.protocols import Locatable
+
+from ophyd_async._docs_parser import ShortenedNamesRenderer  # noqa: PLC2701
+from ophyd_async.core._derived_signal import get_locatable_type  # noqa: PLC2701
+
+# src/ophyd_async/_docs_parser.py:10
+
+
+class DummyRenderer(ShortenedNamesRenderer):
+    def __init__(self):
+        super().__init__(InMemoryDb(), Config())
+
+
+def test_format_annotation_with_annotation():
+    renderer = DummyRenderer()
+    result = renderer.format_annotation("some.module.ClassName")
+    assert "~some.module." in result or "ClassName" in result
+
+
+def test_format_annotation_without_annotation():
+    renderer = DummyRenderer()
+    result = renderer.format_annotation(None)
+    assert result == renderer.format_annotation(None)  # just ensure not crashing
+
+
+# src/ophyd_async/core/_derived_signal.py:330
+
+
+def test_get_locatable_type():
+    class DummyLocatable(Locatable[int]):
+        def set(self, _): ...
+        def locate(self): ...
+
+    class NonLocatable: ...
+
+    class DummyNonLocatable(NonLocatable): ...
+
+    obj_locatable = DummyLocatable()
+    obj_nonlocatable = DummyNonLocatable()
+
+    assert get_locatable_type(obj_locatable) is int
+    assert get_locatable_type(obj_nonlocatable) is None


tests/system_tests/tango/conftest.py
```diff
@@ -38,6 +38,14 @@ def pytest_collection_modifyitems(config, items):
...
                     )
                 )
+            if sys.platform.startswith(
+                "win"
+            ):  # expect "win32", but open to a future change: https://mail.python.org/pipermail/patches/2000-May/000648.html
+                item.add_marker(
+                    pytest.mark.skip(
+                        reason="Ophyd-async is currently not tested on Windows + Tango"
+                    )
+                )
```
DiamondJoseph, line 48: Can we use `@pytest.mark.skipif` instead? https://docs.pytest.org/en/stable/how-to/skipping.html#id1 Means that the skipping expectation is right next to the test and I believe removes the need for most of the `pytest_*` magics

tests/system_tests/epics/eiger/test_eiger_system.py
```diff
@@ -65,6 +120,7 @@ async def test_eiger(RE, ioc_prefixes) -> EigerDetector:
     return test_eiger
 
 
+@pytest.mark.skip_as_not_ready
```
DiamondJoseph, line 123: Can we replace with `@pytest.mark.skip(reason="Not ready, to be implemented in #<issue number>")`?

tests/system_tests/epics/adsim/test_adsim_system.py
```diff
@@ -35,17 +35,23 @@
 TIMEOUT = 10.0
 
 
-@pytest.fixture(scope="module", autouse=True)
+@pytest.fixture(scope="module")
 def with_env():
-    with patch.dict(
```
DiamondJoseph, line 40: Is there a reason this is preferable to patching the environment with clear? It seems more complex to me- was the old way not passing all required environment variables?

tests/conftest.py
```diff
@@ -23,18 +23,52 @@
...
+
+
+def pytest_collection_modifyitems(config, items):
+    for item in items:
+        if "skip_as_not_ready" in item.keywords:
+            item.add_marker(
+                pytest.mark.skip(
+                    reason="This test is skipped as not ready for general use"
+                )
+            )
```
DiamondJoseph, line 69:
```suggestion
```

I believe if we make use of `@pytest.mark.skip(reason="")` later we do not need these magics?

DiamondJoseph (commented):
Generally support the approach, just some questions about specific changes that have been made, thanks for your effort!

.coveragerc
coretl, line 1: Please can you put this in pyproject.toml instead. E.g. we get these from the copier template:
https://github.com/DiamondLightSource/python-copier-template/blob/2aa11776d7af19970b52bcf05dfdbf9262b80ea0/template/pyproject.toml.jinja#L73-L79
```
[tool.coverage.run]
patch = ["subprocess"]
data_file = "/tmp/{{ package_name }}.coverage"

[tool.coverage.paths]
# Tests are run from installed location, map back to the src directory
source = ["src", "**/site-packages/"]
```

tests/system_tests/epics/adsim/test_adsim_system.py
```diff
@@ -99,9 +106,14 @@ def prepare_then_count() -> MsgGenerator[None]:
...
 
+@pytest.mark.insubprocess
+@pytest.mark.skipif(
+    sys.platform.startswith("win"), reason="Services not set up on Windows"
+)
 @pytest.mark.timeout(TIMEOUT + 15.0)
-def test_software_triggering(RE: RunEngine, adsim: SimDetector) -> None:
+def test_software_triggering(RE: RunEngine, adsim: SimDetector, with_env: None) -> None:
     docs = run_plan_and_get_documents(RE, bp.count([adsim], num=2))
+    print("::debug::", docs)
```
coretl, line 116: Please remove debug

tests/system_tests/epics/adsim/test_adsim_system.py
```diff
@@ -35,17 +36,23 @@
...
+    original_env = {key: os.environ.get(key, "") for key in target_env}
+
+    # Set the test values
+    os.environ.update(target_env)
+
+    try:
         yield
+    finally:
+        # Restore original values (or delete if originally unset)
+        os.environ.update(original_env)
```
coretl, line 55: Now we've worked out these need to be run in a subprocess, I think you can go back to using `patch.dict`

pyproject.toml
```diff
@@ -108,6 +109,9 @@ asyncio_default_fixture_loop_scope = "function"
 timeout = 1
 [tool.coverage.run]
 data_file = "/tmp/ophyd_async.coverage"
+branch = true
+parallel = true
+concurrency = ["multiprocessing"]
```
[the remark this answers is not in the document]
hyperrealist: @coretl So, I took a closer look, and
- `branch = true` just makes codecov measure if both branches are executed in `if ... else ...` like branches.
- `parallel = true` and `concurrency = ["multiprocessing"]` do measure subprocesses, but the kind spawned with `multiprocessing` module. 
- Setting `patch = ["subprocess"]` adds `subprocess.Popen`- kind of subprocesses.

Now that I made the other changes, codecov seems to be happy. It doesn't seem to hurt to retain above settings too, but I'll leave that for you to decide. Perhaps they are more relevant to copier than this repo!

hyperrealist:
> I expect these lines should be hit by `test_tutorials.py`. These import code via `importlib`, so is anything in this thread helpful?

I think it is not the issue you linked, but coverage using a different tracing approach when multiprocessing is enabled. Including branches do still lower the coverage, but that is only because it is doing what it is supposed to be doing.

@coretl any thoughts?

PR #1058, opened 2025-09-25
Copier update to uv

coretl:
Updates to latest copier which uses uv to manage the project venv.


coretl:
Issues:
- https://github.com/DiamondLightSource/python-copier-template/issues/294
- https://github.com/DiamondLightSource/python-copier-template/issues/295
- Need to restore the checkout of main branch
- Need to set pytest default paths to be `src docs tests/unit_tests` instead of `src docs tests`

Issue #1060, opened 2025-09-25
StandardDetector is a StreamDetector, there should also be a StepDetector


oliwenmandiamond:
As of current, the only detector that is being implemented by other devices is the `StandardDetector`. This is tailored mostly for fly scanning by hooking up to a file writing mechanism and streaming the data it gets to there. This does not meet all detector needs and is too inflexible. For example, the Electron Analyser Detectors have the following issues when using a `StandardDetector`:

- Forced to use HDF file writer through epics. We cannot use this as our data shape changes between regions. There is no data returned in read method. 
- `TriggerInfo` logic does not apply to this detector as we use `regions` for 'preparing' our detector for data acquisition.
- Is not a fly scanning detector.

Speaking with @coretl IRL, we are proposing the following changes:

1. Create a `BaseDetector` class which has abstract methods. 
2. `StandardDetector` is renamed to `StreamDetector`.
3. `StandardDetector` implements methods from `BaseDetector`.
4. A new `StepDetector` is made which also implements methods from `BaseDetector`. It utilises a `controller`, `config_sigs`, and maybe new class called `ReadDataHandler` class which is responsible for the read and describe methods of the detector.

PR #1062, opened 2025-09-26
Added BaseDetector and StepDetector

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1060

### DIFF
--- a/src/ophyd_async/core/_detector.py
+++ b/src/ophyd_async/core/_detector.py
@@ -49,7 +49,7 @@ class DetectorTrigger(Enum):
 
 
 class TriggerInfo(ConfinedModel):
-    """Minimal set of information required to setup triggering on a detector."""
+    """Minimal set of information required to setup triggering on a flying detector."""
 
     number_of_events: NonNegativeInt | list[NonNegativeInt] = Field(default=1)
     """Number of events that will be processed, (0 means infinite).
@@ -187,6 +187,7 @@ def _ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo
 class BaseDetector(
     Device,
     Stageable,
+    Preparable,
     AsyncConfigurable,
     AsyncReadable,
     Triggerable,
@@ -203,6 +204,7 @@ class BaseDetector(
     ) -> None:
         self._controller = controller
         self._config_sigs = list(config_sigs)
+        self._trigger_info: TriggerInfo | None = None
         super().__init__(name, connector=connector)
 
     @AsyncStatus.wrap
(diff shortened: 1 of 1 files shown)


src/ophyd_async/core/_detector.py
```diff
@@ -231,40 +387,18 @@ def __init__(
...
     async def unstage(self) -> None:
         """Disarm the detector and stop file writing."""
-        await asyncio.gather(self._writer.close(), self._controller.disarm())
-
-    async def read_configuration(self) -> dict[str, Reading]:
-        return await merge_gathered_dicts(sig.read() for sig in self._config_sigs)
-
-    async def describe_configuration(self) -> dict[str, DataKey]:
-        return await merge_gathered_dicts(sig.describe() for sig in self._config_sigs)
+        await asyncio.gather(super().stage(), self._writer.close())
```
Villtord, line 401: super().unstage() perhaps?

oliwenmandiamond: Good spot, I copy and pasted from stage

src/ophyd_async/core/_detector.py
```diff
@@ -214,13 +430,10 @@ def __init__(
...
         connector: DeviceConnector | None = None,
     ) -> None:
-        self._controller = controller
         self._writer = writer
         self._describe: dict[str, DataKey] = {}
-        self._config_sigs = list(config_sigs)
         # For prepare
-        self._arm_status: AsyncStatus | None = None
-        self._trigger_info: TriggerInfo | None = None
+        self._arm_status: AsyncStatus | None = None  # Can this be removed?
```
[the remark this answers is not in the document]
coretl: If it's private and not used in the codebase, please remove it!

src/ophyd_async/core/_detector.py
```diff
@@ -184,13 +187,226 @@ def _ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo
...
+        new_shape: list[int | None] = [x, y]
+        new_size = x * y
+        if current_size != new_size:
+            raise ValueError(
+                f"The size of signal {self._image.name} is of shape "
+                f"{current_shape}. Failed to resize to new shape "
+                f"{new_shape} as new size is {new_size}."
+            )
+        self._set_shape(image_describe, new_shape)
+        return image_describe
```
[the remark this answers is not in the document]
coretl: Yes, if we use the PVA plugin this reshaping is done on the EPICS side for you. We need to talk to Greg to get it added

coretl (changes requested):
Please can we discuss `ReadableDetector` in person. I'm pretty convinced this should share more with `StandardReadable` than with `StandardDetector`.

src/ophyd_async/core/_detector.py
```diff
@@ -184,6 +186,34 @@ def _ensure_trigger_info_exists(trigger_info: TriggerInfo | None) -> TriggerInfo
...
+
+    @AsyncStatus.wrap
+    async def prepare(self, value: TriggerInfo) -> None:
+        await self.controller.prepare(value)
+        self.trigger_info = value
+
+    @AsyncStatus.wrap
+    async def trigger(self) -> None:
+        await self.controller.arm()
+        await self.controller.wait_for_idle()
```
coretl, line 214: I like the encapsulation of `prepare`, but I think you could suck in `ensure_trigger_info` in as well and the software triggering part too. How about:

```suggestion
class _ControllerContext(Device, Stageable, Triggerable, Preparable):
    """Controller and trigger_info as a wrapped device."""

    def __init__(self, controller: DetectorController, name: str = ""):
        self._controller = controller
        self._trigger_info: TriggerInfo | None = None
        super().__init__(name)

    @AsyncStatus.wrap
    async def stage(self) -> None:
        await self._controller.disarm()
        self._trigger_info = None
        
    unstage = stage
    
    @property
    def trigger_info(self) -> TriggerInfo:
        """The TriggerInfo the detector has been prepared with, raising RuntimeError if prepare not called."""
        if self._trigger_info is None:
            raise RuntimeError("Detector has not been prepared with a TriggerInfo.")
        return self._trigger_info

    @AsyncStatus.wrap
    async def prepare(self, value: TriggerInfo) -> None:
        await self._controller.prepare(value)
        self._trigger_info = value

    @AsyncStatus.wrap
    async def trigger(self) -> None:
        if self._trigger_info is None:
            await self.prepare(TriggerInfo())
        await self.controller.arm()
        await self.controller.wait_for_idle()
```

PR #1090, opened 2025-10-17
Add note on using Annotated with ruff

DominicOram:
In doing https://github.com/DiamondLightSource/dodal/pull/1635 we flagged that `Annotated as A` broke N817, it wasn't 100% clear to me whether I should ignore it, rename it etc. until I saw what `ophyd-async` was doing and decided that copying that was probably best. I've added a note to the docs here to try and help future developers make that decision quicker.

If you don't think it's useful feel free to reject


coretl (approved):
Thanks, that's a useful addition

PR #1096, opened 2025-10-18
Allow mock signals to have different readback to setpoint values

### LINKED ISSUE #1095: Add helper mock functions for signals where the setpoint and readback differ
Following from https://github.com/DiamondLightSource/mx-bluesky/issues/1360. We have a device where there is a setpoint and a readback such that we have a `self.x_step_size = epics_signal_rw_rbv(float, f"{prefix}X_STEP_SIZE")`. When we set a value on the setpoint of this some internal logic will happen in the device and the readback will come back rounded. Therefore to set it and wait for it to come back we need to do something like `await set_and_wait_for_value(self.x_step_size, value, match_value=partial(isclose, param_value, abs_tol=0.0001))`. This works fine in production but is hard to test because there is no way to make a mock that when set does not change its value to exactly the setpoint.

Effectively we want to be able to write a test like https://github.com/bluesky/ophyd-async/blob/b81e0b88760488cc09f6d69179227910c66ec17f/tests/unit_tests/core/test_signal.py#L231 but for `set_and_wait_for_value`, not `set_and_wait_for_other_value`

## Acceptance Criteria
- There is a test like `test_set_and_wait_for_value_different_set_and_read_times_out` but for a single RW signal
- This can be easily copied by downstream codebases

DominicOram:
Fixes #1095 

This feels equal parts hacky and elegant. It obviously needs docs and probably a helper function in `_mock_signal_utils.py` but wanted a second pair of eyes on the approach before I tidied it up.


coretl (commented):
Good idea! Needs some though as to how it will interact with `callback_on_mock_put` though. Maybe we still let it return `None` and default to the setpoint in that case? Do you need a more focussed utility function to set this, or is `callback_on_mock_put` sufficient?

DominicOram:
>Maybe we still let it return None and default to the setpoint in that case?

Yh, I like that.

> Do you need a more focussed utility function to set this, or is callback_on_mock_put sufficient?
Yh, `callback_on_mock_put` is probably fine. Just need good docs + tests.

I will tidy this up, thanks for looking.

coretl (approved):
LGTM!

PR #1124, opened 2025-11-04
Use dial limit for motor limit check

### LINKED ISSUE #1121: Take offset into account with checking motor limits
When a motor has an offset set (the `OFF` field) then it is applied to the limit checking, particularly for the logic that a limit of 0, 0 means apply no limits. e.g.
* If OFF is set to 0 then LLM=0, HLM=0 means no limit applied
* If OFF is set to 5 then LLM=5, HLM=5 means no limit applied

We should update the logic added in https://github.com/bluesky/ophyd-async/pull/1052 to take the offset into account. Note that we may end up with some floating point conversion issues here if we don't check they're approx equal.

## Acceptance Criteria
- When a non-zero offset is applied then the limit checking logic takes this into account

teoching0705:
fix #1121


src/ophyd_async/epics/motor.py
```diff
@@ -131,14 +133,23 @@ async def check_motor_limit(self, abs_start_pos: float, abs_end_pos: float):
...
             self.low_limit_travel.get_value(),
             self.high_limit_travel.get_value(),
             self.motor_egu.get_value(),
+            self.dial_low_limit_travel.get_value(),
+            self.dial_high_limit_travel.get_value(),
         )
 
         # EPICS motor record treats limits of 0, 0 as no limit
-        if motor_lower_limit == 0 and motor_upper_limit == 0:
+        # Use DLLM and DHLM to check
```
DominicOram: Should: I think we can be a bit more informative/concise on the comment. Something like:
```python
# EPICS motor record treats dial limits of 0, 0 as no limit
```

DominicOram (approved):
Great, thank you!

PR #1126, opened 2025-11-06
Add intial collect_stream_docs for eiger odin

RJCD-Diamond:
Add collect_stream_docs to eiger odin. Should allow hardware triggering of the eigers to be performed


RJCD-Diamond:
Closed as https://github.com/bluesky/ophyd-async/pull/1128 this one uses better logic (and works reliably and produces nexus files)

PR #1127, opened 2025-11-07
Add fan_dev_shm_enable signal to Odin

jacob720:
Needed for https://github.com/DiamondLightSource/mx-bluesky/issues/1076
We are currently migrating to the ophyd-async Eiger, and trying to use this in Hyperion. We need to be able to turn on dev/shm streaming for GPU analysis during xray centring, and turn it off to clear the memory, this PR adds the relevant PV to the Odin device.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -166,6 +166,7 @@ obj_ignore = [
     "ophyd_async.sim._mirror.TwoJackDerived",
     "0.1",
     "1.0",
+    "bluesky.protocols.T_co",
 ]
 nitpick_ignore = []
 for var in obj_ignore:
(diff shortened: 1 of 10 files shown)


jacob720:
Great, thanks

GDYendell:
The PV isn't temporary. What might change is that we don't rely on toggling it off and on to clear the memory when the consuming process deletes the images as soon as they have been processed. Although it would probably still needed in some cases like scan failure where there might be some images left behind.

DominicOram:
> @DominicOram @jacob720 I thought we were going to delete ADOdin from ophyd-async and switch testing FastCS Odin instead? That would mean this change would be deleted shortly...

We're actively testing the FastCS Eiger + ADOdin in the meantime with this patched on to make sure Hyperion works with FastCS Eiger / StandardDetector generally. Once https://github.com/bluesky/ophyd-async/pull/1158 is merged then happy to drop this

Issue #1117, opened 2025-10-31
EigerDetectorIO nexpi PV should be optional or possibly removed?


RJCD-Diamond:
Our new Eiger detector does not have the nexpi PV, and therefore when trying to connect to nexpi on b21's Eiger it fails.

B21 has two eiger's one newer and one older, so it works on one but not the other.

 I'm not sure this PV is actually used for anything important, so it might be an option just to remove it.

coretl:
@GDYendell or @shihab-dls do you know what this PV is and why it is only on some Eigers?

PR #1128, opened 2025-11-07
Improve Eiger: Add collect_stream_docs for EigerOdin using ADCore Logic to allow hardware triggering, and add plugins 

### LINKED ISSUE #530: FastCS Eiger: Correctly implement describe and collect
As of https://github.com/bluesky/ophyd-async/pull/528 the describe/collect on the `OdinWriter` of the fastCS eiger are placeholders. We should implement them properly.

## Acceptance Criteria
* A full system test is written that uses the describe/collect functionality of the eiger
* This passes against the simulated eiger

RJCD-Diamond:
The Odin in Eiger doesn't have several things that StandardDetector has, such as a collect_stream_docs method, preventing it from being used for hardware triggering, and a way of adding NDAttributes. 

It also contains more PV's  than just the minimum required PV's. Such that when connecting to modern Eigers - it fails. All of this is to allow the Eiger to be used for hardware triggering on b21 but also just generally improves it.

Closes: https://github.com/bluesky/ophyd-async/issues/530 and https://github.com/bluesky/ophyd-async/issues/1117

If someone could do a release after this has been merged into main, I'd really appreciate that


src/ophyd_async/epics/odin/_odin_io.py
```diff
@@ -91,16 +99,26 @@ def __init__(
         path_provider: PathProvider,
         odin_driver: Odin,
         detector_bit_depth: SignalR[int],
+        plugins: dict[str, NDPluginBaseIO] | None = None,
     ) -> None:
         self._drv = odin_driver
         self._path_provider = path_provider
         self._detector_bit_depth = Reference(detector_bit_depth)
+        self._plugins = plugins
```
DiamondJoseph:
```suggestion
        self._plugins = plugins or {}
```

DiamondJoseph (commented):
🤷 From what I understand, Shihab or Tom will need to look

src/ophyd_async/epics/odin/_odin_io.py
```diff
@@ -128,24 +146,84 @@ async def open(self, name: str, exposures_per_event: int = 1) -> dict[str, DataK
             wait_for_value(self._drv.meta_writing, "Writing", timeout=DEFAULT_TIMEOUT),
         )
 
-        return await self._describe()
+        # Add the main data
+        self._datasets = [
+            HDFDatasetDescription(
+                data_key=name,
+                dataset="/data",
```
shihab-dls: I would've assumed this to be `/entry/data/data`?
```suggestion
                dataset="/entry/data/data",
```

shihab-dls (changes requested):

src/ophyd_async/epics/odin/_odin_io.py
```diff
@@ -155,15 +244,38 @@ async def observe_indices_written(
...
         await self._drv.meta_stop.set(True, wait=True)
         if self._capture_status and not self._capture_status.done:
             await self._capture_status
         self._capture_status = None
+
+    def _get_odin_filename_suffix(self) -> str:
+        """Should result in _000001 for the first file created by this OdinWriter.
+
+        If odin creates more frames than max number of frames it "rollsover"
+        If there are 4 nodes, the next file should be _000005, etc.
```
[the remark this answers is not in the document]
RJCD-Diamond: It should work for the many nodes case. If the self._odin_writer_number is anything other than 1, it will not be _000001, and if you had multiple nodes each one would be numbered. I will link to the issue though

Issue #1039, opened 2025-09-05
Provide the ability to inject logic when mock devices are created


DominicOram:
We have a number of devices where for a mock of the device to at all be useful we need to inject some logic. The obvious example is something like:

```python
class Device():
    def __init__(self, prefix):
        self.setpoint = epics_signal_w(prefix+":SP")
        self.readback = epics_signal_w(prefix+":RBV")
```

where to make a mock of the device behave like expected we need to do something like:

```python 
with init_devices(mock=True):
    my_device = Device()

callback_on_mock_put(my_device.setpoint, lambda _: set_mock_value(my_device.readback))
```

This works fine for individual tests but it has some disadvantages:
* We inevitably need the same logic for tests of the device itself and the plans. But plans/devices are normally in separate repos for us. This leads to arguments about where the logic should go (https://github.com/DiamondLightSource/dodal/issues/1455)
* It can be useful on a beamline to be able to quickly switch a device to mock mode. The usual usecase for this is that hardware has failed but to get us close to operational quickly we want to run the same plans as usual just with mock data. In some cases the logic above is required to get to a point where the plan will run without failure.
* Personal preference but I think it's generally cleaner to have the device logic and the "expected ioc logic" (which is essentially what we're writing here) close to each other

I propose that `ophyd-async` should give us the ability to add this behaviour automatically any time the device is instantiated with `mock=True` e.g. some kind of interface like:

```python
class Device():
    def __init__(self,prefix):
        self.setpoint = epics_signal_w(prefix+":SP")
        self.readback = epics_signal_w(prefix+":RBV")
 
   def _default_mock_logic(self):
       callback_on_mock_put(my_device.setpoint, lambda _: set_mock_value(my_device.readback))
```

I guess some that there may be times you don't want that behaviour on `mock=True`. But I think this can be handled on a case by case basis. E.g. if you want to test what happens to the above device when data isn't getting passed through from `setpoint` -> `readback` then our test can just remove the callback as it's first thing.

## Acceptance Criteria
- It is easier to specify logic for mock devices alongside each device.

coretl:
As discussed in person, I agree with adding some generally useful mock behaviour, but we want to be able to opt out of it so tests can implement their own. I will have a think about the best mechanism to support this

coretl:
[...]

PR #1129, opened 2025-11-11
Implement DeviceMock with automatic logic injection for mock devices

vredchenko:
Please note - this PR is a test of Claude Code's capabilitites
_____

This commit addresses issue #1039 by implementing Tom Cobb's proposed solution for automatic mock device logic injection.

Changes:
- Renamed LazyMock to DeviceMock (keeping LazyMock as backward-compatible alias)
- Made DeviceMock generic over Device type using Generic[DeviceType]
- Added connect() method to DeviceMock that can be overridden by subclasses
- Implemented @default_device_mock_for_class decorator to register DeviceMock subclasses
- Updated Device.connect() to automatically look up and use registered DeviceMock classes
- Added InstantMotorMock as an example implementation for Motor
- Added test to verify automatic mock injection works correctly

The implementation allows device developers to define custom mock behavior alongside their device definitions, which is automatically applied when devices are connected in mock mode. This eliminates the need for manual callback setup in test code.

Example usage:
```python
@default_device_mock_for_class
class InstantMotorMock(DeviceMock[Motor]):
    async def connect(self, device: Motor):
        callback_on_mock_put(
            device.user_setpoint,
            lambda v, _: set_mock_value(device.user_readback, v)
        )

# The mock is automatically applied when connecting in mock mode
async with init_devices(mock=True):
    motor = Motor("BLxxI-MO-TABLE-01:X")
# motor now has instant movement behavior without manual setup
```

Resolves: #1039

### DIFF
--- a/docs/tutorials/writing-tests-for-devices.md
+++ b/docs/tutorials/writing-tests-for-devices.md
@@ -102,7 +102,7 @@ Then decorate the original class with [](#default_mock_class) so it is automatic
 
 ```{literalinclude} ../../src/ophyd_async/epics/motor.py
 :language: python
-:start-at: default_mock_class
+:start-at: default_mock_class(
 :end-at: class Motor
 ```
 


src/ophyd_async/core/_utils.py
```diff
@@ -342,7 +347,7 @@ def __call__(self) -> T:
         return self._obj
 
 
-class LazyMock:
+class DeviceMock(Generic[DeviceType]):
```
coretl: Move `DeviceMock` and all other additions to this file into `_device.py` instead

DominicOram (changes requested):
Great, this replaces most of the behaviour we have (but the extra import is annoying, see comment in code). However, I think it has broken adding to parent mocks. I used to be able to do:
```python
async def test_xy_stage():
    from dodal.devices.motors import XYStage

    with init_devices(mock=True):
        xy_stage = XYStage("")

    parent_mock = get_mock(xy_stage)

    # Reset mock to clean up sets done in the test fixture
    parent_mock.reset_mock()
    await xy_stage.x.set(1)
    await xy_stage.y.set(2)

    parent_mock.assert_has_calls(
        [
            call.x.user_setpoint.put(1, wait=True),
            call.y.user_setpoint.put(2, wait=True),
        ]
    )
```
this no longer works.

oliwenmandiamond:
This looks very good and clean! 

However, one concern we have is that mock behaviour will be very well hidden. Is there an argument to not have `default_device_mock_for_class` so that when writing a test, you always have to explicitly state the mock behaviour you want e.g

```Python
await motor.connect(mock=custom_mock)
```
so that when reading a test it is very clear where all the behaviour is coming from?  Otherwise it might be hard to debug and looks like magic!

If it's a device that is used often in tests, you can then very easily create a pytest fixture to reuse with the default custom_mock
```Python
@pytest.fixture
async def my_device() -> Device
     with init_device(mock=True):
           my_device = MyDevice()
     await my_device.connect(mock=custom_mock)
     return my_device
```
or even like this
```Python
@pytest.fixture
def my_device() -> Device
     with init_device(mock=custom_mock):
           my_device = MyDevice()
     return my_device
```
Thus, the `default_device_mock_for_class` seems a little overkill and through a backdoor? Worth a thought

oliwenmandiamond:
The only other thing I would add before Tom and Dom make a decision is using pytest approach means we don’t reinvent the wheel and have a single way to customise mock behaviour. Having too many options to do the same thing can lead to unnecessary complexity and much harder to maintain as we have seen with GDA code

vredchenko:
> As an aside, I'm very happy for us using AI tools to help us write code but the amount of AI responses in the comments on this thread makes me uncomfortable. At best I feel the responses are noisy, at worse I feel they are patronising.

At the flip of a switch we can get it to flatter or even worship us. Goes without saying verbosity can also be adjusted. The point of this specific PR is to showcase what's possible (albeit this is just the beginning of a journey - to do it properly we need to equip it with project-specific structured and searchable docs, tooling, skills it can re-use, etc)

oliwenmandiamond:
> I like the opt-in idea and attaching the mock to the Device, but not via a private variable. How about:
> 
> ```python
> @default_mock_class(InstantMotorMock)
> class Motor(...):
>     ...
> ```

I quite like this approach as when you delve into class code, you can still clearly see where a default mock comes from. If we're allowing multiple mock classes, then I would change this to be 

```Python
@add_default_mock_class(InstantMotorMock)
@add_default_mock_class(...)
...
class Motor(...):
```
and / or
```Python
@default_mock_classes([InstantMotorMock, ...])
```

> Or we could ditch the attaching to Device classes, and be even more explicit with the adding of these mocks:
> 
> ```python
> with init_devices(mock_classes=[InstantMotorMock, ...]):
>     stage = XYZStage(...)
> # stage.x is now connected with mock=InstantMotorMock()
> ```
> 
> Thoughts?

I also really like this as we can then chain on together mock logic which can target specific bits of the device and becomes a list of plugins essentially. However, I can also maybe see issues where mock behaviours conflict so am fine making it a single mock class, happy either way. 

If you use mock_classes=[...] or mock_class=... , it should override the default mock class(es).

coretl:
> I also really like this as we can then chain on together mock logic which can target specific bits of the device and becomes a list of plugins essentially. However, I can also maybe see issues where mock behaviours conflict so am fine making it a single mock class, happy either way.

I meant this as an either/or decision.

Either we decorate each Device class with precisely one DeviceMock class that will be used as the default when we `connect(mock=True)`:
```python
@default_mock_class(InstantMotorMock)
class Motor(...):
    ...
```

Or we temporarily set the DeviceMock class for the Device and its children for the duration of connect. Mechanisms could look like passing the classes to connect:
```python
stage = XYZStage(...)
stage.connect(mock=[InstantMotorMock, ...])
```
or an arg to `init_devices`:
```python
with init_devices(mock_classes=[InstantMotorMock, ...]):
    stage = XYZStage(...)
```

@DominicOram @oliwenmandiamond thoughts?

oliwenmandiamond:
> My preference is:
> 
> ```python
> @default_mock_class(InstantMotorMock)
> class Motor(...):
>     ...
> ```
> 
> I would rather not have to worry in downstream libraries what the intended mock should be unless I want to override that mocks behaviour (which is rare).

I'm fine with this too. How does one override the default if it is ever needed or to just revert back to basic DeviceMock if situation ever arises?

src/ophyd_async/epics/motor.py
```diff
@@ -61,15 +67,45 @@ def __getattr__(name):
...
+        set_mock_value(device.max_velocity, 1000)  # Prevent ZeroDivisionError
+
+        # Motor starts in "done" state (not moving)
+        set_mock_value(device.motor_done_move, 1)
+
+        # When setpoint is written to, immediately update readback and done flag
+        def _instant_move(value, wait):
+            set_mock_value(device.motor_done_move, 0)  # Moving
+            set_mock_value(device.user_readback, value)  # Arrive instantly
+            set_mock_value(device.motor_done_move, 1)  # Done
```
DominicOram: Should: After https://github.com/bluesky/ophyd-async/pull/1096 this should `return value`. We should probably also have a test that would have caught this.

src/ophyd_async/epics/motor.py
```diff
@@ -61,15 +67,45 @@ def __getattr__(name):
...
+
+        callback_on_mock_put(device.user_setpoint, _instant_move)
 
 
+@default_mock_class(InstantMotorMock)
 class Motor(
```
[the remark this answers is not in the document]
oliwenmandiamond: Looks good, my only nit was that the motor limits shouldn't be +/-1000. In dodal, with `patch_motor` we set this to +/-infinity. I can understand why this wasn't desired here and that solution was before the limits were ignored if set to zero and zero. I think the default limits should be left at zero and zero so they are ignored, effectively the same behaviour as +/-infinity rather than some arbitrary number

PR #1134, opened 2025-11-14
Remove wait from Signal.set and replace with epics specific argument

### LINKED ISSUE #1048: Remove `wait` from `SignalW.set` and replace it with EPICS backend specific argument
For EPICS there are 3 ways you can use a signal
1. `await signal.set(wait=True)` will send the value and wait until the device has responded
2. `status = signal.set(wait=True)` will send the value, and return a status that tells when the device has responded
3. `status = signal.set(wait=False)` will send the value but tell the IOC to return as soon as it has got the value, not to wait for the device

Most of the time, if the developer does not want to wait until the device has responded they will use 2. They only use 3 if they required two set operations to happen at the same time (not supported by the IOC).

If Tango developers only need 1 and 2 then you could say that 3 is unsupported for tango, to use 2 instead.

_Originally posted by @coretl in https://github.com/bluesky/ophyd-async/issues/1001#issuecomment-3252858737_
            
I suggest we remove case 3 by taking the `wait` argument out, and replacing it with a value specific flag on the `EpicsSignalBackend`. Maybe we call it `no_wait_when_setting: set[SignalDatatypeT]`.

So instead of:
https://github.com/bluesky/ophyd-async/blob/aeded110f2410caf2e5882554eed2ae2a8951d01/src/ophyd_async/epics/motor.py#L90
https://github.com/bluesky/ophyd-async/blob/aeded110f2410caf2e5882554eed2ae2a8951d01/src/ophyd_async/epics/motor.py#L227
we should do
```python
epics_signal_w(int, prefix + ".STOP", no_wait_when_setting={1})
```
and instead of:
https://github.com/bluesky/ophyd-async/blob/aeded110f2410caf2e5882554eed2ae2a8951d01/src/ophyd_async/epics/adcore/_core_io.py#L26
https://github.com/bluesky/ophyd-async/blob/aeded110f2410caf2e5882554eed2ae2a8951d01/src/ophyd_async/epics/core/_util.py#L89
we should do
```python
A[SignalRW[bool], PvSuffix.rbv("Acquire"), NoWaitWhenSetting(0)]
```

gkalua:
## Breaking change

Removes the `wait` argument to `Signal.set`, it is now set when constructing the EPICS Signal.

It can be one of:
- `True`: Return when server-side operation has completed
- `False`: Return when server-side operation has started
- `callable`: Call with the value being put to decide whether to wait

For procedural style devices:
```python
# old
signal = epics_signal_rw(int, "PV")
signal.set(5, wait=False)

# new
signal = epics_signal_rw(int, "PV", wait=False)
signal.set(5)
```

For declarative style devices:
```python
# old
from ophyd_async.epics.core import PvSuffix

class MyDevice(EpicsDevice):
    signal: A[SignalRW[int], PvSuffix("SUFF")]

my_device.signal.set(5, wait=False)

# new
from ophyd_async.epics.core import PvSuffix, EpicsOptions

class MyDevice(EpicsDevice):
    signal: A[SignalRW[int], PvSuffix("SUFF"), EpicsOptions(wait=False)]

my_device.signal.set(5)
```

If you need a different value for `wait` depending on the value you are setting, like for the busy record:
```python
# old
busy_record_signal = epics_signal_rw(bool, "Acquire")
status = busy_record_signal.set(True, wait=True)
busy_record_signal.set(False, wait=False)
await status

# new
from ophyd_async.core import non_zero

busy_record_signal = epics_signal_rw(bool, "Acquire", wait=non_zero)
status = busy_record_signal.set(True)
busy_record_signal.set(False)
await status
```

Fixes #1048

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Changes here will be overwritten by Copier
-_commit: 5.0.1
+_commit: 5.0.2
 _src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: tom.cobb@diamond.ac.uk
 author_name: Tom Cobb
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -24,4 +24,4 @@ It is recommended that developers use a [vscode devcontainer](https://code.visua
 
 This project was created using the [Diamond Light Source Copier Template](https://github.com/DiamondLightSource/python-copier-template) for Python projects.
 
-For more information on common tasks like setting up a developer environment, running the tests, and setting a pre-commit hook, see the template's [How-to guides](https://diamondlightsource.github.io/python-copier-template/5.0.1/how-to.html).
+For more information on common tasks like setting up a developer environment, running the tests, and setting a pre-commit hook, see the template's [How-to guides](https://diamondlightsource.github.io/python-copier-template/5.0.2/how-to.html).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,7 +147,7 @@ commands = [
     [
         "pyright",
         "--pythonpath",
-        ".venv/bin/python",
+        "{env:VIRTUAL_ENV}/bin/python",
         "src",
         { replace = "posargs", default = [
         ], extend = true },
--- a/src/ophyd_async/epics/adcore/_core_io.py
+++ b/src/ophyd_async/epics/adcore/_core_io.py
@@ -241,6 +241,12 @@ class NDCBFlushOnSoftTrgMode(StrictEnum):
 
 
 class NDPluginCBIO(NDPluginBaseIO):
+    """Plugin that outputs pre/post-trigger NDArrays based on defined conditions.
+
+    This mirrors the interface provided by ADCore//Db/NDCircularBuff.template
+    See HTML docs at https://areadetector.github.io/areaDetector/ADCore/NDPluginCircularBuff.html
+    """
+
     pre_count: A[SignalRW[int], PvSuffix.rbv("PreCount")]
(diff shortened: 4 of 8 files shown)


src/ophyd_async/epics/core/_util.py
coretl: We need to find all the places this function is used, and make sure the signal itself is constructed with `wait=False`

src/ophyd_async/epics/motor.py
```diff
@@ -267,7 +267,7 @@ async def stop(self, success=False):
         self._set_success = success
         # Put with completion will never complete as we are waiting for completion on
         # the move above, so need to pass wait=False
-        await self.motor_stop.set(1, wait=False)
+        await self.motor_stop.set(1)
```
coretl: `motor_stop` needs constructing with `wait=False`

coretl (commented):
Made some comments, all the Tango tests need fixing too...

src/ophyd_async/epics/core/_signal.py
```diff
@@ -78,14 +78,23 @@ def get_signal_backend_type(protocol: EpicsProtocol) -> type[EpicsSignalBackend]
...
     w_protocol, w_pv = split_protocol_from_pv(write_pv)
     protocol = get_unique({read_pv: r_protocol, write_pv: w_protocol}, "protocols")
+
     signal_backend_type = get_signal_backend_type(protocol)
-    return signal_backend_type(datatype, r_pv, w_pv)
+    return signal_backend_type(
+        datatype,
+        r_pv,
+        w_pv,
+        options,  # type: ignore
```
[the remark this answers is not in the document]
gkalua: It complains that only 3 arguments were expected.

PR #1158, opened 2025-12-08
Convert ADOdin to FastCS-Odin

### LINKED ISSUE #551: Eiger FastCS: Use PVI information to fill in PVs

shihab-dls:
closes #551 

This PR implement FastCS-Odin in place of ADOdin.

### DIFF
--- a/tests/unit_tests/epics/eiger/test_odin_io.py
+++ b/tests/unit_tests/epics/eiger/test_odin_io.py
@@ -72,7 +72,7 @@ async def test_when_open_called_then_all_expected_signals_set(
     await writer.open(ODIN_DETECTOR_NAME)
 
     get_mock_put(driver.fp.data_datatype).assert_called_once_with("uint16", wait=ANY)
-    get_mock_put(driver.fp.frames).assert_called_once_with(1, wait=ANY)
+    get_mock_put(driver.fp.frames).assert_called_once_with(0, wait=ANY)
 
     get_mock_put(driver.fp.start_writing).assert_called_once_with(None, wait=ANY)
 


src/ophyd_async/fastcs/odin/_odin_io.py
```diff
@@ -0,0 +1,129 @@
...
+            self._drv.fp.data_compression.set("BSLZ4"),
+            self._drv.fp.frames.set(exposures_per_event),
+            self._drv.fp.file_path.set(str(info.directory_path)),
+            self._drv.mw.directory.set(str(info.directory_path)),
+            self._drv.fp.file_prefix.set(info.filename),
+            self._drv.mw.file_prefix.set(info.filename),
+            self._drv.mw.acquisition_id.set(info.filename),
+        )
+
+        await self._drv.fp.start_writing.trigger(wait=True)
```
coretl: wait=True is the default

src/ophyd_async/fastcs/odin/_odin_io.py
```diff
@@ -0,0 +1,129 @@
...
+    async def open(self, name: str, exposures_per_event: int = 1) -> dict[str, DataKey]:
+        info = self._path_provider(device_name=name)
+        self._exposures_per_event = exposures_per_event
+
+        await asyncio.gather(
+            self._drv.fp.data_datatype.set(
+                f"uint{await self._detector_bit_depth().get_value()}"
+            ),
+            self._drv.fp.data_compression.set("BSLZ4"),
+            self._drv.fp.frames.set(exposures_per_event),
```
coretl: What does fp.frames do?

coretl:
Merged into https://github.com/bluesky/ophyd-async/pull/1177

PR #1169, opened 2025-12-19
Run empty scan when pmac unstage

### LINKED ISSUE #1153: Pmac stage should first zero EQU outputs
At the moment if you halt in the middle of a scan then the EQU outputs may be non-zero. First we should make a trajectory scan with no motors in it, just setting the right user programs to zero the EQU outputs

## Acceptance Criteria
- Change made and test written, and tested on ViSR after abort of a scan halfway through

shihab-dls:
closes #1153 

This PR adds logic to run an empty trajectory when `PmacTrajectoryTriggerLogic` is stopped.

### DIFF
--- a/src/ophyd_async/epics/pmac/_pmac_io.py
+++ b/src/ophyd_async/epics/pmac/_pmac_io.py
@@ -2,7 +2,7 @@ from collections.abc import Sequence
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


shihab-dls:
Tested one visr, and works fine.

Issue #721, opened 2025-01-07
 Rationalize names and behaviour of tutorial devices


coretl:
There are a couple of classes of devices that we have in tutorials:
1. Simulations where the logic is in the ophyd level, no control system access required
2. Simulations where the logic is in the EPICS or Tango level, requiring control system access

Although the devices we would like to simulate are similar, the implementations are very different, and the requirement of control system access makes me and @callumforrester think we should make the distinction obvious in the naming.

## Proposal
We should have the following devices:
- `ophyd_async.sim.SimMotor` (existing): ophyd level simulation of a motor
- `ophyd_async.sim.SimPointDetector` and `SimPointDetectorChannel` (new): N counter channels producing integer count rates proportional to a couple of motors
- `ophyd_async.sim.SimFilewritingDetector` (was `PatternDetector`): Gaussian blobs written to HDF with intensity proportional to a couple of motors
- `ophyd_async.epics.demo.DemoMotor` (was `Mover`): as `SimMotor` but with logic in epics database (not an epics simMotor)
- `ophyd_async.epics.demo.DemoPointDetector` and `DemoPointDetectorChannel` (was `SensorGroup` and `Sensor`): as `SimPointDetector` but with logic in epics database
- `ophyd_async.tango.demo.DemoMotor` (was `Mover`): as `SimMotor` but with logic in tango server
- `ophyd_async.tango.demo.DemoPointDetector` and `DemoPointDetectorChannel`: as `SimPointDetector` but with logic in tango server
- `ophyd_async.fastcs.demo.DemoMotor` (was `Mover`): as `SimMotor` but with logic in fastcs server
- `ophyd_async.fastcs.demo.DemoPointDetector` and `DemoPointDetectorChannel`: as `SimPointDetector` but with logic in fastcs server

## Acceptance Criteria
- Files exist with the right functionality

jsouter:
I have FastCS Controllers that implement the logic from the epics DemoMotor and DemoPointDetector*, do we want to have tests for these, or just provide documentation for how to spin up the FastCS IOC for different transports?

coretl:
> I have FastCS Controllers that implement the logic from the epics DemoMotor and DemoPointDetector*, do we want to have tests for these, or just provide documentation for how to spin up the FastCS IOC for different transports?

[...]

PR #1191, opened 2026-02-03
Tango demo

ajgdls:
Initial Tango demo branch for @coretl to review.

Partial resolution of issue #721


src/ophyd_async/tango/demo/_motor.py
```diff
@@ -54,8 +55,8 @@ async def set(self, value: float, timeout: CalculatableTimeout = CALCULATE_TIMEO
 
         if not (isinstance(timeout, float) or timeout is None):
             raise ValueError("Timeout must be a float or None")
+        # For this server, set returns immediately so this status should not be awaited
         await self.position.set(value, timeout=timeout)
-
         move_status = AsyncStatus(
             wait_for_value(self.state, DevStateEnum.ON, timeout=timeout)
         )
```
coretl: We now have a context manager for AsyncStatus, so now you can do:
```python
async with move_status:
    async for current_position in observe_value(...):
        ...
```

src/ophyd_async/tango/demo/_counter.py
```diff
@@ -12,7 +12,7 @@
 from ophyd_async.tango.core import TangoDevice, TangoPolling
 
 
-class TangoCounter(TangoDevice, StandardReadable):
+class DemoCounter(TangoDevice, StandardReadable):
```
coretl: Please can you just delete this file?

coretl (changes requested):
I've commented on the ophyd side of this, but I can't comment on the Tango side much...

src/ophyd_async/tango/demo/_tango/_servers.py
```diff
@@ -1,30 +1,33 @@
...
-    @attribute(dtype=float, access=AttrWriteType.READ_WRITE)
+    @attribute(dtype=float, access=AttrWriteType.READ_WRITE, format="%6.3f")
     async def position(self):
         return self._position
 
     async def write_position(self, new_position):
+        self.set_state(DevState.MOVING)
         self._setpoint = new_position
-        await self.move()
+        asyncio.create_task(self.move())
```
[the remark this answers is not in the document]
burkeds: Because of the default timeout behaviour of Tango commands, most commands will return immediately regardless of how long it takes to complete their action. Most attribute writes are similar.

burkeds:
The device servers look ok to me.

docs/tutorials/implementing-devices.md
```diff
@@ -130,7 +140,18 @@ There is no introspection of PVs in a device in EPICS, if we tell the IOC to mak
...
+:language: python
+:emphasize-lines: 20-
+```
+
+Here we start the TANGO device server application with a device name prefix and add all of the devices into a `MultiDeviceTestContext` which allows us to run without an external TANGO database.  Starting this device server here is done just for the demo, in production the application would already be running before you started bluesky.
+
+We then pass the same device Tango Resource Locator (TRL) for each Device down using prior knowledge about the attributes that these particular devices will create. For example, we know that there will be a `DemoStage`, and the device names will start with `prefix + "/X"` and `prefix + "/Y"`.
+
+```{note}
+There is no introspection of attributes in a device in TANGO, if we tell the device server to make 3 channels on the point detector, we must also tell the ophyd-async device that the point detector has 3 channels.
```
coretl:
```suggestion
There is no introspection of child devices of a device in TANGO, if we tell the device server to make 3 channels on the point detector, we must also tell the ophyd-async device that the point detector has 3 channels.
```

coretl (approved):
Minor clarification, then good to merge

coretl:
@burkeds I'm ready to merge this, did you want to review before I do?

burkeds (approved):

burkeds:
> @burkeds I'm ready to merge this, did you want to review before I do?

My mistake, please go ahead.

Issue #1204, opened 2026-02-19
Following standard detector rewrite, the `device_name` passed to path provider is the name of the writer


jwlodek:
Trying to write data with a new style `AreaDetector` passes the `writer` name to the `path_provider`:

```
<WatchableAsyncStatus, device: lambda1, task: <coroutine object AsyncStatusBase.__init__.<locals>.wait_with_error_message at 0x7f0ffeef8400>, errored: FileNotFoundError("Path /nsls2/data/pdf/proposals/commissioning/pass-319561/assets/lambda1-writer/2026/02/19/ doesn't exist or not writable!")>
```

Previous behavior was to use the parent device name, which I would prefer.

PR #1205, opened 2026-02-19
Pass detector name + datakey suffix to providers instead of writer name to match filenames to datakeys.

jwlodek:


jwlodek:
Resolves #1204 . I think using `datakey_suffix` if it is set is a cleaner way of allowing the use of the detector name in providers. This way the files would also match datakeys in the docs.

coretl (changes requested):
Thanks, this is much cleaner, just some comments on naming so it is clear that the `DetectorDataLogic` is getting its datakey_name passed into the `prepare_` methods rather than having to calculate it from `detector_name`

PR #1234, opened 2026-04-02
refactor: sum acceleration time and cruise time for PmacTrajectoryTriggerLogic._move_to_start() timeouts

shihab-dls:
closes #1233 

This PR calculates the time required to move from a pmac motor's initial position to the `ramp_up_position`. This is done by calculating the time required for acceleration, deceleration, and constant velocity cruise.


coretl (approved):
Can we reuse the maths from velocity profile for this? If not then this is fine...

shihab-dls:
> Can we reuse the maths from velocity profile for this? If not then this is fine...

Good point. Definitely can; re-requesting review with those changes.

coretl (approved):

PR #1235, opened 2026-04-02
fix: correctly check pmac execution status and messages

### LINKED ISSUE #1107: Check build, execute and append status before running a scan.
Investigate the busy record within the flyscanning trajectory and return the relevant error messages.

### Background:
When a Flyscan is run using a bluesky plan, we notice that if any of these vital commands fail, 'Build Profile', 'Append Points', and 'Execute Profile', expectedly, the scan does not run. However, on the run engine, the failure is not reported and instead waits forever and eventually times out.

### Acceptance Criteria:

- Show error message when build, execute or append fails.
- End the wait if any of the above commands fail.

shihab-dls:
Closes #1107 

Checks PMAC `execute_status` and `execute_message` after `execute_profile` returns, and raises appropriate error messages if not successful.


src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -154,6 +156,39 @@ async def _execute_trajectory(self, path: Path, motor_info: _PmacMotorInfo):
...
+            )
+
+    async def _ensure_trajectory_complete(self):
+        pmac_status = None
+        try:
+            async for pmac_status in observe_value(
+                self.pmac_ref().trajectory.execute_profile, done_timeout=DEFAULT_TIMEOUT
+            ):
+                if pmac_status is False:
+                    break
```
coretl: This was meant to `observe_value` on `execute_status` and raise with the value of `execute_message` if it fails. A similar check is needed where we do `build_profile` and `append_profile`

src/ophyd_async/epics/pmac/_pmac_io.py
```diff
@@ -19,6 +19,12 @@ class PmacExecuteState(SubsetEnum):
     EXECUTING = "Executing"
 
 
+class PmacStatus(SubsetEnum):
+    SUCCESS = "Success"
+    FAILURE = "Failure"
+    UNDEFINED = "Undefined"
```
shihab-dls, line 25: This is the subset of choices that `execute`, `build`, and `append` all have, according to this [db file](https://github.com/DiamondLightSource/pmac/blob/270962096d93d89b4c1c96b98b497ec2505d6e80/pmacApp/Db/pmacControllerTrajectory.template#L271)

src/ophyd_async/epics/pmac/_pmac_trajectory.py
shihab-dls: I think I've implemented what you mean; I'll re-request a review now and head down to ViSR to test this next week.

coretl (approved):

PR #1238, opened 2026-04-09
Add helpers for initializing device vectors from mbbi/mbbo direct record bitmasks. Allow for coercing char/byte/short pv types into 0/1 bools for CA and PVA

### LINKED ISSUE #1168: Allow cleaner handling of mbbiDirect fields
I have a number of PVs that are `mbbiDirect` types where each of the `B0...B16` fields has different meaning e.g. corresponds to a different PLC input/output.

If I add these to my device I currently have to do the following:

```python
self.interlock_disabled = epics_signal_r(int, prefix + "PLC_OUT.B8")
```

Meaning to read/write to it I have to use a 0/1, which is not very descriptive and has no type checking etc. for a user to write a 2 to it.

Instead it would be nice to either use an IntEnum:

```python
class ThisInterlock(IntEnum):
    DISABLED = 1
    ENABLED = 0

self.interlock = epics_signal_r(ThisInterlock, prefix + "PLC_OUT.B8")
```

or a bool:

```python
self.interlock_disabled = epics_signal_r(bool, prefix + "PLC_OUT.B8")
```

My preference is probably the `IntEnum` for the additional description but either is fine.

## Acceptance Criteria
- `mbbiDirect` fields can use more applicable types than ints

jwlodek:
Resolves #1168

* Allows for creating `DeviceVector`s for mbbiDirect and mbboDirect bitmasks.
* Allows for CA/PVA signals explicitly marked as bool with inferred types of byte/char/short to be converted to 0/1 bools.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Changes here will be overwritten by Copier
-_commit: 5.0.2
+_commit: 5.0.3
 _src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: tom.cobb@diamond.ac.uk
 author_name: Tom Cobb
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -24,4 +24,4 @@ It is recommended that developers use a [vscode devcontainer](https://code.visua
 
 This project was created using the [Diamond Light Source Copier Template](https://github.com/DiamondLightSource/python-copier-template) for Python projects.
 
-For more information on common tasks like setting up a developer environment, running the tests, and setting a pre-commit hook, see the template's [How-to guides](https://diamondlightsource.github.io/python-copier-template/5.0.2/how-to.html).
+For more information on common tasks like setting up a developer environment, running the tests, and setting a pre-commit hook, see the template's [How-to guides](https://diamondlightsource.github.io/python-copier-template/5.0.3/how-to.html).
--- a/README.md
+++ b/README.md
@@ -5,15 +5,14 @@
 
 # ![ophyd-async](https://raw.githubusercontent.com/bluesky/ophyd-async/main/docs/images/ophyd-async-logo.svg)
 
-Asynchronous Bluesky hardware abstraction code, compatible with control systems like EPICS and Tango.
-
-|    Source     |     <https://github.com/bluesky/ophyd-async>      |
-| :-----------: | :-----------------------------------------------: |
-|     PyPI      |             `pip install ophyd-async`             |
-| Documentation |      <https://bluesky.github.io/ophyd-async>      |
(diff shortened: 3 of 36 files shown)


coretl (commented):
I definitely agree with the `bool` typed signals, but I'm not sure about the `mbb_direct` `DeviceVector`s. @DominicOram's comment was about giving descriptive names to the bits, and giving a helper to make `DeviceVector`s of them might encourage people to do the opposite, especially as the vector is zero indexed. Do you have a use case for needing the vector?

jwlodek:
Mostly just as a convenience when there is an mbb(i/o)Direct with lots of bits that I want available, but don't necessarily need descriptive names for. My intention with allowing the `bits` parameter to the helper functions to be a tuple of bit numbers or a dict mapping bits to names was to allow for easily creating a structure like:

Pass in {0: "acquire_ok", 1: "temp_ok", 2: "humidity_ok"} and get a nested structure `status` with attributes `acquire_ok`, `temp_ok` and `humidity_ok` as signals. It seems `DeviceVector` only does integer indexing, so that is only somewhat useful (although I still give the `name` to the signals.

Perhaps a default `Bitmask` `Device` that gets returned instead is a better option? Could potentially be useful for non-mbb(i/o)Direct bitmasks as well?

coretl:
This would already work:
```python
class MyStatus(EpicsDevice):
    acquire_ok: A[SignalR[bool], PvSuffix(".B0")]
    temp_ok: A[SignalR[bool], PvSuffix(".B1")]
    humidity_ok: A[SignalR[bool], PvSuffix(".B2")]
```
That's quite a lot of typing, but it does at least match with the existing structure.

Would you mind if we put only the `bool` bits in and left out the `DeviceVector` functions for now? I think I'd like to see what people make first before putting a helper into the library for it.

jwlodek:
Fine by me.

Perhaps some kind of builtin device like this would also work?

```Python
class MbbRecord(EpicsDevice):
    def __init__(self, prefix: str, bits: dict[int, str], name: str = ""):
        [setattr(self, bit_name, epics_signal_r(bool, f"{prefix}.B{bit:X}", name = bit_name) for bit, bit_name in bits.items()]
```

I'll probably put something like this in the NSLS2 equivalent of dodal.

DominicOram (changes requested):
Great, thank you! I agree with @coretl about splitting the `DeviceVector` bit into a new issue. I can see some usecases for it but I think I think generally listing each bit is a lot more descriptive

jwlodek:
Ok, removed the device vector abstractions for now, and left the support for 0/1 -> bool signals.

coretl (approved):

coretl:
@jwlodek code changes look fine but tests are failing

coretl (approved):

PR #1250, opened 2026-04-20
refactor: update EigerDetector and OdinIO to match FastCS changes

shihab-dls:
Update to support features of `fastcs-eiger==0.4.0` and `fastcs-odin==0.9.0`.


src/ophyd_async/fastcs/eiger/_arm_logic.py
```diff
@@ -1,14 +1,23 @@
...
+    wait_for_value,
+)
 
 from ._io import EigerDetectorIO
 
 
 class EigerArmLogic(DetectorArmLogic):
-    def __init__(self, detector: EigerDetectorIO) -> None:
+    def __init__(
+        self, detector: EigerDetectorIO, arm_when_ready: Reference[SignalX]
```
shihab-dls: This shouldn't be a `Reference`, as the arm logic isn't a device.

shihab-dls:
`fastcs-eiger` has a `start_writing` signal that passes compression and datatype, so this should be passed instead of `bit_depth`.

This command does not exist on Jungfrau. I should add it first. Both Eiger and Jungfrau can be launched as `XOdinController`s or `xController`s (i.e., with or without `od` as a subdevice). The ophyd-async layers assume the former is deployed, so I think handling datatype and compression passing at the IOC level is reasonable (so no optional bit depth passing at the ophyd level).

coretl (approved):

shihab-dls:
> `fastcs-eiger` has a `start_writing` signal that passes compression and datatype, so this should be passed instead of `bit_depth`.
> 
> This command does not exist on Jungfrau. I should add it first. Both Eiger and Jungfrau can be launched as `XOdinController`s or `xController`s (i.e., with or without `od` as a subdevice). The ophyd-async layers assume the former is deployed, so I think handling datatype and compression passing at the IOC level is reasonable (so no optional bit depth passing at the ophyd level).

I've decided to merge this in for now, such that we have landed on a working solution. Currently, `JungfrauDetector` [does not have](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/fastcs/jungfrau/_detector.py#L23) a top level `detector` and `odin`, and thus adding a `start_writing` command should be accompanied by rectifying the controller interface. As I have been unable to test against a Jungfrau, this PR leaves us in a state where both `Eiger` and `Jungfrau` detectors are functional, but `OdinDataLogic` must concern itself with retrieving their `bit_depth`. This should be done in a future FastCS sprint.

PR #1253, opened 2026-04-21
EPICS: Allow reading partial waveforms (first N elements)

PierreSchnizer:
Large EPICS waveforms (array-valued signals) are currently always transferred in full, even when only a small subset of the data is needed.

For high-rate acquisitions, data is buffered continuously but read out on trigger. In typical use, users select only a small portion of that buffer. Transferring the full array in these cases creates unnecessary network load and does not scale well.

This has a measurable impact in practice: at BESSY II, reading out all ~120 BPMs doubles measurement time due to the volume of data transferred. As a workaround, users are forced to limit the number of devices instead of using all available data sources.

This PR adds support for reading only the first N elements of a waveform (i.e. partial reads of array/spectrum data). It uses the EPICS Channel Access n_elements feature and exposes it via epics_options when defining a signal.

The implementation is currently EPICS-specific. It supports *1* or more elements for channel access and *2* or more elements for pva, as pva only supports only slices above 2 elements. PVA supports slices from start..end, which is limited to 0..end to be compliant what ca provides.

As an example use case, beam position monitors (e.g. Libera Spark, Brilliance) produce turn-by-turn data at ~1 µs intervals and store it in buffers of up to 65k samples. In many applications, only the first few samples are required, making full data transfer unnecessary.


coretl:
I like the concept, but we need to make sure both CA and PVA can do this. `count` seems the right approach for CA, but there is no easy concept in PVA. Claude suggests the following approach:

> For EPICS 7 IOC waveforms served via QSRV2, array slicing can be requested by the client by appending a [subarray field modifier](https://epics.anl.gov/base/R7-0/9-docs/filters.html) to the channel name in the form `PV_NAME.[start:end]` (e.g. `MY:WF.[0:99]`). This is processed server-side by the IOC's `dbChannel` filter infrastructure, which QSRV2 respects, so no server-side changes are needed — the full waveform is never transmitted. For p4p-based servers such as those created by FastCS, there is no equivalent native mechanism; `SharedPV` broadcasts the same value to all clients and does not interpret the channel name. Support would need to be added explicitly by registering a `DynamicProvider` alongside the existing `StaticProvider`, intercepting channel names with the `.[start:end]` suffix, and posting sliced numpy arrays to per-slice `SharedPV` instances whenever the underlying attribute value updates.
 
Would you be able to try this out for PVA as well please?

PierreSchnizer:
Yes I will try to get it running with P4P

PierreSchnizer:
Sorry I was detoured by other tasks.
I will try to implement the comment given above soon

PierreSchnizer:
Sorry for the delay. I expected much more work to verify what Claude suggested.
Yes it works as suggested for 

```python

from p4p.client.thread import Context

pv_name = "MY:WF"
ctx = Context()

start, end = 10, 15
slice = ctx.get(f"{pv_name}.[{start}:{end}]")
```

I have a look how it could be added.

PierreSchnizer:
Backends p4p and aioca are tested that conistent outputs are made for methods `describe`, `read` and `get_value`

I am not sure if set_callback should be addressed too, as it also uses the readpv: see line

https://github.com/hz-b/ophyd-async/blob/e8f37e0b1f085a89e470327829a74b05e874a889/src/ophyd_async/epics/core/_p4p.py#L473

coretl:
> Backends p4p and aioca are tested that conistent outputs are made for methods `describe`, `read` and `get_value`
> 
> I am not sure if set_callback should be addressed too, as it also uses the readpv: see line
> 
> https://github.com/hz-b/ophyd-async/blob/e8f37e0b1f085a89e470327829a74b05e874a889/src/ophyd_async/epics/core/_p4p.py#L473

Yes please, set_callback should also be updated to do the same.

Also, please could you put the tests in `tests/system_tests/epics/core/test_epics_signal_mechanisms.py` and modify `src/ophyd_async/epics/testing/_epics_test_ca_records.db` to have the records you require for the tests?

PierreSchnizer:
set_callback implemented and tested

Moved tests to tests/system_tests/epics/core/test_epics_signal_mechanisms.py
Needed to implement more records than I would need to be compliant with expectations of other tests.

Now epics.testing._devices.EpicsTestCaDevice derives from StandardReadable. Thus this device needs to have a name when instaniated. It will not work without the name.

It would be nice if StandardReadble could check for that (but I don't know how to implement it)

jwlodek:
Should we also allow "Last N elements"? I suppose whether the first or last elements are the most recent readings is an implementation detail at the IOC level.

PierreSchnizer:
In my experience, the number of waveform elements required is application-dependent, and different applications may need different subsets concurrently. For example, one application may display the full turn-by-turn spectrum (64 * 1024 elements or more) from a single beam position monitor (BPM) of the electron storage ring, while another reads data from all 128 BPMs but needs only the first 1024 elements from each.

This PR intentionally limits the implementation to the first N elements. For Channel Access, the n_elements mechanism provides this behavior. PVA also supports more general slicing, but addressing “last N” or arbitrary ranges is outside the scope of this PR.

I have opened a broader discussion about supporting partial reads at  https://github.com/bluesky/ophyd-async/discussions/1254

PierreSchnizer:
> Should we also allow "Last N elements"? I suppose whether the first or last elements are the most recent readings is an implementation detail at the IOC level.

Sorry, I did not address your suggested. 

AFAIK The last N elements can only be addressed using PVA. CA only allows first N elements. 

Personnally, I'd think also only the first N values could be useful. I am happy to extend the patch. If so I would then suggest to provide what PVA provides: start..end. If start != 0 is requested for CA a error would be raised.

src/ophyd_async/epics/core/_util.py
```diff
@@ -40,6 +40,15 @@ class EpicsOptions(Generic[SignalDatatypeT]):
...
     """
 
+    element_count: None | int = None
+    """Epics allows to specify the maximum number of elements to transfer
+
+    Fast devices provide buffers that acquire large data sets. Typically
+    one only needs the beginning of this buffer.
+
+    None is used as standard argument: transfer as many elements as provided
+    """
```
evvaaaa: NIT:

```suggestion
    """For array valued records, specifies the number of elements to
    receive from the beginning of the array.
    
    Default `None` for receiving all elements.

    This allows for the reading only the initial buffer from devices
    which provide large data sets over EPICS.
    """
```

src/ophyd_async/epics/core/_p4p.py
```diff
@@ -383,6 +385,13 @@ async def _store_initial_value(self, pv: str, timeout: float):
         self.initial_values[pv] = await pvget_with_timeout(pv, timeout)
 
     async def connect(self, timeout: float):
+        # options are only available at connetion time
+        if self.options and self.options.element_count is not None:
+            if self.options.element_count <= 1:
```
evvaaaa: NIT:

```suggestion
        if (
            self.options
            and self.options.element_count is not None
            and self.options.element_count <= 1
        ):
```

src/ophyd_async/epics/core/_p4p.py
evvaaaa:
```suggestion
super().__init__(datatype, read_pv, write_pv, options)
if (
    options
    and options.element_count is not None
    and get_origin(self.datatype) is not Array1D
):
    raise TypeError(
        f"Cannot specify `element_count` on a signal {datatype} which isn't `Array1D`"
    )
```

src/ophyd_async/epics/testing/_devices.py
```diff
@@ -90,6 +92,15 @@ class EpicsTestCaDevice(EpicsDevice):
     int16a: A[SignalRW[Array1D[np.int16]], PvSuffix("int16a")]
     int32a: A[SignalRW[Array1D[np.int32]], PvSuffix("int32a")]
     float32a: A[SignalRW[Array1D[np.float32]], PvSuffix("float32a")]
+    # todo: check if it can be combined with int32a above
+    #       if its user tolerate a length of 5 with elements set
+    float32al5: A[SignalRW[Array1D[np.float32]], PvSuffix("float32al5")]
+    # need a separate entry to be compatible with signal test
+    float32al5o3: A[
+        SignalRW[Array1D[np.float32]],
```
evvaaaa, line 100: As a procedural definition, this has only been added to `def epics_signal_r`, not `def epics_signal_rw` but here it's specified on a `SignalRW`.

The `put` performance isn't tested... I would be inclined towards not allowing `SignalRW` to have an `element_count, even though there is presumably a way to do it through pva 🤔

src/ophyd_async/epics/core/_p4p.py
evvaaaa: If we restrict to `SignalR` on instantiation then I don't really know a nice way to do it in the current interface, only

```suggestion
if (
    options
    and options.element_count is not None
):
    if get_origin(self.datatype) is not Array1D:
        raise TypeError(
            "Cannot specify `element_count` on a signal which isn't `Array1D`"
        )
    if write_pv != "":
        raise TypeError(
            "Cannot specify `element_count` on any signal kind other than `SignalR`"
        )
        
```

src/ophyd_async/epics/core/_p4p.py
```diff
@@ -414,8 +423,14 @@ async def put(self, value: SignalDatatypeT | None):
             wait = self.options.wait
         await context().put(self.write_pv, {"value": write_value}, wait=wait)
 
+    def _get_read_pv(self) -> str:
```
evvaaaa: NIT:

```suggestion
    def _get_read_pv_with_element_count(self) -> str:
```

evvaaaa (changes requested):

src/ophyd_async/epics/core/_aioca.py
evvaaaa, line 286: Ditto validation checks on signal type in CA...

evvaaaa (commented):
> PVA also supports more general slicing, but addressing “last N” or arbitrary ranges is outside the scope of this PR.

We could change `EpicsOptions.element_count` to

``` python
slice_from: int | None = None
"""For 1D array valued data, skip indices up until this index in the record.

`None` to start from the beginning.

Only supported with pv access.
"""


slice_to: int | None = None
"""For 1D array valued data, skip indices up after this index in the record.

`None` to retrieve until the end of the array.

Supported in both channel access and pv access.
"""

and then it would just be a case of

https://github.com/hz-b/ophyd-async/blob/d0c6ec3b6dd914dfefea212f8102d72c29abaf5d/src/ophyd_async/epics/core/_p4p.py#L430

``` python
    def _get_read_pv(self) -> str:
        """Read pv with subarray index when requested."""
        if self.options.slice_to is None and self.options.slice_from is None:
            return self.read_pv
        slice_to_string = self.options.slice_to - 1 if self.options.slice_to else ""
        slice_from_string = self.options.slice_from or 0
        return f"{self.read_pv}.[{slice_from_string}:{slice_to_string}]"
```

PR #1258, opened 2026-04-30
Add ciruclar buffer trigger to ContAcqArmLogic

jacob-williamson:
Fixes an issue with the `ContAcqDetector` where the circular buffer was never triggered leading to a timeout error. This seems to have been introduced by the `StandardDetector` rewrite. 

However, we ran into a different timeout error when working on v0.16, where we were awaiting for a set completion that we never got from the circular buffer trigger PV. I've the fixes for both into this PR.

@jwlodek have you seen similar issues?


jwlodek:
I haven't had a chance to update to the beamlines I was using this with to 0.17.0+ so haven't ran into the no trigger issue. Don't recall seeing the other problem. I'll try this out today and let you know of it works

PR #1262, opened 2026-05-05
Add configurable turnaround and ramp up time for PMAC trajectories

LuisFSegalla:
This PR aims at adding the capabilities for configuring both the turnaround and the ramp up time in a trajectory .

The idea is to create a new structure (`PmacScanInfo`) that will store both the ScanSpec spec and the information for the turnarounds. This information is then used to define the minimum turnaround necessary or the time to ramp into the velocity before reaching the first point in the trajectory. Both values can be set to `None` so that Ophyd will use the pre-defined `MIN_TURNAROUND` as the value for the calculations.

Unit tests were added for testing the feature and I also tested it against real hardware.

This might become outdated once we move to [ScanSpec2](https://github.com/bluesky/scanspec/issues/196). But it's a feature that we're currently using for SWIFT development and might be useful for others.


src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -66,20 +73,25 @@ def __init__(self, pmac: PmacIO, name: str = "") -> None:
         super().__init__(name=name)
 
     @AsyncStatus.wrap
-    async def prepare(self, value: Spec[Motor]):
-        path = Path(value.calculate())
+    async def prepare(self, value: PmacScanInfo):
+        spec = value.spec
+        self._ramp_time = value.ramp_time
```
shihab-dls: `self._ramp_time` is only passed to `calculate_ramp_position_and_duration` during `perpare`, so I don't think we need to keep state here, and can just directly pass `value.ramp_time`.

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -66,20 +73,25 @@ def __init__(self, pmac: PmacIO, name: str = "") -> None:
...
+        self._turnaround_time = value.turnaround_time
+        path = Path(spec.calculate())
         slice = path.consume(SLICE_SIZE)
         path_length = len(path)
         motors = slice.axes()
         motor_info = await _PmacMotorInfo.from_motors(self.pmac_ref(), motors)
         ramp_up_pos, ramp_up_time = calculate_ramp_position_and_duration(
-            slice, motor_info, True
+            slice, motor_info, True, self._ramp_time
         )
```
shihab-dls, line 86: Could you please add to the docstring of `calculate_ramp_position_and_duration` the parameter `ramp_time`

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -66,20 +73,25 @@ def __init__(self, pmac: PmacIO, name: str = "") -> None:
...
         self._prepare_context = PmacPrepareContext(
             path=path, motor_info=motor_info, ramp_up_time=ramp_up_time
         )
         await asyncio.gather(
-            self._build_trajectory(motor_info, slice, path_length, ramp_up_time),
+            self._build_trajectory(
+                motor_info, slice, path_length, ramp_up_time, self._turnaround_time
```
shihab-dls: Nit: Given we set `_turnaround_time` as state, I would prefer if helper function accessed `self._turnaround_time`; so, both build and append use the state instead of passing it into `_build_trajectory`.

src/ophyd_async/epics/pmac/_pmac_trajectory.py
```diff
@@ -162,7 +174,9 @@ async def _execute_trajectory(self, path: Path, motor_info: _PmacMotorInfo):
     async def _append_trajectory(
         self, slice: Slice, path_length: int, motor_info: _PmacMotorInfo
     ):
-        trajectory = await self._parse_trajectory(slice, path_length, motor_info)
+        trajectory = await self._parse_trajectory(
+            slice, path_length, motor_info, None, self._turnaround_time
+        )
```
shihab-dls: Nit: A very tiny nit! Give we use positional args here, I would switch around the order of the last two parameters in `_parse_trajectory`, so you only pass `slice, path_length, motor_info, self._turnaround_time`.

shihab-dls (changes requested):

shihab-dls:
Thank you @LuisFSegalla! I just had a few tiny nits, but this looks good. As you said, I think this will be affected largely by `scanspec` changes in the near future. As a note, this PR merging into main will cause `test_rig_bluesky` (ViSR plans) to fail if the `blueapi` pod restarts, so I'll make that change once this is in.

LuisFSegalla:
Thank you for the review @shihab-dls! I addressed the comments as you suggested.

src/ophyd_async/epics/pmac/_pmac_trajectory_generation.py
```diff
@@ -167,6 +168,7 @@ def from_slice(
             kwargs = {}
             if gap == 0 and ramp_up_time:
                 kwargs["ramp_up_time"] = ramp_up_time
+            kwargs["turnaround_time"] = turnaround_time
```
coretl, line 171: In general I prefer:
```python
@dataclass
class Thing:
    a: int
    def __call__(self, b: float): ...

callbacks.append(Thing(a=1))
```
to
```python
def thing(a: int, b: float): ...

callbacks.append(functools.partial(a=1))
```

But that's not a comment for this PR, just cos I noticed it...

shihab-dls (approved):
Sorry for the delay on the review, Luis; This looks great

PR #1263, opened 2026-05-07
refactor: Replace DetectorArmLogic with DetectorAcquireLogic

### LINKED ISSUE #1255: Add `on_stage` boolean to `DetectorArmLogic.arm()`?
I might be doing some wrong assumptions in this issue so I apologize in advance.

I'm trying to implement a `StandardDetector` capable of both emitting live frames after being staged and kicked off, and after a request from the user is emitted, while the device is still armed, start writing data to disk while still sending an update of the latest acquired frame for visualization.

With a little bit of help from Claude I managed to trace the behavior of a `StandardDetector` to the following:

```
stage   → disarm(on_unstage=False) 
prepare → prepare_internal + prepare_unbounded 
kickoff → arm()
complete→ wait for frame count 
unstage → disarm(on_unstage=True)
```

This forces me to do the following for a plan:

```py
        yield from bps.open_run()
        while True:
            yield from bps.stage_all(*detectors)

            for det in detectors:
                yield from bps.prepare(det, write_info, wait=True)
            if not streams_declared:
                # for the first time, declare the stream
                yield from bps.declare_stream(*detectors, name=stream_name)
                streams_declared = True
            
            # kickoff calls arm, which calls the camera API to start acquiring
            yield from bps.kickoff_all(*detectors, wait=True)

            # live view
            name, current_action = yield from rps.wait_for_actions(
                self.action_map, wait_for="set"
            )

            # enable a write signal
            # to notify that we're now writing to disk
            for det in detectors:
                yield from bps.abs_set(det.write_sig, True, wait=True)

            if write_forever:
                # wait for the toggle off action instead of waiting for completion
                name, current_action = yield from rps.wait_for_actions(
                    self.action_map, wait_for="reset"
                )

            yield from bps.complete_all(*detectors, wait=True)
            yield from bps.collect(*detectors)
            yield from bps.unstage_all(*detectors)
            self.clear_and_notify(name, current_action)
```

`rps.wait_for_actions` is a trick I'm using to wait for an update from a viewer while still emitting new frames (there's an internal `buffer` signal that is used to asynchronously get the latest reading set by the logic).

It would be a little bit tidier to have the possibility of just arming the device without necessarely calling a subsequent `kickoff`, as in my view that should be delegated only when there's a commitment to actually writing something. So to discern this `arm` could be provided with an `on_stage` boolean.

It would also save me from calling `stage()` and `unstage()` inside the loop so that I can simply arm the devices outside the loop and when going outside the loop or when the plan is instructed to be stopped the `RunEngine` automatically calls `unstage` for disarming any pending device.

For context, [here](https://github.com/redsun-acquisition/redsun-mimir/blob/refactor/device-attr/src/redsun_mimir/device/mmcore/_logics.py) is a current implementation I'm working on for controlling a third-party package using these logics.

Maybe I can do better on my side without having to change the current API (I'm pretty sure I can but I'm slowly getting acquainted with ophyd-async) but in general it feels more natural to use `arm` in such a manner.

coretl:
# Replace `DetectorArmLogic` with `DetectorAcquireLogic` and four lifecycle hooks

## Context

`DetectorArmLogic` had three abstract methods:

- `arm()` — start acquiring, called from `prepare()`, `kickoff()`, or `trigger()`
- `wait_for_idle()` — wait for idle after the final collection
- `disarm(on_unstage: bool)` — stop acquiring; the flag distinguished `stage()`
  (reset before a new scan) from `unstage()` (end-of-scan teardown)

The `on_unstage` flag was introduced in ADR 0014 to give implementations a way
to distinguish the two call sites, but the single method still forced those two
concerns into one place. More importantly, there was no hook at `stage()` time
that could perform *different* work from `unstage()`. A detector that should be
armed once at `stage()` and then triggered multiple times (e.g. an Eiger in
step-scan mode, or a continuously-acquiring detector) had no clean way to
express this; the closest workaround was to put the arm call inside `disarm`
and branch on `on_unstage`, which was confusing.

## Decision

`DetectorArmLogic` is renamed `DetectorAcquireLogic` and its interface is
replaced with four named hooks that map directly to the `StandardDetector`
lifecycle:

| Hook | Called from | Purpose |
|---|---|---|
| `ensure_ready()` | `stage()` | Put the detector into a known idle state before a scan. |
| `start_acquiring()` | `prepare()`, `kickoff()`, `trigger()` | Start the detector acquiring. |
| `wait_for_idle()` | after final collection | Wait for the detector to return to idle. |
| `ensure_stopped()` | `unstage()` | Stop the detector and perform end-of-scan cleanup. |

`ensure_ready` has a concrete default that delegates to `ensure_stopped`. This
is correct for the common case where stage-time reset and scan-end teardown are
identical (e.g. calling `stop_busy_record`). Subclasses that need different
behaviour at stage time (such as arming the detector once and keeping it armed
across multiple kickoff/complete cycles) override `ensure_ready` independently.

`AreaDetector.__init__` renames its `arm_logic` keyword argument to
`acquire_logic` for consistency.

## Consequences

All existing `DetectorArmLogic` subclasses must be updated:
- rename the base class to `DetectorAcquireLogic`
- rename `arm()` to `start_acquiring()`
- split `disarm(on_unstage)` into `ensure_stopped()` (and optionally override
  `ensure_ready()` if stage-time behaviour differs)

This is a clean breaking change; the library is in alpha and no compatibility
shim is provided.

Fixes #1255


jwlodek (approved):
LGTM

PR #1264, opened 2026-05-11
Xspress odin dev

LuisFSegalla:
Adding a module for the FastCS-Xspress interface.

This re-implements some of FastCS-Odin's features, but updates them for Xspress's specific needs.


LuisFSegalla:
@shihab-dls I rewrote parts of the code after our conversation yesterday. I think in the end it might be easier to just have Xspress have it's own data_logic, specifically for the `prepare_unbound` method. I believe there are enough differences in Xspress that it's worth having it's own method and just inherit the base FastCS data_logic for the other ones.

src/ophyd_async/fastcs/xspress/_data_logic.py
```diff
@@ -0,0 +1,63 @@
...
+        odin: XspressOdinIO,
+        detector_bit_depth: SignalR[int] | None = None,
+        compression: str | None = None,
+    ):
+        self.path_provider = path_provider
+        self.odin = odin
+        self.detector_bit_depth = detector_bit_depth
+        self.compression = compression
+
+    async def prepare_unbounded(self, datakey_name: str) -> StreamableDataProvider:
```
coretl, line 27: The fact you need to override this method from the baseclass means we didn't get the baseclass method right. When you have this working, before we merge, we should reconsider the baseclass to see if we can provide init params so this method doesn't need overriding.

src/ophyd_async/fastcs/xspress/_detector.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ._acquire_logic import XspressAcquireLogic
+from ._data_logic import XspressOdinDataLogic
+from ._io import XspressDetectorIO
+from ._trigger_logic import XspressTriggerLogic
+from ._xsp_odin_io import XspressOdinIO
+
+
+class XspressTriggerInfo(TriggerInfo):
+    chunk: PositiveInt = Field(default=1)
```
coretl: Please don't specify this via prepare, the Signal should be the ground truth of this value. Use restore_settings to set this value once at the start of the plan. This will be the pattern we use when we have a generic scan plan after 1.0

src/ophyd_async/fastcs/xspress/_detector.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        super().__init__(name=name, connector=connector)
+
+    @AsyncStatus.wrap
+    async def prepare(self, value: XspressTriggerInfo) -> None:  # pyright: ignore[reportIncompatibleMethodOverride]
+        datakey_name = self.name + self._data_logics[0].datakey_suffix
+
+        await self.od.file_prefix.set(
+            self._data_logics[0].path_provider(datakey_name).filename  # pyright: ignore[reportAttributeAccessIssue]
+        )
```
coretl, line 50: @shihab-dls don't all fastcs odin things need to do this? Isn't it in the OdinDataLogic?

coretl (changes requested):

src/ophyd_async/fastcs/xspress/_data_logic.py
LuisFSegalla: So, I think the base class works well for almost all the cases, it's just that for Xspress there are a few things that need to be treated differently, for example:

- Datatype will always be `uint32` and we don't have access to a value for `detector_bit_depth`
- Compression must always be set to blosc
- Due to the way Xspress counts the frames (not in a round robin, but by summing the values for all the FPs) I had to use a different PV to track the actual number of frames that were written.

Not sure if it's worth changing the base class so much or if it's just better to have `XspressDataLogic` rewrite this method.

src/ophyd_async/fastcs/xspress/_detector.py
LuisFSegalla: Sorry, I don't think I understand this comment. What you mean by `the Signal should be the ground truth of this value`?
Maybe a bit more of context might help with my idea for this class. What I wanted here was to give the user a way to configure how many frames will be 'bundled' together before writing to disk. This is something that we already do with Malcolm and it give us a chance of changing how many frames we expect before writing and this value will change with each scan, usually based on the acquisition period. So the idea is that in the plan, the user can add some logic to calculate the chunk size and pass it down in `prepare`.

src/ophyd_async/fastcs/xspress/_detector.py
shihab-dls: [It is](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/fastcs/odin/_data_logic.py#L39):

```python
self.odin.acquisition_id.set(filename)
```
This is unclear though, as you'd need to know that `acquistion_id` internally sets `file_prefix`. However, testing on i15 showed that we must actually clear acquisition_id, and set file_prefix afterwards, for Odin to be happy, so I'll open up a PR to make those changes on `OdinDataLogic`.

shihab-dls: https://github.com/bluesky/ophyd-async/pull/1310 (PR with data logic changes)

PR #1265, opened 2026-05-12
Expose the HDF PandA Logic classes.

jwlodek:
In cases where I want an `HDFPandA`, but with one or more of the logics replaced, I'd like to be able to import the ones I plan to keep as-is.


jwlodek:
No longer necessary - we've taken a different approach, using an ephemeral flyer to do what we need.

PR #1268, opened 2026-05-15
Replace AreaDetector writer parameters with ADWriterFactory

### LINKED ISSUE #1267: Replace `AreaDetector` writer parameters with `ADWriterFactory`
Expanded from a conversation with @oliwenmandiamond 

## Background

`AreaDetector.__init__` currently accepts three writer-related keyword parameters:

```python
det = adaravis.AravisDetector(
    "PREFIX:",
    path_provider=path_provider,
    writer_type=NDFileHDF5IO,
    writer_suffix="HDF1:",
)
```

These three parameters always travel together and are only meaningful as a unit.
Passing them individually obscures the relationship and leaves callers responsible
for keeping them consistent.

## Problems with the current API

1. **Coupled triple parameters** — `path_provider`, `writer_type`, and
   `writer_suffix` form a logical unit but are spelled out separately every time.

2. **No first-class multiple writers** — adding a second HDF writer (e.g. for a
   second ROI) requires calling `add_detector_logics()` after construction and
   duplicating information already present at the call site.

3. **Hardcoded array metadata signals** — the writer logic always reads
   `driver.array_size_{x,y,z}`, `driver.data_type`, and `driver.color_mode`.
   ROI plugins and processing plugins expose their own shape/type signals, which
   are currently ignored.

4. **Deferred driver construction** — subclasses like `AravisDetector` create
   their driver internally.  Any `NDArrayDescription(data_type_signal=driver.data_type)`
   passed at call site would need the driver to already exist, which it does not.

## Proposed solution

Replace the three writer parameters with a `*writer_factories` varargs of
`ADWriterFactory` instances.

### `ADWriterFactory`

```python
@dataclass
class ADWriterFactory(Generic[NDPluginFileIOT]):
    writer_cls: type[NDPluginFileIOT]
    writer_suffix: str
    writer_name: str          # attribute name on the detector: det.hdf, det.hdf1 …
    datakey_suffix: str
    array_description: NDArrayDescription | Callable[[ADBaseIO], NDArrayDescription] | None
    data_logic_factory: Callable[...]

    @staticmethod
    def hdf(path_provider, writer_suffix="HDF1:", writer_name="hdf",
            datakey_suffix="", array_description=None) -> ADWriterFactory[NDFileHDF5IO]: ...
    @staticmethod
    def jpeg(...): ...
    @staticmethod
    def tiff(...): ...
```

### New call site — single writer

```python
det = adaravis.AravisDetector(
    "PREFIX:",
    adcore.ADWriterFactory.hdf(path_provider),
)
det.hdf  # NDFileHDF5IO instance
```

### New call site — multiple ROI writers

Because `AravisDetector` creates its driver internally, `data_type_signal` and
`color_mode_signal` are not yet available at the outer call site.
`array_description` therefore accepts a **callable** that receives the
fully-constructed driver:

```python
roi1 = adcore.NDROIIO("PREFIX:ROI1:")
roi2 = adcore.NDROIIO("PREFIX:ROI2:")

det = adaravis.AravisDetector(
    "PREFIX:",
    adcore.ADWriterFactory.hdf(
        path_provider,
        writer_suffix="HDF1:", writer_name="hdf1", datakey_suffix="-roi1",
        array_description=lambda driver: adcore.NDArrayDescription(
            shape_signals=(roi1.size_y, roi1.size_x),
            data_type_signal=driver.data_type,
            color_mode_signal=driver.color_mode,
        ),
    ),
    adcore.ADWriterFactory.hdf(
        path_provider,
        writer_suffix="HDF2:", writer_name="hdf2", datakey_suffix="-roi2",
        array_description=lambda driver: adcore.NDArrayDescription(
            shape_signals=(roi2.size_y, roi2.size_x),
            data_type_signal=driver.data_type,
            color_mode_signal=driver.color_mode,
        ),
    ),
    plugins={"roi1": roi1, "roi2": roi2},
)
det.hdf1  # NDFileHDF5IO for ROI 1
det.hdf2  # NDFileHDF5IO for ROI 2
```

The callable receives the driver at `AreaDetector.__init__` time—the earliest
point at which all driver signals exist.  This also covers plugins that remap
`data_type` or `color_mode`, since every field of `NDArrayDescription` can be
overridden independently.

### Omitting file writing

Callers that do not want file writing simply omit all factory arguments (no more
`writer_type=None`):

```python
det = adaravis.AravisDetector("PREFIX:")
```

### `prefix` placement

`prefix` is placed as an ordinary positional parameter before `*writer_factories`
so it can be passed by position.  Providing factories without a `prefix` raises a
clear `ValueError`.

coretl:
Fixes #1267 

Note that `det.writer` -> `det.hdf`. This doesn't have to be so

### DIFF
--- a/src/ophyd_async/epics/adcore/_data_logic.py
+++ b/src/ophyd_async/epics/adcore/_data_logic.py
@@ -1,9 +1,8 @@
 import asyncio
-from collections.abc import Mapping, Sequence
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass
-from enum import Enum
 from pathlib import PureWindowsPath
-from typing import Any
+from typing import Any, Generic
 from xml.etree import ElementTree as ET
 
 import numpy as np
@@ -31,6 +30,7 @@ from ._io import (
     NDFileHDF5IO,
     NDPluginBaseIO,
     NDPluginFileIO,
+    NDPluginFileIOT,
 )
 from ._ndattribute import NDAttributeDataType, NDAttributePvDbrType
 
@@ -53,28 +53,38 @@ class PluginSignalDataLogic(DetectorDataLogic):
 
 @dataclass
 class NDArrayDescription:
+    """Signals that describe the shape and data type of an NDArray frame.
+
+    :param shape_signals: Signals providing the frame dimensions (e.g.
+        ``size_y``, ``size_x``).  Zero-valued entries are filtered out, so it
+        is safe to include ``array_size_z`` for 2-D detectors.
+    :param data_type_signal: Signal providing the pixel data type.
+    :param color_mode_signal: Signal providing the colour mode (MONO or RGB1).
+    """
+
     shape_signals: Sequence[SignalR[int]]
     data_type_signal: SignalR[ADBaseDataType]
     color_mode_signal: SignalR[ADBaseColorMode]
 
 
 async def get_ndarray_resource_info(
-    description: NDArrayDescription,
+    array_description: NDArrayDescription,
     data_key: str,
     parameters: dict[str, Any],
     frames_per_chunk: int = 1,
 ) -> StreamResourceInfo:
     # Grab the dimensions and datatype of the NDArray
     shape, datatype, color_mode = await asyncio.gather(
-        asyncio.gather(*[sig.get_value() for sig in description.shape_signals]),
-        description.data_type_signal.get_value(),
-        description.color_mode_signal.get_value(),
+        asyncio.gather(*[sig.get_value() for sig in array_description.shape_signals]),
+        array_description.data_type_signal.get_value(),
+        array_description.color_mode_signal.get_value(),
     )
     # Remove entries in shape that are zero
     shape = [x for x in shape if x > 0]
     if datatype is ADBaseDataType.UNDEFINED:
         raise ValueError(
-            f"{description.data_type_signal.source} is blank, this is not supported"
+            f"{array_description.data_type_signal.source} is blank, "
+            "this is not supported"
         )
     if color_mode == ADBaseColorMode.RGB1:
         shape = [3, *shape]
@@ -155,8 +165,7 @@ async def prepare_file_paths(
 class ADHDFDataLogic(DetectorDataLogic):
     """Data logic for AreaDetector HDF5 writer plugin.
 
-    :param shape_signals: Signals that provide the shape of the NDArray.
-    :param data_type_signal: Signal that provides the data type of the NDArray.
+    :param array_description: Signals describing the NDArray shape and data type.
     :param path_provider: Callable that provides path information for file writing.
     :param driver: The AreaDetector driver instance.
     :param writer: The NDFileHDFIO plugin instance.
@@ -164,7 +173,7 @@ class ADHDFDataLogic(DetectorDataLogic):
     :param datakey_suffix: Suffix to append to the data key for the main dataset
     """
 
-    description: NDArrayDescription
+    array_description: NDArrayDescription
     path_provider: PathProvider
     driver: ADBaseIO
     writer: NDFileHDF5IO
@@ -199,7 +208,7 @@ class ADHDFDataLogic(DetectorDataLogic):
         )
         # Return a provider that reflects what we have made
         main_dataset = await get_ndarray_resource_info(
-            description=self.description,
+            array_description=self.array_description,
             data_key=datakey_name,
             parameters={"dataset": "/entry/data/data"},
             frames_per_chunk=frames_per_chunk,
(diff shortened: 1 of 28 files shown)


DominicOram (commented):
From the perspective of changing how we do instantiation this looks sensible and I'm happy with it. I have no opinion on whether it solves @oliwenmandiamond's usecase or on the details of the code changes in `ophyd-async`. If you would like a more indepth review i'd be happy to but it would be a week or so until I get round to it

jwlodek:
I will have an extensive look at this tomorrow - quick off the top of my head question, is the reason we don't just pass in the data logics directly is to avoid needing to instantiate the IO? Would a classmethod factory function on the data logic classes work for this? Something like `ADHDFDataLogic.factory`? Maybe with a free function for any common portions of the logic?

coretl:
> I will have an extensive look at this tomorrow - quick off the top of my head question, is the reason we don't just pass in the data logics directly is to avoid needing to instantiate the IO? Would a classmethod factory function on the data logic classes work for this? Something like `ADHDFDataLogic.factory`? Maybe with a free function for any common portions of the logic?

That would probably work too, but you'd need 2 factories for the multipart logic (for tiff and jpeg). Probably still makes sense to have an `ADWriterFactory` dataclass or protocol to simplify the type hints for the AreaDetector subclasses, so probably wouldn't make much of a different to the code, and requires people knowing that you need to use a multipartlogic for tiff or jpeg.

src/ophyd_async/epics/adcore/_data_logic.py
```diff
@@ -293,54 +301,201 @@ def get_hinted_fields(self, datakey_name: str) -> Sequence[str]:
...
             )
-        case ADWriterType.TIFF:
-            writer = NDPluginFileIO(f"{prefix}{writer_suffix or 'TIFF1:'}")
-            data_logic = ADMultipartDataLogic(
-                description=description,
+        data_logic = self.data_logic_factory(writer, array_description, driver, plugins)
+        return writer, data_logic
+
+    @staticmethod
+    def hdf(
```
oliwenmandiamond, line 379: Nit: Should these be classmethods?

oliwenmandiamond (approved):
Looks good to me. One question / nit

src/ophyd_async/epics/adcore/_data_logic.py
coretl: Pyright is not happy with classmethods on unbound generics, you can fix it with a type hint, but staticmethod gets around this issue, and we never subclass ADWriterFactory so it doesn't make any difference to us. I had an AI summarize the differences if you're interested:
[classmethod_generic_demo.py](https://github.com/user-attachments/files/28103935/classmethod_generic_demo.py)

tests/unit_tests/epics/adcore/test_detector.py
```diff
@@ -1,23 +1,124 @@
...
+            "PREFIX:",
+            adcore.ADWriterFactory.hdf(static_path_provider),
+            name="det",
+        )
+    set_mock_value(det.driver.array_size_x, 10)
+    set_mock_value(det.driver.array_size_y, 10)
+    set_mock_value(det.driver.data_type, adcore.ADBaseDataType.UNDEFINED)
+    set_mock_value(det.driver.color_mode, adcore.ADBaseColorMode.MONO)
+    set_mock_value(det.get_plugin("hdf", adcore.NDFileHDF5IO).file_path_exists, True)
+    await det.stage()
```
shihab-dls:
```suggestion
    set_mock_value(det.driver.data_type, adcore.ADBaseDataType.UNDEFINED)
    set_mock_value(det.get_plugin("hdf", adcore.NDFileHDF5IO).file_path_exists, True)
```

tests/unit_tests/epics/adcore/test_detector.py
```diff
@@ -1,23 +1,124 @@
...
+            "PREFIX:",
+            adcore.ADWriterFactory.hdf(static_path_provider),
+            name="det",
+        )
+    set_mock_value(det.driver.array_size_x, 10)
+    set_mock_value(det.driver.array_size_y, 10)
+    set_mock_value(det.driver.data_type, adcore.ADBaseDataType.UINT8)
+    set_mock_value(det.driver.color_mode, adcore.ADBaseColorMode.BAYER)
+    set_mock_value(det.get_plugin("hdf", adcore.NDFileHDF5IO).file_path_exists, True)
+    await det.stage()
```
shihab-dls:
```suggestion
    set_mock_value(det.driver.color_mode, adcore.ADBaseColorMode.BAYER)
    set_mock_value(det.get_plugin("hdf", adcore.NDFileHDF5IO).file_path_exists, True)
```

shihab-dls (changes requested):
Implementation looks great; a couple of redundant operations in the added tests, so I'll request changes, push the changes, and then approve!

shihab-dls (approved):

jwlodek (approved):

PR #1273, opened 2026-05-19
Write all frames in first h5 file until VDS is working

### LINKED ISSUE #1272: Workaround VDS issues in Odin
See https://github.com/bluesky/ophyd-async/issues/1274 for context.

Until virtual dataset generation is working properly, we should ensure that all data from a scan goes into a single file. This means it is safe to assume that all the data goes into `*_000001.h5` and the nxs file can point at that.

To do this:
- Ideally there would only be one file writer connected. This can be done by changing `odinDataCount` to 1 inside `values.yaml` in `ixx-services` repos
- `Files Per Block` should be set to a very large number, to ensure that even if there are multiple file writers, only the first will ever write anything.
- ~~Blocks Per File should be set to 0, meaning file writers will never create a new file.~~ Not necessary if `Files Per Block` is large enough

`Files Per Block` is currently set to 1000 inside ophyd-async. Changing this number breaks things because `Block Size` is still 1000. I think we should set `Block Size` instead, or both, depending on if `Block Size` propagates down to `Files Per Block`. If we set these PVs to 100,000 then only the first file writer will ever write anything even if there a multiple file writers connected.

This is a temporary fix and comments should be added linking to this issue to ensure the code is amended once vds generation is working. 

## Acceptance Criteria
- When you run a scan, frames are only written into `*_000001.h5`
- The `.nxs` file produced by the scan contains all the frames.

jacob-williamson:
Fixes #1272

### DIFF
--- a/src/ophyd_async/fastcs/odin/_data_logic.py
+++ b/src/ophyd_async/fastcs/odin/_data_logic.py
@@ -43,7 +43,7 @@ class OdinDataLogic(DetectorDataLogic):
             self.odin.fp.frames.set(0),
             self.odin.block_size.set(
                 100000  # Needed temporarily, see https://github.com/bluesky/ophyd-async/issues/1272
-            ),  # Need to check if this propogates to odin.fp.process_frames_per_block
+            ),
         )
         # Start writing
         await self.odin.fp.start_writing.trigger()
--- a/src/ophyd_async/fastcs/odin/_data_logic.py
+++ b/src/ophyd_async/fastcs/odin/_data_logic.py
@@ -41,7 +41,9 @@ class OdinDataLogic(DetectorDataLogic):
             self.odin.fp.data_compression.set("BSLZ4"),
             self.odin.fp.data_datatype.set(datatype),
             self.odin.fp.frames.set(0),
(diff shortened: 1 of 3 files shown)


shihab-dls (approved):
This looks fine as a stopgap

src/ophyd_async/fastcs/odin/_data_logic.py
```diff
@@ -41,7 +41,9 @@ async def prepare_unbounded(self, datakey_name: str) -> StreamableDataProvider:
             self.odin.fp.data_compression.set("BSLZ4"),
             self.odin.fp.data_datatype.set(datatype),
             self.odin.fp.frames.set(0),
-            self.odin.fp.process_frames_per_block.set(1000),
+            self.odin.block_size.set(
+                100000  # Needed temporarily, see https://github.com/bluesky/ophyd-async/issues/1272
+            ),
```
shihab-dls, line 46: Is this arbitrarily large? Or did something inform this?

jacob-williamson: Arbitrarily large

Issue #1241, opened 2026-04-13
StandardFlyer -> StandardFlyable


coretl:
Convert the existing `StandardFlyer` to a mix-in `StandardFlyable` that has a cached property `flyable_logic: FlyableLogic` the user can fill in (like `StandardMovable`).

It should:
- have a `prepare` method that takes a `scanspec.Path` that calls `FlyableLogic.on_prepare`
- have a `kickoff` method that calls `FlyableLogic.on_kickoff`
- have a `complete` method that calls `FlyableLogic.on_complete`
- have `stage` and `unstage` call `FlyableLogic.stop`
- be used both in ephemeral flyers like `StaticPcompTriggerLogic` and Devices like `Motor`
- have a `FlyableLogic.with_device()` that provides an easy way to make ephemeral Devices in a plan

## Acceptance Criteria
- Code, docs, tests written

coretl:
`stop` shadows a verb, maybe think of another name for this, also in StandardMovable

PR #1275, opened 2026-05-20
Update PmacTrajectoryTriggerLogic to use FlyerController protocol

iain-hall:
PmacTrajectoryTriggerLogic should implement  FlyerController protocol rather than Flyable to be used with StandardFlyer.


iain-hall:
Problem will be resolved in : https://github.com/bluesky/ophyd-async/issues/1241
For now, we will workaround in spectroscopy-bluesky by either using PmacTrajectoryTriggerLogic directly in `prepare` and `kickoff` or use `#ignore` on the offending line to keep PyRight happy.

PR #1280, opened 2026-06-03
SoftSignalBackend wrapping arbitrary callables

### LINKED ISSUE #1279: No clear way to use Ophyd-Async without Tango or Epics
After Ophyd-Async is released, I foresee questions from users/beamline scientists like "I want to add a signal to my device but there is no tango attribute / epics pv" or "We don't use Tango/EPICS, but we want to use Bluesky, can we still use Ophyd-Async?". The latter question has already been asked by attendees of the Bluesky community workshop coming from smaller lab based groups and universities. There is currently no guidance on how this can be done short of asking users to write their own signal backends.

Potential use cases:
- Adding a custom analysis script as a readable signal
- read/write signals to call or subscribe to an AI model
- Build a signal for a matlab device or for a device which has its own python api
- Wrap pre-existing class methods to avoid re-writing device classes after ophyd-async adoption

I propose one of two changes:

1. We add `CallableSignalBackend` to core which is a standard backend to interface with arbitrary read and write callable functions. 

OR

2. We modify `SoftSignalBackend` to support wrapping callables. This will require some sensible defaults for `timestamp`, `alarm severity`, metadata and a subscription interface.

@jacopoabramo @coretl

burkeds:
**Add getter, setter, and poll_period to SoftSignalBackend**

Closes #1279

## Motivation

Users working outside of EPICS/Tango — wrapping third-party Python APIs, calling analysis scripts, or interfacing with devices that have their own Python drivers — currently have no supported path to create signals without writing a full custom `SignalBackend`. This was raised at the Bluesky community workshop and is a known friction point for smaller lab-based groups.

## Changes

Three new keyword-only parameters are added to `SoftSignalBackend`:

- `getter: Callable[[], T | Awaitable[T]]` — called on `get_value` and `get_reading` to fetch the current value from an external source. When `poll_period` is also set, called periodically while a subscription is active.
- `setter: Callable[[Any], T | None | Awaitable[T | None]]` — called on `put` in place of the default internal store update. May return the settled value the device actually reached; if it returns `None` and a getter is configured, the getter is called immediately to refresh the cache.
- `poll_period: float` — interval in seconds at which the getter is called while a subscription is active. Requires `getter` to be set.

All three parameters are optional. When none are provided, behaviour is identical to the existing `SoftSignalBackend`, so there are no breaking changes.

The internal `self.reading` store remains the single source of truth. The getter updates it rather than bypassing it, keeping subscriptions and cached reads coherent.

Modified `soft_signal_rw` and `soft_signal_r_and_setter` to take setter, getter, and poll_period arguments.

## Design notes

- `put` is typed as `Any` to allow heterogeneous setter input types (e.g. a config class, a raw integer count, a command string) that differ from the signal's `SignalDatatypeT`.
- The polling task is started and cancelled in `set_callback`, tied to subscription lifetime.
- Transient getter failures during polling are swallowed and retried on the next interval, consistent with EPICS/Tango backend behaviour.
- `get_setpoint` deliberately does not call the getter — it returns the last requested value, not the live device readback.


src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,18 +144,51 @@ def __init__(
...
         self.set_value(self.initial_value)
         super().__init__(datatype)
 
+    async def _call_getter(self) -> SignalDatatypeT:
+        if self._getter is None:
+            raise RuntimeError("No getter configured")
+        result = self._getter()
+        if isinstance(result, Awaitable):
+            result = await result
+        return self.converter.write_value(result)
```
oliwenmandiamond: Can use bluesky.utils maybe_await here

```python
from bluesky.utils import maybe_await
...
return self.converter.write_value(await maybe_await(self._getter()))
```

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,18 +144,51 @@ def __init__(
...
+
+    async def _call_setter(self, value: Any) -> SignalDatatypeT | None:
+        if self._setter is None:
+            raise RuntimeError("No setter configured")
+        result = self._setter(value)
+        if isinstance(result, Awaitable):
+            result = await result
+        return result
```
oliwenmandiamond: Same here

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -160,19 +205,32 @@ def source(self, name: str, read: bool) -> str:
     async def connect(self, timeout: float):
         pass
 
-    async def put(self, value: SignalDatatypeT | None) -> None:
+    async def put(self, value: Any) -> None:
         write_value = self.initial_value if value is None else value
-        self.set_value(write_value)
+        if self._setter is not None:
+            settled = await self._call_setter(write_value)
```
coretl:
```suggestion
            written_value = await self._call_setter(write_value)
```

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -160,19 +205,32 @@ def source(self, name: str, read: bool) -> str:
...
+            if settled is not None:
+                self.set_value(self.converter.write_value(settled))
+            elif self._getter is not None:
+                self.set_value(await self._call_getter())
+            else:
+                self.set_value(self.converter.write_value(write_value))
```
coretl: set_value will call the converter, don't need to call it here

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,18 +144,51 @@ def __init__(
...
         self.set_value(self.initial_value)
         super().__init__(datatype)
 
+    async def _call_getter(self) -> SignalDatatypeT:
+        if self._getter is None:
+            raise RuntimeError("No getter configured")
+        result = self._getter()
+        if isinstance(result, Awaitable):
+            result = await result
+        return self.converter.write_value(result)
```
coretl: again, no need to run the converter on this, it will be converted in set_value

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,18 +144,51 @@ def __init__(
...
         self.reading: Reading[SignalDatatypeT]
         self.callback: Callback[Reading[SignalDatatypeT]] | None = None
+        self._getter = getter
+        self._setter = setter
+        self._poll_period = poll_period
+        self._poll_task: asyncio.Task | None = None
         self.set_value(self.initial_value)
         super().__init__(datatype)
 
+    async def _call_getter(self) -> SignalDatatypeT:
```
coretl: I would prefer this function be called `_update_value_from_getter` which calls `set_value` internally, and is a no-op if `self._getter is None`

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,18 +144,51 @@ def __init__(
...
+        if self._getter is None:
+            raise RuntimeError("No getter configured")
+        result = self._getter()
+        if isinstance(result, Awaitable):
+            result = await result
+        return self.converter.write_value(result)
+
+    async def _call_setter(self, value: Any) -> SignalDatatypeT | None:
```
coretl: I would put this code in `put`

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -184,4 +242,10 @@ def set_callback(self, callback: Callback[Reading[SignalDatatypeT]] | None) -> N
             raise RuntimeError("Cannot set a callback when one is already set")
         if callback:
             callback(self.reading)
+            if self._poll_period is not None:
+                self._poll_task = asyncio.get_event_loop().create_task(self._poll())
```
coretl:
```suggestion
                self._poll_task = asyncio.create_task(self._poll())
```

oliwenmandiamond (commented):
You will also need to update `soft_signal_r_and_setter` and `soft_signal_rw` to accept the new optional getter, setter and poll arguments to pass on to `SoftSignalBackend`

tests/unit_tests/core/test_soft_signal_backend.py
```diff
@@ -204,3 +204,234 @@ async def test_soft_signal_coerces_numpy_types():
...
+    await backend.connect(timeout=1)
+    await backend.put(5.0)
+    store[0] = 99.0
+    assert await backend.get_setpoint() == 5.0
+    assert await backend.get_value() == 99.0
+
+
+async def test_soft_signal_backend_setter_homogeneous_types():
+    store = [0.0]
+    backend = SoftSignalBackend(float, setter=lambda v: store.__setitem__(0, v))
```
coretl: `__setitem__`?

jacopoabramo:
Looks good so far. Testing it locally with the pymmcore-plus package. It would have definetely saved me a lot of time.

I'm guessing some customization (such as changing the `source`) would still require subclassing but it's definetely an improvement

jacopoabramo:
This PR only addresses procedural device declaration, right? Is a declarative approach achievable somehow?

src/ophyd_async/core/_soft_signal_backend.py
burkeds: Don't we want to avoid explicit dependencies on bluesky in ophyd-async?

burkeds:
> This PR only addresses procedural device declaration, right? Is a declarative approach achievable somehow?

Right, this addresses procedural only. Is a declarative approach even desirable in this case? Seems like you will likely need to write getters and setters anyway.

tests/unit_tests/core/test_soft_signal_backend.py
```diff
@@ -204,3 +204,234 @@ async def test_soft_signal_coerces_numpy_types():
...
+
+    configs_received = []
+    store = [0.0]
+
+    def config_setter(config: MotorConfig) -> float:
+        configs_received.append(config)
+        store[0] = config.velocity
+        return config.velocity
+
+    backend = SoftSignalBackend(float, setter=config_setter)
```
coretl: Hmm, is it a requirement for `Signal.set` to take a different datatype to its internal datatype? I know we do this for type coercion (take a string or a float, convert, store as float), but this example seems to promote this to *requiring* a different datatype. This will break typing...

coretl (changes requested):

coretl:
> This PR only addresses procedural device declaration, right? Is a declarative approach achievable somehow?

I think for declarative we should wait for the FastCS example. Passing setters and getters via a declarative approach gets messy fast...

jacopoabramo:
> Seems like you will likely need to write getters and setters anyway.

In my case, using pymmcore-plus, there is a `Core` object that does the heavy lifting in loading dedicated C++ DLLs and manages communication with the actual devices. I think that with declarative it would avoid some boilerplate code in having to create everytime the core object and dispatching it to a ... `DeviceConnector`, I think? From the declarative example the closest thing I can imagine is something like:

```py
class MMCamera(StandardReadable, MMCoreDevice)
    ...
```

where core creation is delegated to `MMCoreDevice`. But then again for wrapping python packages with a more traditional API design this is fine so I'll just wait for the progress of FastCS, this works fine too

src/ophyd_async/core/_soft_signal_backend.py
oliwenmandiamond: Ophyd-async already imports from bluesky utils

https://github.com/bluesky/ophyd-async/blob/1e07267fb12ff511b248a77a5ed5b6ac50364ad9/src/ophyd_async/plan_stubs/_wait_for_awaitable.py#L4

I don't see why we can't do that again here but correct me if wrong @coretl

src/ophyd_async/core/_soft_signal_backend.py
jacopoabramo: EDIT: moved comment in upper section

src/ophyd_async/core/_soft_signal_backend.py
jacopoabramo: I would agree that ophyd-async should be usable stand alone without having to necessarely depend on bluesky as a dependency. If the things currently used are type annotations it should be fine to declare them locally but from the looks of it it's not much of an option, [there's a lot of imports from bluesky](https://github.com/search?q=repo%3Abluesky%2Fophyd-async%20from%20bluesky&type=code).

src/ophyd_async/core/_soft_signal_backend.py
burkeds: What exactly is the meaningful change here? Do you prefer `written_value` as the variable name?

coretl: yes

src/ophyd_async/core/_soft_signal_backend.py
burkeds:
```python
    async def _update_value_from_getter(self) -> SignalDatatypeT | None:
        if self._getter is None:
            return
        result = self._getter()
        if isinstance(result, Awaitable):
            result = await result
        self.set_value(result)
```

So something like this? You would prefer not to raise an error if called when `_getter` is None?

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: Yeah, using a helper method is cleaner. changed.

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: I don't think I quite understand. How does it promote requiring something?

If `CallableSignalBackend.put` calls any callable then it must be typed `Any` or perhaps `T`. This requires that `Signal.set` also be typed `Any` or `T`.

src/ophyd_async/core/_soft_signal_backend.py
coretl: yes, then you can remove the `if self._getter is not None` check from the call sites

tests/unit_tests/core/test_soft_signal_backend.py
coretl: Example:
```python
sig = soft_signal_rw(float, setter=config_setter)
await sig.set(32.4)  # pyright says ok, runtime says ok
value = await sig.get_value()  # value is float
await sig.set(MotorConfig(...))  # pyright says bad, runtime says ok
value = await sig.get_value()  # value is float
```
I don't think I want to promote a pattern where the set argument type is different from the value type

src/ophyd_async/core/_soft_signal_backend.py
burkeds: Ok since there is already lots of bluesky dependency I will make the change.

src/ophyd_async/core/_soft_signal_backend.py
burkeds: done

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: @coretl What do you suggest then? If we say that the setter can only take one argument of the same type as the value carried by the function, then that strictly limits the types of operations we can reasonably support.

jacopoabramo: I have viewed a few third-party libraries for device control, not enough to constitute as an expert on the matter, but I haven't found so far the situation where an entire data structure is passed down to a function controlling a device. I've seen more use case of APIs with multiple parameters, but in that case using `functools.partial` to bind other arguments should be sufficient.

@burkeds have you encountered these situations before?

burkeds: Yes. One of my beamlines creates a "scan config" object like this:

```python
class Config:
  detector: Any
  start: Any
  stop: Any
  int_time: Any

config = Config(...)
do_stuff(config)
```

This config object is then passed around to scan scripts and other scripts.

Or lets say someone has a script which initializes a detector. The script probably takes a half dozen arguments. Since we are only supporting one argument setters, the argument might be a dataclass. This is how the `prepare` function is meant to be used. The only way to incorporate event specific analysis or statistics into the Bluesky document model is to somehow read it as a device signal.

This means that ALL event specific results must be readable from a device signal. Right now if the beamline has any function that produces a useful point of data as part of a scan which isn't tied directly to a tango attribute, there is no easy way to get it into the Bluesky document model.

jacopoabramo: Got it. Then an argument could be made that the accepted input types could be relaxed? Or provide an additional hook for parsing the type to create a representable datakey in case an unexpected type from `SignalDataTypeT` is provided?

I did find myself wanting some more flexibility in type hinting, in particular in the case of using `np.ndarray[some-dtype]`, but I'm not sure how easy it would be

coretl: > Or lets say someone has a script which initializes a detector. The script probably takes a half dozen arguments. Since we are only supporting one argument setters, the argument might be a dataclass. This is how the `prepare` function is meant to be used. The only way to incorporate event specific analysis or statistics into the Bluesky document model is to somehow read it as a device signal.
> 
> This means that ALL event specific results must be readable from a device signal. Right now if the beamline has any function that produces a useful point of data as part of a scan which isn't tied directly to a tango attribute, there is no easy way to get it into the Bluesky document model.

Ok, this is a different question then, and I'm going to tag @danielballan for this.

It seems like you want to get an arbitrary structure into the document model. The options I think we support at the moment are:
- Put an arbitrary structure in the start document metadata
- Put something of type [SignalDatatype](https://github.com/bluesky/ophyd-async/blob/fab63a8a846acabfc1e2d3df4e29c7f57529c18f/src/ophyd_async/core/_signal_backend.py#L20-L41) into the event document

The reason we have SignalDatatype is so that we can limit what downstream consumers could expect to see in events, ensuring it has a numpy dtype associated with it.

My suggestion is either:
- Put it in the start document (we will do that with the `scanspec` that I think is the equivalent of your scan config
- JSON encode it and put the string in the event document

Would either of those work?

burkeds: No, that's not what I am saying.

Let's  say you have a script that takes x data and returns y data.

You can easily split that return data into however many signals you need to represent all the important components of Y.

What you can't do is go the other way. If we want to support generic callables, we need a way to pass any single argument to it. You can't split the input into config signals because there could be many things you want to call but the protocol limits your ability to call functions.

burkeds: This isn't really the primary reason to implement this though. I think the more important point is that we probably want a way to create a signal that does not rely on Tango or EPICS. This means that, theoretically, this interface could take any form. Therefore write actions need to be very flexible and we need a way to map data from some non-control system source to one or more `Signal` values.

We could rely on a combination of `Command` to write and then reading back signals that are populated with the results but this is feels like a more awkward implementation, especially if the `Command` only results in a single important value. Because `Command` can't be read, every `Command` will need at least one `Signal` associated with it to read a result into a datastream.

Now that I think more about it I think I would be happy with the following.

Case A: You have a multi-argument callable which returns an important value.

1. Write a `Command` signal to call it.
2. Write a `Signal` that will hold the return value.
2. Write the value returned by `Command` to the `Signal`.

Case B: You have a callable with one argument that takes a different input type to its output type. 
Same steps as case A. It is more cumbersome then a input/output flexible `CallableSoftSignal` but not overly so I suppose.

Case C: You have a multi-argument Callable with multiple or complex returns.
1. Write a new function that calls your callable and maps the results to some number of `Signal`
2. Write a `Command` signal backed by your new function.

Case D: You have a callable with one argument that is the same type as the output.
1. Write a `CallableSoftSignal` to write/read.

Here, the role of `CallableSoftSignal` is a more convenient alternative to the `Command`/`Signal` approach and is not strictly necessary. It preserves signals requiring the same input/output types.

What do you think about this @coretl @jacopoabramo ?

jacopoabramo: @burkeds thanks, I think that your summary covers pretty much all possible case combinations I could also think of myself.

I'm only concerned about the steps of case A / B.

> 1. Write a Command signal to call it.
> 2. Write a Signal that will hold the return value.
> 3. Write the value returned by Command to the Signal.

Are these steps responsability of the final user? Or you want to bake this in somehow in a "one class, one job" approach?

burkeds: Here, the user would have to be responsible for writing these. When I say "user", I usually mean beamline scientists as they are the ones writing control scripts here but it could also mean a visiting end user if they are the technically inclined type.

burkeds: I very confusingly used `CallableSoftSignal` in my explanation here because in my head I was thinking of it as its own class. Within the context of this thread, please consider `CallableSoftSignal` as shorthand for the `SoftSignalBackend` extended to support callables.

coretl: I agree with your suggestions, and I think a combination of soft `Command` and soft `Signal`s can be made to handle all your use cases. Remember that you can call the signal set value within the command function, so the user only has to write one function that pokes the hardware and also changes a signal on the return path.

There is also the pattern we use in EPICS, make a `Signal` for every input value, a `Signal` for every output value, then a single `TriggerableCommand` to execute. For this you wouldn't need any callables passed to the signals, the callable passed to the `TriggerableCommand` would tie it all together.

burkeds: @coretl that sounds good. That pattern is common in EPICS but Tango users are more accustomed to tango commands which which can take different input/output types as remote RPCs or writing a setpoint and reading back from the same signal (a poor design pattern in my opinion).

In this case, are there any further changes that should be made?

coretl: Just restrict the type hint for the setter to take `SignalDatatypeT` and I think we are done. Please can you write an ADR to go with this that summarizes this conversation.

tests/unit_tests/core/test_soft_signal_backend.py
```diff
@@ -204,3 +204,238 @@ async def test_soft_signal_coerces_numpy_types():
...
+    cfg.velocity = 2.5
+    cfg.acceleration = 0.1
+    cfg.units = "mm/s"
+    await backend.put(cfg)
+
+    assert len(configs_received) == 1
+    assert configs_received[0].velocity == 2.5
+    assert configs_received[0].acceleration == 0.1
+    assert configs_received[0].units == "mm/s"
+    assert await backend.get_value() == pytest.approx(2.5)
```
coretl: Can we delete this test now then?

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: I will do this.

What should `get_setpoint` do? Return the last value set with `SoftSignalBackend.put`?

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: Yes

tests/unit_tests/core/test_soft_signal_backend.py
burkeds:
```python
    async def put(self, value: Any) -> None:
        write_value = self.initial_value if value is None else value
        self._setpoint = write_value
        ...

    async def get_setpoint(self) -> SignalDatatypeT:
        # For a soft signal, the setpoint and readback values are the same.
        if self._setter is None and self._getter is None:
            return self.reading["value"]
        else:
            return self._setpoint
```

Something like this?

coretl: Good point, we probably need the concept of setpoint and reading:
- initial value goes to self._setpoint and self.reading
- put value goes through converter, then stored as self._setpoint
- setter called on self._setpoint, if it returns a value that goes in self.reading, otherwise call getter to get that value

burkeds: A possible issue. Currently, the value is not converted until `self.set_value(...)`.

If we put `self._setpoint = self.converter.write_value(value)` in `set_value` that works provided there is not `setter`. If there is a `setter` then the converted write value would not be accurate to assign as `self._setpoint` because the value is not converted before it is passed to the setter.

Should we pass the converted write value to the `setter`?

```python
    def set_value(self, value: SignalDatatypeT):
        """Set the current value, alarm and timestamp."""
        self.reading = Reading(
            value=self.converter.write_value(value),
            timestamp=time.time(),
            alarm_severity=0,
        )
        if self.callback:
            self.callback(self.reading)

    async def put(self, value: Any) -> None:
        write_value = self.initial_value if value is None else value
        if self._setter is not None:
            written_value = await maybe_await(self._setter(value))
            if written_value is not None:
                self.set_value(written_value)
            elif self._getter is not None:
                await self._update_value_from_getter()
            else:
                self.set_value(write_value)
        else:
            self.set_value(write_value)
```

coretl: > Should we pass the converted write value to the `setter`?

I think that's probably best

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -166,6 +166,7 @@ def __init__(
         self._poll_period = poll_period
         self._poll_task: asyncio.Task | None = None
         self.set_value(self.initial_value)
+        self._setpoint = None
```
coretl:
```suggestion
        self._setpoint = self.initial_value
```

docs/how-to/how-to-use-soft-signals.md
coretl, line 1: Assuming you're using AI to write this, please could you turn this into a "How to use soft signals" guide?

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -29,9 +31,6 @@
 
 
 class SoftConverter(Generic[SignalDatatypeT]):
-    # This is Any -> SignalDatatypeT because we support coercing
```
coretl, line 32: Please could you put this comment back, it is still true

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,22 +149,46 @@ def __init__(
...
+            try:
+                await self._update_value_from_getter()
+            except Exception:
+                continue
+
     def set_value(self, value: SignalDatatypeT):
         """Set the current value, alarm and timestamp."""
         self.reading = Reading(
-            value=self.converter.write_value(value),
+            value=value,
```
coretl: Please could you put the convert back here, I remembered that we call this code from outside so it's safer to have a double convert on put than lose the convert on `set_value`...


```suggestion
            value=self.converter.write_value(value),
```

docs/how-to/how-to-use-soft-signals.md
burkeds: The format is AI but otherwise I wrote it.

I renamed the file to "how to use soft signals". Did you want any other changes?

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: Because the setter is possibly async, it can't go in set_value. That means to pass the converted value to the setter we need to convert the value twice, once in put and once in set_value. I didn't want to do that.

Instead, I thought we could pass the unconverted value to the setter (as I am sure a user would expect) and the setpoint is then recorded as the value passed to set_value.

docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
```diff
@@ -0,0 +1,38 @@
...
+
+### **Design Choices**
+- All three parameters are **optional**. If none are provided, behavior remains identical to the existing `SoftSignalBackend`.
+- The internal `self._reading` store remains the **single source of truth**. The `getter` updates this store rather than bypassing it, preserving coherence for subscriptions and cached reads.
+- The `put` method accepts `SignalDatatypeT` (the same type as the signal's stored value) to maintain type safety and consistency.
+- Polling tasks are used for subscriptions, starting in `set_callback` and canceling when subscriptions end.
+- `get_setpoint()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of.
+
+### **Factory Function Updates**
+The convenience functions `soft_signal_rw` and `soft_signal_r_and_setter` were updated to accept `getter`, `setter`, and `poll_period` arguments, passing them through to `SoftSignalBackend`.
```
oliwenmandiamond: Currently, `soft_signal_r_and_setter` does not have a setter argument.

src/ophyd_async/core/_signal.py
```diff
@@ -375,6 +397,9 @@ def soft_signal_r_and_setter(
     name: str = "",
     units: str | None = None,
     precision: int | None = None,
+    *,
+    getter: Getter[SignalDatatypeT] | None = None,
+    poll_period: float | None = None,
```
oliwenmandiamond, line 402: The documentation above says that the setter method should be available here. Is this a mistake that it's left out?

burkeds: It was a mistake in the explanatory document. `soft_signal_r_and_setter` is meant to return a read-only signal so it does not take a `setter` argument. This has been corrected.

burkeds:
@coretl For some reason, `test_observe_value_times_out_with_no_external_task` is seeing an elapsed time of 0.25s instead of ~0.07s consistently in python 3.13 but not in other versions. It is not clear to me why, some sort of change in how async tasks are scheduled?

If we relax the equality in the assertion to `>=` that would solve the issue but I am not sure if we want to accept this delay.

docs/how-to/how-to-use-soft-signals.md
coretl: I meant put it in the `docs/how-to`, and add a section on using soft signals (without callables) as ways to store data. Then the rest of the document you wrote follows that says how to add callables to make the set/get have side effects and introducing soft command too.

coretl:
> @coretl For some reason, `test_observe_value_times_out_with_no_external_task` is seeing an elapsed time of 0.25s instead of ~0.07s consistently in python 3.13 but not in other versions. It is not clear to me why, some sort of change in how async tasks are scheduled?
> 
> If we relax the equality in the assertion to `>=` that would solve the issue but I am not sure if we want to accept this delay.

I've noticed this on other PRs, I'll investigate...

https://github.com/bluesky/ophyd-async/pull/1295

docs/how-to/how-to-use-soft-signals.md
burkeds: done

tests/unit_tests/core/test_soft_signal_backend.py
coretl: I'm not sure I agree, let's go with an example of a float signal with a setter:
- I call `signal.set("34.56")`
- First this converts the value to `34.56`
- Then it calls `setter(34.56)`
- Imagine that `setter` returns `34.6`

In this example I would suggest that the setpoint is `34.56` and the value is `34.6`. Would you agree?

If so then this becomes:
```python
    async def put(self, value: Any) -> None:
        self._setpoint = self.initial_value if value is None else self.converter.write_value(value)
        if self._setter is not None:
            readback = await maybe_await(self._setter(value))
            if readback is not None:
                self.set_value(readback)
            elif self._getter is not None:
                await self._update_value_from_getter()
            else:
                self.set_value(self._setpoint)
        else:
            self.set_value(self._setpoint)
```

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -132,22 +150,48 @@ def __init__(
...
     def set_value(self, value: SignalDatatypeT):
         """Set the current value, alarm and timestamp."""
+        setp = self.converter.write_value(value)
+        self._setpoint = setp
         self.reading = Reading(
-            value=self.converter.write_value(value),
+            value=setp,
             timestamp=time.time(),
             alarm_severity=0,
         )
```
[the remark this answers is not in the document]
coretl: Agree

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -160,28 +204,44 @@ def source(self, name: str, read: bool) -> str:
...
+        if self._setter is not None:
+            returned_by_setter = await maybe_await(self._setter(write_value))
+            if returned_by_setter is not None:
+                self.set_value(returned_by_setter)
+            elif self._getter is not None:
+                await self._update_value_from_getter()
+            else:
+                self.set_value(write_value)
+        else:
+            self.set_value(write_value)
```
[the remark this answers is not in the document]
coretl: Agree, as per https://github.com/bluesky/ophyd-async/pull/1280#discussion_r3404687587

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -114,6 +114,14 @@ def make_converter(datatype: type[SignalDatatype]) -> SoftConverter:
     raise TypeError(f"Can't make converter for {datatype}")
 
 
+Setter = (
+    Callable[[SignalDatatypeT | None], SignalDatatypeT | None]
+    | Callable[[SignalDatatypeT | None], Awaitable[SignalDatatypeT | None]]
+    | None
+)
+Getter = Callable[[], SignalDatatypeT | Awaitable[SignalDatatypeT]]
+
```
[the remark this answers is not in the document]
coretl: Disagree, https://github.com/bluesky/ophyd-async/pull/1280#discussion_r3404687587 would pass the converted value to the setter

src/ophyd_async/core/_signal.py
```diff
@@ -353,6 +353,10 @@ def soft_signal_rw(
     name: str = "",
     units: str | None = None,
     precision: int | None = None,
+    *,
+    getter: Getter[SignalDatatypeT] | None = None,
+    setter: Setter[Any] | None = None,
+    poll_period: float | None = None,
```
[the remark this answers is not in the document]
coretl: agree

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: @coretl this works but it means we need to convert the value twice. `set_value` can be called directly instead of `put` which means that `set_value` must also convert the value.

Are we ok with converting twice?

docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
```diff
@@ -0,0 +1,38 @@
...
+2. **`setter`**: A callable (`Callable[[SignalDatatypeT], SignalDatatypeT | None | Awaitable[SignalDatatypeT | None]]`) invoked during `put()`. It may return a `SignalDatatypeT`; if it returns `None`, the `getter` (if configured) is called immediately to refresh the cache.
+3. **`poll_period`**: A float representing the interval (in seconds) at which the `getter` is polled while a subscription is active. Requires `getter` to be set.
+
+### **Design Choices**
+- All three parameters are **optional**. If none are provided, behavior remains identical to the existing `SoftSignalBackend`.
+- The internal `self._reading` store remains the **single source of truth**. The `getter` updates this store rather than bypassing it, preserving coherence for subscriptions and cached reads.
+- The `put` method accepts `SignalDatatypeT` (the same type as the signal's stored value) to maintain type safety and consistency.
+- Polling tasks are used for subscriptions, starting in `set_callback` and canceling when subscriptions end.
+- `get_setpoint()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of.
+
```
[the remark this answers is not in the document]
burkeds: @coretl Should put actually be typed as SignalDatatypeT instead of Any? We don't want to encourage heterogeneous typing for signals.

tests/unit_tests/core/test_soft_signal_backend.py
coretl: I think this is fine, put is relatively rare compared with get and monitor

docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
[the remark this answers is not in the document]
coretl: You're right, it should be `SignalDatatypeT`

burkeds:
@coretl All the copilot suggestions have been addressed.

tests/unit_tests/core/test_soft_signal_backend.py
burkeds: Ok done

coretl (approved):

Issue #1281, opened 2026-06-04
Names for signals in DeviceVector should be allowed to override for read


oliwenmandiamond:
I am porting some logic from GDA. It is a device that can be configured with channels and then give those channels names when being read.

```Python
class MyDevice(StandardReadable):
    def __init__(self, prefix: str, name: str = ""):
        with self.add_children_as_readables():
            self.hm3amp20 = epics_signal_r(float, prefix + f".S{2}")
            self.sm5amp8 = epics_signal_r(float, prefix + f".S{3}")
            self.smpmamp39 = epics_signal_r(float, prefix + f".S{4}")
            self.rfdamp10 = epics_signal_r(float, prefix + f".S{5}")
        super().__init__(name)
```
So this translated to the above, however this has hardcoded the channels. I wanted to use the device vector for this instead and have it be configurable:

```Python
class MyDevice(StandardReadable):
    def __init__(self, prefix: str, config: dict[int, str], name: str = ""):
        with self.add_children_as_readables():
            self.channels = DeviceVector(
                {
                    key: epics_signal_r(float, prefix + f".S{key}", name=value)
                    for key, value in config.items()
                }
            )
        super().__init__(name)
```

```Python
@pytest.fixture
def my_device() -> MyDevice:
    with init_devices(mock=True):
        my_device = MyDevice(
            "TEST:", {2: "hm3amp20", 3: "sm5amp8", 4: "smpmamp39", 5: "rfdamp10"}
        )
    return my_device
```

However, the read outputs is this:
```python
        {
            "my_device-channels-2": partial_reading(0),
            "my_device-channels-3": partial_reading(0),
            "my_device-channels-4": partial_reading(0),
            "my_device-channels-5": partial_reading(0),
        }
```
rather than
```python
        {
            "my_device-channels-hm3amp20": partial_reading(0),
            "my_device-channels-sm5amp8": partial_reading(0),
            "my_device-channels-smpmamp39": partial_reading(0),
            "my_device-channels-rfdamp10": partial_reading(0),
        },
```

So the name that I provided to the signal was ignored and overwritten by key value int instead. I think we should be able to customise the signals name so that the read outputs the desired users format. We can do this very easily by adding this small change to `DeviceVector`

from this
```Python
    def children(self) -> Iterator[tuple[str, Device]]:
        for key, child in self._children.items():
            yield str(key), child
        yield from super().children()
```

to this
```Python
    def children(self) -> Iterator[tuple[str, Device]]:
[...]

PR #1198, opened 2026-02-10
Fix DeviceVector introspection

shihab-dls:
Closes #1197 

This PR mainly creates a `PviTree` data structure, to recursively explore a PVI structure, providing enough information for`Device/DeviceVector` type selection when creating devices.

PR #1282, opened 2026-06-05
Add DeviceMap and declarative sub-devices (#1252, #1182)

### LINKED ISSUE #1182: `StandardReadable`: support declarative subdevices
The declarative approach does not currently supports sub devices:

The below gives the following error:

``` python
class SomeSubDevice(StandardReadable, EpicsDevice):
    signal_r_1: A[SignalR[int], Fmt.HINTED_SIGNAL]
    signal_r_2: A[SignalR[Array1D[np.float64]], Fmt.CHILD]
    config_signal: A[SignalR[Array1D[np.float64]], Fmt.CONFIG_SIGNAL]
    enum_signal: A[SignalR[Sequence[SomeEnum]], Fmt.CHILD]


class SomeDevice(StandardReadable, EpicsDevice):
    signal_r_1: A[SignalR[int], Fmt.CHILD]
    signal_r_2: A[SignalR[Array1D[np.float64]], Fmt.UNCACHED_SIGNAL]
    signal_rw_1: A[SignalRW[bool], Fmt.CHILD]
    signal_rw_1: A[SignalRW[bool], Fmt.CHILD]
    sub_device_1: A[SomeSubDevice, Fmt.CHILD]
    sub_device_2: SomeSubDevice
```

```
  File "/home/a5497/Code/bluesky_nexus/venv/lib/python3.12/site-packages/ophyd_async/epics/core/_epics_connector.py", line 70, in create_children_from_annotations
    list(self.filler.create_devices_from_annotations())
  File "/home/a5497/Code/bluesky_nexus/venv/lib/python3.12/site-packages/ophyd_async/core/_device_filler.py", line 231, in create_devices_from_annotations
    device = child_type(connector=connector)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: EpicsDevice.__init__() got an unexpected keyword argument 'connector'
```

because all ophyd-async device annotations (whether explicitly given a `StandardReadableFormat` or not) are assumed to be signals:

A `Device` is added here:

https://github.com/bluesky/ophyd-async/blob/f96a9fa0973bd13dc30d354cd6981c49c51c5164/src/ophyd_async/core/_device_filler.py#L166-L167

and then attempted to be connected to as if it were a signal here:

https://github.com/bluesky/ophyd-async/blob/f96a9fa0973bd13dc30d354cd6981c49c51c5164/src/ophyd_async/core/_device_filler.py#L231

### 1 (Preferred)

We allow these devices to be connected to as if they were signals. The subdevice can only be specified declaritavely if it takes default values for this kind of device, in this case the `PvSuffix` for prefix and name from the annotation key.

### 2

We add logic here so that only signals, or `DeviceVector` of signals are connected to in this way.

### LINKED ISSUE #1252: `DeviceVector` for string-indexed device children
I often find myself in the situation of wanting to define a map of devices indexed by a string - in particular for motors that may have a variable number of axis and to control them from an upper layer I find uncomfortable to use `getattr`.

Would it be ok to either

- add a `DeviceMap` which behaves as a `DeviceVector` but with `str` as key rather than `int`
- add a more general annotation to `DeviceVector`, i.e.

```py

K = TypeVar("K")

class DeviceVector(MutableMapping[K, DeviceT], Device):
    ...
```

which would allow to do something like

```py
    self.axis = DeviceVector({
        "x": MotorAxisDevice("x")
        "y": MotorAxisDevice("y")
    )}
```

oliwenmandiamond:
Fixes #1252
Fixes #1182


src/ophyd_async/core/_device.py
```diff
@@ -282,41 +283,36 @@ async def connect(
 }
 
 
-class DeviceVector(MutableMapping[int, DeviceT], Device):
+class DeviceMap(MutableMapping[T, DeviceT], Device):
     """Defines a dictionary of Device children with arbitrary integer keys.
```
jacopoabramo, line 353: Docstring should be updated to reflect the change

jacopoabramo:
What kind of types should the key allow? So far I could only think of `int` and `str`, is there the expectation to allow more?

oliwenmandiamond:
@coretl and @shihab-dls, I've updated the core logic for `DeviceMap`, however I'm a bit stuck on the `DeviceFiller` and `PviConnector` parts. It's especially tricky because there is little code coverage in this area so hard for me to make changes without knowing if I've broken something. Can we please lay out exactly how this should behave here, should `DeviceMap` be even supported in this area? Thanks

oliwenmandiamond:
The more I continue to delve into the depths of device filler and pvi connector code, the more I'm convinced `DeviceMap` shouldn't be supported here, only `DeviceVector`. 

From my understanding, `DeviceFiller` reads the device python class annotations, create the objects from the annotations, then attach to the device. `EpicsDevice` does this by adding some additional helpers for the annotations to correctly give the PVs to signals. However for something like odin, it doesn't use `EpicsDevice` and instead use the PVI mechanism, 

So `PviTree` is a representation of of the IOC. `PviConnector` will read this and populate a device annotation using `DeviceFiller` and the pvi tree structure . I think it would ever make sense to support `DeviceMap` here if the IOC will represent the structure with keys which are not integers which I don't think we have a use case for yet?

So therefore, I think I only need to add the `DeviceMap` part and then remove all the other changes I've made? Do you agree?

oliwenmandiamond:
Okay, reading more of this original ticket https://github.com/bluesky/ophyd-async/issues/1252 there is discussion on DeviceFiller and the PVI structure with how this works with it and is more involved like I thought. Sorry I missed this as I was looking at my issue https://github.com/bluesky/ophyd-async/issues/1281.

jacopoabramo:
Hi @oliwenmandiamond sorry for pitching in, what's the status on this? I just wanted to know if there's any blocker I can help with...

oliwenmandiamond:
Hey @jacopoabramo , currently the test coverage is very poor in areas I'm changing for the pvi logic. This means it's very hard to make progress as I will probably break the intended behaviour without anyone knowing and hard for me to know what the intended behaviour is to begin with. @coretl has created a few tickets to up the code coverage. I will try and pick this back up soon(ish) after these are complete, but am also busy with current beamline support and am not an expert in control systems, only data acquisition. Feel to continue work on it if it is urgent.

jacopoabramo:
No worries and no rush on my part, just making sure I was up to speed

coretl:
The unit tests are being done as part of #1321, when this is done then the PVI tests will exist which should help here

src/ophyd_async/core/_device.py
```diff
@@ -398,6 +398,58 @@ def __hash__(self):  # to allow DeviceVector to be used as dict keys and in sets
...
+        # make a non-string named child
+        if not isinstance(key, str):
+            msg = f"Expected str, got {key}"
+            raise TypeError(msg)
+        if not isinstance(value, Device):
+            msg = f"Expected Device, got {value}"
+            raise TypeError(msg)
+        self._children[key] = value
+        value.parent = self
+
```
coretl, line 434: Also need to ban `Device` attrs. Similar code as to what was deleted from `DeviceVector` in https://github.com/bluesky/ophyd-async/pull/1198

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -91,8 +94,12 @@ def _command_backend_factory(
...
         if isinstance(device, DeviceVector):
-            self.filler.create_device_vector_entries_to_mock(
-                self.mock_device_vector_len
+            self.filler.create_device_dict_entries_to_mock(
+                self.mock_device_vector_children
+            )
+        elif isinstance(device, DeviceMap):
+            self.filler.create_device_dict_entries_to_mock(
+                self.mock_device_map_children
             )
```
coretl: I don't think we should mock any children here. Let's wait for a use-case before adding anything here

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -359,28 +388,32 @@ async def build_device_tree(cls, pvi_pv: str, timeout: float) -> PviTree:
...
+        # "__N" entries here, i.e. a DeviceVector of commands).
         for processed_entries in (sub_trees, signal_details, commands):
             for child_name in list(processed_entries):
                 if m := re.match(r"^__(\d+)$", child_name):
-                    sub_tree = processed_entries.pop(child_name)
-                    vector_children[int(m.group(1))] = sub_tree
+                    vector_children[int(m.group(1))] = processed_entries.pop(child_name)
+
+                elif m := re.match(r"^__([A-Za-z]\w*)$", child_name):
+                    map_children[m.group(1)] = processed_entries.pop(child_name)
```
coretl: No, this is not what we want. There will only ever be `__N` entries where `N` is an integer. Never a string. The only time we create a `DeviceMap` is if we explicitly are annotated as one. For instance, consider the tree:
```
parent:
  child:
    a: rw: pv1
    b: rw: pv2
```
If you fill a bare Device with no type hints you should get:
```
parent (Device):
  child (Device):
    a: SignalRW
    b: SignalRW
``
However if you fill the following:
```python
class Parent(Device):
  child: DeviceMap[SignalRW[int]]
```
Then you would get
```
parent (Device):
  child (DeviceMap):
    "a": SignalRW[int]
    "b": SignalRW[int]
``
and the types of "a" and "b" would be checked as `SignalRW[int]`. This needs adding to the pvi tests

File: src/ophyd_async/tango/core/_base_device.py
Comment (coretl) [inline src/ophyd_async/tango/core/_base_device.py:1]: I think we need to support:
- `connector`
- or `trl`
- but not both

**Review by coretl — CHANGES_REQUESTED** [review]

File: src/ophyd_async/core/_device.py (hunk 2 above)
Comment (coretl) [reply src/ophyd_async/core/_device.py:434]: Done in dec9922: added `DeviceMap.__setattr__` that raises `AttributeError` if a `Device` is set as an attribute (children must go via `device_map[key] = child`), mirroring the guard #1198 removed from `DeviceVector`. Unit test `test_device_map_bans_device_attributes`.

File: src/ophyd_async/epics/core/_pvi_connector.py (hunk 3 above)
Comment (coretl) [reply src/ophyd_async/epics/core/_pvi_connector.py]: Done in dec9922: removed `mock_device_map_children` and the map branch in `connect_mock` — a `DeviceMap` now stays empty in mock mode (its entries come from the served tree). Only `DeviceVector` keeps its [1, 2] mock fabrication.

File: src/ophyd_async/epics/core/_pvi_connector.py (hunk 4 above)
Comment (coretl) [reply src/ophyd_async/epics/core/_pvi_connector.py]: Reworked in dec9922 exactly as described. Removed the `__str`->`map_children` auto-detection (`__N` is int-only now), the `map_children` field and `is_signal_map`/`is_command_map`. A `DeviceMap` is created only from an explicit `DeviceMap[...]` annotation and filled from its node's *normal* named entries (`_fill_device_map`), keyed by name and element-type-checked by the filler. Added live-IOC PVI tests in `test_pvi_nested.py`: `test_device_map_of_signals` (your exact `a`/`b` example) and `test_device_map_of_devices`, served under an isolated `mapd:` group in `_pvi_nested_records.db`.

File: src/ophyd_async/tango/core/_base_device.py
Comment (coretl) [reply src/ophyd_async/tango/core/_base_device.py:1]: Done in dec9922: `TangoDevice.__init__` now raises `ValueError` if both `trl` and `connector` are given. Unit test `test_tango_device_rejects_both_trl_and_connector`.

File: src/ophyd_async/core/_device.py [hunk 5]
```diff
@@ -398,6 +398,68 @@ def __hash__(self):  # to allow DeviceVector to be used as dict keys and in sets
...
+    def __iter__(self) -> Iterator[str]:
+        yield from self._children
+
+    def __len__(self) -> int:
+        return len(self._children)
+
+    def children(self) -> Iterator[tuple[str, Device]]:
+        for key, child in self._children.items():
+            yield str(key), child
```
coretl:
```suggestion
            yield key, child
```

src/ophyd_async/core/_device_filler.py
```diff
@@ -426,24 +443,36 @@ def check_filled(self, source: str):
...
-    def _ensure_device_vector(self) -> DeviceVector:
-        if not isinstance(self._device, DeviceVector):
-            self._raise(self._device.name, f"Expected DeviceVector, got {self._device}")
+    def _ensure_device_dict(self) -> DeviceVector | DeviceMap:
+        if not isinstance(self._device, (DeviceVector, DeviceMap)):
+            self._raise(
+                self._device.name,
+                f"Expected DeviceVector or DeviceMap, got {self._device}",
+            )
         return self._device
```
coretl: I think we should delete this...

src/ophyd_async/core/_device_filler.py
```diff
@@ -426,24 +443,36 @@ def check_filled(self, source: str):
...
 
+    @staticmethod
+    def _set_dict_child(
+        device_dict: DeviceVector | DeviceMap, map_key: int | str, value: Device
+    ) -> None:
+        # DeviceVector (int keys) and DeviceMap (str keys) each validate the key
+        # type at runtime in __setitem__; the union of their two signatures can't
+        # be expressed statically, so cast to the combined mapping type.
+        cast("MutableMapping[int | str, Device]", device_dict)[map_key] = value
```
coretl: ...then make this do:
```
if we are a `DeviceVector` then check the `map_key` is an int
elIf we are a `DeviceMap` then check the `map_key` is a str
else fail

self._device[map_key] = value
```

src/ophyd_async/epics/core/_epics_connector.py
```diff
@@ -100,11 +128,15 @@ def _command_backend_factory(
             self.filler = DeviceFiller(
                 device,
                 signal_backend_factory=get_signal_backend_type(protocol),
-                device_connector_factory=DeviceConnector,
+                # Declarative sub-devices get their own EpicsDeviceConnector so
+                # their signals are created under a PvSuffix-derived prefix.
+                device_connector_factory=lambda: EpicsDeviceConnector(""),
```
coretl: Make `EpicsDeviceConnector` have a default `prefix = ""` then we don't need the lambda here

src/ophyd_async/epics/core/_epics_device.py
```diff
@@ -1,15 +1,30 @@
...
+        with_pvi: bool = False,
+        name: str = "",
+        connector: DeviceConnector | None = None,
+    ):
+        if connector is None:
+            if with_pvi:
+                connector = PviDeviceConnector(prefix)
+            else:
+                connector = EpicsDeviceConnector(prefix)
         super().__init__(name=name, connector=connector)
```
coretl, line 35: Tighten up all these inits so you get prefix xor connector

src/ophyd_async/core/_device.py
coretl: Applied in 9c21e7f. `DeviceMap.children()` now yields its (already-str) keys directly — ruff (UP028) collapses it to `yield from self._children.items()`, which is the equivalent of your suggestion without the redundant `str()`.

src/ophyd_async/core/_device_filler.py
coretl: Done in 9c21e7f — `_ensure_device_dict` deleted.

src/ophyd_async/core/_device_filler.py
coretl: Done in 9c21e7f. `_set_dict_child(map_key, value)` now does exactly this: DeviceVector -> require int key, DeviceMap -> require str key, else raise, then `self._device[map_key] = value`. The `isinstance`/key-type narrowing lets the assignment type-check, so the `MutableMapping` cast is gone. Callers updated to `_get_device_dict_child_datatype(self._device)`.

src/ophyd_async/epics/core/_epics_connector.py
coretl: Done in 9c21e7f. `EpicsDeviceConnector.__init__` now defaults `prefix=""`, so the factory is just `device_connector_factory=EpicsDeviceConnector` (no lambda).

src/ophyd_async/epics/core/_epics_device.py
coretl: Done in 9c21e7f. `EpicsDevice.__init__` now raises `ValueError` if both `prefix` and `connector` are given (matching `TangoDevice`). Unit test `test_epics_device_rejects_both_prefix_and_connector`.

src/ophyd_async/core/_device_filler.py
```diff
@@ -570,20 +598,19 @@ def fill_child_command(
...
             backend, expected_command_type = self._filled_command_backends[logical_name]
 
-        # Handle DeviceVector case
-        elif vector_index is not None:
-            vector = self._ensure_device_vector()
+        # Handle DeviceVector/DeviceMap case
+        elif map_key is not None:
             vector_command_type = (
                 self._vector_device_type.get(logical_name) or command_type
             )
```
shihab-dls: In `fill_child_device` you've changed this variable to `dict_device_type`. So this should be renamed as well as `self._vector_device_type`.

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -53,137 +54,139 @@ def __init__(self, prefix: str = "", error_hint: str = "") -> None:
...
+            device_connector_factory=PviDeviceConnector,
+            command_backend_factory=command_backend_factory,
+        )
+        # Address the type-hinted signals/commands now; sub-devices get an
+        # unfilled connector. The rest is filled from the PVI tree at connect.
+        for backend, extras in self.filler.create_signals_from_annotations(
+            filled=False
+        ):
+            fill_backend_with_prefix(self.prefix, backend, extras)
+        for backend, extras in self.filler.create_commands_from_annotations(
```
shihab-dls: nit: I'm not sure why we'd want to call these `extras` here, given we only pass them as `annotations` in our filler create methods. I would keep them as `annotations`

tests/system_tests/epics/core/test_pvi_nested.py
```diff
@@ -142,3 +143,47 @@ async def test_undeclared_device_vector_of_devices(
...
+
+
+@pytest.fixture
+async def map_device(nested_ioc_and_prefix) -> EpicsTestPviMapDevice:
+    _, prefix = nested_ioc_and_prefix
+    # The DeviceMap structure is served under its own "mapd:" group (see
+    # _pvi_nested_records.db) so it doesn't perturb EpicsTestPviNestedDevice.
+    device = EpicsTestPviMapDevice(f"{prefix}mapd:", with_pvi=True, name="map_device")
+    await device.connect(timeout=TIMEOUT)
+    return device
```
shihab-dls: nit: I'd put all my fixtures near the top of the file, not between tests

shihab-dls (approved):
This looks good, just a couple of nits but nothing too importand, so I'm approving. I will comment some behaviour I found when testing against an IOC.

shihab-dls:
Testing this against an IOC such as:
```python
class ParentController(Controller):
    pass

class ChildController(Controller):
    pass

    class ChildVector(ControllerVector):
        vector_attribute: AttrR = AttrR(Int())

sub_controller = ChildVector(
        {
            1: ChildController(
                description="some sub controller",
            ),
            2: ChildController(
                description="another sub controller",
            ),
        },
        description="some child vector",
    )

controller.add_sub_controller("child", sub_controller)

fastcs = FastCS(controller, [p4p_options])
fastcs.run()
```

with an ophyd-async device such as:

```python
class TestDevice(Device):
    child: DeviceMap[SignalR[int]]

    def __init__(self, name="", connector=None):
        connector = fastcs_connector("P4P_TEST_DEVICE:", self)
        super().__init__(name, connector)
```
will result in a `DeviceMap` that only has the `vector_attribute` signal (in a dict). The vector children of the underlying `ControllerVector` are lost due to this check:

```python
        if not is_map:
            self._fill_vector_children(device, tree)
```
This means that if you redefine what would've been a `DeviceVector` as a `DeviceMap`, you may still connect, but lose the sub-devices. This is probably fine, but I'm flagging this in case we'd rather raise at this point.

coretl:
> This means that if you redefine what would've been a `DeviceVector` as a `DeviceMap`, you may still connect, but lose the sub-devices. This is probably fine, but I'm flagging this in case we'd rather raise at this point.

Good spot, this should raise, I'll add it

src/ophyd_async/epics/core/_pvi_connector.py
```diff
@@ -53,137 +54,139 @@ def __init__(self, prefix: str = "", error_hint: str = "") -> None:
...
+            filled=False
+        ):
+            fill_command_with_prefix(self.prefix, backend, extras)
+        list(self.filler.create_devices_from_annotations(filled=False))
```
coretl: given that we fill the signals and commands with prefix, we should do the same for the devices here like in EpicsConnector. In fact this whole body looks very similar so we can pull it all into a function rather than calling `fill_*_with_prefix`

tests/system_tests/epics/core/test_pvi_nested.py
coretl, line 1: We also need a test that if the PvSuffix given as an annotation to a declarative signal/device is different to the one provided by PVI, that we raise

src/ophyd_async/core/_device_filler.py
coretl: Good catch, and it was worse than a naming inconsistency: `self._vector_device_type` is written **nowhere** (on this branch or on `main`), so that `.get()` always returned `None` and the lookup was dead.

It dates from when the parent's filler owned the `DeviceVector` as a plain attribute and recorded its element type per attribute name (7f3c2d5). Once a `DeviceVector` became a `Device` with its own filler, the element type moved to `_get_device_dict_child_datatype(self._device)` and every caller was ported to it except `fill_child_command`.

So rather than rename dead state, I made it live the way it was always meant to: `fill_child_command` now reads the hinted element type, as `fill_child_device` does, and the dict is gone. One wrinkle -- a hint can be a parameterized alias (`DeviceVector[Command[[], int]]`), which `issubclass` rejects, so the final check compares origin classes exactly like `fill_child_signal` already does. 828aae47b

src/ophyd_async/epics/core/_pvi_connector.py
coretl: Agreed. They are gone entirely now: both connectors walked the same three generators, so that whole body is pulled into one `fill_children_with_prefix` in `_epics_connector.py`, where they are `annotations` throughout (qualified per loop as `signal_annotations`/`command_annotations`/`device_annotations`, since a bare `annotations` shadows the `from __future__ import annotations` import). bd079dc2d

tests/system_tests/epics/core/test_pvi_nested.py
coretl: Done -- `map_device` now sits with `nested_device` at the top of the file. bd079dc2d

src/ophyd_async/epics/core/_pvi_connector.py
coretl: Done in bd079dc2d -- the whole body is now one `fill_children_with_prefix(prefix, filler, filled)` in `_epics_connector.py`, called by both connectors.

You were right that devices were being missed here, and it was a real bug rather than just an inconsistency: this line dropped the generator with `list(...)`, so a `PvSuffix` on a declarative sub-device of a PVI device silently did nothing.

The one thing the two callers genuinely differ on is whether a sub-device *must* carry a `PvSuffix`. That turns out to be exactly the existing `filled` flag -- static EPICS is `filled=True` (nothing else will address it, so no `PvSuffix` is an error), PVI is `filled=False` (PVI addresses it later, so it is optional) -- so the function takes `filled` and passes `required=filled` down. The two connectors also now share a `_PvPrefixDeviceConnector` base, so the shared function can set a prefix on either; `PviDeviceConnector` overrides `set_prefix` to keep `pvi_pv` derived from it.

tests/system_tests/epics/core/test_pvi_nested.py
coretl: Done in bd079dc2d, but it needed two existing behaviours reversed, so flagging them explicitly.

Only two places in the codebase actually had a `PvSuffix` disagreeing with PVI, and both were deliberate:

**1. `EpicsTestPviDevice.overridden_float`** existed *only* to prove the opposite rule, via a passing test called `test_pvi_wins_over_static_pv_suffix`. That test is now inverted (`test_pvi_disagreeing_with_static_pv_suffix_raises`). `overridden_float` had to move to a device of its own, because a Signal that can no longer connect would otherwise stop the shared `pvi_device` fixture connecting at all.

**2. `EpicsTestCaDevice.longstr2`** is the interesting one. It is annotated `PvSuffix("longstr2.VAL$")` -- the CA idiom for reading an `lsi` record in full -- while PVI served the bare record name, which silently truncates at 40 chars. So the new rule immediately caught a latent bug: `EpicsTestCaDevice(prefix, with_pvi=True)` was quietly reading a truncated `longstr2`, and nothing asserted on it. Rather than exempt it, the IOC now serves that PVI entry from a stringin's `VAL` holding the field-qualified name (the same indirection the `d` pointers use), so the annotation and PVI agree and the ca-pvi path reads the whole string.

Worth knowing that this rule means a `PvSuffix` can no longer *refine* a PVI-supplied address (field modifiers like `.VAL$`), only duplicate it -- the address has to be fixed on the IOC side instead, as above. That is the trade I read you as asking for, but say the word if you would rather it only applied to sub-devices.

Coverage: a disagreeing Signal raises (`test_pvi_disagreeing_with_static_pv_suffix_raises`), a disagreeing sub-device raises (`test_sub_device_pv_suffix_disagreeing_with_pvi_raises`), and an agreeing `PvSuffix` on both a signal and a sub-device still connects and reads (`test_pv_suffix_agreeing_with_pvi_connects`), so the rule is not just "any PvSuffix raises".

coretl:
@shihab-dls thanks for testing this against a real IOC -- the `DeviceMap`-over-a-`ControllerVector` case you found now raises, as @coretl called for.

`PviDeviceConnector.connect_real` refuses to fill a `DeviceMap` from a node whose PVI entries are integer-keyed `__N`:

```
<pvi_pv>: <name> is annotated as a DeviceMap, but PVI serves integer-keyed
entries [1, 2] here, which only a DeviceVector can hold
```

Rather than mock it, it is covered end-to-end against the test IOC (`test_device_map_over_vector_children_raises` in `tests/system_tests/epics/core/test_pvi_nested.py`), reusing the `nested:` tree's existing `device_vector` group -- which is served exactly like your `ChildVector`: a node whose children are `__1`/`__2`. The error arrives wrapped in the parent's `NotConnectedError`, since the map is filled while connecting as a child.

Your other three comments are addressed too -- the `_vector_device_type` one turned out to be a genuinely dead lookup rather than just a stale name, see the thread. bd079dc2d

coretl:
> @shihab-dls thanks for testing this against a real IOC -- the `DeviceMap`-over-a-`ControllerVector` case you found now raises, as @coretl called for.

Not sure I like the fact the AI is talking to actual people as if it was me. I'll see if I can get it to stop

coretl (approved):

PR #1290, opened 2026-06-08
Fix derived signal all read only throwing error with rd plan

### LINKED ISSUE #1291: Unable to use RW derived signals from `Motor` axes
We are runnign into an issue where `SignalTransformer` requires its input signals to be Locatable for a RW derived signal even if the input signals are RO.

This wasn't previously an issue, however recent changes to ophyd-async have meant that previously where motors could be specified as inputs to derived signals, they no longer can be and instead the user_readback signal must be specified.

In this particular case, the beamstop device presents a RW StrictEnum  derived from motor positions, so I don't think having the derived setpoints from Locatable are terribly useful, although I suppose it is possible this might be more interesting in other cases.

I've also had to make some changes to i21 ToolPointMotion where this might possibly be of more interest. However I'm not really sure what the utility is of exposing the setpoints of the inputs of derived signals.

For dodal to use the latest ophyd-async we will need either a fix to derived signal to allow RW derived signals from RO inputs or an updated fix to Motor to allow motors to be used as derived signal inputs - which possibly was not something previously supported officially but it seems we were relying on it anyway.

The above is currently manifesting itself in 
https://github.com/DiamondLightSource/dodal/pull/2081

where we get errors along the lines of

```
FAILED tests/devices/mx_phase1/test_beamstop.py::test_beamstop_pos_read_selected_pos[0-0-0-Unknown] - TypeError: <ophyd_async.core._signal.SignalR object at 0x7f05a2ded890> is not an instance of <class 'ophyd_async.core._protocol.AsyncLocatable'>
```

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1291

### DIFF
--- a/tests/unit_tests/core/test_single_derived_signal.py
+++ b/tests/unit_tests/core/test_single_derived_signal.py
@@ -381,7 +381,7 @@ def derived_signal_r_example() -> DerivedSignalWithSignalR:
 @pytest.mark.parametrize(
     "fixture_name", ["derived_signal_rw_example", "derived_signal_r_example"]
 )
-async def test_derived_signal_rw_with_bluesky_read(
+async def test_derived_signal_with_bluesky_rd_plan(
     RE: RunEngine,
     request: pytest.FixtureRequest,
     fixture_name: str,
--- a/tests/unit_tests/core/test_single_derived_signal.py
+++ b/tests/unit_tests/core/test_single_derived_signal.py
@@ -4,11 +4,14 @@ from unittest.mock import MagicMock, call, patch
 
 import numpy as np
 import pytest
+from bluesky import RunEngine
+from bluesky.plan_stubs import rd
 from bluesky.protocols import Reading, Subscribable
 
 from ophyd_async.core import (
     Array1D,
     Callback,
+    Device,
     Signal,
     SignalBackend,
     SignalDatatype,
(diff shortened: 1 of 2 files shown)


tests/unit_tests/core/test_single_derived_signal.py
```diff
@@ -324,3 +330,69 @@ def _total(spectrum: Array1D[np.float64]) -> float:
...
+        self.sig1 = epics_signal_r(float, "")
+        self.sig2 = epics_signal_r(float, "")
+        self.sig3 = derived_signal_rw(self._get, self._set, v1=self.sig1, v2=self.sig2)
+        super().__init__(name)
+
+    async def _set(self, value: float):
+        pass
+
+    def _get(self, v1: float, v2: float) -> float:
+        return v1 + v2
```
coretl, line 364: I can't see the difference between these 2 classes?

oliwenmandiamond: I think I overrode it in my testing. Added it back. First one is rw, second is r

coretl:
@rtuck99 now @oliwenmandiamond has fixed #1291, do we still need this? If not then I'm inclined to drop it until we have a usecase for it

coretl:
I've got some comments in the implementation, but the resolution depends on the usecase. Do we have a usecase for this? If not then let's pause it until we do.

PR #1293, opened 2026-06-09
Make DeviceProcessor publicly available

### LINKED ISSUE #1277: Make the `DeviceProcessor` class public
While `init_devices` is pretty useful in general, it is also appropriate in some situations to customize the behavior of the callback function it uses.

For instance, I have an use-case in which I need to take all devices that successfully connected, and add them to a centralized registry of devices for later retrieval. In this scenario, I want to take the inner `process_devices` function inside `init_devices`, and modify it to add these devices to the registry after connection, with some extra exception handling thrown in as well.

This is pretty easy to make with the `DeviceProcessor` class, but by virtue of it being a class only exposed inside a private module, there's no guarantee about the API stability of it (not that there's any API stability for now, since the project is still in version 0.x, but considering it part of the public package's interface would be a good first step!).

I don't know the history of this implementation, so I have no idea if there's a particular reason for it being the way it is now, so I though of making an issue first to gather some opinions on it!

## Acceptance Criteria
Be able to import `DeviceProcessor` without relying on a private interface (currently `ophyd_async.core._device`)

flowln:
This allows downstream users to use `DeviceProcessor` to make modified versions of `init_devices`. A test has been added with one such case, simplified for testing purposes.

Resolves #1277

### DIFF
--- a/src/ophyd_async/core/__init__.py
+++ b/src/ophyd_async/core/__init__.py
@@ -37,6 +37,7 @@ from ._device import (
     Device,
     DeviceConnector,
     DeviceMock,
+    DeviceProcessor,
     DeviceVector,
     LazyMock,
     default_mock_class,
@@ -177,6 +178,7 @@ __all__ = [
     "DeviceFiller",
     "DeviceAnnotation",
     "DeviceVector",
+    "DeviceProcessor",
     "init_devices",
     # Movable
     "MovableLogic",
--- a/tests/unit_tests/core/test_device.py
+++ b/tests/unit_tests/core/test_device.py
@@ -10,6 +10,7 @@ from ophyd_async.core import (
     DEFAULT_TIMEOUT,
     Device,
     DeviceFiller,
+    DeviceProcessor,
     DeviceVector,
     NotConnectedError,
     Reference,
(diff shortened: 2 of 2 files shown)


flowln:
I have no idea why the Windows pipeline failed, it seems totally unrelated to the changes in this PR. Given how very inexperienced I am with Tango, there's not much I can do about it, the only clue I found is in [this pytango test skip](https://gitlab.com/tango-controls/pytango/-/blob/develop/tests/test_test_context.py?ref_type=heads#L439), which seems somewhat close to what ophyd-async does in `src/ophyd-async/tango/testing/_device_server.py` (when `green_mode` is `None`).

oliwenmandiamond:
> I have no idea why the Windows pipeline failed, it seems totally unrelated to the changes in this PR. Given how very inexperienced I am with Tango, there's not much I can do about it, the only clue I found is in [this pytango test skip](https://gitlab.com/tango-controls/pytango/-/blob/develop/tests/test_test_context.py?ref_type=heads#L439), which seems somewhat close to what ophyd-async does in `src/ophyd-async/tango/testing/_device_server.py` (when `green_mode` is `None`).

Think it was just a flakey test / connection lost for the system test, I've reran it and passes fine now

jwlodek (approved):
LGTM

coretl (approved):

coretl:
Thanks!

PR #1289, opened 2026-06-05
adds a check for motor limit

gkalua:
closes #1283 

Adds motor and velocity limit validation to `Trajectory.from_slice`

Adds validation to `Trajectory.from_slice` that raises `ValueError` before any trajectory computation if the slice would violate hardware constraints; motor position limits or max velocity. 

Tests added for both checks across x and y axes.

Issue #1283, opened 2026-06-05
Check velocity and positions when generating trajectory.


gkalua:
The velocities and positions in a trajectory collection window are not currently being checked against motor limits. 

## Acceptance Criteria
- Update logic in _pmac_trajectory_generation.py to check the velocities from the collection window before they are added to the arrays.
- Update the _PmacMotorInfo class to include velocity and position limits for each motor
- Use unambiguous error messaging that is clear about the motor hitting the limit

PR #1296, opened 2026-06-11
Add safeguards for trajectory

LuisFSegalla:
The idea of this PR is to add some "safety" checks before passing the trajectory down to the motion controller. Essentially we're checking if at any point the desired velocity exceeds the maximum speed in the motor or if any position in the trajectory violates either low or high limits.

Unit tests were added and the code was also tested against real hardware.


coretl:
@gkalua this looks similar to #1289, please can you review this and comment

gkalua:
As this PR is similar to #1283, this PR will be closed when 1283 is closed.

PR #1300, opened 2026-06-17
Bug: Update typing to allow for None for calculate_timeout for MovableLogic

oliwenmandiamond:
When trying to use `StandardMovable`, I have a device that is similar to a motor but has no velocity or acceleration. Therefore, I am unable to calculate a valid timeout value. I want to specify None as a valid return value for `StandardMovable.move` suggests I can do (so there is no timeout), however the typing for `StandardMovable.calculate_timeout` only allows you to return a float (with the default being 10 seconds).

I have updated this to allow for None and the default to be None (though preferred, I can revert back to 10 seconds as the default).

### DIFF
--- a/docs/explanations/decisions/0017-standard-movable.md
+++ b/docs/explanations/decisions/0017-standard-movable.md
@@ -60,7 +60,7 @@ methods with safe defaults:
 | `readback: SignalR[T]` | required |
 | `stop()` | no-op |
 | `check_move(new)` | no-op |
-| `calculate_timeout(old, new)` | `DEFAULT_TIMEOUT` |
+| `calculate_timeout(old, new)` | `None` |
 | `get_units_precision()` | reads from `readback.describe()` |
 | `move(new_position, timeout)` | `set_and_wait_for_other_value(setpoint, new_position, readback)` |
 
--- a/docs/explanations/decisions/0017-standard-movable.md
+++ b/docs/explanations/decisions/0017-standard-movable.md
@@ -60,7 +60,7 @@ methods with safe defaults:
 | `readback: SignalR[T]` | required |
 | `stop()` | no-op |
 | `check_move(new)` | no-op |
-| `calculate_timeout(old, new)` | `None` |
+| `calculate_timeout(old, new)` | `DEFAULT_TIMEOUT` |
 | `get_units_precision()` | reads from `readback.describe()` |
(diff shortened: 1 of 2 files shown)


coretl:
But surely there is some reasonable value for a timeout? Even if it is 10 minutes, it's better to timeout at somepoint rather than hang forever if the device doesn't respond...

What is this device?

oliwenmandiamond:
I'm happy to keep the default as 10 seconds, I can change this back. I would like the option to be float or None like the rest of the API. 

The device is a piezo electric motor, not sure how long a sensible timeout option is to specify but the current code I am porting from doesn't have a timeout so wanted to also reflect that here

coretl:
> I'm happy to keep the default as 10 seconds, I can change this back. I would like the option to be float or None like the rest of the API.
> 
> The device is a piezo electric motor, not sure how long a sensible timeout option is to specify but the current code I am porting from doesn't have a timeout so wanted to also reflect that here

10 seconds is ample for a piezo. I think we should be asking the question for every device "how long does it take". If we give people the option to say forever then we'll never get any timeouts anywhere...

oliwenmandiamond:
> I think we should be asking the question for every device "how long does it take". If we give people the option to say forever then we'll never get any timeouts anywhere...

I've reverted it back to use the default 10 second timeout. Only updated the typing.

If we make it a mandatory timeout, then someone could return make `calculate_timeout` return a very large arbitrary number if there is no sensible way to calculate a timeout if there's a device that takes a long time which has the same effect as forever. I would rather just have the option to return None as the rest of the API accepts e.g `checker.wait_for_value(...)` and `asycio.timeout(...)` allows as valid options for timeout=None. 

Maybe we could log the total calculate timeout value so it is clear if something is waiting for long time / forever, we know because we can see it used X value for timeout?

oliwenmandiamond:
Something like this

```Python
        if timeout == CALCULATE_TIMEOUT:
            move_timeout = await self.movable_logic.calculate_timeout(
                old_position, new_position
            )
        else:
            move_timeout = timeout

        self.log.debug(f"Moving {self.name} with a total timeout of {move_timeout}.")

        async with AsyncStatus(
            self.movable_logic.move(new_position=new_position, timeout=move_timeout)
        ) as move_status:
```

coretl (approved):

PR #1302, opened 2026-06-18
StandardMovable logic more accurate calculate_timeout

### LINKED ISSUE #1301: StandardMovable calculate_timeout doesn't behave as expected for custom move
When trying to create an implementation of StandardMovable, the `calculate_timeout` doesn't work as expected. Lets compare two different types.

Motor

```Python
    async def calculate_timeout(
        self, old_position: float, new_position: float
    ) -> float:
        (
            velocity,
            acceleration_time,
        ) = await asyncio.gather(
            self.velocity.get_value(),
            self.acceleration_time.get_value(),
        )
        try:
            return (
                abs((new_position - old_position) / velocity)
                + 2 * acceleration_time
                + DEFAULT_TIMEOUT
            )
        except ZeroDivisionError as error:
            msg = f"Motor {self.readback.name} has zero velocity."
            raise ValueError(msg) from error

    async def move(self, new_position: float, timeout: float | None) -> None:
        """Move by setting the setpoint and waiting for put completion."""
        await self.setpoint.set(new_position, timeout)
```

the timeout is passed to the move method, this works as expected because you have a single set to do and thus only one timeout operation. Now lets look at this example with custom move logic:

```Python
    async def move(self, new_position: float, timeout: float) -> None:
        await self.setpoint.set(new_position, timeout=timeout)
        await wait_for_value(self.setpoint, new_position, timeout=timeout)
        await wait_for_value(self.within_tolerance, True, timeout=timeout)
```

The timeout is passed to three separate methods. This effectively means the timeout can be roughly 3x as long as it was actually calculated for because each bit that waits reuses the same timeout. The calculate_timeout should apply to the entire move method.

Instead we should pass in some sort of object that allows us to calculate a new timeout by subtracting the time remaining each time it is called so it is the true calculate_timeout.

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1301

### DIFF
--- a/src/ophyd_async/core/_movable.py
+++ b/src/ophyd_async/core/_movable.py
@@ -43,7 +43,7 @@ class MoveTimeout:
         return max(0.0, self.timeout - elapsed)
 
 
-Timeout = Callable[[], float | None]
+CalculatedTimeout = Callable[[], float | None]
 
 
 @dataclass
@@ -82,7 +82,9 @@ class MovableLogic(Generic[SignalDatatypeT]):
         datakey = (await self.readback.describe())[self.readback.name]
         return datakey.get("units"), datakey.get("precision")
 
-    async def move(self, new_position: SignalDatatypeT, timeout: Timeout) -> None:
+    async def move(
+        self, new_position: SignalDatatypeT, timeout: CalculatedTimeout
+    ) -> None:
         """Move the device, waiting for the readback to reach the correct position.
 
         ```{note}
--- a/src/ophyd_async/core/_movable.py
+++ b/src/ophyd_async/core/_movable.py
@@ -1,5 +1,7 @@
 import asyncio
+import time
 from abc import abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass
 from functools import cached_property
 from typing import Generic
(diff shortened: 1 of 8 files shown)


src/ophyd_async/core/_movable.py
```diff
@@ -26,6 +28,24 @@
...
+
+    def __call__(self) -> float | None:
+        """Remaining time for a calculated timeout left for the move to use."""
+        if self.timeout is None:
+            return None
+        elapsed = time.monotonic() - self.start_time
+        return max(0.0, self.timeout - elapsed)
+
+
+CalculatedTimeout = Callable[[], float | None]
```
coretl: How about `TimeoutFactory`?

coretl (commented):
Neat! Just a suggestion on naming...

src/ophyd_async/core/_movable.py
oliwenmandiamond: I liked using `CalculatedTimeout` because it shows to the user that this involves the `movable_logic.calculated_timeout` method and made sense to me. I don't have too strong of an opinion either way though

coretl: I see your point, and I like the tying together of the names, but I'm concerned we are about to export `CalculatedTimeout`, `CalculatableTimeout` and `CALCULATE_TIMEOUT`. I'm tempted by the name `TimeoutCalculator` instead, what do you think?

oliwenmandiamond: Yeah, I see why now why you wanted to rename XD. `TimeoutCalculator` seems reasonable

coretl (approved):

PR #1310, opened 2026-06-23
fix: set aquisition ID to  before setting file prefix

shihab-dls:
Upon testing on i15, we found that setting `acquisition_id` results in no meta data being written. This is because this fanout signal does not put to `EF.acqid`, so images received from the fan are not set to the right ID. Further testing showed that setting `EF.acqid` to `filename` (alongside IDs on `FP`s and `MW`) results in FP errors, where:
```
FP.FrameProcessorPlugin ERROR - Received image message with series_number 2067 - expecting 2065 - ignoring
```
Thus, acquisition ID must be set to a specific series number on all plugins. Acquisition ID is not used on i15 and i19, and I believe it is not used anywhere else. Therefore, we can simply set acquisition ID to `""`, which uses default writers on all plugins, such that anything popped off of `EF` is processed. This needs to be done before setting `file_name` and `file_prefix` as `MW` will set `acquisition_ID` to `None`, which must be unset before putting to `MW.file_name` or `MW.file_prefix`, and `acquisition_ID` being a fanout into `file_prefix` attributes means it needs to be set first to not override.


coretl (approved):

coretl:
Merge when you're ready

PR #1320, opened 2026-07-06
StandardMovable cancels move task when stopped

### LINKED ISSUE #1324: StandardMovable does not cancel move task when stopped
When trying to use `StandardMovable`, the expected behaviour is that when stop is called during a move, it will stop the move. However, for epics `Motor` the stop is not instant. It will wait for the time out for the move to finish, then raise a `RuntimeError` telling the user it has been stopped. 

However, for `SimMotor` it explicitly cancels the move task so that stop is instant. 
https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/sim/_motor.py#L37

All `StandardMovable` behaviour should do this.

When implementing a StandardMovable with a threshold, e.g 

```python
class MovableWithToleranceLogic(MovableLogic[float]):
    within_tolerance: SignalR[bool]
    motor_stop: SignalW[int]

    async def stop(self) -> None:
        await self.motor_stop.set(1)

    async def move(self, new_position: float, timeout: float | None) -> None:
        await self.setpoint.set(new_position, timeout=timeout)
        await wait_for_value(self.setpoint, new_position, timeout=timeout)
        # Once setpoint is at new position, we can check for tolerance signal to see if
        # true now as the within_tolerance window has updated to the new setpoint
        # position. Now it doesn't matter if motor steps are small or large.
        await wait_for_value(self.within_tolerance, True, timeout=timeout)

def _within_tolerance_read(setpoint: float, readback: float, tolerance: float) -> bool:
     return abs(setpoint - readback) < abs(tolerance)
```

If we try to stop this device, it won't stop the move so therefore the move may never complete until the timeout is done (which may be never). This is proven with this test:

```python
 async def test_piezo_motor_stop_while_moving(piezo_motor: PiezoElectricMotor) -> None:
    await piezo_motor.connect(mock=DeviceMock())
    setpoint = 10
    piezo_motor.set(setpoint)
    await asyncio.sleep(0.1)
    set_mock_value(piezo_motor.user_readback, 5)
    assert await piezo_motor.within_tolerance.get_value() is False
    with pytest.raises(RuntimeError, match="Device piezo_motor was stopped."):
        await piezo_motor.stop()
```
If a motor is stopped, we should expect it to raise the `RuntimeError` with error message. This tests shows it is not behaving as intended because the `RuntimeError` is never raised because the move continues to complete in the background and therefore never raises the error.

## Acceptance Criteria
- StandardMovable when stopped cancels move task and raises the expected error message. Tests are added to prove this behaviour.

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1324


src/ophyd_async/core/_movable.py
```diff
@@ -168,31 +171,39 @@ async def set(
...
-                    unit=units,
-                    precision=precision,
-                )
-
-        if not self._set_success:
-            raise RuntimeError(f"Device {self.name} was stopped.")
+        try:
+            # Suppress CancelledError and instead we will raise our own
+            # with more useful information.
+            with contextlib.suppress(asyncio.CancelledError):
```
coretl: any reason not to use try/except rather then suppress?

oliwenmandiamond: I was adopting what was used for `SimMotor`. However, I've now redone it and think it looks much better. Instead, we only suppress the error now if the movable was deliberately stopped, else raise like normal

coretl (approved):

PR #1325, opened 2026-07-07
Fix DeviceVector-of-signals type check in PviDeviceConnector

coretl:
## Summary
- `DeviceFiller.fill_child_signal` compared the PVI-inferred signal type against the annotated one with `is`. For a `DeviceVector[SignalRW[float]]` annotation the expected type is the parameterized generic `SignalRW[float]`, but a real connection only ever reports the bare class `SignalRW` (inferred from the "r"/"w"/"rw" PVI designator) — so the comparison always failed.
- Practical effect: **any `DeviceVector` of signals connected via `PviDeviceConnector` over a real (non-mock) network connection raised**, regardless of whether the shapes actually matched. Mock-mode connections use a separate code path (`connect_mock`/`create_device_vector_entries_to_mock`) that never exercised this comparison, which is why it went undetected until now.
- Fix: compare `get_origin_class(...)` on both sides (an existing utility already used identically elsewhere in this file), which correctly unifies bare and parameterized forms.

## Context
Found while building real-IOC-backed structural PVI test coverage (nested devices / `DeviceVector` support) in a follow-up PR. That PR is stacked on top of this one and depends on it — it's where the dedicated regression coverage for this fix lands, since exercising it requires a real IOC rather than a mocked connection.

## Test plan
- [x] Full unit test suite unaffected (778 passed, 2 skipped, 2 xfailed, 6 xpassed — unchanged from baseline)
- [x] Regression coverage added in the follow-up stacked PR (`DeviceVector[SignalRW[float]]` connected over a real IOC)

### DIFF
--- a/src/ophyd_async/core/_device_filler.py
+++ b/src/ophyd_async/core/_device_filler.py
@@ -466,7 +466,13 @@ class DeviceFiller(Generic[SignalBackendT, DeviceConnectorT, CommandBackendT]):
             backend = self._signal_backend_factory(None)
             expected_signal_type = signal_type
             setattr(self._device, name, signal_type(backend))
-        if signal_type is not expected_signal_type:
+        # Compare origin classes (e.g. SignalRW) rather than the raw objects:
+        # `expected_signal_type` may be a parameterized generic alias like
+        # `SignalRW[float]` (from a `DeviceVector[SignalRW[float]]` annotation),
+        # while a real (non-mock) connection reports `signal_type` as the bare
+        # class inferred from the "r"/"w"/"rw" PVI designator -- these should
+        # be considered equal as long as they agree on the Signal kind.
+        if get_origin_class(signal_type) is not get_origin_class(expected_signal_type):
             self._raise(
                 name,
                 f"is a {signal_type.__name__} not a {expected_signal_type.__name__}",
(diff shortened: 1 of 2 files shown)


src/ophyd_async/core/_device_filler.py
```diff
@@ -466,7 +466,13 @@ def fill_child_signal(
...
             expected_signal_type = signal_type
             setattr(self._device, name, signal_type(backend))
-        if signal_type is not expected_signal_type:
+        # Compare origin classes (e.g. SignalRW) rather than the raw objects:
+        # `expected_signal_type` may be a parameterized generic alias like
+        # `SignalRW[float]` (from a `DeviceVector[SignalRW[float]]` annotation),
+        # while a real (non-mock) connection reports `signal_type` as the bare
+        # class inferred from the "r"/"w"/"rw" PVI designator -- these should
+        # be considered equal as long as they agree on the Signal kind.
+        if get_origin_class(signal_type) is not get_origin_class(expected_signal_type):
```
shihab-dls, line 475: Should: This is correct, but there doesn't seem to be a test that checks when this evaluates to false. A test that checks a generic alias of type `SignalRW[float]` does not fail this evaluation against a `SignalRW`, and another that checks different signal classes fail this check, would be good.

shihab-dls (approved):
I've left a comment about tests, but I'll approve anyways as they should be trivial to implement.

src/ophyd_async/core/_device_filler.py
coretl: There is a test for this in #1327

PR #1326, opened 2026-07-07
Add DeviceVector-of-commands support to PviDeviceConnector

coretl:
## Summary
- `PviTree.build_device_tree` grouped `"__N"`-keyed sub-devices and signals into `vector_children`, but never routed `"__N"`-keyed command (`"x"`) entries there. A `DeviceVector` of `TriggerableCommand` was silently mis-filled as literal attributes named `"__1"`/`"__2"` on the parent instead of populating the vector by index.
- This wasn't a comparison bug like #1325 — this shape (a PandA-style vector of commands served over real PVI) had **no support at all**.
- Adds a `str`-valued `vector_children` case (the command's execute PV), an `is_command_vector` classifier alongside the existing `is_signal_vector`, and the corresponding fill branch in `PviDeviceConnector.connect_real`.

## Context
Found in the same investigation as #1325 (building real-IOC-backed structural PVI test coverage for a follow-up PR). Stacked on top of #1325 since both surfaced from the same work, though they're independent fixes to different mechanisms. The follow-up PR (stacked on this one) is where dedicated regression coverage lands, since exercising this requires a real IOC rather than a mocked connection.

## Test plan
- [x] `tests/unit_tests/epics/pvi/` and `tests/unit_tests/fastcs/` unaffected (mock-mode `Block1`-`Block5` tests use a different code path, untouched by this change)
- [x] Regression coverage added in the follow-up stacked PR (`DeviceVector[TriggerableCommand]` connected over a real IOC)

### DIFF
--- a/src/ophyd_async/epics/core/_pvi_connector.py
+++ b/src/ophyd_async/epics/core/_pvi_connector.py
@@ -138,6 +138,18 @@ class PviDeviceConnector(DeviceConnector):
                         "Failed to fill DeviceVector. "
                         f"Expected SignalDetails, got {type(vector_child)}"
                     )
+            elif self.pvi_tree.is_command_vector:
+                # DeviceVector of commands
+                if isinstance(vector_child, str):
+                    backend = self.filler.fill_child_command(
+                        device.name, TriggerableCommand, vector_index
+                    )
+                    backend.write_pv = vector_child
+                else:
+                    raise TypeError(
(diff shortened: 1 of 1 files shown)


shihab-dls (approved):
This all seems reasonable; currently fastcs does not support serving vectors other than controller vectors, so I assume we will be adding facilitating changes to base fastcs next?

PR #1329, opened 2026-07-07
Update to allow for using process plugin to apply filter and adjust exposures per collection

jwlodek:


src/ophyd_async/epics/adcore/_detector.py
```diff
@@ -31,6 +31,11 @@ def __init__(
             for plugin_name, plugin in plugins.items():
                 setattr(self, plugin_name, plugin)
         if trigger_logic:
+            # If the trigger logic has a process_plugin attribute, automatically 
+            # set it to the first NDProcessIO plugin found in the plugins dictionary (if any).
+            # This allows the trigger logic to use the process plugin for preparing exposures per collection.
+            if hasattr(trigger_logic, "process_plugin"):
+                setattr(trigger_logic, "process_plugin", next(self.get_plugins_by_type(NDProcessIO), None))
```
coretl, line 38: This seems too magic to me, can we explicitly pass the correct ProcessPlugin to the TriggerLogic, then we don't need this?

src/ophyd_async/epics/adcore/_io.py
```diff
@@ -78,7 +82,7 @@ class NDArrayBaseIO(EpicsDevice):
 # Classes for drivers
 
 
-class ADImageMode(SubsetEnum):
+class ADImageMode(IntersectEnum):
     SINGLE = "Single"
     MULTIPLE = "Multiple"
     CONTINUOUS = "Continuous"
```
coretl, line 92: What does the IOC have in this case to require an `IntersectEnum`?

jwlodek: ADPerkinElmer, which has an additional "Average" Image mode.

jwlodek:
@coretl I think this is now ready for review, we've tested it with a few detectors and it works as advertised.

src/ophyd_async/epics/adcore/_io.py
```diff
@@ -277,13 +275,144 @@ class NDCircularBuffIO(NDPluginBaseIO):
...
+    enable_flat_field: A[SignalRW[bool], PvSuffix.rbv("EnableFlatField")]
+    valid_flat_field: A[SignalR[bool], PvSuffix("ValidFlatField_RBV")]
+    scale_flat_field: A[SignalRW[float], PvSuffix.rbv("ScaleFlatField")]
+    enable_offset_scale: A[SignalRW[bool], PvSuffix.rbv("EnableOffsetScale")]
+    auto_offset_scale: A[SignalRW[bool], PvSuffix("AutoOffsetScale")]
+    offset: A[SignalRW[float], PvSuffix.rbv("Offset")]
+    scale: A[SignalRW[float], PvSuffix.rbv("Scale")]
+    enable_low_clip: A[SignalRW[bool], PvSuffix.rbv("EnableLowClip")]
+    # low_clip_thresh: A[SignalRW[float], PvSuffix.rbv("LowClipThresh")]
+    # low_clip_value: A[SignalRW[float], PvSuffix.rbv("LowClipValue")]
```
jwlodek, line 322: The PV names of these signals changed from version 3-13 onwards. 3-13 is new enough that I don't think we want to make that a minimum required version, so I just leave them commented out. I was thinking, maybe we should create a new device connector for areaDetector stuff, in which we can allow for doing things based on the ADCore version? We could specify a pin via `__init__`, and then have a way to tag signals with a minimum supported version. Ophyd v1 solved this problem with subclassing with version suffixes (something like `NDProcessIO_V33`) but I really don't want to go down that road again.

If I can get https://github.com/areaDetector/ADCore/pull/607 merged, we could also use the AD specific connector to create plugin objects on the detector automatically via introspection, like with the PandA and it's blocks - but we'd need a way to version gate on ADCore version since obviously it would be only available in 3-15 plus. @coretl @thopkins32 what do you think?

coretl: I'd love to get PVI into ADCore. Need any help with that?

coretl: My thoughts would be:
- Per plugin fields go in a `$(P)$(R)PVI` PV with the database records in it. We can do that entirely database side (there are tests in ophyd-async that do this), just needs an info tag in each record.
- Per chain listing could either go in your inventory plugin as you've written, or you could hardcode `$(P)PVI` as the top level and put a single extra database record in each plugin database that publishes to it (there are tests in ophyd-async that do this too)

So you could do the whole thing in just database if you're willing to use the PVI format

jwlodek: > just needs an info tag in each record.

While this would probably be doable in the builtin plugins, I think it would be hard to enforce for downstream / third party plugins, and I think for drivers it would also be difficult to enforce across the board. I wonder if there is something that could be done entirely in the C++ of the base class like what I did with the inventory to get the API back from asyn parameters, so that no modifications downstream are required?

By the way, did you have a chance to review this PR?

PR #1330, opened 2026-07-07
Add additional epics motor record fields for motor/encoder resolution

jwlodek:
This information would be useful for dynamically calculating things like flyscanning trigger positions which are done in encoder counts rather than EGUs.


coretl (approved):

jwlodek:
It seems the one check is failing on an unrelated test, re-running doesn't fix it. Can't seem to merge because the test is failing.

coretl:
https://github.com/bluesky/ophyd-async/pull/1341, but that is hitting another flaky test! I'll add another PR for that...

PR #1334, opened 2026-07-09
Update standard detector watcher update behavior to match ophyd v1, report in collections instead of events for step scans.

jwlodek:
When working with ophyd-async standard detectors & the bluesky progress bars, I noticed the watcher update behavior was not consistent with ophyd v1 - essentially it was displaying updates in units of events, instead of collections. This meant that for step scans you would never get any progress indicator at all, since each trigger would create a new progress bar, and within the trigger the number of events was always 1. The behavior also didn't handle the path of preserving detector state. 

I've tested this PR with a step scan and the behavior now matches ophyd v1. Will test a flyscan next, although I believe the behavior there should be unchanged.


src/ophyd_async/core/_detector.py
```diff
@@ -662,7 +662,7 @@ async def trigger(self) -> AsyncIterator[WatcherUpdate[int]]:
             data_providers=ctx.streamable_data_providers,
             trigger_info=ctx.trigger_info,
             initial_collections_written=ctx.collections_written,
-            collections_requested=1,
+            collections_requested=ctx.trigger_info.collections_per_event,
```
tacaswell, line 665: I'm a bit confused by this change.

tacaswell: Ah, this is specifically in trigger so this is how many times we expect the detector to collect before the data for this step is done.

tacaswell, line 665:
```suggestion
            collections_requested=ctx.trigger_info.collections_per_event,
            watch_divisor=1,
```

Even though it works with the default, I think it is clearer to pass it here for future readers.

jwlodek: `collections_requested` should basically be the number of collections * number of events. Previously, both were always 1, so this was hard coded here. Now with the preserve detector state option, the collections per event is no longer 1, but number of events still is, hence collections requested being equal to collections per event.

Here is the equivalent setting for the flyscanning path: https://github.com/bluesky/ophyd-async/blob/f3df2f900bb39f3224469df64c3672ed408ecf18/src/ophyd_async/core/_detector.py#L681

tacaswell:
We are sure these are the only two call sites?

tacaswell (approved):
Looks good to me, but it should be verified it works with flyscanning before merging.

jwlodek:
> We are sure these are the only two call sites?

Yes, confirmed with `grep`, and it's a private method so shouldn't be any downstream hits.

```
jwlodek@alma10:~/Workspace/ophyd-async$ grep -rin _wait_for_index src/
grep: src/ophyd_async/core/__pycache__/_detector.cpython-312.pyc: binary file matches
grep: src/ophyd_async/core/__pycache__/_detector.cpython-311.pyc: binary file matches
src/ophyd_async/core/_detector.py:516:    async def _wait_for_index(
src/ophyd_async/core/_detector.py:661:        async for update in self._wait_for_index(
src/ophyd_async/core/_detector.py:709:        async for update in self._wait_for_index(
```

coretl (approved):

src/ophyd_async/core/_detector.py
```diff
@@ -662,7 +662,8 @@ async def trigger(self) -> AsyncIterator[WatcherUpdate[int]]:
...
```
thopkins32, line 665: @jwlodek I'm fairly certain this fix is larger than you gave it credit for. If `NumImages_RBV` is > 1 and the driver completes acquisition, then `trigger()`'s status completes early. With `WaitForPlugins` off, a long plugin chain or slow file writing, then `NumCaptured_RBV` could still be less than `NumImages_RBV` by the time you `collect_asset_docs()`. Which means you fail to emit `StreamResource` and `StreamDatum` for the first `Event`.

Just wanted to point this out in case anyone else runs into this bug like we did at CDI.

jwlodek: Are you saying this PR fixed what you are describing in addition to the watcher update issue, or are you saying a follow up PR is needed?

thopkins32: No follow-up needed. This PR fixes what I described above. The follow-up PR tagged below is a test to make sure this behavior doesn't revert.

PR #1340, opened 2026-07-10
Wrap Tango device-proxy connect failures as NotConnectedError

### LINKED ISSUE #1336: Tango: connecting to a nonexistent device raises raw tango.DevFailed instead of NotConnectedError
## Description

`TangoDeviceConnector.connect_real` (`src/ophyd_async/tango/core/_base_device.py`)
starts with:

```python
async def connect_real(self, device: Device, timeout: float, force_reconnect: bool):
    if not self.trl:
        raise RuntimeError(...)
    proxy = await AsyncDeviceProxy(self.trl)  # type: ignore
    ...
```

If `self.trl` is well-formed (valid host:port syntax) but no device server
is actually serving that device name, `AsyncDeviceProxy(self.trl)` raises a
raw `tango.DevFailed`/`tango.ConnectionFailed` (a PyTango-internal exception
type), which propagates out of `Device.connect()` completely unwrapped.

This is inconsistent with every other connection failure in ophyd-async,
which surfaces as `ophyd_async.core.NotConnectedError`:

- A bad *attribute* TRL (device exists, attribute doesn't) *does* get
  wrapped into `NotConnectedError`, because it fails later, inside the
  per-child `wait_for_connection` dispatch in the same method (after the
  device-level proxy already connected successfully).
- EPICS has no equivalent "whole-device proxy" connect step (CA/PVA connect
  per-signal), so it doesn't share this gap -
  `test_signals.py::test_non_existent_errors` reliably gets
  `NotConnectedError` for a bad PV today.

## Suggested fix

Wrap the `AsyncDeviceProxy(self.trl)` call (and/or the whole body of
`connect_real`) in a try/except that re-raises Tango proxy-connection
failures as `NotConnectedError`, consistent with the per-child wrapping
`wait_for_connection` already does.

## Where this was found

`tests/system_tests_tango/test_tango_signal_lifecycle.py::test_signal_error_paths`
(added in #1332) originally asserted `NotConnectedError` for a well-formed
but nonexistent device TRL, expecting this to behave like the EPICS
equivalent. It doesn't - the test currently asserts the actual behaviour
(`tango.DevFailed`) with a comment explaining why, rather than the intended
one. Once this is fixed, that test should be updated back to
`pytest.raises(NotConnectedError)`.

## References

- Discovered in bluesky/ophyd-async#1332
- Part of the system-test rewrite epic, bluesky/ophyd-async#1321

coretl:
## Summary
- `TangoDeviceConnector.connect_real` now catches `tango.DevFailed` (and its `ConnectionFailed` subclass) around the `AsyncDeviceProxy(self.trl)` call and re-raises it as `ophyd_async.core.NotConnectedError`, matching every other connection-failure path in ophyd-async (EPICS CA/PVA, and Tango's own per-child attribute/command dispatch).
- Updated `test_tango_signal_lifecycle.py::test_signal_error_paths` to assert the intended `NotConnectedError` for a well-formed-but-nonexistent device TRL, removing the comment documenting the old unwrapped behaviour.

## Test plan
- [x] `pytest tests/system_tests_tango/` (168 passed)
- [x] `pyright src --pythonpath <venv-python>` (0 errors)

Fixes #1336

https://claude.ai/code/session_01TN8jcKUkoeSzCv1sYF3neN


burkeds (commented):
Looks good

PR #1341, opened 2026-07-10
Fix flaky timing budget in test_many_individual_device_connects_not_slow

coretl:
## Summary
- CI run [29095956743](https://github.com/bluesky/ophyd-async/actions/runs/29095956743) failed `test (ubuntu-latest, 3.12)` on a routine push to `main` with `assert 1.0255489799999964 < 1.0` in `test_many_individual_device_connects_not_slow[0-True]`. Tom reports this fails roughly 1 in 10 CI runs, never reproduced locally.
- Investigated whether this was a real perf regression or CI noise on a too-tight budget.

## Investigation
- The triggering commit (`116a68ea1`, "Narrow Array1D to genuinely 1-D at the type-checker level") only edits a `TypeAlias` in `_datatypes.py` — no runtime code path is touched, ruling out a regression from that specific push.
- Pulled `--durations=20` output (via `gh api repos/.../actions/jobs/<id>/logs`) from ~10 unrelated green `ubuntu-latest` CI runs spanning 2026-06-24 to 2026-07-10. The sequential-connect branch (`[0-True]`) has consistently measured **0.79s–0.99s against the 1.0s budget** the entire time — i.e. as little as ~1% headroom on a normal *passing* run, well before this failure. This rules out a recent regression and points to long-standing runner noise hitting a razor-thin budget.
- Locally, the same 500-mock-connect workload completes in ~0.15–0.25s (24-core dev sandbox), even under artificial CPU contention (`taskset` pinned to 2 cores + background busy loops). CI's GitHub-hosted runners are simply much slower/noisier than dev hardware, and the code itself shows no sign of slowing down.
- Checked recent history of `_device.py`, `_soft_signal_backend.py`, and the `Motor`/mock-connect path for anything that could add per-connect overhead: the only relevant recent change (`e9a9e1987`, caching setattr dispatch) is a **perf improvement**, not a regression.
- Confirmed the repo has no existing convention for retrying flaky wall-clock tests (`pytest-rerunfailures` is a listed dependency but not wired into `addopts`, and no test uses `@pytest.mark.flaky` anywhere) — introducing one here would be a bigger, unilateral architectural change, so I didn't.

## Conclusion
Genuine CI-runner noise against a too-tight, long-standing hardcoded budget — not a code regression.

## Fix
- Raised the non-Windows `expected_duration` from 1.0s to 1.5s, justified by the real CI duration distribution above (comfortable margin over all observed durations, including the 1.0255s failure, while still catching a real quadratic/serial-connect regression).
- Added inline comments: the `parallel=True` parametrize value confusingly takes the *sequential* connect-in-a-loop branch, and `parallel=False` is the one that actually calls `wait_for_connection` concurrently. Flagging this because it's easy to misread when triaging future flakes — left the parametrize values as-is to avoid changing test IDs, just clarified in comments.

## Test plan
- [x] `ruff check` / `ruff format --check` clean on the touched file
- [x] `pytest tests/unit_tests/core/test_device.py` passes (19 passed)
- [x] Full `pytest tests/unit_tests` passes (777 passed, 2 skipped, 2 xfailed, 6 xpassed)
- [ ] CI green (this fix is itself timing-based so could theoretically still flake rarely; will monitor)

Found via CI run https://github.com/bluesky/ophyd-async/actions/runs/29095956743.

https://claude.ai/code/session_01TN8jcKUkoeSzCv1sYF3neN

### DIFF
--- a/tests/unit_tests/core/test_device.py
+++ b/tests/unit_tests/core/test_device.py
@@ -221,30 +221,30 @@ class MotorBundle(Device):
         super().__init__(name)
 
 
-@pytest.mark.parametrize("parallel", (False, True))
+@pytest.mark.parametrize("serial", (False, True))
 @pytest.mark.parametrize("execution_number", range(1))
-async def test_many_individual_device_connects_not_slow(parallel, execution_number):
+async def test_many_individual_device_connects_not_slow(serial, execution_number):
     start = time.monotonic()
     bundles = [MotorBundle(f"bundle{i}") for i in range(100)]
-    if parallel:
-        # NOTE: despite the parameter name, this connects each bundle
-        # sequentially (500 individual mock connects, one after another).
+    if serial:
+        # Connect each bundle sequentially (500 individual mock connects,
+        # one after another). This is the slower, tighter-budget case.
         for bundle in bundles:
             await bundle.connect(mock=True)
     else:
-        # ...whereas this is the branch that actually connects in parallel,
-        # via wait_for_connection gathering all the coroutines at once.
+        # Connect all bundles in parallel, via wait_for_connection
+        # gathering all the coroutines at once.
         coros = {bundle.name: bundle.connect(mock=True) for bundle in bundles}
         await wait_for_connection(**coros)
     duration = time.monotonic() - start
     # Windows runners on GitHub are slow...
     # On Linux, shared/throttled GitHub-hosted runners routinely take
-    # 0.8-1.0s for the sequential-connect case (500 mock connects), leaving
-    # almost no headroom against a 1.0s budget: CI history on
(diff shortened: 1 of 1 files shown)


coretl:
Follow-up: renamed the confusing \`parallel\` parametrize dimension to \`serial\` (bc448ce51) so it actually describes each branch (\`serial=True\` = sequential connect-in-a-loop, \`serial=False\` = \`wait_for_connection\` gather). Parametrized test IDs are unchanged (\`[0-True]\` is still the sequential/tight-budget case), so the comments/budget from the first commit still line up. Re-ran \`ruff check\`/\`ruff format --check\` and the full \`tests/unit_tests\` suite (777 passed) after the rename.

Issue #1102, opened 2025-10-22
Add missing signals to SimMotor and create Motor Protocol


oliwenmandiamond:
When using a motor in plans for testing and live, we want to specify a motor device but don't care about underlying implementation. At the moment, we're having to extend `SimMotor` to add missing signals and then use type: ignore when used in test plans. We should do the following things:

- Add the following missing signals to SimMotor 
```Python
def __init__(
    self,
    ...
    low_limit: float = float("-inf"),
    high_limit: float = float("inf"),
    max_velocity: float = 100,
    ...
):
    ...
    self.max_velocity = soft_signal_rw(float, max_velocity
    self.low_limit_travel = soft_signal_rw(float, low_limit)
    self.high_limit_travel = soft_signal_rw(float, high_limit)
   ...
```
- Create a `Motor` protocol which defines the the base signals a device needs to define so that any device or plan can specify Motor protocol so it doesn't matter if a `SimMotor` or epics `Motor`
- Possibly rename `Motor` to `EpicsMotor`?

## Acceptance Criteria
- Generic interface for Motor for plans and devices created so it doesn't matter on underlying implementation

jwlodek:
I suggest if we rename `Motor`, to use `AsyncEpicsMotor` (and maybe `AsyncSimMotor`, since those same class names are used in ophyd v1. Would be good to make them easily distinguishable.

oliwenmandiamond:
I would say make a protocol `AsyncEpicsMotor` and then have EpicsMotor and SimMotor (rename to SimEpicsMotor?) implement it. This way, I can provide my plan and / or device and it doesn't matter if it's live or sim. I don't see the point in distinguishing between SimMotor and Motor as the point of this is to make that a plan or device it doesn't matter if Sim or Live, I just declare the type as AsyncEpicsMotor

coretl:
I'm inclined to go the opposite direction:
- Remove from `SimMotor` anything that is not directly related to the simplest possible simulated motor, and make the signals intentionally different from `epics.Motor` to differentiate them
- Make a more useful mock mode for `epics.Motor` to allow you to test your plan (refs #1039)

What extra mock behaviour would you need to make testing plans easier?

oliwenmandiamond:
At the moment, we are checking the upper and lower limits of a motor in some of our plans. This doesn't work when using a `SimMotor` as it doesn't define `low_limit_travel` and `high_limit_travel` so we have to extend `SimMotor` to add those signals and then either

[...]

PR #1344, opened 2026-07-10
Add a device mock for the epics motor that respects velocity and acceleration time.

jwlodek:
Useful for more representative CI. Also used to confirm progress bars were functioning correctly, since the instant motor mock would immediately be at the target.

1. Connect to an EPICS motor with the new mock class
2. Perfrom a move and ensure it blocks for the correct amount of time.

### Checks for reviewer
- [X] Would the PR title make sense to a user on a set of release notes
- [X] If the change requires a bump in an IOC version, is that specified in a `##Changes` section in the body of the PR
- [X] If the change requires a bump in the PandABlocks-ioc version, is the `ophyd_async.fastcs.panda._hdf_panda.MINIMUM_PANDA_IOC` variable updated to match

### DIFF
--- a/src/ophyd_async/epics/motor.py
+++ b/src/ophyd_async/epics/motor.py
@@ -27,6 +27,7 @@ from ophyd_async.core import (
     callback_on_mock_put,
     default_mock_class,
     error_if_none,
+    set_mock_put_proceeds,
     set_mock_value,
 )
 from ophyd_async.core import StandardReadableFormat as Format
@@ -179,6 +180,42 @@ class InstantMotorMock(DeviceMock["Motor"]):
             set_mock_value(device.motor_done_move, 1)  # Done
 
         callback_on_mock_put(device.user_setpoint, _instant_move)
+
+
+class VelocityRespectingMotorMock(DeviceMock["Motor"]):
+    """Mock behaviour that respects motor velocity and acceleration time."""
+
+    async def connect(self, device: Motor) -> None:
+        """Mock signals to simulate a move respecting velocity and acceleration."""
+        set_mock_value(device.velocity, 10)
+        set_mock_value(device.max_velocity, 100)
+        set_mock_value(device.acceleration_time, 1)
+
+        # Motor starts in "done" state (not moving)
+        set_mock_value(device.motor_done_move, 1)
+
+        async def _do_move(target: float):
+            current = await device.user_readback.get_value()
+            velocity = await device.velocity.get_value()
+            acceleration_time = await device.acceleration_time.get_value()
+            move_time = abs(target - current) / velocity + 2 * acceleration_time
+            set_mock_value(device.motor_done_move, 0)
+            elapsed = 0.0
+            while elapsed < move_time:
+                await asyncio.sleep(min(0.1, move_time - elapsed))
+                elapsed += 0.1
+                fraction = min(elapsed / move_time, 0.1)
+                position = current + (target - current) * fraction
+                set_mock_value(device.user_readback, position)
+            set_mock_value(device.user_readback, target)
+            set_mock_value(device.motor_done_move, 1)
+            set_mock_put_proceeds(device.user_setpoint, True)
+
+        def _on_setpoint_write(value):
+            set_mock_put_proceeds(device.user_setpoint, False)
+            asyncio.ensure_future(_do_move(value))
+
+        callback_on_mock_put(device.user_setpoint, _on_setpoint_write)
 
 
 @default_mock_class(InstantMotorMock)
--- a/tests/unit_tests/epics/test_motor.py
+++ b/tests/unit_tests/epics/test_motor.py
@@ -21,7 +21,7 @@ from ophyd_async.core import (
     set_mock_value,
     soft_signal_rw,
 )
-from ophyd_async.epics.motor import Motor, MotorLimitsError
+from ophyd_async.epics.motor import Motor, MotorLimitsError, VelocityRespectingMotorMock
 from ophyd_async.testing import (
     StatusWatcher,
     wait_for_pending_wakeups,
@@ -44,6 +44,13 @@ async def motor():
     set_mock_value(motor.high_limit_travel, 21)
     set_mock_value(motor.dial_low_limit_travel, -11)
     set_mock_value(motor.dial_high_limit_travel, 21)
+    yield motor
+
+
+@pytest.fixture
+async def velocity_respecting_motor():
+    motor = Motor("BLxxI-MO-TABLE-01:X", name="motor")
+    await motor.connect(mock=VelocityRespectingMotorMock())
     yield motor
 
 
@@ -565,3 +572,27 @@ async def test_instant_motor_mock_preserves_parent_mock_tracking():
     # Verify the mock calls include the child operations
     assert any("x" in str(call) for call in parent_mock_obj.mock_calls)
     assert any("y" in str(call) for call in parent_mock_obj.mock_calls)
+
+
+async def test_velocity_respecting_motor_mock_behavior(
+    velocity_respecting_motor: Motor,
+):
+    """Test that move time matches distance / velocity + 2 * acceleration_time."""
+    await velocity_respecting_motor.velocity.set(50.0)
+    await velocity_respecting_motor.acceleration_time.set(0.05)
+
+    # Expected: abs(10 - 0) / 50 + 2 * 0.05 = 0.3s
+    start = asyncio.get_event_loop().time()
+    status = velocity_respecting_motor.set(10.0)
+    await status
+    assert status.success
+    assert await velocity_respecting_motor.user_readback.get_value() == 10.0
+    assert asyncio.get_event_loop().time() - start == pytest.approx(0.3, abs=0.01)
+
+    # Expected: abs(-5 - 10) / 50 + 2 * 0.05 = 0.4s
+    start = asyncio.get_event_loop().time()
+    status = velocity_respecting_motor.set(-5.0)
+    await status
+    assert status.success
+    assert await velocity_respecting_motor.user_readback.get_value() == -5.0
+    assert asyncio.get_event_loop().time() - start == pytest.approx(0.4, abs=0.01)


coretl:
Can we use SimMotor for this instead? Otherwise we're going to end up making lots of conflicting mock devices with different sets of logic for everyone's different test purposes. I would prefer:
- Mocked devices just do the bare minimum as quickly as possible so you can use them in test plans and they don't error. Different tests will augment them in different ways to test the calling code
- Sim devices are sensible "real time" simulations of the device, meant for tutorial code, demos, and UI testing

What do you think?

oliwenmandiamond:
> Can we use SimMotor for this instead? Otherwise we're going to end up making lots of conflicting mock devices with different sets of logic for everyone's different test purposes. I would prefer:
> 
>     * Mocked devices just do the bare minimum as quickly as possible so you can use them in test plans and they don't error. Different tests will augment them in different ways to test the calling code
> 
>     * Sim devices are sensible "real time" simulations of the device, meant for tutorial code, demos, and UI testing
> 
> 
> What do you think?

I think the issue with using `SimMotor` at the moment is it doesn't have all of the same matching signals as epics `Motor`. So when you type your plan or device to use `Motor`, it is very awkward to simulate it as you either have to do `SimMotor | Motor` everywhere or use `# type:ignore` when forcing through a SimMotor.

I think this also circles back to this issue I raised before https://github.com/bluesky/ophyd-async/issues/1102. If we have a `Motor` protocol, it is then much easier to type your plan / devices to use a `Motor` and not care about the underlying behaviour. Mock instant, real backend, simulate real backend move etc, doesn't matter.

Maybe something like this

```python
class Motor(StandardMovable[float])
        readback: SignalR[float
        setpoint: SignalRW[float]
        velocity: SignalRW[float]
        acceleration: SignalRW[float]
        ...
```

Then when static typing your thing, you can use `Motor` (epics.motor.Motor should be renamed to EpicsMotor).

and then you use 
```python
def my_thing(motor: Motor):
     ...

my_thing(motor=SimMotor(...))
my_thing(motor=EpicsMotor(...))
my_thing(motor=DemoMotor(...))
```

With no issue or additional work. We just have to agree what the bare minimum of what a Motor is and then each implementation comply with it.

coretl:
> I think this also circles back to this issue I raised before #1102. If we have a `Motor` protocol, it is then much easier to type your plan / devices to use a `Motor` and not care about the underlying behaviour. Mock instant, real backend, simulate real backend move etc, doesn't matter.

I don't have enough of an idea of what plans expect a `Motor` to be to comment on a `Motor` protocol. If we have some clear examples of what plans require this protocol (e.g. do they just set `velocity`, or need `acceleration_time` too?) then we can discuss further in a different issue. Either way, I think adding behaviour to `SimMotor` is useful even without it, for plans like `mv` that just want a `Movable`

jwlodek:
The one thing I think extending SimMotor wouldn't be able to do would be to be a stand-in automatically with `init_devices(mock=True)`. I've set this up as our default motor mock in our beamline `profiles` (equivalent to your dodal connect), and you can enable mock mode with a flag. Having the motors behave closer to the real ones is valuable in that case. Is there a way we could have the epics motor initialize to a SimMotor instead for this use case?

If we don't want this upstream, that's OK too, I originally had it in a beamline's repo, but thought it might be more broadly useful, hence the PR.

oliwenmandiamond:
@jwlodek that makes more sense. We also have a similar use case for testing [fly scanning](https://github.com/DiamondLightSource/sm-bluesky/pull/100#discussion_r2358375012). We need the motor in mock mode to still move from position A to B and not be instantaneous. If we have devices that have Motors as sub devices, then it becomes awkward to test as you have to sub it out with a SimMotor or recreate the device 

I personally think having multiple mock behaviours to choose from isn't a bad idea and we have enough use cases for it. @coretl thoughts on adding this to ophyd-async?

coretl:
> I personally think having multiple mock behaviours to choose from isn't a bad idea and we have enough use cases for it. @coretl thoughts on adding this to ophyd-async?

Ok, I think I'm convinced, but it needs some docs so that people know it is here. Please could you add those @jwlodek?

src/ophyd_async/epics/motor.py
```diff
@@ -181,6 +182,42 @@ def _instant_move(value):
...
+                fraction = min(elapsed / move_time, 0.1)
+                position = current + (target - current) * fraction
+                set_mock_value(device.user_readback, position)
+            set_mock_value(device.user_readback, target)
+            set_mock_value(device.motor_done_move, 1)
+            set_mock_put_proceeds(device.user_setpoint, True)
+
+        def _on_setpoint_write(value):
+            set_mock_put_proceeds(device.user_setpoint, False)
+            asyncio.ensure_future(_do_move(value))
```
coretl, line 216: `callback_on_mock_put` can take sync or async callbacks, so please can you merge these 2 functions.

I'll make a PR making the docs clearer on this.

coretl (changes requested):

oliwenmandiamond:
I have also come across a use case that I developed myself for a custom device

https://github.com/DiamondLightSource/dodal/pull/2189
https://github.com/DiamondLightSource/sm-bluesky/pull/363

I need to test fly scanning with the device but currently my device always instantly completes, making the test not very good. I have opted to make by default the device to not complete instantly. Therefore, if you need tests to be quick you have to manually make it instant. This also means if I connect in mock mode via blueapi, it also simulate the beamline as I can still do fly scanning without actually do anything special to connect in mock mode.

So the way this can done is 

```python
# default mock, not instant
with init_devices(mock=True):
    my_device = MyDevice(...)
```

or I specifically make it instant which is useful for tests

```python
my_device = MyDevice(..., name="my_device")
# Configure to be instant
await my_device.connect(mock=MockMyDevice(...))
```
What I can foresee being awkward with this is if I have another complicated device which has children as `Motor`s for example. How do I configure the child devices to be instant or not? Do I have to create a DeviceMock for this device that can propagate this configuration to the child devices? Do I have to connect each individual device with the mock behaviour I need e.g


```python
with init_devices(mock=True):
    my_device = MyDevice(...)

my_device.motor_a.connect(mock=InstantMockDevice(...)) # ?
my_device.motor_b.connect(mock=InstantMockDevice(...)) # ?
```

jwlodek:
The way I've worked around that is having a fixture that updates the default mock for a device class prior to instantiation. Then I can include that fixture for tests where I want specific behavior. Probably if you want to isolate a specific child device to have different mock behavior than other children that are of the same type, probably subclassing is the best bet? Would that work for you?

oliwenmandiamond:
> The way I've worked around that is having a fixture that updates the default mock for a device class prior to instantiation. Then I can include that fixture for tests where I want specific behavior. Probably if you want to isolate a specific child device to have different mock behavior than other children that are of the same type, probably subclassing is the best bet? Would that work for you?

That is actually a really elegant way to customise it for tests using a fixture! Yeah, I think that would work for me

coretl:
> The way I've worked around that is having a fixture that updates the default mock for a device class prior to instantiation.

That sounds neat, do you think we should turn the decorator into a context manager to help this pattern?

```python
@default_mock_class(InstantMotorMock)
class Motor(StandardMovable[float]):
    ...

# this could also be a fixture
def test_velocity_motor():

    with default_mock_class(Motor, VelocityRespectingMotorMock):
        async with init_device(mock=True):
            velocity_device = MyComplicatedDevice("foo", "bar")
    # Now velocity_device's mock is VelocityRespectingMotorMock

    async with init_device(mock=True):
        normal_device = MyComplicatedDevice("foo", "bar")
    # But normal_device's mock is InstantMotorMock
```

jwlodek:
That would be nice, or maybe even an arg to init_devices? Like a dict mapping type to default mock to use?

coretl:
> That would be nice, or maybe even an arg to init_devices? Like a dict mapping type to default mock to use?

That's interesting, we could overload `connect(mock=...)` even more so mock can be:
- `False`: real connect
- `True`: use default mock class
- `type[DeviceMock]`: use this mock class
- `dict[type[Device]: type[DeviceMock]]`: for each child, if it is an instance of supplied device then use given mock class

Then `init_devices` could be of the same type

jwlodek:
I like that. Limits the number of args, but makes things super flexible.

coretl:
> I like that. Limits the number of args, but makes things super flexible.

ok, I'll add in another PR. Are you happy to finish this one up?

jwlodek:
@coretl Done

src/ophyd_async/epics/motor.py
```diff
@@ -261,6 +269,39 @@ def _instant_move(value):
...
+        """Mock signals to simulate a move respecting velocity and acceleration."""
+        set_mock_value(device.velocity, 10)
+        set_mock_value(device.max_velocity, 100)
+        set_mock_value(device.acceleration_time, 0.1)
+
+        # Motor starts in "done" state (not moving)
+        set_mock_value(device.motor_done_move, 1)
+
+        async def _velo_and_accl_respecting_move(value):
+            set_mock_put_proceeds(device.user_setpoint, False)
```
coretl: I'm wondering why you had to do this? The puts should be blocked while `_velo_and_accl_respecting_move` runs anyway?

src/ophyd_async/epics/motor.py
```diff
@@ -261,6 +269,39 @@ def _instant_move(value):
...
+            current = await device.user_readback.get_value()
+            velocity = await device.velocity.get_value()
+            acceleration_time = await device.acceleration_time.get_value()
+            move_time = abs(value - current) / velocity + 2 * acceleration_time
+            set_mock_value(device.motor_done_move, 0)
+            elapsed = 0.0
+            while elapsed < move_time:
```
coretl: This appears to move in the correct time, but at a constant velocity rather than accelerating and decelerating. The code for doing that is here:
https://github.com/bluesky/ophyd-async/blob/b400cb1e767ce01ad1d5043f5e254ef90613370c/src/ophyd_async/sim/_motor.py#L37-L104

Should we extract it to a function in core and use in both places?

oliwenmandiamond: Yes, think we should do that! It does beg the question though if SimMotor is needed now that epics Motor can do both?

coretl: yes, because the sim devices are there for tutorials and tests, where you need a real device rather than a mock, and because they hook into the pattern generator detector too

PR #1349, opened 2026-07-13
result to abstract method from property to keep with protocol - resolves #1306

### LINKED ISSUE #1306: Update `AsyncStatusBase` to follow protocol
https://github.com/bluesky/ophyd-async/blob/abc4ecdde1aeb3d40d303e231ebfe8118d9108ae/src/ophyd_async/core/_status.py#L96-L99

needs to look like

https://github.com/bluesky/bluesky/pull/2030

tomkane-dls:
this seems like a breaking change that needs to be documented:

the test cases needed to be updated from `result` to `result()` 

using status.result like this:
```
if status.result:
   ...
```
would now always return `true` and silently fail if the syntax wasn't updated


gkalua (commented):
Looks good

coretl (approved):

PR #1357, opened 2026-07-14
Fix incorrect type for epics motor SREV

jwlodek:


jwlodek:
Going to self merge as this is a trivial change and CI passes.

PR #1359, opened 2026-07-15
StandardFlyer -> StandardFlyable

### LINKED ISSUE #1220: Convert PmacTrajectoryTriggerLogic from a device class to a logic class
`PmacTrajectoryTriggerLogic` was converted from a FlyerController to a full Device (implementing Stageable, Preparable, Flyable directly).

 It became necessary for the PMAC, where stage must also run an empty trajectory to reset the EQU on the Panda Brick.

## Acceptance Criteria
- `PmacTrajectoryTriggerLogic` is not a Device 
- Architecture Overhaul
- Rework how StandardFlyer handles stage and unstage.

### LINKED ISSUE #1241: StandardFlyer -> StandardFlyable

coretl:
Fixes #1241.
Fixes #1220.

Converts `StandardFlyer`/`FlyerController` into a `StandardMovable`-style pair — a `StandardFlyable` mix-in plus a fill-in `FlyableLogic` — and migrates every in-repo consumer (the three PandA logics, the Pmac trajectory logic, and `Motor`).

## What changed

- **`FlyableLogic[PrepareT, CtxT]`** — the minimum logic to fly. Fill in `on_prepare` / `on_kickoff` / `on_complete` (plus optional `stop`/`on_stage`/`on_unstage`). State that must survive between stages is threaded through an explicit **context object** (`on_prepare` returns it, `on_kickoff` receives+returns it, `on_complete` receives it) rather than stashed on the logic; stateless flyers use `CtxT=None`.
- **`StandardFlyable[PrepareT, CtxT]`** — the device mix-in. It owns the context, enforces `prepare -> kickoff -> complete` ordering (so logics don't self-guard), and composes `stage()`/`unstage()` with the other Standard mix-ins via `_StandardBase` instead of shadowing them.
- **Watchable `complete()`** — `complete()` is a `WatchableAsyncStatus`. When the logic is also a `MovableLogic` (e.g. a flying `Motor`), it reports progress by observing the readback, reusing the exact same watcher-update stream as `StandardMovable.set` (shared `observe_watcher_updates` helper). Non-movable flyers just block.
- **`FlyableLogic.with_device(name="")`** — wraps a logic in an ephemeral `StandardFlyable` for use directly in a plan, e.g. `flyer = StaticPcompFlyableLogic(pcomp).with_device(name="pcomp")`.
- **`Motor`** now flies via a single combined `MotorFlyableMovableLogic` backing both `movable_logic` and `flyable_logic`; `MotorMoveLogic` folded in. Only writes `STOP` when actually moving, so a set+staged motor isn't double-stopped.
- **Hard rename, no deprecated aliases.** `*TriggerLogic` flyable classes renamed to `*FlyableLogic` to disambiguate from the detector `DetectorTriggerLogic` family.
- Docs (`how-to/choose-right-baseclass`, `explanations/devices-signals-backends` diagram) updated to teach the new pair.

## Validation

Full CI gate green locally: `tox -e pre-commit type-checking docs tests` — pre-commit OK, pyright 0 errors, docs (`--fail-on-warning`) OK, **852 passed** (2 skipped, 2 xfailed, 6 xpassed, 82% cov).


src/ophyd_async/core/_flyable.py
coretl, line 1: This should be called _flyable.py

src/ophyd_async/core/_movable.py
```diff
@@ -47,6 +47,39 @@ def __call__(self) -> float | None:
...
+    """
+    async for current_position in observe_value(readback, done_status=done_status):
+        yield WatcherUpdate(
+            current=current_position,
+            initial=initial,
+            target=target,
+            name=name,
+            unit=units,
+            precision=precision,
+        )
```
coretl: I don't think this function proves its worth, inline it in the 2 places it is used

src/ophyd_async/core/_flyable.py
coretl: Done in 4cfd28eb6 — renamed `_flyer.py` to `_flyable.py` (updated the `core/__init__` import and `conf.py` obj_ignore).

src/ophyd_async/core/_movable.py
coretl: Done in 4cfd28eb6 — inlined the observe/yield loop back into `StandardMovable.set` and `StandardFlyable.complete` and removed the helper.

tests/unit_tests/epics/pmac/test_pmac_trajectory.py
coretl, line 1: All the tests in this file should act on the public surface, i.e. `PmacTrajectoryFlyableLogic.with_device()`. That should cut out most of the changes in this file

src/ophyd_async/epics/motor.py
```diff
@@ -159,6 +188,52 @@ async def move(self, new_position: float, timeout: TimeoutCalculator) -> None:
...
+        """Begin moving the motor from the prepared position to the final position."""
+        acceleration_time = await self.acceleration_time.get_value()
+        target = ctx.fly_info.ramp_down_end_pos(acceleration_time)
+        if ctx.fly_info.timeout == CALCULATE_TIMEOUT:
+            initial = await self.readback.get_value()
+            timeout = await self.calculate_timeout(initial, target)
+        else:
+            timeout = ctx.fly_info.timeout
+        ctx.status = AsyncStatus(self.move(target, lambda: timeout))
+        return ctx
```
coretl, line 233: I'm sure at some point this did `sleep(acceleration_time)` so that we were up to speed before kickoff. I think it should do we can kickoff motors, then kickoff internal triggered detectors, and it will be more accurate for long acceleration times

coretl: Done in fa37cb95f — `on_kickoff` now `await asyncio.sleep(acceleration_time)` after starting the move, so `kickoff()` returns once the motor is up to speed (fly tests use a small acceleration_time to keep the wait negligible).

tests/unit_tests/epics/pmac/test_pmac_trajectory.py
coretl: Done in fa37cb95f — the lifecycle tests now drive `PmacTrajectoryFlyableLogic.with_device()` (prepare/kickoff/complete/stage/unstage), dropping the manual on_*/context threading. The remaining logic-object uses are internal-helper tests (`_move_to_start`, `_check_profile_status`, and the `stop` primitive), which have no public equivalent.

gkalua (approved):

src/ophyd_async/epics/motor.py
```diff
@@ -159,6 +188,57 @@ async def move(self, new_position: float, timeout: TimeoutCalculator) -> None:
...
+        await asyncio.gather(
+            self.check_move(ramp_up_start_pos), self.check_move(ramp_down_end_pos)
+        )
+        # Move to the run-up start at maximum velocity
+        await self.velocity.set(abs(max_speed))
+        old_position = await self.readback.get_value()
+        timeout = await self.calculate_timeout(old_position, ramp_up_start_pos)
+        await self.move(ramp_up_start_pos, lambda: timeout)
+        # Set the velocity we will use for the fly scan
+        await self.velocity.set(abs(value.velocity))
```
shihab-dls: nit: The assumption here is that `velocity: SignalRW[float]` is expected to be positive, but `FlyMotorInfo` may return a velocity that is negative if `end_position` is lesser than `start_position`. This is fine, but given we always evaluate on `abs(value.velocity)`, I'm wondering if it's better to call `abs()` in `FlyMotorInfo.velocity` instead (I guess in all cases we seem to need speed and not velocity).

tests/unit_tests/epics/test_motor.py
```diff
@@ -287,77 +284,80 @@ async def test_prepare_valid_limits(motor: Motor):
...
+    set_mock_value(motor.acceleration_time, 0.2)
+    set_mock_value(motor.max_velocity, 100)
+    await motor.prepare(
+        FlyMotorInfo(start_position=12, end_position=2, time_for_move=1)
     )
     await motor.kickoff()
-    mock_set.assert_called_once_with(-3.0, timeout=CALCULATE_TIMEOUT)
+    # kickoff starts the move to the ramp-down end position (1.0)
+    await motor.complete()
+    assert await motor.user_setpoint.get_value() == 1.0
```
shihab-dls: should: I don't think it's right to call `motor.prepare()` or `motor.complete()` in this test, as future issues in those hooks will show up as exceptions in this kickoff test. I think we should mock what we need to in order to make sure we only test `kickoff()`s scope.

tests/unit_tests/epics/test_motor.py
```diff
@@ -287,77 +284,80 @@ async def test_prepare_valid_limits(motor: Motor):
...
-    assert motor._fly_status.done
+    with pytest.raises(RuntimeError, match="kickoff.* before complete"):
+        await motor.complete()
+    set_mock_value(motor.acceleration_time, 0.2)
+    set_mock_value(motor.max_velocity, 100)
+    await motor.prepare(
+        FlyMotorInfo(start_position=0, end_position=10, time_for_move=1)
+    )
+    await motor.kickoff()
+    status = motor.complete()
```
shihab-dls: same here, as we are calling `prepare` and `kickoff` on a test about `complete`

tests/unit_tests/core/test_flyer.py
```diff
@@ -0,0 +1,272 @@
...
+        self.calls.append("on_complete")
+
+    async def stop(self) -> None:
+        self.calls.append("stop")
+
+
+async def test_ephemeral_flyable_drives_logic_hooks():
+    logic = RecordingFlyableLogic()
+    async with init_devices(mock=True):
+        flyer = logic.with_device(name="flyer")
```
shihab-dls: nit: looks like we make this flyer a lot; I would make a fixture for it at the top.

tests/unit_tests/core/test_flyer.py
```diff
@@ -0,0 +1,272 @@
...
+        return {"prepared": value}
+
+    async def on_kickoff(self, ctx: dict) -> dict:
+        self.seen.append(("kickoff", dict(ctx)))
+        ctx["kicked"] = True
+        return ctx
+
+    async def on_complete(self, ctx: dict) -> None:
+        self.seen.append(("complete", dict(ctx)))
+
```
shihab-dls, line 59: nit: I would put this dataclass near the top with the other logic class.

tests/unit_tests/core/test_flyer.py
```diff
@@ -0,0 +1,272 @@
...
+async def test_flyable_complete_before_kickoff_raises():
+    logic = RecordingFlyableLogic()
+    async with init_devices(mock=True):
+        flyer = logic.with_device(name="flyer")
+
+    await flyer.prepare(5)
+    with pytest.raises(RuntimeError, match="kickoff.* before complete"):
+        await flyer.complete()
+    assert logic.calls == ["on_prepare"]
```
shihab-dls: should: no need to call `prepare` in this test
```suggestion
async def test_flyable_complete_before_kickoff_raises():
    logic = RecordingFlyableLogic()
    async with init_devices(mock=True):
        flyer = logic.with_device(name="flyer")

    with pytest.raises(RuntimeError, match="kickoff.* before complete"):
        await flyer.complete()
    assert logic.calls == []
```

tests/unit_tests/core/test_flyer.py
```diff
@@ -0,0 +1,272 @@
...
+        return MovableFlyableLogic(setpoint=self.setpoint, readback=self.readback)
+
+    @cached_property
+    def movable_logic(self) -> MovableFlyableLogic:
+        return self._logic
+
+    @cached_property
+    def flyable_logic(self) -> MovableFlyableLogic:
+        return self._logic
+
```
shihab-dls: nit: Again I would put this near the top with the other logic classes.

shihab-dls (changes requested):
The logic implementation is really good, but I'm requesting some changes regarding the tests.

src/ophyd_async/epics/motor.py
coretl: Good call. I think we should change `FlyMotorInfo.velocity` to `.speed` and make it do the `abs` there

tests/unit_tests/epics/test_motor.py
coretl: I think I prefer the opposite approach, merge the `prepare`, `kickoff` and `complete` tests into a single "happy path" test using only the public interface, then having small "unhappy path" tests that are majority public interface, but can use mocks in order to shorten sequences or inject errors.

This approach should be noted in `CLAUDE.md` too so that this testing strategy is always followed.

src/ophyd_async/epics/motor.py
coretl: Done in 97ea7b5bf — `FlyMotorInfo.velocity` is now `.speed` and does the `abs` there, so callers drop the `abs()`. The ramp-up/down helpers still need the signed direction, so that is factored into a private `_ramp_distance` (behaviour unchanged).

tests/unit_tests/epics/test_motor.py
coretl: Done in 31fd54914 — merged `test_prepare`/`test_kickoff`/`test_complete` into a single public-interface `test_fly_scan` happy path (parametrised both directions), plus small `test_kickoff_before_prepare_raises`/`test_complete_before_kickoff_raises` for the ordering errors. Also noted this happy-path/unhappy-path strategy in `CLAUDE.md`.

tests/unit_tests/epics/test_motor.py
coretl: Done in 31fd54914 — folded into the single `test_fly_scan` happy path (see above); the complete-before-kickoff error is now its own small `test_complete_before_kickoff_raises`.

tests/unit_tests/core/test_flyer.py
coretl: Done in 1522565d0 — added a `recording_flyer` fixture near the top that returns `(logic, flyer)`.

tests/unit_tests/core/test_flyer.py
coretl: Done in 1522565d0 — moved `CtxFlyableLogic` up with the other logic classes at the top of the file.

tests/unit_tests/core/test_flyer.py
coretl: Done in 1522565d0 — dropped the `prepare()` call; the test now just asserts `complete()` raises with `logic.calls == []`.

tests/unit_tests/core/test_flyer.py
coretl: Done in 1522565d0 — moved `MovableFlyableLogic`/`MovableFlyer` up with the other logic classes.

shihab-dls (approved):
Looks good to me! Tests seem reasonable, and I like the idea of a single happy path test followed by isolated unhappy path tests. Approving.

Issue #1248, opened 2026-04-18
Can StandardDetector be collectable instead of writing asset streams?


canismarko:
I have a question that might also be a feature request. I am trying to apply the new StandardDetector() to some new devices that can be used in fly scanning, but cannot write data to disk.

- Struck SIS3820 scaler
- MeasComp USB CTR-08 counter
- TetrAmm electrometer

These devices have inputs for hardware triggering and we use them for fly scanning, so I need to be able to pass these devices to `bps.collect()` and have it produce the right event docs (I think they would be event docs in this case, right?). 

With some custom data provider and data logic classes, I have the devices working for regular step-scan triggering, However, since the providers and logics don't seem to have a way to collect for fly scanning, I tried implementing `def collect(self)` or `def collect_pages(self)` on the `StandardDetector` class. In either case, it doesn't seem like these methods are getting called (`assert False` doesn't raise an exception).

I suspect this is because the [bluesky bundler](https://github.com/bluesky/bluesky/blob/4390e3dce0b04472157664b7de3111b18469f59c/src/bluesky/bundlers.py#L1148) only uses `collect()` if the device is not a `WritesStreamAssests`, which StandardDetector is.

I could do this without using StandardDetector, but I still want all the triggering, arming, preparing support in the StandardDetector. The only option I see right now  is to copy the entire StandardDetector definition and omit `def collect_asset_docs()`.

I'm wondering if there's a recommended way to make an ophyd-async detector that implements `EventCollectable` or `EventPageCollectable` but not `WritesStreamAssets`?

I wrote a test to check whether any event documents are created, and only `"start"`, `"descriptor"`, and `"stop"` documents are emitted: https://github.com/spc-group/haven/blob/3e23a1a52a75f1ca3c7da87d5810902aa277ad40/tests/haven/devices/test_ion_chamber_scaler.py#L40

Here is the current (incomplete) implementation of the device: https://github.com/spc-group/haven/blob/3e23a1a52a75f1ca3c7da87d5810902aa277ad40/src/haven/devices/detectors/ion_chamber_scaler.py#L120

## Acceptance Criteria

- A StandardDetector() device can be created that will produce event documents when given to `bps.collect()` without writing to disk.

coretl:
[...]

PR #1364, opened 2026-07-16
Only show a warning, rather than raise if a detector data logic doesn't support unbounded mode.

jwlodek:
Currently, mixing a streamamble and non-streamable data logic will not allow the detector to be used in streaming mode. Instead of raising, just show a warning and remove the dl from the prepare context, to still allow for streaming mode even if a single only dl is present.

Reproducible with an AD + PluginSignalDataLogic.


coretl:
This is going to be relevant for #1248 too, when I want to consider how to support things like the time series arrays from stats plugins.

Can I ask for your use case? Is it that you have an HDF writer and a Signal from the stats plugin, and in step scans you would like to see both, but you are fine with flyscans only having the HDF files?

jwlodek:
> Can I ask for your use case? Is it that you have an HDF writer and a Signal from the stats plugin, and in step scans you would like to see both, but you are fine with flyscans only having the HDF files?

Exactly. Re-creating the object that is identical minus the step scan only data logics seems unnecessary if we can just filter them out at run time. I've tested this PR with a Pilatus4 object setup to write HDF and read stats (step scan only) and it works as advertised.

coretl:
If I support time series arrays for the stats plugin, will you still need this?

jwlodek:
I think so. We use the signal data in the event stream to show live plots / live tables when step scanning. I don't think the time series array solution would allow that, since presumably it would be `dtype: array`?

coretl:
> I think so. We use the signal data in the event stream to show live plots / live tables when step scanning. I don't think the time series array solution would allow that, since presumably it would be `dtype: array`?

I plan to make it expand its pages into single events with `dtype: number` for the step scan case, so this should still be supported. However I guess we still need this for something that doesn't have a time series equivalent

coretl:
Do you also stamp your NDArrays with stats data and put it in the HDF file?

jwlodek:
> Do you also stamp your NDArrays with stats data and put it in the HDF file?

In some cases, yes, but a lot of our beamlines are step-scan only, and those just rely on the signal data in events.

jwlodek:
@coretl Any objections with merging this as an interim solution until #1367 gets finished up and merged?

coretl (approved):

Issue #1016, opened 2025-08-18
Roadmap for 1.0


coretl:
List of upcoming releases on the way to 1.0.

## 0.17

- https://github.com/bluesky/ophyd-async/issues/1255

## 0.18

- #1200 - Needs addition of `check_value` to `StandardMovable` (`Checkable` protocol)

## 0.19

- #1138 - Needs conversion of `epics_signal_x` to `epics_triggerable_command`

## 0.20

- #1182 - Needs implementation

## 0.21

- #1241 - Needs implementation
- Update to depend on scanspec 1.0.0

## 1.0

Remove all the deprecated code

Issue #888, opened 2025-05-08
Change StandardDetector DetectorWriterT to be optional


oliwenmandiamond:
When trying to implement an instance of the StandardDetector class, it forces you to use a file writer plugin along with it. I think this should be optional as this doesn't apply to all detector, for example electron analysers do not use the HDF plugin.

```
    def __init__(
        self,
        controller: DetectorControllerT,
        writer: DetectorWriterT,
        config_sigs: Sequence[SignalR] = (),
        name: str = "",
        connector: DeviceConnector | None = None,
    ) -> None:
```
The writer should default to None or default to a dummy file writer that doesn't do anything.

coretl:
Your detector still produces data though doesn't it? Just it is being read from PVs rather than HDF files. I wonder if we should extend `DetectorWriter` so that it supports not just file writing detectors, but also detectors that produce their data via PVs. We should revisit this once the initial electron analyser code is written and see if we can converge on a solution.

Issue #825, opened 2025-03-17
StandardDetector.stage behavior with writer and trigger info


thomashopkins32:
Hello, while implementing #726, I ran into an issue where you can't really set the `TriggerInfo` for a simple `count` plan easily.

For example, if I try to do
```python
trigger_info = TriggerInfo(number_of_events=1, exposures_per_event=5)
# Prepare the detector to use capture 5 exposures in a single event
RE(bps.prepare(det, trigger_info))
# EXPECTED: Gather data for 5 events. 5 events x 5 exposures per event = 25 total
# ACTUAL: 5 total since trigger_info is unset in stage()
RE(bp.count([det], num=5))
```

The only way around this is to create a custom plan that calls `stage()` before `bps.prepare` and make it into a single RE call.

This begs the question, should we really be unsetting `TriggerInfo` during `stage()`?
Should we be closing the file writer during `stage()`?

https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/_detector.py#L230-L231

## Acceptance Criteria
- My example above runs and produces 25 exposures as output

coretl:
Hmm, I'm not sure about this. We use `stage` at the moment to say "put into a known good state", which should close the file writer ready for a new file, but I'm not sure whether it should unset the `TriggerInfo`.

- Pro: There is prior art in that `TriggerInfo(livetime=None)` means "take the last configured exposure from the detector", which will also survive `stage`
- Con: If you `prepare` once to try things out, then someone else goes down to run the `count` plan, it might be surprising that it now does something different, especially as there's nothing corresponding to `exposures_per_event` in EPICS to inspect

@danielballan @tacaswell @jwlodek @mrakitin thoughts?

An alternative would be to add support for `prepare` to `count` in someway, either with an extra `prepare` argument, or a custom `per_shot` plan...

thomashopkins32:
> so we wouldn't want to read that back for use as `exposures_per_event` for the next step scan...

For step scans I think we have to be able to read it back from EPICS. If someone manually configures the `self.driver.num_images` they would expect that value to stick around. I see why this can't be the case for a fly scan.

For example,

```python
def custom_count(det, exposures_per_event = 5):
    # Set the exposures_per_event at the EPICS level
    yield from bps.mv(det.controller.driver.num_images, exposures_per_event)
    # EXPECTED: Collect `exposures_per_event` number of images for a single event
    # ACTUAL: Always collects 1 image for a single event
[...]

PR #1367, opened 2026-07-17
Bounded DetectorDataLogic tier, capability-based dispatch, and scan-period sizing

### LINKED ISSUE #1309: Discuss `chunk size` configuration
Based on how some devices are configured we'll need to configure their chunk size/data flush rate based on user configured parameters (exposure time, number of frames). This issue can serve as a place where we can discuss how to better implement this type of logic.
Examples are:

1. Xspress using chunk size to configure how many frames to batch before writing it to disk.
2. PandA flush rate

coretl:
Adds a third `DetectorDataLogic` tier and capability-based tier selection, with the ADR (0020) landing in the same PR as the code it describes.

## Why these issues are one design

The 0.22 roadmap (#1016) lists #1248, #888, #1309 and #1364 separately, but they bottom out in the same two gaps:

- a `DetectorDataLogic` is told nothing about timing or extent at prepare time
- `StandardDetector` inherits `WritesStreamAssets` unconditionally, so the bluesky bundler will never call `collect_pages` on it

The devices that cannot write files (MeasComp CTR-08, SIS3820 scaler — #1248/#888) and the ones that want rate-based chunking (Xspress, PandA — #1309) both need to be told what the scan looks like before it starts. One `prepare_bounded` tier closes both.

## What it does

- **Three data-logic tiers, selected by capability.** A `DetectorDataLogic` implements exactly one of:
  - `prepare_single(datakey_name) -> ReadableDataProvider` — one collection only (e.g. a scalar PV)
  - `prepare_bounded(datakey_name, num_collections, period) -> PageableDataProvider` — a finite buffer sized upfront, emitting event pages (`collect_pages`)
  - `prepare_unbounded(datakey_name, period) -> StreamableDataProvider` — any number of collections, emitting stream assets (`collect_asset_docs`)

  Core does not choose between tiers; it only asks whether the tier a logic implements can serve the requested `n = number_of_collections`. A logic that cannot is dropped with a warning rather than failing the scan, which is what lets an AD detector carry an HDF writer plus a `PluginSignalDataLogic` scalar and still fly (**#1364**).

- **The frame period is threaded to prepare (#1309).** `prepare_bounded`/`prepare_unbounded` receive the frame period (`livetime + deadtime`). `TriggerInfo.livetime == 0` ("use current") is resolved in `prepare()` from the trigger logic, so no data logic ever sees a period of 0. `ADHDFDataLogic` gains an opt-in `flush_period` that sizes the HDF chunk from the period (e.g. 400 Hz frames with a 0.5 s flush period → 200 frames/chunk); left unset it reads the chunk size back from the IOC as before, so existing detectors are unchanged. The period also joins the provider-reuse key, so an exposure change starts a new file.

- **`StandardDetector` no longer inherits `WritesStreamAssets`.** `collect_asset_docs` and `collect_pages` are exposed via `__getattr__`, keyed off the data logics present, so the bundler's structural `isinstance` sees exactly the one that applies. Mixing bounded and unbounded logics on one detector raises. No bluesky change needed.

- **A writer-less stats time series is now expressible.** `NDStatsIO` gains the `TS:*` time-series signals and a bounded `StatsTimeSeriesDataLogic` reads them as event pages, for detectors with no file writer. (A file-writing detector should pull stats into the file as NDAttributes instead.)

- **`add_detector_logics()` raises on ambiguity** — an object satisfying two logic protocols would otherwise register as the first match only and drop its other roles silently.

- **`StandardDetector` inherits `StandardFlyable`** (on top of #1400, which makes it a `StandardReadable`). The blocker was a state machine rather than typing: the detector allows many `kickoff()`/`complete()` cycles per `prepare()` — pinned by `test_kickoff_respects_prepare_bounds` and by the "file stays open between cycles" test — while `StandardFlyable` returns to `IDLE` on complete and so rejects the second kickoff. `FlyableLogic` gains an opt-in `supports_multiple_kickoffs`, defaulting `False` so `Motor`/PMAC/PandA keep erroring on a stray re-kickoff instead of silently repeating the move. It also gains a guarded `_prepared_fly_ctx` accessor, for the five detector verbs that read the prepare context from *outside* the fly pipeline (`describe_collect`, `collect_asset_docs`, `get_index`, `describe()`/`read()`, `trigger()`); without it those would read the `None` stand-in *typed as* `CtxT` and fail with an `AttributeError` on `None` — the failure mode ADR 0022 exists to remove. `_PrepareCtx` and `_KickoffCtx` collapse into one dataclass with the kickoff-produced fields `| None`, which is already the shape of `MotorFlyCtx` and `PmacFlyCtx`, so no third `CtxT` TypeVar is needed. The flag is a **deliberate exception** to capability-discovery-by-override-detection: override detection answers *"did you implement this behaviour?"* and needs a method with a body, whereas multi-kickoff is a property of whether the prepared state is reusable.

## The CTR-08 shaped the design

#1248 is **not** about areaDetector, and it is the case that puts pressure on the trigger/acquire/data split. A CTR-08 has one control that both starts acquisition and erases the buffer (`MCS:EraseStart`), and one PV that is both the exposure count and the buffer size (`MCS:NuseAll`). The split still holds — it is a split of *concerns*, not of hardware — but it forces the progress baseline in `trigger()` to be **stated (zero) rather than derived**: a buffer erased by the acquire control is cleared *after* the data logic is prepared, so a post-prepare readback would be stale and the detector would wait for twice the buffer length and hang.

## Review notes

- **areaDetector TS PV suffixes are template-derived** (from ADCore's `NDStats`/`NDPluginTimeSeries` templates) and are exercised only by mock unit tests here — they want verifying against a live IOC. Flagged in the code.
- **`test_step_scan_keep_numimages` changed**: per the ADR's reuse rule, changing the exposure now opens a fresh `StreamResource` rather than reusing the file.
- **#1248/#888 are `Refs`, not `Fixes`.** The framework limitation is removed and the AD stats case is delivered, but the concrete non-AD device (CTR-08 / scaler) is a follow-up, so I have not auto-closed them — happy to add `Fixes` lines if you'd rather.
- **#825 is already fixed** by ADR 0013 and needs no part of this design; verified against its acceptance criteria.

Fixes #1309
Fixes #1364

Refs #1016, #1248, #888

## AI Disclaimer

### DIFF
--- a/docs/explanations/decisions/0020-bounded-data-logic-tier.md
+++ b/docs/explanations/decisions/0020-bounded-data-logic-tier.md
@@ -1,0 +1,283 @@
+# 20. Add a bounded `DetectorDataLogic` tier and select tiers by capability
+
+Date: 2026-07-17
+
+## Status
+
+Accepted
+
+## Context
+
+ADR 0012 split data production out into `DetectorDataLogic`, with two tiers that a subclass
+opts into by overriding the corresponding method:
+
+- `prepare_single(datakey_name) -> ReadableDataProvider` — a single event only, feeding
+  `read()` and `describe()`
+- `prepare_unbounded(datakey_name) -> StreamableDataProvider` — any number of collections,
+  feeding `collect_asset_docs()` and `describe_collect()`
+
+Four open issues turn out to bottom out in the same two gaps.
+
+### The data logic is not told enough at prepare time
+
+Issue #1309 wants chunk size and flush rate configured from the scan parameters. An Xspress
+running at 400 Hz with a 2 Hz flush rate wants 200 frames per chunk; the PandA wants a flush
+period. Neither is expressible today, because a data logic is prepared for an *unbounded* number
+of collections and so is told nothing about timing. Today's implementations paper over this:
+`ADHDFDataLogic` reads `num_frames_chunks` back from the IOC and forces it to 1 if unset, the
+PandA hardcodes `chunk_shape=(1024,)` behind a TODO, and Odin hardcodes `(1, *data_shape)`.
+
+The parameter that is actually needed is the frame **period** (`livetime + deadtime`), not the
+number of frames — the requirement is a flush *rate*. This matters, because the period is known
+for step scans as well as fly scans, whereas the total frame count is not.
+
+Two workarounds were considered and rejected. Subclassing `TriggerInfo` and `prepare()` to carry a
+chunk size (as Malcolm did) puts a data-writing concern into the trigger path and defeats ADR
+0012's rule that logic classes are handed only the fields they need. Computing the chunk size in
+the `DetectorTriggerLogic` puts a data concern in the trigger logic.
+
+### `StandardDetector` unconditionally claims to write stream assets
+
+Issues #1248 and #888 both come from detectors that produce data but cannot write files: a Struck
+SIS3820 scaler in a VME crate, a MeasComp CTR-08, a TetrAmm electrometer, an electron analyser.
+These want the trigger, arm, prepare and stage machinery of `StandardDetector`, but need to emit
+event pages rather than stream assets.
+
+These are worth describing concretely, because they are not areaDetectors and they put more
+pressure on the trigger/acquire/data split than an areaDetector does. A MeasComp CTR-08 exposes:
+
+| PV | role |
+|---|---|
+| `MCS:EraseStart` | clears the arrays *and* starts counting — one control, no separate erase |
+| `MCS:StopAll` | stop |
+| `MCS:Acquiring` | idle status |
+| `MCS:NuseAll` | number of channels to use — the buffer size, and the point at which it stops |
+| `MCS:CurrentChannel` | how many points have been collected |
+| `mca<i>.VAL` | the growing waveform arrays that hold the data |
+
+The difference from an areaDetector is that there is no driver/plugin split in the hardware. An
+areaDetector acquires with `driver.acquire` and captures with a *separate* `TSAcquire` or HDF
+`capture`, so the acquire and data concerns land on different PVs and the split falls out of the
+hardware. Here one control both gates acquisition and clears the data buffer, and one PV
+(`NuseAll`) is both the exposure count and the buffer size.
+
+They cannot, because `StandardDetector` inherits `WritesStreamAssets` unconditionally, and the
+bluesky bundler only calls `collect_pages()` on a device that is *not* a `WritesStreamAssets`.
+The two protocols are mutually exclusive by design, to avoid the case where both produce blocks
+of different sizes.
+
+#888 asks for the writer to be optional. The rewrite already made it optional in the literal
+sense — `StandardDetector()` takes no writer at all. But the underlying request ("no HDF writer,
+because I have PVs instead") is #1248, so #888 is a duplicate.
+
+### These are the same gap
+
+Both groups of devices must be told *how many* collections to expect at prepare time, because
+they own a finite buffer that has to be sized before acquisition: the scaler's MCA array length
+*is* the frame count, and the areaDetector stats plugin's `TSNumPoints` must be set before the
+time series starts. A device with a finite buffer also naturally produces its data as pages at
+the end of an event rather than as stream datums as it goes.
+
+So one new tier serves both, and #1309's period rides along with it.
+
+### The areaDetector stats duality
+
+An `NDPluginStats` exposes each statistic twice: as a scalar (`Total_RBV`) and as a time series
+array (`TS:TSTotal`). Only the scalar is modelled today, by `PluginSignalDataLogic`. The scalar
+can only ever serve a single collection; the time series can serve any number. Issue #1364 arises
+directly from this: an AD detector with an HDF writer *and* a stats scalar cannot fly, because
+the scalar is single-only.
+
+## Decision
+
+### A third tier: `prepare_bounded`
+
+```python
+class DetectorDataLogic:
+    async def prepare_single(
+        self, datakey_name: str
+    ) -> ReadableDataProvider: ...
+    async def prepare_bounded(
+        self, datakey_name: str, num_collections: int, period: float
+    ) -> PageableDataProvider: ...
+    async def prepare_unbounded(
+        self, datakey_name: str, period: float
+    ) -> StreamableDataProvider: ...
+```
+
+`prepare_bounded` and `prepare_unbounded` are told the frame period; `prepare_single` is not,
+since a single collection has no rate to size against.
+
+`num_collections` is always the **total** for the scan (`TriggerInfo.number_of_collections`),
+never a per-kickoff figure. A finite buffer is fed by data callbacks and does not care how many
+kickoffs span it, so per-kickoff sizing would mean resizing the buffer mid-scan.
+
+### `PageableDataProvider`, with readings derived from pages
+
+`prepare_bounded` returns a single provider type that emits pages. Readings are derived from
+pages rather than being a separate code path: a step-scan prepare has `number_of_events == 1`, so
+its page contains exactly one event, which extracts to a single reading. `collections_per_event`
+appears in the datakey shape exactly as it does for `StreamableDataProvider`.
+
+### The period is resolved in `prepare()`
+
+`TriggerInfo.livetime` of 0 means "whatever is currently set on the detector", so the period
+cannot simply be forwarded from the `TriggerInfo` — an explicit
+`prepare(TriggerInfo(number_of_events=10))` would hand the data logic a period of 0.
+
+Instead `prepare()` resolves it. `DetectorTriggerLogic.default_trigger_info()` is extended to
+return the current `livetime` and `deadtime` as well as the frame count, and `prepare()` uses it
+to fill in a zero `livetime` after the trigger logic has been prepared. ADR 0012 already requires
+that ordering, so the hardware holds the real values by that point. Data logics therefore never
+see a period of 0, and the awkward case is handled once in core rather than by every implementer.
+
+### Tier selection is by capability, with no precedence rule
+
+Each data logic implements exactly one tier, so core does not choose between tiers — it only
+asks whether the tier a logic implements can serve the requested number of collections
+`n = TriggerInfo.number_of_collections`:
+
+| tier | serves |
+|---|---|
+| `prepare_unbounded` | always |
+| `prepare_bounded` | `n != 0` (finite) |
+| `prepare_single` | `n == 1` |
+
+A logic whose tier cannot serve `n` is dropped from the prepare context with a warning, rather
+than raising. This generalises the rule agreed in #1364 to all three tiers: it is what lets an
+AD detector carry an HDF writer plus a `PluginSignalDataLogic` on some unrelated PV (a
+temperature, a ring current) and still fly, with only the scalar dropped.
+
+### Bounded providers are never reused
+
+`_update_prepare_context` currently reuses providers when `collections_per_event` is unchanged,
+to avoid reopening files on every step-scan point. Two changes follow.
+
+The reuse check gains the period, since the period now determines chunk shape, and chunk shape is
+baked into the `StreamResource` at provider construction. A period change therefore starts a new
+file. This is inherent rather than incidental: areaDetector's `num_frames_chunks` can only be set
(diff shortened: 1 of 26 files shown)


src/ophyd_async/epics/adcore/_trigger_logic.py
```diff
@@ -31,9 +31,24 @@ async def prepare_exposures(
 
 
 async def trigger_info_from_num_images(driver: ADBaseIO) -> TriggerInfo:
```
jwlodek: Perhaps a different name makes sense here? We're not just reading `num_images` anymore

src/ophyd_async/core/_detector.py
```diff
@@ -333,18 +353,48 @@ async def stop(self) -> None:
...
+    Tiers are discovered by method-override detection, the same mechanism used
+    for trigger logic. A data logic must override exactly one.
+    """
+    if _data_logic_supported(dl.prepare_unbounded):
+        return "unbounded"
+    if _data_logic_supported(dl.prepare_bounded):
+        return "bounded"
+    if _data_logic_supported(dl.prepare_single):
+        return "single"
+    raise RuntimeError(f"DataLogic hasn't overridden any prepare_* methods {dl}")
```
jwlodek: Maybe worth having an enum for data logic tier, rather than just str? Allows for tighter typing.

src/ophyd_async/core/_detector.py
```diff
@@ -417,6 +493,39 @@ def add_detector_logics(
...
+        )
+        has_unbounded = any(
+            _data_logic_supported(dl.prepare_unbounded) for dl in self._data_logics
+        )
+        if has_bounded and has_unbounded:
+            raise TypeError(
+                f"Detector {self.name} has both bounded and unbounded data logics; "
+                "these cannot be combined on one detector"
+            )
+
```
jwlodek: It may be helpful to list the logics and their types. in this message.

src/ophyd_async/core/_detector.py
```diff
@@ -417,6 +493,39 @@ def add_detector_logics(
...
+        # never recurses.
+        if name == "collect_asset_docs" and any(
+            _data_logic_supported(dl.prepare_unbounded) for dl in self._data_logics
+        ):
+            return self._collect_asset_docs
+        if name == "collect_pages" and any(
+            _data_logic_supported(dl.prepare_bounded) for dl in self._data_logics
+        ):
+            return self._collect_pages
+        raise AttributeError(name)
```
jwlodek: shouldn't this fall back to `super().__getattr__(name)`?

jwlodek (changes requested):
Read through the ADR, and overall it sounds good to me. Some suggestions/comments listed below.

src/ophyd_async/epics/adcore/_trigger_logic.py
coretl: Agreed — renamed to `trigger_info_from_driver`, since it now reads `num_images`, `acquire_time` and `acquire_period` back from the driver. Updated all AD subclasses, the adcore export, and the how-to. (Left ADR 0013's point-in-time reference as historical.) dac43478

src/ophyd_async/core/_detector.py
coretl: Done — added a `_DataLogicTier` enum (UNBOUNDED/BOUNDED/SINGLE); `_data_logic_tier` now returns it and `_tier_can_serve` plus all the `tier is _DataLogicTier.X` call sites are typed against it. dac43478

src/ophyd_async/core/_detector.py
coretl: Good call — the mix TypeError now lists each side by class name, e.g. `has both bounded data logics (BoundedOnlyDataLogic) and unbounded data logics (StreamableOnlyDataLogic)`. dac43478

src/ophyd_async/core/_detector.py
coretl: There's nothing to delegate to here: no base class in `StandardDetector`'s MRO defines `__getattr__` (`Device`/`object` don't), so `super().__getattr__(name)` would itself raise `AttributeError: 'super' object has no attribute '__getattr__'`. Instead I now raise the same message Python's default lookup would — `'StandardDetector' object has no attribute '<name>'` — so `hasattr()`/`getattr()` and tracebacks behave normally. Added `test_missing_attribute_raises_standard_attribute_error` to lock this in. dac43478

src/ophyd_async/core/_data_providers.py
```diff
@@ -58,6 +59,59 @@ async def make_stream_docs(
...
+        extracts to a single reading per key.
+        """
+        collections_written = await self.collections_written_signal.get_value()
+        readings: dict[str, Reading] = {}
+        async for page in self.make_pages(collections_written, collections_per_event):
+            times = page["time"]
+            for key, values in page["data"].items():
+                for value, timestamp in zip(values, times, strict=True):
+                    readings[key] = Reading(value=value, timestamp=timestamp)
+        return readings
```
coretl, line 112: I prefer this to be in `StandardDetector` until we have a case for overriding it in a provider.

src/ophyd_async/core/_data_providers.py
```diff
@@ -58,6 +59,59 @@ async def make_stream_docs(
...
+        A step-scan prepare has a single event, so the page holds one event
+        whose per-key value is the `collections_per_event`-length array; this
+        extracts to a single reading per key.
+        """
+        collections_written = await self.collections_written_signal.get_value()
+        readings: dict[str, Reading] = {}
+        async for page in self.make_pages(collections_written, collections_per_event):
+            times = page["time"]
+            for key, values in page["data"].items():
+                for value, timestamp in zip(values, times, strict=True):
```
coretl, line 110: If a page has more than one value in it then this will silently pick the last value. It should error instead

src/ophyd_async/epics/adcore/_stats_time_series.py
coretl, line 1: We should always capture the `TSTimestamp` field, and use that as the timestamp source

src/ophyd_async/epics/adcore/_stats_time_series.py
```diff
@@ -0,0 +1,143 @@
...
+        events = collections_written // collections_per_event
+        if events <= self.last_emitted:
+            return
+        new = range(self.last_emitted, events)
+        # A single timestamp per event: the plugin does not expose a per-point
+        # time, so use the read time.
+        now = time.time()
+        values = {
+            datakey: await signal.get_value() for datakey, signal in self.arrays.items()
+        }
```
coretl, line 66: Should use gather_dict

src/ophyd_async/epics/adcore/_io.py
```diff
@@ -179,6 +207,18 @@ class NDStatsIO(NDPluginBaseIO):
     compute_statistics: A[SignalRW[bool], PvSuffix.rbv("ComputeStatistics")]
     bgd_width: A[SignalRW[int], PvSuffix.rbv("BgdWidth")]
     total: A[SignalR[float], PvSuffix("Total_RBV")]
+    # Time series. NDStats.template includes NDPluginTimeSeries.template with an
+    # inner "TS:" record prefix, so each statistic is also exposed as a fixed-
+    # length array. ts_num_points sizes the buffer, ts_control erases and starts
+    # it, ts_current_point reports progress and ts_total holds the Total series.
+    # These suffixes are derived from the ADCore templates and want verifying
+    # against a live IOC.
+    ts_control: A[SignalRW[NDStatsTSControl], PvSuffix("TS:TSControl")]
```
coretl, line 216: This PV doesn't exist, it should be `ts_acquire`

src/ophyd_async/epics/adcore/_stats_time_series.py
```diff
@@ -0,0 +1,143 @@
...
+            self.stats.enable_callbacks.set(EnableDisable.ENABLE),
+            self.stats.compute_statistics.set(True),
+            self.stats.ts_num_points.set(num_collections),
+            self.stats.ts_acquire_mode.set(NDStatsTSAcquireMode.FIXED_LENGTH),
+        )
+        # Erase and start clears the arrays and resets ts_current_point to 0, so
+        # the buffer is armed and empty before the detector's frames arrive. This
+        # is the data logic performing the erase itself, which is why trigger()'s
+        # zero baseline is correct (see ADR 0020).
+        await self.stats.ts_control.set(NDStatsTSControl.ERASE_START)
```
coretl, line 134:
```suggestion
        await self.stats.ts_acquire.set(1)
```

canismarko:
I applied this design to the [counter devices](https://github.com/spc-group/haven/blob/main/src/haven/devices/detectors/counter.py) I was working on (USB CTR-08 and Struck SIS3820). I tested it on real hardware for step scanning and it worked great! I did not test on hardware for fly scanning yet, but [this test](https://github.com/spc-group/haven/blob/dafc3090a8e9e98e914640bc7097794a9967326c/tests/haven/devices/test_counter.py#L245) passes now which was the basis for #1248.

I had to add `wait=False` to the signals for `start_all`, `stop_all`, `start_erase`, and `erase` or else the triggering would timeout. Seems like it's an oddity of the EPICS support more than anything to do with ohpyd-async?

Regarding…
> I prefer this to be in StandardDetector until we have a case for overriding it in a provider.

…I didn't run into a case for overriding this. Everything fit within `make_datakeys()` and `make_pages()`.

canismarko:
I tested this with fly scanning using internal triggering for both a CTR08 and SIS3820 counter at the same time, and it worked as expected.

docs/explanations/decisions/0020-bounded-data-logic-tier.md
```diff
@@ -0,0 +1,283 @@
...
+
+Each data logic implements exactly one tier, so core does not choose between tiers — it only
+asks whether the tier a logic implements can serve the requested number of collections
+`n = TriggerInfo.number_of_collections`:
+
+| tier | serves |
+|---|---|
+| `prepare_unbounded` | always |
+| `prepare_bounded` | `n != 0` (finite) |
+| `prepare_single` | `n == 1` |
```
davidpcls, line 144: Why both prepare_bounded and prepare_single? Prepare_single is just prepare_bounded when n==1.

davidpcls: Is there any difference? 

At a more general level:

Is a flyscan always unbounded?
Is a stepscan always bounded?

davidpcls:
@coretl asked me to take a look here after I made this issue:https://github.com/bluesky/ophyd-async/issues/1412#issuecomment-5572697469 (adding as context)

I am trying to learn this code base so i can provide step scanning for pandabox.  What I need is to be able to detect when a scan is started as a step scan or a flyscan, apply changes to the detector, then either kickoff or trigger and read. I don't think this is any different than the standard generic case for a detector. When a scan is called the detector class should be informed what type of scan and allow it to become prepared for it (or if it's not supported, raise an error).

docs/explanations/decisions/0020-bounded-data-logic-tier.md
coretl: The difference:
- `prepare_single` is now handled by `StandardReadable` baseclass, so will disappear. It is a set of signals to `read()` at each point of a step scan. It does not contribute to `collect()` for flyscans
- `prepare_bounded` is handled in this PR (or will be when I've done the code split). It is for detectors that put their data in a growing array with a maximum bound (like a binned scaler card or a NDStats plugin timeseries array). For step scans the array will be populated at each point, then `read()` returns it all. For fly scans, the new items from the array will be returned from each `collect()`
- `prepare_unbounded` is the existing behaviour of publishing `StreamResource` document that point to data files in `collect_asset_docs`. It works for step and fly scans

Having written this I'm even more convinced the names are wrong, so I'll have a think about some better names for these methods...

coretl:
> I am trying to learn this code base so i can provide step scanning for pandabox. What I need is to be able to detect when a scan is started as a step scan or a flyscan, apply changes to the detector, then either kickoff or trigger and read. I don't think this is any different than the standard generic case for a detector. When a scan is called the detector class should be informed what type of scan and allow it to become prepared for it (or if it's not supported, raise an error).

I *think* this works as is without any of this PR, but I'm not sure. I'll comment in matrix as a back and forth is easier there.

PR #1369, opened 2026-07-17
Add opt-outs for the reserved-attr check (#1028 follow-up)

### LINKED ISSUE #1368: Provide a test-facing opt-out for the reserved-attr check (follow-up to #1028)
### Follow-up to #1028

#1028 made `Device.__setattr__` raise a `NameError` whenever an attribute name collides with a bluesky protocol method (`set`, `read`, `trigger`, `stage`, `complete`, ... — the full set is `DEVICE_RESERVED_ATTRS` in `core/_device.py`). This is a good guard against accidentally shadowing a protocol method with a signal.

The side effect is that it breaks a common, legitimate testing pattern used in downstream repos, where a test mocks out a device method directly:

```python
mydevice.set = AsyncMock()   # after #1028, raises NameError
```

There is currently only an undiscoverable escape hatch — dropping down to `object.__setattr__`:

```python
object.__setattr__(mydevice, "set", AsyncMock())
```

We use exactly this in our own tests (`tests/unit_tests/epics/test_motor.py`), but a downstream user who hits the `NameError` has no signposted way to opt out.

### Plan

We will do two things:

1. **An environment variable to turn the check off.** This gives downstream repos a quick, zero-code-change way to get their existing tests passing again while they migrate — flip the env var and the reserved-name check is disabled.
2. **An `ophyd_async.testing` helper that does the `object.__setattr__`.** A discoverable, first-class way to override a reserved-name method on a `Device` in a test, e.g. `set_mock_attr(device, "set", AsyncMock())`. This matches the existing `ophyd_async.testing` surface (`set_mock_value`, `assert_has_calls`) and is the recommended per-test approach. Our own two `object.__setattr__` call sites in `test_motor.py` will switch to it.

The accidental-shadowing guard from #1028 stays on by default in production code; both of these are explicit, test-facing opt-outs.

## AI Disclaimer

coretl:
Fixes #1368

Follow-up to #1028, which made `Device.__setattr__` raise `NameError` when an attribute name collides with a bluesky protocol method (`set`, `read`, `trigger`, ...). That guards against accidentally shadowing a verb with a Signal, but it breaks the common test pattern of mocking a verb — `device.set = AsyncMock()` — and the only escape was the undiscoverable `object.__setattr__`.

This adds the two opt-outs agreed in the issue:

1. **`ophyd_async.core.set_mock_attr(device, attr, mock) -> mock`** — overrides a reserved-name attribute in a test and returns the mock it just set, so the override and the assertion are one expression:
   ```python
   mock_set = set_mock_attr(motor, "set", MagicMock())
   await motor.kickoff()
   mock_set.assert_called_once_with(-3.0)
   ```
   It lives next to the other mock helpers (`set_mock_value`, `callback_on_mock_put`, ...) in `ophyd_async.core`.
2. **`OPHYD_ASYNC_ALLOW_RESERVED_ATTRS=YES`** — disables the check globally, so a downstream suite that mocks verbs this way can be turned back on without editing every call site.

The `NameError` message now points at both. The two `object.__setattr__` sites in `tests/unit_tests/epics/test_motor.py` switch to the helper.

Docs: a "Mocking out a verb" section in the writing-tests tutorial, and the env var in the environment-variables reference.

## AI Disclaimer


oliwenmandiamond (commented):
This looks good and have no strong opinions on which way we go, but I personally think we should stick with one approach rather than have several.

DominicOram:
> I personally think we should stick with one approach rather than have several.

I don't particularly mind having multiple ways of doing it. I think the `ophyd-async` docs should give an opinionated order of preference though, which I would say is:
1. Implement sufficient logic in the device mock hook so that it "just works" (won't let you easily catch logic of what a plan does if a `set` fails)
2. Use the `set_mock_attr` above
3. Use the env variable

I'm not adverse to removing the env variable, as @coretl suggested in https://diamondlightsource.slack.com/archives/C04RRAM68LD/p1784709965349119?thread_ts=1784284243.842139&cid=C04RRAM68LD but it would require downstream changes in `xx-bluesky`. So it may be valuable as a temporary migration path. We don't use it much in SCM + X so I would be happy not to have the env variable but @rtuck99 may think differently

coretl:
> I don't particularly mind having multiple ways of doing it. I think the `ophyd-async` docs should give an opinionated order of preference though, which I would say is:
> 
>     1. Implement sufficient logic in the device mock hook so that it "just works" (won't let you easily catch logic of what a plan does if a `set` fails)
> 
>     2. Use the `set_mock_attr` above
> 
>     3. Use the env variable

I like this. I've put in some docs to the effect, keeping the environment variable for migration. I'll delete the env var before I make 1.0

oliwenmandiamond (approved):
> I like this. I've put in some docs to the effect, keeping the environment variable for migration. I'll delete the env var before I make 1.0

Deleting the env var before 1.0 I think is good compromise. Looks good, thanks!

PR #1370, opened 2026-07-20
Allow PvaCommandBackend to have '?' specifiers

olliesilvester:
Makes PvaCommandBackend compatible with scalar booleans.

Before this change, connecting an ophyd-async device to a FastCS IOC containing boolean command type attributes would give the error:
`TypeError: pva://{pv_address} is not a scalar numeric PV (typeid='epics:nt/NTScalar:1.0', specifier='?'); PvaCommandBackend requires a scalar numeric PV`


src/ophyd_async/epics/core/_p4p.py
```diff
@@ -483,14 +483,17 @@ async def connect(self, timeout: float) -> None:
         typeid = value.getID()
         specifier = _get_specifier(value)
 
+        is_bool = typeid == "epics:nt/NTScalar:1.0" and specifier == "?"
         is_int = typeid == "epics:nt/NTScalar:1.0" and specifier in _int_specifiers
```
shihab-dls: nit: I would do this so we don't specify the typeid more than once

```suggestion
        is_scalar = typeid == "epics:nt/NTScalar:1.0"
        is_bool = is_scalar and specifier == "?"
        is_int = is_scalar and specifier in _int_specifiers
```

tests/unit_tests/epics/signal/test_pva_command_backend.py
```diff
@@ -0,0 +1,17 @@
...
+
+async def test_pva_command_backend_accepts_bool_pv(
+    monkeypatch: pytest.MonkeyPatch,
+):
+    async def pvget_with_timeout(pv: str, timeout: float):
+        return P4PValue(NTScalar.buildType("?"), {"value": True})
+
+    monkeypatch.setattr(_p4p, "pvget_with_timeout", pvget_with_timeout)
+
+    await PvaCommandBackend("BOOL").connect(timeout=1.0)
```
shihab-dls: nit: This tests the same code paths as the system tests, so I wouldn't include it, but it's correct regardless.

shihab-dls (approved):
Thank you! It's like you never left... Approving with some comments, so its up to your judgement on those.

PR #1371, opened 2026-07-20
refactor so that signal backend uses get location instead of setpoint

### LINKED ISSUE #815: Refactor `SignalBackend.get_setpoint` to be `SignalBackend.get_location`
At the moment we call `SignalBackend.get_setpoint` from `SignalRW`:
https://github.com/bluesky/ophyd-async/blob/b8f73bbb5d4e1e3b21066ecdee62fd444fe1e0b1/src/ophyd_async/core/_signal.py#L289-L292

This adds caching for the readback, but causes a double fetch if the setpoint and readback are coming from the same PV:
https://github.com/bluesky/ophyd-async/blob/b8f73bbb5d4e1e3b21066ecdee62fd444fe1e0b1/src/ophyd_async/epics/core/_aioca.py#L336-L342

This also causes a double transform in derived signal.

This ticket is to change the baseclass method from `SignalBackend.get_setpoint` to `SignalBackend.get_location`, then change each subclass to return the `Location` directly.

e.g. for `CaSignalBackend` something like:
```python
    async def get_location(self) -> Location[SignalDatatypeT]:
        if self.write_pv != self.read_pv:
            raw_setpoint, raw_readback = await asyncio.gather(
                self._caget(self.write_pv, FORMAT_RAW),
                self._caget(self.read_pv, FORMAT_RAW)
            )
            setpoint = self.converter.value(raw_setpoint)
            readback = self.converter.value(raw_readback)
        else:
            raw_value = await self._caget(self.write_pv, FORMAT_RAW)
            setpoint = readback = self.converter.value(raw_value)
        return Location(setpoint=setpoint, readback=readback) 
```

## Acceptance Criteria
- `SignalBackend` and all subclasses are updated as above, tests pass, and docs updated

tomkane-dls:
Replace SignalBackend.get_setpoint with get_location

Closes issue #815 

SignalRW.locate() previously called backend.get_setpoint() and get_value() separately. Where the setpoint and readback come from the same PV this meant two fetches for one value, and in the derived signal backend it meant running the transform twice.

The backend now returns a Location directly, so each implementation can fetch both halves in whatever way suits its protocol.

Updated:

- SignalBackend.get_setpoint → get_location
- CaSignalBackend / PvaSignalBackend — one fetch when read_pv == write_pv, asyncio.gather otherwise
- DerivedSignalBackend — returns the whole Location from get_locations()
- SoftSignalBackend, MockSignalBackend, TangoSignalBackend 

Notes:

So i see locate() used to fetch the readback via _backend_or_cache(), so a monitored signal served it from the cache. A single backend.get_location() can't see that cache, so the readback now always comes from the backend. Is this what we are intending or do I need a rethink to carry on looking for cached values?

Breaking for anything implementing SignalBackend outside the repo, though loudly

Please checkj that TangoSignalBackend change looks okay as I have no experience with tango and it is entirely AI

### DIFF
--- a/docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
+++ b/docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
@@ -22,7 +22,7 @@ To address this, `SoftSignalBackend` was extended to support arbitrary callables
 - The internal `self._reading` store remains the **single source of truth**. The `getter` updates this store rather than bypassing it, preserving coherence for subscriptions and cached reads.
 - The `put` method accepts `SignalDatatypeT` (the same type as the signal's stored value) to maintain type safety and consistency.
 - Polling tasks are used for subscriptions, starting in `set_callback` and canceling when subscriptions end.
-- The setpoint half of `get_location()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of. The readback half does invoke the `getter`, as it must.
+- `get_setpoint()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of.
 
 ### **Factory Function Updates**
 The convenience functions `soft_signal_rw` and `soft_signal_r_and_setter` were updated to accept `getter`, and `poll_period` arguments. `soft_signal_rw` additionally accepts a `setter` argument. These additional arguments are passed to `SoftSignalBackend`.
--- a/docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
+++ b/docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
@@ -22,7 +22,7 @@ To address this, `SoftSignalBackend` was extended to support arbitrary callables
 - The internal `self._reading` store remains the **single source of truth**. The `getter` updates this store rather than bypassing it, preserving coherence for subscriptions and cached reads.
 - The `put` method accepts `SignalDatatypeT` (the same type as the signal's stored value) to maintain type safety and consistency.
 - Polling tasks are used for subscriptions, starting in `set_callback` and canceling when subscriptions end.
-- `get_setpoint()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of.
+- The setpoint half of `get_location()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of. The readback half does invoke the `getter`, as it must.
 
 ### **Factory Function Updates**
 The convenience functions `soft_signal_rw` and `soft_signal_r_and_setter` were updated to accept `getter`, and `poll_period` arguments. `soft_signal_rw` additionally accepts a `setter` argument. These additional arguments are passed to `SoftSignalBackend`.
--- a/src/ophyd_async/core/_soft_signal_backend.py
+++ b/src/ophyd_async/core/_soft_signal_backend.py
@@ -10,7 +10,7 @@ from functools import lru_cache
 from typing import Any, Generic, get_args
 
 import numpy as np
-from bluesky.protocols import Reading
+from bluesky.protocols import Location, Reading
 from bluesky.utils import maybe_await
 from event_model import DataKey
 
@@ -234,8 +234,9 @@ class SoftSignalBackend(SignalBackend[SignalDatatypeT]):
         await self._update_value_from_getter()
         return self.reading["value"]
 
-    async def get_setpoint(self) -> SignalDatatypeT:
-        return self._setpoint
+    async def get_location(self) -> Location[SignalDatatypeT]:
+        await self._update_value_from_getter()
+        return Location(setpoint=self._setpoint, readback=self.reading["value"])
 
     def set_callback(self, callback: Callback[Reading[SignalDatatypeT]] | None) -> None:
         if callback and self.callback:
--- a/src/ophyd_async/epics/core/_aioca.py
+++ b/src/ophyd_async/epics/core/_aioca.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import os
 import sys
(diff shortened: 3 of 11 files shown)


docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
```diff
@@ -22,7 +22,7 @@ To address this, `SoftSignalBackend` was extended to support arbitrary callables
 - The internal `self._reading` store remains the **single source of truth**. The `getter` updates this store rather than bypassing it, preserving coherence for subscriptions and cached reads.
 - The `put` method accepts `SignalDatatypeT` (the same type as the signal's stored value) to maintain type safety and consistency.
 - Polling tasks are used for subscriptions, starting in `set_callback` and canceling when subscriptions end.
-- `get_setpoint()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of.
+- The setpoint half of `get_location()` **does not invoke the `getter`**; it returns the last value written to the `setter` or the initial value of. The readback half does invoke the `getter`, as it must.
```
coretl, line 25: We don't rewrite historical ADRs

src/ophyd_async/epics/core/_aioca.py
```diff
@@ -370,9 +371,21 @@ async def get_value(self) -> SignalDatatypeT:
...
+            setpoint = self.converter.value(raw_value)
+            readback = self.converter.value(raw_value)
+        else:
+            # otherwise retrieve PVs separately
+            raw_setpoint, raw_readback = await asyncio.gather(
+                self._caget(self.write_pv, FORMAT_RAW),
+                self._caget(self.read_pv, FORMAT_RAW),
+            )
+            setpoint = self.converter.value(raw_setpoint)
+            readback = self.converter.value(raw_readback)
```
coretl, line 387: Nit: can be shorter

```suggestion
        if self.write_pv == self.read_pv:
            # one fetch when read and write PVs are identical
            setpoint = readback = await self.get_value()
        else:
            # otherwise retrieve PVs separately
            raw_setpoint, readback = await asyncio.gather(
                self._caget(self.write_pv, FORMAT_RAW),
               self.get_value(),
            )
            setpoint = self.converter.value(raw_setpoint)
```

src/ophyd_async/epics/core/_p4p.py
```diff
@@ -433,10 +433,22 @@ async def get_value(self) -> SignalDatatypeT:
         value = await context().get(self.read_pv, request=request)
         return self.converter.value(value)
 
-    async def get_setpoint(self) -> SignalDatatypeT:
+    async def get_location(self) -> Location[SignalDatatypeT]:
         request = _pva_request_string(self.converter.value_fields)
-        value = await context().get(self.write_pv, request=request)
-        return self.converter.value(value)
+        if self.read_pv == self.write_pv:
```
coretl, line 438: ditto

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -962,14 +962,20 @@ async def get_value(self) -> SignalDatatypeT:
...
+        if read_proxy is None:
+            raise NotConnectedError(f"Not connected to {self.read_trl}")
+        write_proxy = self.proxies[self.write_trl]
+        if write_proxy is None:
             raise NotConnectedError(f"Not connected to {self.write_trl}")
-        w_value = await proxy.get_w_value()
-        return cast(SignalDatatypeT, w_value)
+        value, w_value = await asyncio.gather(
+            read_proxy.get(), write_proxy.get_w_value()
+        )
```
coretl, line 974: ditto, could use get_value here

src/ophyd_async/epics/core/_aioca.py
tomkane-dls: This was a design choice to assign the value separately, in case the fetched value is a mutable object so they don't point to the same data. But if this is never a concern i will go with the shorter version, thanks!

docs/explanations/decisions/0019-soft-signals-backed-by-arbitrary-callables.md
tomkane-dls: this is now reverted

src/ophyd_async/epics/core/_p4p.py
tomkane-dls: shortened similarly to _aioca.py

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -234,8 +234,9 @@ async def get_value(self) -> SignalDatatypeT:
         await self._update_value_from_getter()
         return self.reading["value"]
 
-    async def get_setpoint(self) -> SignalDatatypeT:
-        return self._setpoint
+    async def get_location(self) -> Location[SignalDatatypeT]:
+        await self._update_value_from_getter()
+        return Location(setpoint=self._setpoint, readback=self.reading["value"])
```
coretl, line 239: Nit

```suggestion
        return Location(setpoint=self._setpoint, readback=await self.get_value())
```

src/ophyd_async/tango/core/_tango_transport.py
```diff
@@ -962,14 +962,18 @@ async def get_value(self) -> SignalDatatypeT:
...
             raise NotConnectedError(f"Not connected to {self.write_trl}")
-        w_value = await proxy.get_w_value()
-        return cast(SignalDatatypeT, w_value)
+        value, w_value = await asyncio.gather(
+            self.get_value(),
+            write_proxy.get_w_value(),
+        )
+        return Location(
+            setpoint=cast(SignalDatatypeT, w_value),
+            readback=cast(SignalDatatypeT, value),
```
coretl, line 975: Nit


```suggestion
            readback=value,
```

coretl (approved):
Couple of nits, then good to merge

PR #1376, opened 2026-07-24
Update scanspec to 1.0.0

coretl:
## What changed

- Bumped the runtime pin in `[project].dependencies` from `scanspec>=0.8` to `scanspec>=1.0.0`, now that scanspec's final 1.0.0 is released on PyPI.
- Removed the redundant `scanspec>=1.0a1` entry from `[dependency-groups].dev` — it only existed to let dev/CI pull the 1.0 alpha ahead of release, and is now subsumed by the runtime pin.
- Relocked `uv.lock`, which now resolves `scanspec` to `1.0.0` (previously `1.0a2`).
- Fixed the one place in the test suite that constructs a `scanspec.specs.Spiral` directly (`tests/unit_tests/epics/pmac/test_pmac_trajectory_generation.py::test_spiral_trajectory_from_slice`). scanspec 1.0.0 changed `Spiral`'s constructor from a `(x_axis, y_axis, x_start, y_start, x_range, y_range, num)` positional API to a `(x_axis, x_centre, x_diameter, x_step, y_axis, y_centre, y_diameter)` keyword API — `num` points is gone, replaced by a radial `step` size. I chose `x_step=3.0` with `x_diameter=y_diameter=5`, which reproduces the same underlying spiral geometry as the old call (`_estimate_num` resolves to the same `num=3`), so the recorded trajectory/velocity assertions are unchanged.

1. `uv run --locked tox -e tests -- tests/unit_tests`
2. Confirm the suite passes (aside from 4 pre-existing, unrelated failures — see below).

### Checks run and results
- `tests/unit_tests` (865 passed, 2 skipped, 2 xfailed, 6 xpassed). 4 failures remain, all pre-existing and unrelated to this change — they're `RuntimeError: Address family not supported by protocol` from PVA connection attempts, an IPv6 restriction in the sandbox this PR was prepared in, not something introduced here. I confirmed `test_spiral_trajectory_from_slice` specifically regresses on scanspec 1.0.0 without this fix (passes cleanly against the old 1.0a2 pin) and passes with it.
- `pyright src` (ad hoc, matching `CLAUDE.md` guidance): 0 errors, 0 warnings.
- `pre-commit run --all-files`: all hooks passed (ruff lint/format, uv.lock sync check, import-linter, gitleaks, etc).

### Deliberately left untouched
- No other dependency pins were touched.
- System tests and container tests were not run (need live IOCs, out of scope for this change).
- The 4 pre-existing IPv6-related unit test failures were left as-is (not caused by this change).

### Checks for reviewer
- [x] Would the PR title make sense to a user on a set of release notes
- [ ] If the change requires a bump in an IOC version, is that specified in a `##Changes` section in the body of the PR — N/A, no IOC version change
- [ ] If the change requires a bump in the PandABlocks-ioc version, is the `ophyd_async.fastcs.panda._hdf_panda.MINIMUM_PANDA_IOC` variable updated to match — N/A, no PandA IOC change

## AI Disclaimer

---
_Generated by [Claude Code](https://claude.ai/code/session_01GfcSNgEdtKGgFh2xAru4uA)_

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "pydantic>=2.0",
     "pydantic-numpy",
     "stamina>=23.1.0",
-    "scanspec>=0.8",
+    "scanspec>=1.0.0",
     "velocity-profile",
 ]
 dynamic = ["version"]
@@ -63,7 +63,6 @@ dev = [
     "pytest-rerunfailures",
     "pytest-timeout",
     "ruff",
-    "scanspec>=1.0a1",
     "sphinx-autobuild",
     "sphinx-autodoc2",
     "sphinxcontrib-mermaid",
--- a/tests/unit_tests/epics/pmac/test_pmac_trajectory_generation.py
+++ b/tests/unit_tests/epics/pmac/test_pmac_trajectory_generation.py
@@ -150,7 +150,15 @@ async def test_spiral_trajectory_from_slice(
     sim_motors: tuple[PmacIO, Motor, Motor], motor_info: _PmacMotorInfo
 ):
     _, sim_x_motor, sim_y_motor = sim_motors
-    spec = Spiral(sim_x_motor, sim_y_motor, 0, 0, 5, 5, 3)
+    spec = Spiral(
+        x_axis=sim_x_motor,
+        x_centre=0,
+        x_diameter=5,
+        x_step=3.0,
+        y_axis=sim_y_motor,
+        y_centre=0,
+        y_diameter=5,
+    )
     slice = Path(Fly(2.0 @ spec).calculate()).consume()
 
     trajectory, exit_pvt = Trajectory.from_slice(slice, motor_info, ramp_up_time=2.0)
--- a/uv.lock
+++ b/uv.lock
@@ -1625,7 +1625,6 @@ dev = [
     { name = "pytest-rerunfailures" },
     { name = "pytest-timeout" },
     { name = "ruff" },
-    { name = "scanspec" },
     { name = "sphinx-autobuild" },
     { name = "sphinx-autodoc2" },
     { name = "sphinx-copybutton" },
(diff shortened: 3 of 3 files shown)


coretl:
The `test (ubuntu-latest, 3.11)` failure and the resulting `codecov/project` drop are both unrelated to this change:

- The only test failure is `tests/unit_tests/core/test_signal.py::test_set_and_wait_for_value_same_set_as_read` — a `TimeoutError: signal didn't match 1 in 0.1s` race in the `set_and_wait_for_value` EPICS-signal timing test, nothing to do with scanspec. It's the same flaky family as the just-landed `tests: harden set_and_wait error test against slow-loop flake (#1375)` — that fix apparently doesn't cover this sibling test on a busy runner.
- The codecov drop is fallout from that job failing before finishing its coverage report, not a real regression from this diff.

I tried to re-run the failed job but the overall workflow run was still in progress (`403 This workflow is already running`). I'll re-run it once the run completes, and will only push a fix here if the failure turns out to reproduce consistently rather than being a timing flake.


---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Same failure reproduced on rerun (`test_set_and_wait_for_value_same_set_as_read`, `TimeoutError: signal didn't match 1 in 0.1s`).

Root cause confirmed: this is a sibling of the exact issue fixed in the base commit this PR is built on, `9c3836e` ("tests: harden set_and_wait error test against slow-loop flake", #1375) — that PR bumped `test_set_and_wait_for_value_waits_for_error`'s timeout from 0.1s to 1.0s because a slow event loop can make the fixed 0.1s timeout fire before the real completion. `test_set_and_wait_for_value_same_set_as_read` has the identical shape (0.1s timeout racing real work) but wasn't covered by that fix.

This is pre-existing test-infrastructure flakiness, unrelated to the scanspec bump, and out of this PR's declared scope (dependency pin + lockfile only). Leaving it as-is per that scope rather than patching test timeouts here — happy to open a follow-up/fix it in a separate PR if useful, just flagging so it doesn't block review of the actual scanspec change.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1379, opened 2026-07-25
Add regression test to cover consistency between trigger() and collect_asset_docs() observed collections_written

thopkins32:
Following up from my comment in #1334 (https://github.com/bluesky/ophyd-async/pull/1334#discussion_r3650416286), I wanted to ensure coverage in the test suite that asserts `StreamResource` and `StreamDatum` are emitted following `trigger()`, in `collect_asset_docs()`.

While trying to figure this out, I think there could be a subtle bug in how ophyd-async does two *distinct* checks for `dl.collections_written_signal` hitting the target (`collections_per_event`) at two different times; first during `trigger()`, second during `collect_asset_docs()` (i.e. `get_index()`). It could be the case that `dl.collections_written_signal` changes between those two calls which is definitely not something we would want. The condition should only be checked once. Not sure how this would work for fly-scans. I added a test with `XFAIL` for this scenario where `trigger()` observes one thing and `collect_asset_docs()` observes something different (like the signal being reset to 0 by some extraneous process or user input). If there's agreement about this issue, I'll create one.

An example with `collections_per_event = 5` would be that `trigger()` observes `dl.collections_written_signal` go from 0 -> 5 and it's status is completed, but before `collect_asset_docs()`, `dl.collections_written_signal` is reset to 0.

### DIFF
--- a/tests/unit_tests/core/test_standard_detector.py
+++ b/tests/unit_tests/core/test_standard_detector.py
@@ -535,7 +535,6 @@ async def test_collect_asset_docs_uses_trigger_observed_event(
     set_collections_written(5)
     await status
 
-    assert await det.read() == {}
     docs = [doc async for doc in det.collect_asset_docs()]
     assert [name for name, _ in docs] == ["stream_resource", "stream_datum"]
     assert docs[1][1]["indices"] == {"start": 0, "stop": 1}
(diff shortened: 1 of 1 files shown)


jwlodek (approved):
LGTM

PR #1380, opened 2026-07-26
Add an asyncio crash course to the tutorials

### LINKED ISSUE #1319: Add "crash course on asyncio" to the docs page
In docs and the project in general, we assume that the user has a working knowledge of asyncio. This might be true for large facilities that have dedicated support staff for controls, however many facilities (like DESY) rely on the end users to design many parts of the orchestration system.

Additionally, smaller lab users are typically not equipped with an async skillset as they are often pure scientists or those with modest computational science knowledge. I suggest we add a short (1-2 page) document explaining the basics of asyncio, why we use it over threads, what the keywords are, and why they're needed.

I will include a link to the discussion thread from a new user who recently joined the Bluesky matrix:

https://matrix.to/#/!v4kfmFGW4x0r_OXkC4j8y0ojCZmOcbJJQJujquFth3U/$Y2PwZLon2R1BhabSbhuImz_gWwNJZnE3LYBJkm08XAg?via=matrix.org&via=helmholtz.cloud&via=epics-controls.org

As well as relevant comments.

```
Hi Scott Feister , thank you for your feedback! I am primarily a Tango user so I can't speak much towards the EPICS side of things. But regarding your experience picking up ophyd_async, I know it can be a big change to adjust to asynchronicity in python. Ophyd-Async relies heavily on asyncio and keywords such as await, coroutines, event-loop, etc.

Do you think it would be worth adding a one page crash course on what these are in the ophyd-async documentation? As it stands, we assume that the reader already has a working knowledge of python asyncio.

"Yes, I think it would be helpful! I had never used asyncio unfortunately - or, maybe even just directing the reader to go read the basics on \"asyncio\" (point them to a good YouTube video for example) before continuing the tutorial?",

JosephDemarest:
Fixes #1319

Adds a short, beginner-facing asyncio tutorial before the existing device tutorial. It covers:

- why ophyd-async uses asyncio for concurrent hardware I/O instead of coordinating threads;
- the event loop, coroutines, `async def`, `await`, `async with`, and `async for`;
- sequential awaits versus independent operations with `asyncio.gather`;
- the correct async entry point for scripts, IPython, pytest, and Bluesky plans; and
- common mistakes such as un-awaited coroutines and blocking the event loop.

The first coroutine example in "Using Devices" now links back to the primer.

1. Run `tox -e docs` and confirm the warning-as-error Sphinx build passes.
2. Open `build/html/tutorials/asyncio.html` and confirm the new tutorial appears after Installation in the Tutorials navigation.
3. Follow the links to the design goals, Using Devices, signal interaction guide, test tutorial, and Python asyncio documentation.

### Checks for reviewer

- [x] Would the PR title make sense to a user on a set of release notes
- [x] If the change requires a bump in an IOC version, is that specified in a `##Changes` section in the body of the PR (not applicable)
- [x] If the change requires a bump in the PandABlocks-ioc version, is the `ophyd_async.fastcs.panda._hdf_panda.MINIMUM_PANDA_IOC` variable updated to match (not applicable)


JosephDemarest:
CI note — resolved: the initial Windows/Python 3.12 unit-test cell had one timing-dependent failure in the unchanged test_observe_signals_value_timeout_message test (recv1 was [0.0] instead of [0.0, 10.0]; 871 tests passed in that cell). A maintainer reran the cell, and it passed. The corresponding unit suites also passed on Windows 3.11, 3.13, and 3.14 and on all Linux versions; docs, lint, packaging, system tests, and Codecov passed as well. This PR changes Markdown only.

jacopoabramo:
I don't know if this might be considered in scope but: when using 3rd party python libraries that directly interact with hardware, some operations might block the event loop because often the api is synchronous (as in, no async / await syntax). It might be worth mentioning the [asyncio.to_thread](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread) api for blocking I/O.

JosephDemarest:
Thanks @jacopoabramo — good point, especially for hardware libraries that expose only synchronous APIs. I added a focused `asyncio.to_thread()` section in the latest commit, including argument passing, preferring a native async API, checking worker-thread safety, and the I/O-bound limitation.

jacopoabramo:
@JosephDemarest thanks, looks good on my end!

coretl (approved):
Excellent, thanks for this!

PR #1382, opened 2026-07-29
Fix StandardReadable to work with DeviceMap

### LINKED ISSUE #1381: StandardReadable doesn't support DeviceMap
When trying to add `DeviceMap` to the readables of a `StandardReadable`, the child devices all get ignored. 

`StandardReadable.add_children_as_readables()` currently flattens `DeviceVector` children when a `DeviceVector` is added within the context manager, but does not do the same for `DeviceMap`.

This means that `DeviceVector` and `DeviceMap` behave differently when used with `add_children_as_readables()`, despite `DeviceMap` being documented as:

> Like `DeviceVector` but indexed by `str` rather than `int`

### Example

Given:

```python
    self.channel = DeviceVector(
        {
            0: epics_signal_r(float, "TEST1:"),
            1: epics_signal_r(float, "TEST2:"),
        }
    )
```
the signals are correctly flattened and added as readables.

However, changing this to:

```python
    self.channel = DeviceMap(
        {
            "dev1": epics_signal_r(float, "TEST1:"),
            "dev2": epics_signal_r(float, "TEST2:"),
        }
    )
```

does not add the signals as readables. The `DeviceMap` itself is passed to `add_readables()`, but since `DeviceMap` is not `AsyncReadable`, its children are not included in `read()` or `describe()`.

The relevant code currently appears to be here in StandardReadable:

```python
flattened_values = []
for value in new_values:
    if isinstance(value, DeviceVector):
        flattened_values.extend(value.values())
    else:
        flattened_values.append(value)
```

### Expected behaviour

`DeviceMap` should behave equivalently to `DeviceVector` in this context, with its mapped child devices being added as readables.

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1381

### DIFF
--- a/src/ophyd_async/core/_readable.py
+++ b/src/ophyd_async/core/_readable.py
@@ -1,5 +1,5 @@
 import warnings
-from collections.abc import Awaitable, Callable, Generator, MutableMapping, Sequence
+from collections.abc import Awaitable, Callable, Generator, Sequence
 from contextlib import contextmanager
 from enum import Enum
 from typing import Any, cast
@@ -7,7 +7,7 @@ from typing import Any, cast
 from bluesky.protocols import HasHints, Hints, Reading
 from event_model import DataKey
 
-from ._device import Device
+from ._device import Device, DeviceMap, DeviceVector
 from ._protocol import AsyncConfigurable, AsyncReadable, AsyncStageable
 from ._signal import SignalR
 from ._standard_base import _StandardBase
--- a/src/ophyd_async/core/_readable.py
+++ b/src/ophyd_async/core/_readable.py
@@ -186,7 +186,7 @@ class StandardReadable(_StandardBase, AsyncReadable, AsyncConfigurable, HasHints
         flattened_values = []
         for value in new_values:
             # DeviceVector and DeviceMap case.
-            if isinstance(value, MutableMapping):
+            if isinstance(value, (DeviceVector, DeviceMap)):
                 flattened_values.extend(value.values())
             else:
                 flattened_values.append(value)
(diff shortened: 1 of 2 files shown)


src/ophyd_async/core/_readable.py
```diff
@@ -185,7 +185,8 @@ def add_children_as_readables(
 
         flattened_values = []
         for value in new_values:
-            if isinstance(value, DeviceVector):
+            # DeviceVector and DeviceMap case.
+            if isinstance(value, MutableMapping):
```
jacopoabramo:
```suggestion
            # DeviceVector and DeviceMap case.
            if isinstance(value, (DeviceVector, DeviceMap)):
```

I would probably be more strict about the the check and not have it on a broader MutableMapping.

oliwenmandiamond: I personally lean in the other direction and make it less strict. The reason this bug crept through was because it was too specific on what it was checking. If another use case comes up for Device with mapping like behaviour, we won't have to update this method again. If someone wants to make their own device with mapping behaviour, it will also be supported out the box with StandardReadable with no additional changes needed.

coretl: I think I'm going to agree with @jacopoabramo here. We do the filter of `Device` isinstance below, so otherwise we would end up supporting:
```
    self.should_add = DeviceVector({1: Device()})
    self.shouldnt_add = {1: Device(), 2: "not_at_device"}
```
Then you get something that shows up in `read()` but not in `children()` which seems wrong.

oliwenmandiamond: Surely the solution to the above then is check it is of class Device AND MutableMapping?

PR #1384, opened 2026-07-30
Stop two async operations outliving the event loop they were made on (3.14 CI failures)

coretl:
Fixes two unrelated intermittent failures on the Python 3.14 CI jobs, in run
[30454704323](https://github.com/bluesky/ophyd-async/actions/runs/30454704323) on `main`.
Both are races that only lose on a machine slow enough, which is why they showed up
there and nowhere else. Neither is related to the scanspec 1.0.0 bump they first appeared
on.

## 1. `test_status_of_set_and_wait_for_value` — test-only fix

`RuntimeError: Not all tasks closed during test` from `fail_test_on_unclosed_tasks`
([windows job](https://github.com/bluesky/ophyd-async/actions/runs/30454704323/job/90585194312)).

`asyncio.gather()` propagates the first exception without cancelling the other awaitable,
so when the match wait times out, `set_and_wait_for_other_value()` returns with the set
still in flight. The test's expected timeouts each leave one behind and it then ends
without waiting for them.

**The library is deliberately unchanged.** The put *is* the operation the control system
is performing — `ADAcquireLogic.start_acquiring()` puts to `driver.acquire`, and that put
is the acquisition. Stopping it means telling the hardware: `unstage()` →
`ensure_stopped()` → `stop_busy_record()`. Cancelling the task client-side would abandon
the put with the detector still acquiring, and on that path `self.acquire_status` was
never assigned, so `ensure_stopped()` is the only thing that can stop it. The task is not
an unbounded leak — it ends when the put ends, because it is the put.

So the tests now wait for the sets they leave running (`wait_for_sets_left_running()`,
which tracks tasks started during the test rather than reaching into `AsyncStatus`), and
a new test pins the behaviour: the put is still in flight after the match timeout and
completes on its own afterwards.

## 2. Tango subscription leak — library fix

`PytestUnraisableExceptionWarning: ... RuntimeError: Event loop is closed`
([ubuntu system-test job](https://github.com/bluesky/ophyd-async/actions/runs/30454704323/job/90585194299)).

`AttributeProxy.subscribe_callback()` subscribes in a background task, so an
`unsubscribe_callback()` landing first finds no event id to remove; the subscribe task
then stores one that nothing will ever unsubscribe. Tango keeps pushing events at it, and
once the per-test loop closes, pytango's dispatch raises into `sys.unraisablehook` —
failing whichever test the GC happens to be in, not the one that leaked.

In `_tango_transport.py`:
- `_subscribe_to_event()` checks `self._callback` when it lands and undoes its own
  subscription if an unsubscribe got there first; `unsubscribe_callback()` clears
  `_callback` up front so the in-flight task sees it.
- `connect()`'s probe subscription unsubscribes in a `finally`, synchronously — a
  cancelled connect previously escaped between the subscribe and the awaited unsubscribe.
- The subscribe task is held on `self` so it can't be GC'd mid-flight.

## Verified

Both failures reproduced locally and fixed: the dangling-task one by making the set slower
than the match timeout (removing the new test's wait reproduces `Not all tasks closed`
every time; restoring a cancel makes it fail the other way, with the put abandoned), the
Tango one with a plugin counting `subscribe_event`/`unsubscribe_event` pairs — 1 leak
before, 0 after. `pytest tests/system_tests/tango` 201 passed ×3; unit tests green;
ruff, import-linter and pyright clean. `windows-latest` can't be run in my sandbox, so
that job is verified by this PR's CI.

## AI Disclaimer

### DIFF
--- a/src/ophyd_async/core/_signal.py
+++ b/src/ophyd_async/core/_signal.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import functools
 import inspect
 import time
@@ -368,14 +369,26 @@ def soft_signal_rw(
     :param precision: The precision of the signal.
     :param getter:
         Optional callable returning the current device value, called on
-        get_value/get_reading and periodically if poll_period is set.
+        get_value/get_reading and periodically if poll_period is set. May be
+        sync or async; an async getter is awaited.
     :param setter:
-        Optional callable performing the set action. May return the settled
+        Optional callable performing the set action. May be sync or async; an
+        async setter is awaited before the put completes. May return the settled
         value; if it returns None and a getter is configured, the getter is
         called to refresh the cache.
     :param poll_period:
         How often (seconds) to call the getter while a subscription is active.
         Requires getter to be set.
+
+    :example:
+    ```python
+    def read_position() -> float: ...
+
+    async def move_to(position: float) -> float: ...
+
+    # getter and setter can independently be sync or async
+    signal = soft_signal_rw(float, getter=read_position, setter=move_to)
+    ```
     """
     backend = SoftSignalBackend(
         datatype,
@@ -412,7 +425,8 @@ def soft_signal_r_and_setter(
     :param precision: The precision of the signal.
     :param getter:
         Optional callable returning the current device value, called on
-        get_value/get_reading and periodically if poll_period is set.
+        get_value/get_reading and periodically if poll_period is set. May be
+        sync or async; an async getter is awaited.
     :param poll_period:
         How often (seconds) to call the getter while a subscription is active.
         Requires getter to be set.
@@ -633,21 +647,6 @@ async def wait_for_value(
     await _ValueChecker(match).wait_for_value(signal, timeout)
 
 
-async def _cancel_and_wait(task: asyncio.Task) -> None:
-    """Cancel a task and wait for it to finish, discarding its outcome.
-
-    For use on an error path, where the exception about to be reraised is the
-    one that matters: whatever these tasks raise on the way down (including the
-    timeout that got us here) is noise. `await task` would reraise it, so use
-    `asyncio.wait`, which doesn't, then retrieve the outcome so asyncio doesn't
-    log it as never retrieved.
-    """
-    task.cancel()
-    await asyncio.wait([task])
-    if not task.cancelled():
-        task.exception()
-
-
 async def set_and_wait_for_other_value(
     set_signal: SignalW[SignalDatatypeT],
     set_value: SignalDatatypeT,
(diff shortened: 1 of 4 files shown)


src/ophyd_async/core/_signal.py
```diff
@@ -634,6 +633,21 @@ async def wait_for_value(
...
+    For use on an error path, where the exception about to be reraised is the
+    one that matters: whatever these tasks raise on the way down (including the
+    timeout that got us here) is noise. `await task` would reraise it, so use
+    `asyncio.wait`, which doesn't, then retrieve the outcome so asyncio doesn't
+    log it as never retrieved.
+    """
+    task.cancel()
+    await asyncio.wait([task])
+    if not task.cancelled():
+        task.exception()
```
coretl, line 648: Why this and not the `suppress` pattern it replaced?

coretl: Because the `suppress` pattern silently skips whatever cleanup comes after it, and this handler now has two steps instead of one.

On the path that actually matters, `wait_task` is **already done, with the `TimeoutError`** — that timeout is what got us into the `except` block. So `task.cancel()` is a no-op, and `await task` re-raises the `TimeoutError`, which `suppress(asyncio.CancelledError)` does not catch. It propagates straight out of the `except` block.

That was harmless in the old code: the wait was the only thing being cleaned up, and the exception escaping happened to be the same one the trailing `raise` was about to raise anyway. It stops being harmless the moment there is a second step:

```python
except Exception:
    await _cancel_and_wait(wait_task)      # re-raises TimeoutError here...
    if status is not None:
        await _cancel_and_wait(status.task)  # ...so this never runs
    raise
```

which is the whole point of the fix. I hit this for real — my first version of `_cancel_and_wait` was the `suppress` pattern, and `test_set_and_wait_for_other_value_cancels_a_set_that_outlives_the_match` still failed with the set task pending. I just re-confirmed it on this branch: swap those three lines back to `suppress(asyncio.CancelledError)` + `await task` and that test fails, everything else in `test_signal.py` still passes.

`asyncio.wait([task])` doesn't re-raise the task's outcome at all, so the cleanup runs to completion; it still propagates a `CancelledError` aimed at *us*, which is what we want. The `if not task.cancelled(): task.exception()` afterwards is only there to mark the exception retrieved so asyncio doesn't log "Task exception was never retrieved" at GC.

`suppress(asyncio.CancelledError, Exception)` would also work, but it swallows a cancellation targeting this coroutine as well as the task's, which seemed worse than being explicit.

Happy to reword the docstring — it says `await task` "would reraise it", but not the consequence, that the set below then never gets cancelled. That's the bit worth spelling out. Say the word and I'll push it.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl: We should not cancel the set task. That would stop the wait for completion of the hardware, but not the actual hardware. This is what unstage is for.

PR #1389, opened 2026-08-03
Support explicit derived_datatypes for generic derived_signal

### LINKED ISSUE #1388: Support explicit datatypes for generic derived_signal_rw
`derived_signal_rw` currently infers the datatype of the derived signal from the return type annotation of `raw_to_derived`:

```python
raw_to_derived_datatype = _get_return_datatype(raw_to_derived)
set_derived_arg_datatype = _get_first_arg_datatype(set_derived)
```

This works when the callbacks use a concrete datatype, for example:

```python
def _get(value: int) -> int:
    ...

async def _set(value: int) -> None:
    ...
```

However, it does not work when the callbacks use a generic `TypeVar`:

```python
T = TypeVar("T", bound=float | int)

def _get(value: T) -> T:
    ...

async def _set(value: T) -> None:
    ...
```

At runtime, the type annotation resolves to the `TypeVar` itself rather than the concrete type used to instantiate the generic class. This means the derived signal is created with a `TypeVar` as its datatype.

When the signal is subsequently described, `_datakey_dtype_numpy()` attempts to use the datatype in `issubclass()`:

```python
issubclass(datatype, EnumTypes)
```

This raises:

```text
TypeError: issubclass() arg 1 must be a class
```

### Example

A generic device may legitimately need to expose a derived signal whose datatype is determined by the concrete generic type:

```python
T = TypeVar("T", bound=float | int)


class GenericExample(Device, Generic[T]):
    def __init__(self, datatype: type[T]):
        self.sig = soft_signal_rw(datatype, 5.0)
        self.offset = soft_signal_rw(datatype, 1.0)

        self.value = derived_signal_rw(
            self._get,
            self._set,
            sig=self.sig,
            offset=self.offset,
        )

    def _get(self, sig: T, offset: T) -> T:
        return self.datatype(sig + offset)

    async def _set(self, value: T) -> None:
        ...
```

The concrete datatype is known at runtime:

```python
GenericExample(int)
GenericExample(float)
```

but cannot be inferred from the `T` annotations alone.

### Proposed solution

Add an optional `datatype` argument to `derived_signal_rw`:

```python
def derived_signal_rw(
    raw_to_derived: Callable[..., SignalDatatypeT],
    set_derived: Callable[[SignalDatatypeT], Awaitable[None]],
    derived_units: str | None = None,
    derived_precision: int | None = None,
    datatype: type[SignalDatatypeT] | None = None,
    **raw_devices_and_constants: Device | Primitive,
) -> SignalRW[SignalDatatypeT]:
```

When supplied, this provides the concrete runtime datatype of the derived signal:

```python
self.value = derived_signal_rw(
    self._get,
    self._set,
    datatype=datatype,
    sig=self.sig,
    offset=self.offset,
)
```

The existing validation that the return datatype of `raw_to_derived` matches the argument datatype of `set_derived` should remain unchanged. The explicit `datatype` only provides the concrete runtime datatype when it cannot be inferred from generic type annotations.

For example:

```python
raw_to_derived: T -> T
set_derived: T -> None
datatype: int
```

The two callback datatypes still match (`T` and `T`), while `datatype=int` provides the concrete runtime type required by the signal backend.

### Expected behaviour

* If `datatype` is not provided, continue to infer it from the callback annotations as currently.
* If `datatype` is provided, use it as the concrete datatype of the derived signal.
* Continue validating that the return datatype of `raw_to_derived` and the first argument datatype of `set_derived` match.
* Generic derived signals should be able to be described successfully when an explicit concrete `datatype` is provided.

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1388.

I've encountered several errors when using generically typed functions with derived signals. Many of these errors were unclear and difficult to diagnose. I originally encountered this while converting a device to use `StandardMovable`. `StandardMovable` calls `describe()` on the `readback` signal to obtain its units and precision, and when the `readback` was a derived signal using generic typing, this resulted in confusing errors that were difficult to trace back to the underlying cause.

This change makes derived signals more flexible and improves their handling of generic types. Where possible, the concrete datatype is now resolved automatically. If the datatype cannot be resolved, an optional datatype override can be provided. The supplied datatype is validated for compatibility with the generic types used by the get/set functions, with clear errors raised if it is incompatible.

If no `derived_datatype` is provided and the type cannot be resolved, a clearer error is now raised explaining the problem and how to resolve it using the `derived_datatype` argument.

This should hopefully make derived signals using generic typing easier to use and significantly easier to debug.

### DIFF
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -302,15 +302,14 @@ def _validate_datatype(
 
 
 def _validate_datatype_compatibility(
-    actual_datatype: type[SignalDatatypeT],
+    datatype: type[SignalDatatypeT],
     expected_datatype: type[SignalDatatypeT],
-    *,
     context: str,
 ) -> None:
-    if not _is_datatype_compatible(actual_datatype, expected_datatype):
+    if not _is_datatype_compatible(datatype, expected_datatype):
         raise TypeError(
-            f"{actual_datatype} is not compatible with {expected_datatype} "
-            f"for {context}."
+            f"{datatype} is not compatible with "
+            f"{_datatype_description(expected_datatype)} for {context}."
         )
 
 
--- a/src/ophyd_async/core/_derived_signal.py
+++ b/src/ophyd_async/core/_derived_signal.py
@@ -125,11 +125,19 @@ class DerivedSignalFactory(Generic[TransformT]):
     def _make_signal(
         self,
         signal_cls: type[SignalT],
-        datatype: type[SignalDatatypeT],
+        datatype: type[SignalDatatypeT] | TypeVar,
         name: str,
         units: str | None = None,
         precision: int | None = None,
     ) -> SignalT:
+        if isinstance(datatype, TypeVar):
+            # Cannot use name here to help identify as it defaults to "value" at this
+            # stage
+            raise TypeError(
+                f'Cannot determine signal datatype from TypeVar "{datatype}". '
+                "If this is a generic derived signal, provide a concrete datatype "
+                'using the "derived_datatype" argument.'
+            )
         # Check up front the raw_devices are of the right type for what the signal_cls
         # supports
         if issubclass(signal_cls, SignalR):
@@ -266,15 +274,15 @@ def _is_datatype_compatible(
 
 
 def _is_typevar_compatible(
-    typevar: TypeVar, datatype: type[SignalDatatypeT] | TypeVar
+    typevar: TypeVar, derived_datatype: type[SignalDatatypeT] | TypeVar
 ) -> bool:
     if typevar.__constraints__:
         return any(
-            _is_datatype_compatible(datatype, constraint)
+            _is_datatype_compatible(derived_datatype, constraint)
             for constraint in typevar.__constraints__
         )
     if typevar.__bound__ is not None:
-        return _is_datatype_compatible(datatype, typevar.__bound__)
+        return _is_datatype_compatible(derived_datatype, typevar.__bound__)
     return True
 
 
@@ -313,7 +321,7 @@ def derived_signal_r(
     raw_to_derived: Callable[..., SignalDatatypeT],
     derived_units: str | None = None,
     derived_precision: int | None = None,
-    datatype: type[SignalDatatypeT] | None = None,
+    derived_datatype: type[SignalDatatypeT] | None = None,
     **raw_devices_and_constants: Device | Primitive,
 ) -> SignalR[SignalDatatypeT]:
     """Create a read only derived signal.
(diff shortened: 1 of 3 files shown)


src/ophyd_async/core/_derived_signal.py
```diff
@@ -262,25 +313,43 @@ def derived_signal_r(
...
     """Create a read only derived signal.
 
     :param raw_to_derived:
         A function that takes the raw values as individual keyword arguments and
         returns the derived value.
-    :param derived_units: Engineering units for the derived signal
-    :param derived_precision: Number of digits after the decimal place to display
+    :param derived_units: Engineering units for the derived signal.
+    :param derived_precision: Number of digits after the decimal place to display.
+    :param datatype:
```
coretl: Please can we spell it `derived_datatype` so there's less likelihood of a collision with the raw_devices_and_constants dict

src/ophyd_async/core/_signal_backend.py
```diff
@@ -145,6 +145,12 @@ def _datakey_dtype(datatype: type[SignalDatatype]) -> Dtype:
 def _datakey_dtype_numpy(
     datatype: type[SignalDatatypeT], value: SignalDatatypeT
 ) -> np.dtype:
+    if isinstance(datatype, TypeVar):
```
coretl: Please can we catch this earlier, probably inserting an exhaustive check like that in `SoftSignalBackend` `make_converter` inside `DerivedSignalBackend` or `_make_signal` would catch all this at signal creation time, then the type hints would be correct from then on

coretl:
You've got a nice example of a generic class in https://github.com/bluesky/ophyd-async/issues/1388, please could you put it in the tests?

src/ophyd_async/core/_derived_signal.py
oliwenmandiamond: Done

src/ophyd_async/core/_signal_backend.py
oliwenmandiamond: I've added to `_make_signal`

oliwenmandiamond:
> You've got a nice example of a generic class in #1388, please could you put it in the tests?

I can add a generic device example and I did this originally, but I'd prefer to build it using the existing small helper functions. The original `GenericExample` was useful for reproducing the issue, but it was quite restrictive for tests, whereas using the small functions in the tests let us cover the generic typing behaviour more directly and flexibly for each combination. 

I've applied all of the the other feedback

coretl:
> I can add a generic device example and I did this originally, but I'd prefer to build it using the existing small helper functions. The original `GenericExample` was useful for reproducing the issue, but it was quite restrictive for tests, whereas using the small functions in the tests let us cover the generic typing behaviour more directly and flexibly for each combination.

I'm happy the bulk of the tests are done with what you've done, but I like the pattern of `test_xxx_use_cases.py` files that are recognisable use cases you might actually want. Then if we ever change the code in the future, we can see if it breaks internal tests (changes possible) or use cases (needs wider audience). It also answers the question I had when just reading this PR:

Why can't we solve it like the `Locatable[T]` detection we already have? Because `T` isn't defined in the class heirarchy but inferred from an `__init__` argument.

If you don't have a use case example where `T` is inferred then we might forget this in the future.

oliwenmandiamond:
Okay, done this now

coretl (approved):
Excellent, thanks

PR #1392, opened 2026-08-11
Loosen type hints for protocols from AsynStatus to AsynStatusBase to allow using WatchableAsyncStatus returns as well

jwlodek:


src/ophyd_async/core/_protocol.py
```diff
@@ -84,25 +84,25 @@ class AsyncStageable(Protocol):
...
+        :return: An `Status` that is marked done when the device is done staging.
         """
 
     @abstractmethod
-    def unstage(self) -> AsyncStatus:
+    def unstage(self) -> AsyncStatusBase:
         """Clean up the device after acquisition.
 
-        :return: An `AsyncStatus` that is marked done when the device is done unstaging.
+        :return: An `Status` that is marked done when the device is done unstaging.
```
jacopoabramo:
```suggestion
        :return: A `Status` that is marked done when the device is done unstaging.
```

Just small typos.

src/ophyd_async/core/_protocol.py
```diff
@@ -84,25 +84,25 @@ class AsyncStageable(Protocol):
     """Async implementation of the sync [](#bluesky.protocols.Stageable)."""
 
     @abstractmethod
-    def stage(self) -> AsyncStatus:
+    def stage(self) -> AsyncStatusBase:
         """Set up the device for acquisition.
 
-        :return: An `AsyncStatus` that is marked done when the device is done staging.
+        :return: An `Status` that is marked done when the device is done staging.
```
jacopoabramo:
```suggestion
        :return: A `Status` that is marked done when the device is done staging.
```

src/ophyd_async/core/_standard_base.py
```diff
@@ -3,7 +3,7 @@
 
 from ._device import Device
 from ._protocol import AsyncStageable
-from ._status import AsyncStatus
+from ._status import AsyncStatus, AsyncStatusBase
```
jacopoabramo, line 6:
```suggestion
from ._status import AsyncStatusBase
```

Unless I'm mistaken AsyncStatus doesn't need to be imported again

jacopoabramo (approved):
Looks good to me. Another set of eyes with a broader understanding of the code might be warranted (if there might be side effects for this change in other context at least).

coretl (approved):

PR #1393, opened 2026-08-12
Support np.ndarray with Any shape for SoftSignalBackend

### LINKED ISSUE #1335: SoftSignalBackend has no mock-mode support for np.ndarray[Any, np.dtype[T]]-annotated signals
## Description

**Blocked by #1338**: fix `Array1D` to be genuinely type-checker-narrowed to 1-D first - today it isn't (`tuple[int, ...]` is variadic, not a fixed 1-tuple), which means it's only different from the "any-shape, fixed-dtype" spelling below by *spelling*, not by what it actually means statically. Doing this issue first would be fixing a near-duplicate rather than adding real new capability.

`SoftSignalBackend.make_converter` (`src/ophyd_async/core/_soft_signal_backend.py`)
only accepts two shapes of ndarray-typed signal:

- bare `np.ndarray` (any shape, any dtype)
- `Array1D[dtype]` (`np.ndarray[tuple[int, ...], np.dtype[T]]` - *intended* to be 1-D, dtype-narrowed, but see #1338: not actually narrowed to 1-D at the type-checker level today)

There's no support for `np.ndarray[Any, np.dtype[T]]` - dtype-narrowed but
*not* shape-restricted to 1-D (the natural annotation for e.g. a 2-D "image"
attribute where the pixel dtype matters but the shape doesn't). Connecting
such a signal in mock mode (`connect(mock=True)`) raises:

```
TypeError: Expected Array1D[dtype], got numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]]
```

from `make_converter`'s dtype-narrowed-ndarray branch, which checks
`datatype not in get_args(SignalDatatype)` - and `SignalDatatype`'s closed
union only lists the finite set of `Array1D[dtype]` variants, not this
"any-shape, fixed-dtype" spelling.

## Where this was found

`ophyd_async.tango.testing.TangoTestDevice.float64_image` is annotated
`SignalRW[np.ndarray[Any, np.dtype[np.float64]]]`. This annotation is
load-bearing for *real* Tango connects - `TangoSignalBackend`'s
`_verify_datatype_matches`/`_type_match_array` needs the dtype-narrowed form
to validate against the server's declared attribute type; switching it to
bare `np.ndarray` was tried and breaks real `connect()` instead (confirmed
empirically in #1332). So the annotation can't just be changed to work
around this - mock mode itself needs to learn this shape.

Every other image-typed signal in this codebase (e.g.
`EpicsTestPvaDevice.ntndarray`) is annotated bare `np.ndarray`, so this
specific combination (dtype-narrowed + shape-`Any`) hadn't been exercised in
mock mode before `TangoTestDevice` existed - no test caught it until #1332's
mock-parity test tried to `connect(mock=True)` the whole device and hit it
(worked around there by using a local device subclass without this field,
not by fixing this).

## Suggested fix

Extend `make_converter`'s dtype-narrowed-ndarray branch (and/or
`SignalDatatype`'s union, and `NDArraySoftConverter`) to also accept
`np.ndarray[Any, np.dtype[T]]` shapes, not just `Array1D[T]` - likely by
relaxing the origin/shape-parameter check rather than requiring literal
membership in the closed `SignalDatatype` union.

## References

- Blocked by bluesky/ophyd-async#1338
- Discovered in bluesky/ophyd-async#1332
- Part of the system-test rewrite epic, bluesky/ophyd-async#1321

jacopoabramo:
Closes #1335 

Had some help with Claude.

At first import, SignalDatatype is inspected to check all the supported dtypes for any np.ndarray type (which includes both Array1D and the general np.ndarray itself).

Adds also a couple of parametrization to existing unit tests.

I haven't checked all data types as I was worried of making the parametrization too verbose but I can address that if needed.

EDIT 29-08-2026
- restricts EPICS signals to allow two possible shapes:
  - Array1D[np.xxx]: a waveform with specified data type
  - np.ndarray: an NDArray of unknown shape and dtype (area detectors can change these dynamically at runtime)


tests/unit_tests/core/test_soft_signal_backend.py
```diff
@@ -42,7 +42,7 @@ def enum_d(value):
 
 
 def waveform_d(value):
-    return {"dtype": "array", "shape": [len(value)]}
+    return {"dtype": "array", "shape": list(np.shape(value))}
```
jacopoabramo, line 46: This is to extend the descriptor to n-dimensional arrays

jacopoabramo:
@coretl 

I'm not sure if the CI fail is a flake or an actual bug but doesn't feel like it...

tests/unit_tests/core/test_soft_signal_backend.py
```diff
@@ -76,6 +84,20 @@ def table_d(value):
         (Array1D[np.uint64], np.array([]), np.array([995444]), waveform_d, "<u8"),
         (Array1D[np.float32], np.array([]), np.array([1.0]), waveform_d, "<f4"),
         (Array1D[np.float64], np.array([]), np.array([0.2]), waveform_d, "<f8"),
+        (
+            np.ndarray[Any, np.dtype[np.uint16]],
```
coretl:
```suggestion
            npt.NDArray[np.uint16],
```

src/ophyd_async/core/_soft_signal_backend.py
coretl, line 1: Please can we have code and tests that record whether we wanted a 1D or ND array, then checks the shape in the converter

coretl (changes requested):
We also need to ban these particular options in EPICS, and add some docs on the 3 different spellings of arrays we now support:
- `Array1D[np.xxx]`
- `npt.NDArray[np.xxx]`
- `np.ndarray`

Please could you get claude to look at those too

jacopoabramo:
> `np.ndarray`

Is there a reason to still support a plain `np.ndarray`?

jacopoabramo:
> We also need to ban these particular options in EPICS

Which particular options exactly?

coretl:
> Is there a reason to still support a plain `np.ndarray`?

Yes, an epics areadetector NTNDArray can change both datatype and shape at runtime

> Which particular options exactly?

EPICS support:
- `Array1D[np.xxx]`: is a `waveform` record
- `np.ndarray`: is an areadetector NTNDArray
- `npt.NDArray[np.xxx]`: is unsupported

so we want to check if anyone is asking for variable shape but fixed datatype and raise if so.

@burkeds can comment on which of the 3 options Tango supports

tests/unit_tests/epics/pmac/test_pmac_trajectory.py
```diff
@@ -372,8 +372,8 @@ async def test_pmac_trajectory_stage(sim_motors: tuple[PmacIO, Motor, Motor]):
     assert mock_pmac_trajectory_io.mock_calls[
         len(logic.pmac.trajectory.use_axis) :
     ] == [
-        call.time_array.put(np.array(0)),
-        call.user_array.put(np.array(8)),
+        call.time_array.put(np.array([0])),
+        call.user_array.put(np.array([8])),
```
jacopoabramo, line 376: This was caught by claude, which justified it as a consistency behavior in putting an Array1D rather than a 0-dim array

jacopoabramo:
I split each requested change in separate commits:

> Please can we have code and tests that record whether we wanted a 1D or ND array, then checks the shape in the converter

[2900640](https://github.com/bluesky/ophyd-async/pull/1393/commits/290064081586d5da34376588a93341453c4c24eb)

> We also need to ban these particular options in EPICS, and add some docs on the 3 different spellings of arrays we now support:
> 
> Array1D[np.xxx]
> npt.NDArray[np.xxx]
> np.ndarray
> Please could you get claude to look at those too

[6a72a80](https://github.com/bluesky/ophyd-async/pull/1393/commits/6a72a80efb4b7dd58c716c05db4772922e499dad)

I'm waiting on @burkeds so that I can apply the same to Tango signals if necessary. I'll update the PR description since it's doing more than what's actually described right now.

tests/unit_tests/epics/pmac/test_pmac_trajectory.py
coretl: claude is right

coretl (approved):
Happy with the soft and EPICS changes, thanks!

Please could you check whether Tango has examples of all 3 sorts of these annotations in its system tests? If not then please can you get claude to expand the tests to include all of them and see if any errors happen. Then we can merge.

jacopoabramo:
I'm traveling today, I'll try to work on that tomorrow (or as soon as possible anyway).

jacopoabramo:
> Please could you check whether Tango has examples of all 3 sorts of these annotations in its system tests? If not then please can you get claude to expand the tests to include all of them and see if any errors happen. Then we can merge.

[3e691e6](https://github.com/bluesky/ophyd-async/pull/1393/commits/3e691e6519fb168ef539b6c5e1286ec2a37d1050) addresses this. It removes a pre-existing mock test device that was implemented because the spellings this PR covers were not there yet.

I'm still waiting on confirmation on what kind of formats TANGO actually supports, overall this should be ready.

burkeds:
Hi all, sorry for the delay. I had a week of beamtime then work to catch up on. Trying my best to follow what's going on here.

So the Tango side of ophyd-async needs to support the enumerated pytango data types: https://tango-controls.readthedocs.io/projects/pytango/en/latest/api/data_types.html

The iterable types returned by pytango are:

np.ndarray[dtype=np.type]
list[type]
Sequence[(np.ndarray[dtype=np.int32] | Sequence[int], Sequence[str])
Sequence[(np.ndarray[dtype=np.float64] | Sequence[int], Sequence[str])
bytearray

Pytango types have SPECTRUM and IMAGE flavors but those are checked for and the dtype returned by the tango backend is static. These are all fixed shape and type if I understand correctly.

ophyd-async tango does not currently support bytearrays or the DEV_ENCODED pytango type. 

`_tango_transport.py` currently maps all pytango types (except DEV_ENCODED) to one of the project defined enum and table types, a python scalar, a Sequence of strings, or an Array1D[np.type].

jacopoabramo:
> Hi all, sorry for the delay.

No worries, thank you for taking the time to answer - apologies for the annoyance.

> ophyd-async tango does not currently support bytearrays or the DEV_ENCODED pytango type.

If I'm understanding this correctly, these are the only gaps. But please confirm if I'm correct.

For the sake of simplicity I would try and address them as a separate PR. Right now these changes seem to work (test-wise at least, which doesn't say much...) and I rather not make this contribution bigger than what already is.

I will make a second PR of course. I'm interested in getting a better understanding of both control systems from ophyd-async perspective anyway.

burkeds:
Supporting bytearrays will likely be the subject of a future PR. I don't think we need to be too concerned about it for a 1.0 release. DEV_ENCODED is the only Pytango datatype which is not supported by _tango_transport.

docs/explanations/decisions/0020-array-signal-spellings.md
```diff
@@ -0,0 +1,59 @@
...
+accepted a scalar or a nested list and stored it unchanged.
+
+## Decision
+
+Soft signals accept `npt.NDArray[np.<dtype>]` as a fixed datatype array whose
+number of dimensions is unconstrained.
+
+The soft converter records the number of dimensions the annotation asks for and
+raises `ValueError` on a write of a different rank. `Array1D[np.float64]` accepts
+only 1D arrays, `np.ndarray[tuple[int, int], np.dtype[np.float64]]` only 2D, and
```
coretl: This promises to support 2D arrays. I don't think we should do that. I think we should stick with exactly:
- `Array1D[np.xxx]`
- `npt.NDArray[np.xxx]`
- `np.ndarray`

src/ophyd_async/core/_signal_backend.py
coretl: I was expecting to see `npt.NDArray[np.xxx]` entries here...

src/ophyd_async/core/_soft_signal_backend.py
```diff
@@ -92,6 +99,13 @@ def write_value(self, value: Any) -> TableT:
             raise TypeError(f"Cannot convert {value} to {self.datatype}")
 
 
+_SUPPORTED_NDARRAY_DTYPES = frozenset(
+    get_dtype(datatype)
+    for datatype in get_args(SignalDatatype)
+    if cached_get_origin(datatype) is np.ndarray
```
coretl: ... so this wouldn't be needed?

coretl (commented):
Just did a closer review and found a couple of places we should be stricter...

jacopoabramo:
> Just did a closer review and found a couple of places we should be stricter...

I'll try to review and adjust these next week, I'll be a little bit busy for personal reasons these few days

Issue #1394, opened 2026-08-12
Allow for changing standard readable format at run-time


jwlodek:
Currently there is no way to change the readable format at run-time. This differs from the equivalent feature in ophyd v1 (i.e. `kind`) which can be changed at run-time. The ability to change signals or types from config to normal or hinted, or from hinted to normal etc. is valuable in several contexts.

For example, if a beamline has two techniques, one of which scans a monochromater's energy, while another keeps it fixed, it makes sense to have the energy signal be hinted for the first, while config for the second. This is easy to achieve in ophyd v1 by just updating the `kind` attribute as needed, but in ophyd async requires editing the device definition, or having multiple identical devices minus one signal's format changed.

## Acceptance Criteria
- Readable format can be changed at run-time via an attribute update or a helper method.

PR #1395, opened 2026-08-12
Rework StandardDetector to extend from StandardReadable, to follow a more ophyd-v1-like pattern for read signals for step scanning and config signals.

jwlodek:
`StandardDetector` now inherits `StandardReadable`, and areaDetector IO classes
(`ADBaseIO`, `NDStatsIO`, `NDROIStatNIO`, `NDROIIO`) declare their own signal
formats using `StandardReadableFormat` annotations. This means detector plugins
and drivers follow the same pattern as any other `StandardReadable` device —
they self-describe which signals are configuration, which are hinted reads, and
which are uncached.

## Motivation

**Readables and detectors now follow the same pattern for signal data.** Before this change,
`StandardDetector` had its own separate mechanism for configuration and reading,
while `StandardReadable` had `add_readables()` with format annotations. Now
both use the same `StandardReadable` machinery. This makes adding step-scan signal reading functionality
to a file-writing detector identical to how it is handled for just StandardReadable. A plugin like `NDStatsIO` is
just a `StandardReadable` that declares its signals, and adding it to a detector
with `Format.CHILD` is identical to composing any two `StandardReadable` devices.
Data logics are still the required approach for any file-based data.
Adding a data logic that is readable only is also still supported.

**IO classes own their signal roles.** Previously, the detector or a separate
`PluginSignalDataLogic` class had to know which plugin signals were
configuration vs. read vs. hinted. Now `ADBaseIO` annotates `acquire_time`,
`model`, etc. as `Format.CONFIG_SIGNAL`, and `NDStatsIO` annotates `total`,
`mean_value` as `Format.HINTED_UNCACHED_SIGNAL` — the knowledge lives where
the signals are defined, not in an external wiring class. Signals or other readables
can also be added outside of the constructor via `det.add_readables`, just as with
`StandardReadable`.

**Closer to ophyd v1.** In ophyd v1, a detector's `Component` declarations
determined what appeared in `read()` and `read_configuration()`, and plugins
composed naturally as sub-devices. `StandardReadable` already followed this pattern,
but `StandardDetector` did not, which caused confusion in some cases during transitions.
This change restores that composability:
adding a plugin to a detector automatically pulls in its read and config
signals, just as adding a component did in v1. Users migrating from ophyd v1
will find the mental model familiar.

**Automatic `wait_for_plugins` management.** `AreaDetector._update_prepare_context`
now uses `has_child_readables(ignore_filter=[self.driver])` to detect whether
any plugin has read signals that need to complete before the detector reads.
During step scans (`number_of_events == 1`), it sets `wait_for_plugins = True`
automatically. Fly scans skip this since they stream data asynchronously. An explicit
`PluginSignalDataLogic` (which was a somewhat verbose solution) is no longer required.

- **Add helper function to standard readable that checks if there are any registered child readables**
- **Make standard detector extend from standard readable. Allow for pulling read/config/describe from children automatically**
- **Update EPICS areaDetector classes to use standard readable child components for config/read**
- **Update implementations to move config_sigs -> deadtime_sigs for trigger logics**


src/ophyd_async/epics/adcore/_io.py
```diff
@@ -156,19 +164,27 @@ class NDPluginBaseIO(NDArrayBaseIO):
...
+    min_y: A[SignalRW[int], PvSuffix.rbv("MinY")]
     size_x: A[SignalR[int], PvSuffix.rbv("SizeX")]
     size_y: A[SignalR[int], PvSuffix.rbv("SizeY")]
     size_z: A[SignalR[int], PvSuffix.rbv("SizeZ")]
 
+    def __init__(self, prefix: str = "", with_pvi: bool = False, name: str = ""):
+        super().__init__(prefix, with_pvi, name)
+        self.add_readables(
+            [self.min_x, self.min_y, self.size_x, self.size_y], Format.CONFIG_SIGNAL
+        )
```
thopkins32, line 184: Why add the readables in `__init__` for these driver IO classes? Can't you do it the type annotation way like in `ADBaseIO`?

src/ophyd_async/epics/adcore/_detector.py
```diff
@@ -48,11 +55,19 @@ def __init__(
...
+            self.add_config_signals(*config_sigs)
         super().__init__(name=name)
 
+    async def _update_prepare_context(self, trigger_info: TriggerInfo) -> None:
+        # Step scans (trigger+read) need plugins to finish before reading scalars
+        if trigger_info.number_of_events == 1 and self.has_child_readables(
+            ignore_filter=[self.driver]
+        ):
+            await self.driver.wait_for_plugins.set(True)
+        await super()._update_prepare_context(trigger_info)
```
thopkins32, line 69: This makes the assumption that every `AreaDetector` wants `WaitForPlugins = true` if they have child readables. I don't really want to have to subclass `AreaDetector` to turn this off. I have step scans where I want more custom waiting than a generic wait for all plugins. 

For example, a very long plugin chain that is only for display purposes, I don't want displays to slow down my scans. `WaitForPlugins` is too coarse for this case.

The data logic abstraction allows you to add this without subclassing. Is the same thing possible to achieve with a data logic? Where the `IO` classes are still `StandardReadable` (that part I like a lot).

src/ophyd_async/epics/adcore/_io.py
jwlodek: Originally I had more signals with formats set, and there were a lot of  `Format.*` repeated, so this was more concise. I removed most and I think I can go back to using the declarative style.

coretl:
I need to understand the use cases more before I can do a decent review of this.

The way we use plugins, we might have a big set of plugins that are instantiated but not actually wired up. E.g. we have 16 ROIs and 16 stats for every detector that are rarely used. This means with this change we would end up reporting the values of all those plugins on every step scan, with no way for the user to know whether they are useful values or not.

Also I agree with https://github.com/bluesky/ophyd-async/pull/1395#discussion_r3768330879 and think that setting `WaitForPlugins` inside `AreaDetector` is too hidden.

My stance so far was to make all the detector things explicit, you had to add a logic for everything you used in both step and fly scans. I avoided tracing plugin chains because that's a complicated route to go down. That's why `StandardDetector` doesn't inherit from `StandardReadable`, because of the need to select what to read from the child plugins, rather than get everything that may or may not be used.

I can offer a couple of directions we could go with this:
1. Go all in on chain parsing, and do sizing, datatypes, and plugin reading based on whether the plugin is wired up
2. Stick with the current approach, and add some helpers to allow people to opt-in to the plugins they know are contributing in the same way we do for multiple HDF streams: https://blueskyproject.io/ophyd-async/main/how-to/implement-ad-detector.html#multiple-hdf-writers-for-different-rois

I'm wary of the amount of work required for 1, and what goes wrong if we start using gather/scatter and PVA transfer plugins between instances, as well as the requirement then to expose *every* plugin to ophyd. 2 is a bit more verbose, but I think it's better to be explicit. It also gives us a place to opt into `WaitForPlugins`.

I think should all be considered with #1367, when you mix step scan only signals like minX, their timeseries equivalents for flyscans, and the ability to stamp them on NDArrays, then we need to decide when to use each, and what to do about things like `WaitForPlugins`. 

Thoughts?

jwlodek:
> The way we use plugins, we might have a big set of plugins that are instantiated but not actually wired up. E.g. we have 16 ROIs and 16 stats for every detector that are rarely used.

We use AD in the same way, but we generally only tell ophyd about the ones we are interested in. In ophyd v1, we can also "turn off" a plugin at run-time by updating it's `kind` attribute, if we want to turn off recording stats1 for example. 

```Python
det.stats1.kind = Kind.omitted
```

So depending on the technique we are running, we'll enable/disable the plugins as needed, which avoids lots of plugins reporting information that isn't relevant. Would you include all 16 ROI plugins in your ophyd object instance, even if most of the time you don't plan to use them?

I have a few issues with the current approach, which prompted the construction of this PR:

As-is, StandardDetector's in ophyd async behave differently than all other readable devices, and all ophyd v1 devices when it comes to getting config/reading data from children.  I think this PR allows for them to behave closer to other devices, while retaining the composition over inheritance approach, since you can always `add_readables`, i.e.

```Python
my_det = Detector()
my_plugin = Plugin()
my_det.add_readables(my_plugin)
```

For example, to define config signals, there are two divergent ways of doing it at the moment - either via StandardReadableFormat (for Readables), or in the TriggerLogic/__init__ for detectors. At least from my perspective, the first is more elegant - I can add some extra stats plugins with no modifications to my logic class and still get their configuration stored (again, via composition, not necessarily by making them an attribute of the parent class). Specifying detector level config signals alongside the IO definitions (i.e. ADBaseIO or AravisDriverIO), feels more natural to me as well. The trigger logic doesn't have access to the plugins (only the driver), so any plugin-level config must be passed at `__init__`, and since the config signals must be spelled out one-by-one (there's no way to ask a plugin about it's configuration), if you have many plugins it becomes very verbose (and it has to be repeated for every instantiation). It's a similar issue with specifying which signals I'm interested in reading - the current `PluginSignalDataLogic` needs to be called for every signal. There's also no way to adjust at run time, as best as I can tell. There's also no way to specify that I want to read all the interesting signals for a plugin, they must be listed out one by one.

> Go all in on chain parsing, and do sizing, datatypes, and plugin reading based on whether the plugin is wired up.

I don't think this is strictly necessary, again provided we could accommodate #1394 . That way it would be possible to have the user specify what they're interested in manually if they are switching techniques (just like with explicitly requiring the data logics), by just adjusting the format of each plugin.

Alternatively, if we don't want the plugins included by default, what if we didn't add plugins as readables automatically when passed in at `__init__`, and instead explicitly required an `add_readables` call to include them? That would make things behave functionally the same as now, just with the added feature of being able to select an entire plugin at a time (and nothing would stop you from `add_readables` for individual signals if you only wanted a subset of the default configuration, or if you wanted something extra)? I think that way the behavior of `StandardDetector` is much closer to all other devices, without adding all plugin info to step scans automatically.

I do agree that breaking back out the `WaitForPlugins` makes sense. I think we can make a data logic for that and have the user explicitly pass it in if the behavior is desired. I guess it should also have a prepare_unbounded where it turns off `WaitForPlugins`?

PR #1396, opened 2026-08-14
Document that mock/soft callbacks accept sync or async callables

coretl:
It wasn't clear from the docs that `callback_on_mock_put` accepts an `async def` callback as well as a sync one — the only way to find out was to read the `MockPutCallback` alias or the tests.

I swept the codebase for the other APIs with the same behaviour, checking how each callable is actually invoked rather than trusting the type alias:

| API | Callable | How it's dispatched |
|---|---|---|
| `callback_on_mock_put` | `callback` | `AsyncMock` side effect awaits a coroutine result |
| `callback_on_mock_execute` | `callback` | same |
| `soft_signal_rw` / `soft_signal_r_and_setter` / `SoftSignalBackend` | `getter`, `setter` | `bluesky.utils.maybe_await` |
| `soft_command` / `SoftCommandBackend` | `command_cb` | `inspect.isawaitable` check in `execute()` |

`soft_command` already said "may be sync or async"; none of the rest did.

## Changes

- Each `:param:` now states sync-or-async and that an async callback is awaited before the put/execute/set completes, with sync + async `:example:` blocks.
- `SoftCommandBackend` had no param docs at all — added them.
- `docs/tutorials/writing-tests-for-devices.md`: the "Setting side effects on mocks" section gets an async example and a note that the return-value-becomes-readback rule holds either way.
- `docs/how-to/how-to-use-soft-signals.md`: Case B gets an all-async variant plus a note that `getter`/`setter` are independent.

`callback_on_mock_execute` had no async-callback test, so `test_soft_command_mock_side_effect_overrides_func` is now parametrized over both rather than documenting untested behaviour. The other three already had async tests.

**Not sync-or-async**, checked and left alone: `set_derived` on derived signals is awaited directly so it's async-only; `wait_for_value`'s `match`, the EPICS `wait=` predicate and adcore's `array_description` are sync-only.

## A note on `:example:` rendering

While checking the built HTML I found that `:example:` doesn't nest the way the syntax suggests — because the fence sits at the same indent as the field, MyST emits an empty `<dd>` and the code block lands as a sibling after the `</dl>`. It reads fine (bold **Example:** label, code beneath), and all 14 pre-existing `:example:` sites do the same, so these new ones match. Indenting the fence fixes it properly (verified — nests inside the field, no empty `<p>`), but that's a repo-wide change for another PR.

`tox -e docs` builds clean (that env runs `--fail-on-warning` with `nitpicky = True`), `pyright src` is 0 errors, `pytest tests/unit_tests/core` passes.

## AI Disclaimer


jwlodek (approved):
LGTM. I actually ran into this earlier this week, and found the issue about `get_mock_value`, where it was noted that `callback_on_mock_put` worked with async functions.

PR #1399, opened 2026-08-18
Mocksignal filter

### LINKED ISSUE #1348: Add `callback_filter` hook to `MockSignalBackend`
PRs like #1235 and #1342 both have to patch `observe_value` in order to simulate "takes a forever to callback on the current value". We would like `MockSignalBackend` to support this usecase.

Initially I thought we could put an async hook into the callback so initial values were delayed. However the callback chain is sync, so can't do this. Instead I suggest we add a `set_callback_filter(signal: SignalR[SignalDatatypeT], filter: Callable[[SignalDatatypeT], SignalDatatypeT | None]` that can modify or veto values from making it up to the person who subscribed to the Signal. For the use cases here we would:

```python
from ophyd_async.core import set_callback_filter

# Drop every monitor update
set_callback_filter(sig, lambda v: None)
with pytest.raises(TimeoutError):
    # Now we should timeout waiting for the initial value
    wait_for_value(sig, 32)
```

## Acceptance Criteria
- Function, docs, tests written
- All patching of `observe_value` replaced with it

tomkane-dls:
Add set_callback_filter for mock signals

Fixes #1348 

Adds a hook to MockSignalBackend that intercepts values before they reach subscribers, replacing observe_value patching used to simulate signals whose monitors never deliver

- set_callback_filter(signal, filter) in ophyd_async.core. The filter runs on each value bound for a subscriber: return a replacement, or None to drop it. Dropping every update (including the initial value emitted at subscribe time) makes waits time out for real
- documents the two non-obvious behaviours: set the filter before anything subscribes (a live cache serves the first value straight past it), and use get_value(cached=False) to read the real value
- observe_value patching replaced — removed in test_acquire_logic.py, test_adpilatus.py, and test_pmac_trajectory.py, along with their duplicated bad_observe_value stubs. test_signal.py keep patch as it tests observe_value's own timeout-race behaviour, not a dead monitor, so the filter doesn't apply

- Replacing the patches made timeouts genuinely elapse rather than raise instantly - two hardcoded timeouts are now injectable; ADAcquireLogic, PmacTrajectoryFlyableLogic._check_profile_status

### DIFF
--- a/src/ophyd_async/core/_mock_signal_utils.py
+++ b/src/ophyd_async/core/_mock_signal_utils.py
@@ -5,7 +5,7 @@ from unittest.mock import AsyncMock, Mock
 
 from ._command import Command, CommandConnector, MockCommandBackend, MockExecuteCallback
 from ._device import Device, DeviceMock
-from ._mock_signal_backend import MockPutCallback, MockSignalBackend
+from ._mock_signal_backend import MockCallbackFilter, MockPutCallback, MockSignalBackend
 from ._signal import Signal, SignalConnector, SignalR
 from ._signal_backend import SignalDatatypeT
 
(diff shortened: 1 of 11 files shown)


src/ophyd_async/core/_mock_signal_utils.py
```diff
@@ -194,6 +194,38 @@ async def on_put(value: float) -> None:
     return _unset_side_effect_cm(backend)
 
 
+@contextmanager
+def _unset_callback_filter_cm(backend: MockSignalBackend):
```
shihab-dls: nit: super tiny nit but `_cm` seems redundant on the name

PR #1400, opened 2026-08-18
Make readable format changeable at runtime

### LINKED ISSUE #1394: Allow for changing standard readable format at run-time

coretl:
Makes the `StandardReadableFormat` of a child a runtime setting rather than something
fixed at class-definition time, so a Device can be switched between techniques without
being redefined. This is the ophyd v1 `kind` capability.

Draft, for @jwlodek to comment on — particularly the load/save design in the last section.

Fixes #1394

## Why the registry had to change

`add_readables(devices, format)` appended bound methods into four tuples:

```python
case StandardReadableFormat.CONFIG_SIGNAL:
    self._describe_config_funcs += (signal.describe,)
    self._read_config_funcs += (signal.read,)
```

Once appended, nothing recorded which device a contribution came from, so it could not be
found, removed or re-formatted. The registry is now a `{device: format}` dict with the
callables derived on demand, so `set_readable_format(device, format)` is a single
assignment and `format=None` drops the device. `get_readable_formats()` returns a copy.

Keying by `Device` is identity keying: `Device` does not override `__eq__`, and while
`DeviceVector`/`DeviceMap` inherit a value based `__eq__` from `Mapping`, they hash by
`id()`, which is injective over live objects — so two distinct devices never share a hash
and `__eq__` is never reached. It is a `@cached_property`, matching `Device._child_devices`,
because `AreaDetector` and `add_children_as_readables` register children *before* calling
`super().__init__()`.

`add_readables` stays as a thin wrapper, so all 85 in-tree call sites and every downstream
Device are unaffected. It gains "set" semantics: registering the same device twice now
replaces rather than contributing twice.

Deriving on demand also fixes staging, which previously captured `signal.stage` at
registration — `stage()` now walks the current registry, so a format changed between runs
takes effect on the next one.

## `StandardDetector` is a `StandardReadable`

Needed so `set_readable_format` works on a detector at all. `read()`, `describe()` and
`hints` merge the registered children on top of whatever the data logics produce, and
`add_config_signals` becomes a deprecated wrapper.

Following @jwlodek's last comment on #1395, plugins are **not** registered automatically.
`AreaDetector` registers only the driver's `acquire_time`/`acquire_period` plus whatever
the caller passes as `config_sigs`; opting a plugin in is a `set_readable_format()` call
at the call site. That is the explicit opt-in @coretl asked for, now expressible per
plugin and reversible at runtime.

## One `standard_logic` property

`Motor` had to expose the same logic object as both `movable_logic` and `flyable_logic`.
Each mix-in now declares `standard_logic` with its own required type, so one
implementation satisfies both and pyright checks it against each base independently — the
intersection is expressed by the logic class's own MRO. `movable_logic` and
`flyable_logic` are removed rather than deprecated — see Breaking changes.

This also fixes an abstractness hole: `functools.cached_property` does not forward
`__isabstractmethod__`, so `StandardMovable.__abstractmethods__` was empty and a Device
that forgot its logic failed with `AttributeError: 'NoneType' object has no attribute
'readback'` from inside `Device.__init__`. `abstract_cached_property` sets the flag on
the descriptor, so it is a `TypeError` at instantiation instead.

## Load/save — one file, one apply

Formats are part of a Device's state and change on the same cadence as values, so
`store_settings` stores both and `apply_settings` applies both. Values stay flat; formats
go under a reserved key:

```yaml
energy: 7.0
temperature: 20.0
<READABLE_FORMATS>:
  <ROOT>:
    energy: CONFIG_SIGNAL
    temperature: CONFIG_SIGNAL
  stats:
    stats.total: HINTED_UNCACHED_SIGNAL
```

**No version marker.** A file written before formats existed simply lacks the key, which
means "change no formats". So existing saved files load unchanged, with no shim and no
deprecation warning. Each part being optional also makes a hand-written formats-only
profile work, switching technique without writing a single value to hardware.

**Applying merges rather than replaces.** A child the file does not mention keeps whatever
format it has, matching how values behave — apply an old file to a Device that has gained
a signal since, and the new signal is left alone rather than silently unregistered.
Dropping a child is an explicit `null`; nothing writes those for you, since storing records
only what is currently registered. The trade is that round-tripping A -> B -> A no longer
drops automatically.

**The reserved keys cannot collide.** `<` cannot begin a Python identifier, so no attribute
assignment can produce a colliding path and no store-time guard is needed. `<` is also not
a YAML indicator, so unlike `*FORMATS*` or `%FORMATS%` the key needs no quoting. `<ROOT>`
replaces `""` as the root owner for the same reason, and because an empty-string key makes
PyYAML emit its hard-to-read `? '' :` explicit key form. `retrieve_settings` drops *every*
`<RESERVED>` key rather than only the one it understands, so a key written by a newer
version is ignored rather than reported as an unknown signal, and adding one later needs
no migration and no constant reserved up front.

**Formats are two levels while values are flat**, because a value is intrinsic to one path
whereas a format is a fact about an (owner, child) *pair* — and the same child can be
registered on two owners with different formats, which #1395 creates directly once
`NDStatsIO` declares `total` on itself while a detector may also register `stats.total`.
Recording the owner rather than inferring it also means the layout needs no "must be a
direct child" rule, so `AreaDetector` keeps registering `driver.acquire_time` and
`DeviceVector` elements keep working — no dependency on #1395, and no second migration
later.

**`Settings.partition` copies formats to both halves.** `apply_panda_settings` partitions
and applies only the halves, never the original `Settings`, so putting formats on one half
would let that plan silently decide whether they were restored at all. Applying them more
than once touches no hardware and is idempotent. `apply_settings` applies formats after
values, so a failed value write leaves the readable registry untouched.

## Breaking changes

- **`ConfigSignal` and `HintedSignal` are removed.** Use `StandardReadableFormat.CONFIG_SIGNAL`,
  `.HINTED_SIGNAL` and `.HINTED_UNCACHED_SIGNAL`. They have been deprecated since well
  before this change, and they worked by comparing equal to their target — which relied on
  the `match` statement running inside `add_readables`. Deriving the callables on demand
  moves that comparison to the first `read()`, so keeping them would have fired the warning
  somewhere unhelpful.
- **`movable_logic` and `flyable_logic` are removed**, replaced by `standard_logic`. No
  deprecation shim: a subclass that still *defines* `movable_logic` would override the shim
  and leave `standard_logic` unimplemented, so it would fail at instantiation regardless.
- **A `StandardDetector` with no hinted fields now reports `hints` as `{}`** rather than
  `{"fields": []}`, because it no longer special cases what `StandardReadable` already
  does. `Hints` is a `total=False` TypedDict and `StandardReadable` already behaved this
  way, so consumers using `.get("fields", [])` are unaffected.
- **Stored settings files gain a `<READABLE_FORMATS>` key.** Reading an old file is
  unaffected; anything parsing a *newly written* file needs to expect it.

## Deliberately out of scope

- **The `DetectorLogic(FlyableLogic)` restructure** — `StandardDetector` still
  re-implements `prepare`/`kickoff`/`complete` and its own context threading rather than
  inheriting `StandardFlyable`. #1367 rewrites `core/_detector.py` heavily, so doing it
  here would make that rebase painful. Worth its own PR and ADR, since it revisits ADR
  0017's no-back-reference rule.
- **`StandardTriggerable`** — only two demo users, and `StandardDetector.trigger()` is
  watchable and bound to the prepare context so it could not use it anyway. The name would
  also collide with `DetectorTriggerLogic` (which configures trigger *modes*) and
  `TriggerableCommand`.
- **Anything unique to #1395 or #1367** — no `prepare_bounded`, and no AD plugin or driver
  IO classes turned into `StandardReadable`s. Both PRs should rebase cleanly on top.

## Notes for review

- ADRs are numbered **0021** and **0022**, leaving 0020 for the unmerged ADR in #1367.
- **Review comments addressed** in 5b83a833b, with the threads resolved. `standard_logic`
  vs `logic` is left open as it was put to the room — my vote is `logic`, and the rename is
  free right now since nothing downstream uses either name yet.
- **@oliwenmandiamond raised** whether class-declared formats should be immutable by
  default. Replied on the thread rather than acting on it: static-by-default would block
  #1394's own motivating case, since that is precisely a class-declared format being
  changed per technique. Suggested a "what differs from the class defaults" diff as a way
  to get the discoverability without removing the capability.
- **#1402** came out of reviewing this: a readable registered *outside* a Device's tree is
  reported by `read()` but never captured by `store_settings`, so it silently fails to
  round-trip. Pre-existing, not introduced here. `walk_readable_formats` warns and skips
  such a child, and that warning now says the value is not stored either. The fix needs a
  decision about where the check lives, since annotations are applied before children are
  parented — so it is deliberately left to its own PR.
- Tests that asserted on `_read_funcs` / `_has_hints` are rewritten onto the public verbs.
  Those assertions were already against the testing conventions in `CLAUDE.md`, and the
  derived-on-demand design makes them impossible rather than merely discouraged.
- `tox -e docs` already failed on `main` with three warnings. Retargeting the
  `write-a-flyer` literalinclude fixes one; the other two are TypeVars documented on
  `StandardFlyable` in #1359, added to the existing nitpick ignore list. `tox -p` is green
  on this branch.

## AI Disclaimer


oliwenmandiamond:
So this looks very promising and welcome the additional flexibility this is providing. However I have one question / concern. 

So for `StandardReadable`, say I have the following device

```python

       def __init__(self, prefix: str, name: str = ""):
            with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
                 self.sig_a = epics_signal_rw(float, prefix + "A")
             self.sig_b = epics_signal_rw(float, prefix + "B")
```
So I have defined `sig_a` as the hinted signal. However, at runtime I can now change to be `sig_b`. When looking at the class, this could be quite confusing that this has changed at runtime. 

I think changing it dynamically at runtime defiantly has a place and of course we have the use case mentioned above, but I would like one or both of these safeguards added.

**1. Make `with self.add_children_as_readables` not dynamic.**
I think if you write a class definition, it should be clear the read, config, and hinted signals are fixed and should not be changable by default unless specified otherwise.

I think this can be done by maybe passing in a flag which by default makes it not dynamic e.g `dynamic = False` or `static = True` by default and can be overwritten.

Or we could make it so `self.add_children_as_readables()` is the static case entry point and `set_readable_format` is the dynamic entry point. If you try to override the hinted signal of a device which was added via self.add_children_as_readables() for example, it raises error because it is intentionally static. 

If you want something dynamic, it must be explicit and obviously defined it is dynamic.

**2. Make `StandardDyanmicReadable`**
We could instead make a sub class which adds the flexibility of readables to be dynamic, then it is very obvious when looking at a device if the signals should be static or dynamic via the type of readable the device implements.

I think my preference is option 1 though. Any thoughts on this? @jwlodek @coretl

docs/explanations/decisions/0022-single-standard-logic-property.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Decision
+
+Each mix-in declares the **same** attribute name, `standard_logic`, with **its own**
+required type:
+
+```python
+class StandardMovable(Device, ..., Generic[SignalDatatypeT]):
+    @abstract_cached_property
+    def standard_logic(self) -> MovableLogic[SignalDatatypeT]: ...
```
coretl, line 38: Do we like `standard_logic`? Or would we prefer `logic` (and accept the increased likelihood of name collision with a signal)?

docs/how-to/change-readable-format.md
```diff
@@ -0,0 +1,108 @@
...
+energy: 7.0
+temperature: 20.0
+<READABLE_FORMATS>:
+  <ROOT>:
+    energy: CONFIG_SIGNAL
+    temperature: CONFIG_SIGNAL
+  stats:
+    stats.total: HINTED_UNCACHED_SIGNAL
+```
+
```
coretl: This entire section should be in store-and-retrieve.md

src/ophyd_async/core/_detector.py
```diff
@@ -455,9 +464,10 @@ async def get_trigger_deadtime(
     @AsyncStatus.wrap
     async def stage(self) -> None:
```
coretl: Rather than replacing the superclass stage, would this be better off added to self._stage_funcs?

src/ophyd_async/core/_flyable.py
```diff
@@ -165,37 +169,48 @@ def __init__(self, *args, **kwargs) -> None:
...
+        raise NotImplementedError
+
+    # Back compat - delete before 1.0
+    @property
+    def flyable_logic(self) -> FlyableLogic[PrepareT, CtxT]:
+        warnings.warn(
+            DeprecationWarning("Use `standard_logic` instead of `flyable_logic`"),
+            stacklevel=2,
         )
+        return self.standard_logic
```
coretl: Does this back compat add anything? If someone implements `flyable_logic` then it will no longer be called. If this adds nothing then delete it.

src/ophyd_async/core/_movable.py
```diff
@@ -142,20 +142,35 @@ class StandardMovable(
...
+        raise NotImplementedError
+
+    # Back compat - delete before 1.0
+    @property
+    def movable_logic(self) -> MovableLogic[SignalDatatypeT]:
+        warnings.warn(
+            DeprecationWarning("Use `standard_logic` instead of `movable_logic`"),
+            stacklevel=2,
+        )
+        return self.standard_logic
```
coretl: Same as `flyable_logic`

src/ophyd_async/core/_flyable.py
```diff
@@ -241,5 +256,20 @@ async def complete(self) -> AsyncIterator[WatcherUpdate]:
...
+    `StandardFlyable.standard_logic` is abstract, so the class itself cannot be
+    instantiated; `FlyableLogic.with_device` needs something concrete to attach
+    a logic object to.
+    """
+
+    @cached_property
+    def standard_logic(self) -> FlyableLogic[PrepareT, CtxT]:
+        raise NotImplementedError(
+            f"{self.name}: no logic set; create this via FlyableLogic.with_device()"
+        )
```
coretl: Prefer an `__init__` method taking a `FlyableLogic`, then this can be concrete...

src/ophyd_async/core/_readable.py
```diff
@@ -9,11 +10,47 @@
...
+    serialised and reported later. The deprecated markers only announce
+    themselves when compared, so the comparison has to happen here, at
+    registration, rather than each time a verb is called.
+    """
+    if isinstance(format, StandardReadableFormat):
+        return format
+    for member in StandardReadableFormat:
+        if format == member:  # _WarningMatcher.__eq__ raises the DeprecationWarning
+            return member
+    raise TypeError(f"{format} is not a StandardReadableFormat")
```
coretl: Remove HintedSignal and ConfigSignal as part of this, they've been deprecated for long enough. Note in Breaking changes section of PR description

coretl:
> I think my preference is option 1 though. Any thoughts on this? @jwlodek @coretl

This is tricky. From a simplicity point of view I would prefer to either have everything be changeable, or nothing. I think ophyd sync was always mutable, so I'd err on that side.

That said, I don't have enough experience here, so I'll leave @jwlodek to comment

docs/how-to/change-readable-format.md
coretl: Moved to `store-and-retrieve.md` in 5b83a833b, rewritten for the merge semantics. `change-readable-format.md` now just cross-references it.

src/ophyd_async/core/_detector.py
coretl: Yes, much better — done in 5b83a833b. `stage()`/`unstage()` are no longer overridden; the detector registers `_stage_detector`/`_unstage_detector` into `_stage_funcs`/`_unstage_funcs` the way `StandardFlyable` already does.

I gave `read()`/`describe()`/`hints` the same treatment, since they had the same problem — each override was re-deriving the registered children's contributions by hand. `StandardReadable` now has `_extra_funcs_for(verb)` and `_extra_hint_sources()` for contributions from outside the registry, and the detector overrides only those.

One visible consequence: a detector with no hinted fields reports `hints` as `{}` rather than `{"fields": []}`, because it no longer special cases what `StandardReadable` already does. Noted in Breaking changes.

src/ophyd_async/core/_flyable.py
coretl: You're right that it adds nothing — a subclass that still *defines* `flyable_logic` overrides the shim and leaves `standard_logic` unimplemented, so it fails at instantiation anyway. It only ever helped code that *reads* the attribute. Deleted in 5b83a833b.

src/ophyd_async/core/_movable.py
coretl: Deleted too, same reasoning — 5b83a833b.

src/ophyd_async/core/_flyable.py
coretl: Done in 5b83a833b — `_EphemeralFlyable.__init__(logic, name="")` sets `self._logic`, and `standard_logic` is a plain `cached_property` returning it. `with_device` is now a one-liner and there's no assigning into a `cached_property` from outside.

src/ophyd_async/core/_readable.py
coretl: Removed in 5b83a833b, along with `_WarningMatcher` and the `_normalise_format` helper that existed only to resolve them. Added to a Breaking changes section in the PR description.

Worth noting they were about to get quietly worse anyway: they announce themselves by comparing equal to their target, which relied on the `match` running inside `add_readables`. Deriving the callables on demand moves that comparison to the first `read()`, so the warning would have fired somewhere unhelpful.

docs/explanations/decisions/0022-single-standard-logic-property.md
coretl: Leaving this one open as you put it to the room, but my vote is `logic`.

It reads better everywhere it appears — `motor.logic.velocity`, `self.logic.readback` — and the rename is free right now: `standard_logic` is new in this PR and nothing downstream uses it, while `movable_logic`/`flyable_logic` are being deleted anyway. If we are going to change it, this is the cheapest moment there will ever be.

On the collision risk, I think it is narrower than it first looks. The name only exists on Devices that mix in `StandardMovable`/`StandardFlyable` — a plain `StandardReadable` has no `logic` attribute at all — so the clash needs a Device that both composes a logic object *and* wants a child signal called `logic`. Plausible for something PLC-flavoured, but not common, and it is the kind of thing that shows up immediately rather than silently: the child assignment lands on the `cached_property` name and the Device stops working on first use.

`standard_logic` also ages slightly badly in one direction — it is named after the `Standard*` family rather than after what it holds, so it reads oddly if a non-`Standard` mixin ever wants one.

Happy either way, and it is a one-commit mechanical change whenever someone says go.

coretl:
@oliwenmandiamond thanks — this is the right thing to worry about, and I want to lay out one tension before @coretl and @jwlodek weigh in, because I think it decides between your two options.

**Static-by-default would block the case the PR exists for.** #1394's motivating example is a monochromator whose `energy` is declared in the class as `CONFIG_SIGNAL`, and which one technique wants hinted instead. Under option 1, that signal is locked precisely *because* it was declared in the class, so the beamline would have to get the device class changed before it could switch technique. Option 2 has the same shape one level up: the author of a generic `AreaDetector` cannot know that some beamline wants to retune it, so the decision to allow retuning ends up being made by whoever is furthest from the technique.

So I think any "static by default" rule needs an answer to: how does a beamline change a class-declared format without editing the class? If the answer is "declare it dynamic in the class", we have moved the decision to the wrong person; if it is "call an escape-hatch method", the safeguard is advisory rather than enforced, which may still be worth having but is a different thing.

**On the readability concern itself, I think you are right and it is worth addressing another way.** Two things in the PR already help a bit, and I would rather lean on those than on a lock:

- Formats are now part of stored settings, so the *file* is the record of what a technique changed, and it is diffable. Reading the class plus the profile tells you the whole story.
- `get_readable_format(device)` reports the current format, so what a live Device is doing is inspectable rather than having to be inferred.

What is missing is the diff between the two — "what has this Device had changed away from its class defaults". That is cheap to add (capture the registry after `__init__` and compare) and gives you the discoverability without taking away the capability. If that would address your concern I am happy to add it here.

The one variant of your option 1 I would push back on hardest is making `add_children_as_readables` static while `set_readable_format` stays dynamic, because they write to the same registry — so the rule would have to be per-entry, and the error would fire at the call site of a perfectly reasonable technique switch.

Deferring to @coretl and @jwlodek on whether a safeguard is wanted at all.

## AI Disclaimer

jwlodek:
Will give a more detailed review when I have a bit more time, but at a cursory glance, this looks good to me.

> I think ophyd sync was always mutable, so I'd err on that side.

That's correct, and I think I'd prefer the behavior to stay like this. As we are planning to have more and more sync & async side-by-side, the fewer behavior differences/pitfalls we have the better from an NSLS-II perspective.

oliwenmandiamond:
> The one variant of your option 1 I would push back on hardest is making `add_children_as_readables` static while `set_readable_format` stays dynamic, because they write to the same registry — so the rule would have to be per-entry, and the error would fire at the call site of a perfectly reasonable technique switch.

I personally think having static configuration is still a good idea. If something can be dynamic, it should be opt in rather than opt out. Were now going from one extreme of it being static to the other of it being fully dynamic. I think a hybrid approach is best but understand the additional complexity with this (though I don't think it would be too difficult, you just need to wrap each bit with an associated boolean of if this readable added is static or not).

@jwlodek I am not familiar with ophyd as not done any development or used the API much compared to ophyd-async. I agree parity is probably a good thing, but why does ophyd-async have to match ophyd. Can ophyd not match ophyd-async? 

> What is missing is the diff between the two — "what has this Device had changed away from its class defaults". That is cheap to add (capture the registry after `__init__` and compare) and gives you the discoverability without taking away the capability. If that would address your concern I am happy to add it here.
> 

However, if we are set on making it fully dynamic I think adding an additional method would help to address this. Name to be determined, but `reset_to_defaults()` will reset any runtime configuration that has changed to what the class originally defined would be good. Then any devices that need to alter the Readable methods at runtime can reset the state back once finished

src/ophyd_async/core/_readable.py
```diff
@@ -194,6 +282,46 @@ def add_children_as_readables(
...
+                _as_signal_r(device)
+        kept = tuple((d, f) for d, f in self._readables if d is not device)
+        self._readables = kept if format is None else (*kept, (device, format))
+
+    def get_readable_format(self, device: Device) -> StandardReadableFormat | None:
+        """Return the format of a child Device, or `None` if it does not contribute."""
+        for registered, format in self._readables:
+            if registered is device:
+                return format
+        return None
```
coretl: This only seems used in tests, delete it...

src/ophyd_async/core/_readable.py
```diff
@@ -194,6 +282,46 @@ def add_children_as_readables(
...
+
+    def readable_children(self) -> Iterator[tuple[Device, StandardReadableFormat]]:
```
coretl: Prefer `get_readable_formats() -> dict[Device, StandardReadableFormat]`

jwlodek:
@oliwenmandiamond I'm not opposed to modifications to ophyd as well to align the two projects, just at this point we have 20+ beamlines with extensive ophyd v1 based codebases, so changes in behavior there requires  a lot of intervention from our pov, vs ophyd async is still pre 1.0 and only extensively used at three beamlines for now, so a bit easier to make changes imo. A lot of these ophyd v1 codebases we manage make extensive use of the dynamic adjustment of the equivalent of `Format`, hence the desire to support it with async as well.

I like the idea of `restore_formats` that you propose by the way. Allows for quickly reverting any live changes, while still allowing for dynamic changes. That's something we could probably backport to ophyd v1 as well.

src/ophyd_async/core/_readable.py
```diff
@@ -52,31 +73,10 @@ def __call__(self, parent: Device, child: Device):
...
-
-
-ConfigSignal = _compat_format("ConfigSignal", StandardReadableFormat.CONFIG_SIGNAL)
-HintedSignal: Any = _compat_format("HintedSignal", StandardReadableFormat.HINTED_SIGNAL)
-HintedSignal.uncached = _compat_format(
-    "HintedSignal.uncached", StandardReadableFormat.HINTED_UNCACHED_SIGNAL
+#: The formats whose devices take part in `stage()`/`unstage()`, if they are
+#: `Stageable` at all.
+_STAGED_FORMATS = frozenset(
+    {StandardReadableFormat.HINTED_SIGNAL, StandardReadableFormat.CHILD}
```
jwlodek, line 79: Do we need to include `HINTED_UNCACHED` here?

jwlodek (approved):
This all looks good to me, will try and take it for a spin later this week.

src/ophyd_async/core/_readable.py
coretl: No, the uncached signals don't need monitoring on stage, so shouldn't be in this list

coretl:
> Do we like `standard_logic`? Or would we prefer `logic` (and accept the increased likelihood of name collision with a signal)?

I think I prefer `logic`. Any other opinions before I make the change?

> I like the idea of restore_formats that you propose by the way. Allows for quickly reverting any live changes, while still allowing for dynamic changes. That's something we could probably backport to ophyd v1 as well.

I've put in `reset_readable_formats` to do this. But then I thought, is this the right model? Should we think of a way to put this in per-scan? Maybe we make `stage` do the stamping of formats then always restore them in `unstage`. What do you think?

EDIT: I think `reset_readable_formats` as written is the right model. A modal `set_readable_format` that behaves differently before and after `stage` seems too surprising to me.

Also I think an environment variable that optionally stops `set_readable_format` being called after `__init__` would be a good idea to mitigate @oliwenmandiamond's concerns

canismarko:
@coretl Here is how we're using `prepare_single()` right now. 

I have a TetrAmm electrometer (based on QuadEM) for our split ion chamber set.

Right now I use it only for step scanning, so I implemented [`prepare_single()`](https://github.com/spc-group/haven/blob/6632e18f27014724f3fd517d7c6c26454806fc59/src/haven/devices/detectors/split_ion_chamber.py#L61) to just measure each point.

Once `prepare_bounded()` is available, I could (and probably should) write that to use the time series plugin. So far, though, I haven't figured out how to get the time series plugin to acquire properly.

jwlodek:
@canismarko I believe after StandardDetector becomes a StandardReadable, you can achieve what that data provider is doing by marking the appropriate signals with `Format.HINTED_UNCACHED_SIGNAL` or `Format.UNCACHED_SIGNAL` (if you don't want hints) in your IO class. Then `StandardReadable` will handle making the data key on describe, and generating the readings dict on `read`.

canismarko:
Thanks, @jwlodek

> marking the appropriate signals with Format.HINTED_UNCACHED_SIGNAL or Format.UNCACHED_SIGNAL

That sounds great!

docs/explanations/decisions/0022-single-standard-logic-property.md
davidpcls: We still have things like DetectorDataLogic, datalogic, etc:
```
    _trigger_logic: DetectorTriggerLogic | None = None
    _acquire_logic: DetectorAcquireLogic | None = None
    _data_logics: Sequence[DetectorDataLogic] = ()
```
As in the standard detector. Would those also get renamed to `logic` for consistency? If not, what differentiates them from "logic"?

coretl: There would be a new `DetectorLogic` that had `trigger_logic`, `acquire_logic`, `data_logics`, etc as children. This has all got rather tangled up with #1367 and #1395, so I'm going to finish the work, then split it back into separate commits on each individual PR so they're actually reviewable. I've also removed claude's permissions so it can't make comments in my name on repos outside my personal area...

PR #1401, opened 2026-08-18
Add pixel mask link to odin

### LINKED ISSUE #115: I15-1: Add strings in the nexus file to where eiger metadata can be found
For analysis we need to know metadata for the eiger, such as the pixel mask. There is currently no way to VDS link to these in Bluesky. Ultimately the answer to this is probably something like https://github.com/DiamondLightSource/fastcs-odin/issues/115 but this has complexities. As a quick workaround we can add into the file a string that describes where the data can be found in the metafile

## Acceptance Criteria
- There is a string in the nexus file for where to find metadata

DominicOram:
Fixes https://github.com/DiamondLightSource/crystallography-bluesky/issues/115

See https://github.com/DiamondLightSource/fastcs-odin/issues/115 for a more correct solution but this gives us a path forward in the short term.

### DIFF
--- a/src/ophyd_async/fastcs/eiger/_detector.py
+++ b/src/ophyd_async/fastcs/eiger/_detector.py
@@ -1,4 +1,10 @@
-from ophyd_async.core import PathProvider, SignalR, StandardDetector, TriggerableCommand
+from ophyd_async.core import (
+    PathProvider,
+    SignalR,
+    StandardDetector,
+    TriggerableCommand,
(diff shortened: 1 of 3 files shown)


shihab-dls (approved):
Sorry for the delay. This seems perfectly fine as a temporary fix. Alternatively, we can provide you a soft signal at the `fastcs-eiger` level so we don't create it here, but given this should all be removed with a proper fix, I'll approve.

PR #1408, opened 2026-08-27
Detector stack: readable formats, detector-as-readable, bounded tier, flyable

### LINKED ISSUE #1309: Discuss `chunk size` configuration

### LINKED ISSUE #1394: Allow for changing standard readable format at run-time

coretl:
**Draft, for CI only. Supersedes nothing, closes nothing, and is not for review as a
single PR** — it is #1400 + #1403 + #1367 + the StandardFlyable rework built as one linear
branch so CI can exercise the union. The intended split is four PRs at the cut points
below, each of which is green locally on its own.

None of #1400, #1395, #1403 or #1367 has been touched.

## The four cut points

| | branch | commits | content |
|---|---|---|---|
| G1 | `detector-stack-1-readable-formats` | 15 | Runtime `StandardReadableFormat`, load/save, one `logic` property, ADRs 0021+0022. **No detector.** `Fixes #1394` |
| G2 | `detector-stack-2-detector-readable` | +11 | `StandardDetector` is a `StandardReadable`; AD IO classes declare their own formats; the `prepare_single` data logic tier goes. Supersedes #1395/#1403 |
| G3 | `detector-stack-3-bounded-tier` | +17 | Bounded `DetectorDataLogic` tier, one `make_data_provider`/`start` split, period plumbing, stats TS, ADR 0020. `Fixes #1309 #1364` |
| G4 | `detector-stack-4-flyable` | +1 | `StandardDetector` inherits `StandardFlyable`, ADR 0023 |

Each boundary passes `pytest tests/unit_tests`, `pyright src --pythonpath`, `ruff`, and
`tox -e docs`; the tip passes all four `tox` envs.

## What changed in response to review

@coretl's review on this branch cut the scope back, and the four branches have been
rewritten in place rather than patched on top, so each gate is coherent on its own.

**The single-event data logic tier is gone (G2).** A `StandardDetector` is a
`StandardReadable`, so a source producing one value per event is a registered signal, not a
data logic: `set_readable_format` puts it in `read()` for any number of collections, where a
`prepare_single` logic dropped itself with a warning as soon as a scan asked for more than
one. `prepare_single`, `ReadableDataProvider` and `SignalDataProvider` are deleted, and
`DetectorDataLogic` has two tiers, not three. Rationale is recorded in ADR 0021.

**A data logic says what it would make before anything starts it (G3).** `prepare_bounded`
and `prepare_unbounded` merge into one `make_data_provider`, which describes the data without
touching hardware, plus a `start` that does the writes and is called only for the providers
the detector settles on. Which tier a logic is becomes the type of the provider it returns,
so the tier enum, the override detection and `_tier_can_serve` all go; a logic sits a scan
out by returning `None`, silently, which also replaces the `contributes()` hook the modes
work was going to need. `ADHDFDataLogic` and `StatsTimeSeriesDataLogic` gain an
`enable_callbacks` flag, defaulting to today's force-on, which when `False` follows whatever
the plugin is set to.

**`DetectorLogic` owns the three logics and holds no back-reference (G4).** The trigger,
acquire and data logics are now children of one `DetectorLogic`, which a detector builds in
`__init__` and returns from `logic`:

```python
class MyDetector(StandardDetector):
        self.driver = MyDriverIO(prefix)
        self._logic = DetectorLogic(
            MyTriggerLogic(self.driver),
            MyAcquireLogic(self.driver),
            publish_collect_methods=self._publish_collect_methods,
        )
        super().__init__(name=name)

    @cached_property
    def logic(self) -> DetectorLogic:
        return self._logic
```

`StandardDetector` itself takes no `__init__` arguments, since a `Standard*` class that did
would interfere with multiple inheritance; `StandardDetector.with_logics(*logics, name=...)`
builds both the logic and a Device for an ad-hoc detector. `add_detector_logics()` is
removed rather than kept as a forwarder. The orchestration moved with the logics, so ADR
0017's rule against back-references now holds unqualified. The two things the logic cannot
get from its own fields are handled narrowly: the datakey namespace is pushed in by
`set_name`, and the collect verb through a required `publish_collect_methods` callback,
called from prepare so it follows what the logics actually produce.

**Mixing a file writer and a finite buffer is allowed when the file covers it (G4).** Where
every datakey a bounded logic would produce is also written by the stream assets — an
areaDetector stats total pulled into the HDF file as an NDAttribute — the durable copy wins
and the buffer sits the scan out, unarmed. Uncovered keys still raise, naming them.

**`events_to_kickoff` and multi-kickoff are gone (G4).** One `prepare()` serves one
kickoff/complete cycle, like every other flyer; `FlyableLogic.supports_multiple_kickoffs`
was never added. Note this removes `events_to_kickoff`, which is on `main` today — nothing
in tree drove it, but downstream is worth a thought.

**Docstrings.** The multi-paragraph docstrings that argued design decisions have been cut
back to what the code does, with the reasoning left in the ADRs.

**Docs.** ADR 0017 gained the forward reference to 0022; the `reset_readable_formats` note
says what it actually affects; `ROOT_PATH = "<ROOT>"` became `ROOT_DEVICE_KEY =
"<ROOT_DEVICE>"` so the constant and the yaml key match; the sections describing a readable
detector moved from G1 to G2, which is the branch that makes them true.

## Behaviour changes

- **`kickoff()` twice without an intervening `prepare()` is now rejected**, where the old
  detector allowed it up to the prepared bounds. The RunEngine completes what it kicked off.
- `events_to_kickoff` is removed from the public API.
- Unprepared detector: `Prepare not run` → `<name>: prepare() must be called first`.
- A `StandardDetector` with no hinted fields reports `hints` as `{}` not `{"fields": []}`
  (from G2).
- Stored settings files gain a `<READABLE_FORMATS>` key, and the root device is `<ROOT_DEVICE>`
  (from G1).
- `add_detector_logics()` is removed; build a `DetectorLogic` in the subclass `__init__`.
- `DetectorDataLogic` implementations must be rewritten onto `make_data_provider`/`start`.
- A detector satisfies neither `WritesStreamAssets` nor `EventPageCollectable` until it has
  been prepared, since which one applies depends on what its data logics produce.

## Found along the way

- **#1367 is not bisectable.** `61ad2d204` deletes `PageableDataProvider.make_readings`
  without updating `test_stats_time_series.py`, which still calls it; the branch is red at
  that commit and green again at the next. Reproducible on `detector-tiers-impl`.
- **`tox -e docs` fails on `detector-tiers-impl`** — `NDWriterPluginT` has no
  `docs/conf.py` nitpick entry. Fixed in G3.
- **`docs/how-to/implement-ad-detector.md` documented `PluginSignalDataLogic`** after it had
  been deleted. Now a `set_readable_format` call.
- **#1403 already answers both of @thopkins32's review threads** on #1395: `c239654d0`
  converts the IO classes to `Format.CONFIG_SIGNAL` annotations, and `fe787ce0c` moves
  `WaitForPlugins` out of `AreaDetector`.
- **One deliberate departure from #1403**, needs @jwlodek: `64f4cd32b` narrows
  `NDROIIO.size_x/y/z` from `SignalRW` to `SignalR`. ROI size is settable on a real
  NDPluginROI and the narrowing is unrelated to that commit's purpose, so they were kept
  `SignalRW` with only the format annotation applied.

## Deferred

- The `_extra_funcs_for` private-API leak, and that it raises before the registry is
  consulted so registered children are unreadable until `prepare()`. Both still want doing.
- The rest of the modes work: computing the NDAttributes XML from the stats logics' datakey
  suffixes so the shadowing holds by construction rather than by the user remembering a plan
  stub, and giving `Settings` enough provenance that storing a mode does not freeze computed
  values into the file.

## AI Disclaimer


docs/tutorials/implementing-detectors.md
```diff
@@ -235,6 +237,18 @@ The [](#StreamableDataProvider) returned from `prepare_unbounded()` contains:
...
 
+#### Choosing a data logic tier
+
+A [](#DetectorDataLogic) implements exactly one of three `prepare_*` methods, and the detector picks a logic by which one it implements — there is no precedence to configure. Which tier fits depends on how the underlying source produces its data:
+
+- `prepare_unbounded(datakey_name, period)` — for a source that works for any number of collections, like a file writer. It returns a [](#StreamableDataProvider) and the detector exposes `collect_asset_docs`. `BlobDataLogic` uses this tier.
+- `prepare_bounded(datakey_name, num_collections, period)` — for a source that holds a *finite buffer* which must be sized before acquisition, like an areaDetector stats time series or a scaler whose array length is the frame count. It returns a [](#PageableDataProvider) that emits event pages, and the detector exposes `collect_pages` instead. See [](#StatsTimeSeriesDataLogic).
+- `prepare_single(datakey_name)` — for a source that can only ever produce one collection, like reading a scalar PV. It returns a [](#ReadableDataProvider). `PluginSignalDataLogic` uses this tier.
```
coretl: Now we inherit `StandardReadable` in `StandardDetector` we no longer need `prepare_single` or any `ReadableDataProvider`. Delete them.

src/ophyd_async/core/_detector.py
```diff
@@ -335,53 +367,194 @@ async def stop(self) -> None:
...
-class _PrepareCtx:
+class _FlyCtx:
+    """The context threaded prepare -> kickoff -> complete for a detector.
+
+    One dataclass rather than two, with the kickoff-produced fields `| None`
+    until `on_kickoff` fills them. That is already the in-tree convention --
+    `MotorFlyCtx(fly_info, status=None)` and `PmacFlyCtx(prepare,
+    trajectory_status=None)` -- and it avoids a third `KickoffCtxT` TypeVar on
+    `StandardFlyable` that all seven other flyers would have to carry. "Has
+    kickoff happened" is already a runtime `None`, never a type.
```
coretl: What is this stream of consciousness in a docstring? Stop doing that. Be concise.

src/ophyd_async/core/_detector.py
```diff
@@ -335,53 +367,194 @@ async def stop(self) -> None:
...
+
+
+class DetectorLogic(WatchableFlyableLogic[TriggerInfo, _FlyCtx]):
+    """Drives a `StandardDetector`'s fly scan through its three logic objects.
+
+    Unlike every other `FlyableLogic`, this one holds a back-reference to its
+    Device rather than the individual signals it needs. ADR 0017 rejected
+    back-references for `MovableLogic` -- "an early prototype passed the parent
+    device into the logic class, creating a circular dependency" -- and the
+    reasons it gave still hold for a logic wired from a handful of signals.
```
coretl: ditto, make concise

src/ophyd_async/core/_detector.py
```diff
@@ -335,53 +367,194 @@ async def stop(self) -> None:
...
-    trigger_info: TriggerInfo
-    data_providers: Sequence[StreamableDataProvider]
-    collections_written: int
-    collections_requested: int
-    is_last_kickoff: bool
+    supports_multiple_kickoffs = True
+    """A detector prepared for N events is kicked off a few at a time."""
+
+    def __init__(self, detector: "StandardDetector") -> None:
+        self.detector = detector
```
coretl: yuck. No back reference. move all the datalogics, triggerlogics, acquirelogics to be children of this class. Ditch events_to_kickoff and the ability to multi-kickoff.

docs/explanations/decisions/0020-bounded-data-logic-tier.md
coretl: Reorder these ADRs so they are added in commit order. Make sure links between them are clickable

docs/explanations/decisions/0020-bounded-data-logic-tier.md
```diff
@@ -0,0 +1,277 @@
...
+    async def prepare_unbounded(
+        self, datakey_name: str, period: float
+    ) -> StreamableDataProvider: ...
+```
+
+Both tiers are told the frame period, so either can size a chunk or a buffer against it.
+
+`num_collections` is always the **total** for the scan (`TriggerInfo.number_of_collections`),
+never a per-kickoff figure. A finite buffer is fed by data callbacks and does not care how many
+kickoffs span it, so per-kickoff sizing would mean resizing the buffer mid-scan.
```
coretl: No per-kickoff anymore. Rewrite this ADR to be shorter and correct

jwlodek:
One convenience feature that I'd like added - could we have `AreaDetector` have an `add_readables` override that if any passed in readables are an instance of `NDPluginBaseIO` to register them as an attribute like with passing plugins at `__init__`? Just to avoid having to pass the same object twice?

coretl:
> One convenience feature that I'd like added - could we have `AreaDetector` have an `add_readables` override that if any passed in readables are an instance of `NDPluginBaseIO` to register them as an attribute like with passing plugins at `__init__`? Just to avoid having to pass the same object twice?

Not sure I understand, when `AreaDetector` is a `StandardReadable` then doesn't `add_readables` just work? A code snippet might help explain...

jwlodek:
I think `add_readables` just works for getting the read / describe / config data, but I don't think it attaches the readable as an attribute? If I pass a plugin via the `plugins` kwarg, I can get it back with `get_plugins`, and access it with `my_det.my_plugin`. I think with `add_readables` there is no `setattr(self, readable.name, readable)` or something along those lines.

Basically, I'd still like this:
```Python
        if plugins is not None:
            for plugin_name, plugin in plugins.items():
                setattr(self, plugin_name, plugin)
```
Without having to pass the plugin in at `__init__` and then pass it in again with `add_readables`.

Generally, it would be really nice if AD had something like PVI so we could just ask at runtime for a plugin list like we do with panda blocks...

Then we could do away with the plugins parameter entirely, and just do `add_readables` for any you're interested in...

PR #1413, opened 2026-09-04
Allow passing strings for directory paths to path providers

jwlodek:
While this runs the risk of the user passing a windows path semantics path on a linux host and getting an invalid result, it is convenient to allow this for cases where the path semantics of the ioc host and client are the same. I think in the name of convenience it's worth allowing it, with the caveat that specifying a particular semantics is preffered.

### DIFF
--- a/src/ophyd_async/core/_path_providers.py
+++ b/src/ophyd_async/core/_path_providers.py
@@ -61,7 +61,7 @@ class PathInfo(ConfinedModel):
     @classmethod
     def validate_directory_path(cls, directory_path: PurePath) -> PurePath:
         """Ensure that the provided directory path is absolute.
-
+    
         :param directory_path: The directory path to validate
         :return: The validated directory path if it is absolute
         :raises ValueError: If the directory path is not absolute
@@ -178,13 +178,13 @@ class StaticPathProvider(PathProvider):
     def __init__(
         self,
         filename_provider: FilenameProvider,
-        directory_path: PurePath,
+        directory_path: PurePath | str,
         directory_uri: str | None = None,
         create_dir_depth: int = 0,
     ) -> None:
         self._filename_provider = filename_provider
-        self._directory_path = directory_path
-        self._directory_uri = directory_uri or generate_directory_uri(directory_path)
+        self._directory_path = PurePath(directory_path) if isinstance(directory_path, str) else directory_path
+        self._directory_uri = directory_uri or generate_directory_uri(self._directory_path)
         self._create_dir_depth = create_dir_depth
 
     def __call__(self, datakey_name: str | None = None) -> PathInfo:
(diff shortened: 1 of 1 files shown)


src/ophyd_async/core/_path_providers.py
```diff
@@ -178,13 +178,13 @@ class StaticPathProvider(PathProvider):
...
         filename_provider: FilenameProvider,
-        directory_path: PurePath,
+        directory_path: PurePath | str,
         directory_uri: str | None = None,
         create_dir_depth: int = 0,
     ) -> None:
         self._filename_provider = filename_provider
-        self._directory_path = directory_path
-        self._directory_uri = directory_uri or generate_directory_uri(directory_path)
+        self._directory_path = PurePath(directory_path) if isinstance(directory_path, str) else directory_path
```
coretl, line 186: I think `PurePath(directory_path)` is safe even if it's already a `PurePath`?


```suggestion
        self._directory_path = PurePath(directory_path)
```

src/ophyd_async/core/_path_providers.py
```diff
@@ -306,9 +304,9 @@ def __init__(
         base_name: str | None = None,
     ) -> None:
         self._filename_provider = filename_provider
-        self._base_directory_path = base_directory_path
+        self._base_directory_path = PurePath(base_directory_path) if isinstance(base_directory_path, str) else base_directory_path
```
coretl, line 307: ditto

src/ophyd_async/core/_path_providers.py
```diff
@@ -354,15 +352,15 @@ class YMDPathProvider(PathProvider):
...
         filename_provider: FilenameProvider,
-        base_directory_path: PurePath,
+        base_directory_path: PurePath | str,
         base_directory_uri: str | None = None,
         create_dir_depth: int = -3,  # Default to -3 to create YMD dirs
         datakey_name_as_base_dir: bool = False,
     ) -> None:
         self._filename_provider = filename_provider
-        self._base_directory_path = base_directory_path
+        self._base_directory_path = PurePath(base_directory_path) if isinstance(base_directory_path, str) else base_directory_path
```
coretl, line 361: ditto

coretl (changes requested):
Also, please can I have a test

src/ophyd_async/core/_path_providers.py
jwlodek: Surprisingly, it actually isn't, since putting a `PurePath` into the constructor of another `PurePath` will change it to match the path semantics of the current host:

```
In [1]: from pathlib import PurePath, PureWindowsPath

In [2]: p = PureWindowsPath("Z:\\test")

In [3]: y = PurePath(p)

In [4]: y
Out[4]: PurePosixPath('Z:/test')
```

Which also brings up the question, perhaps we should attempt to deduce the path semantics of the passed in string? I.e. does it have `/` or `\\`?

coretl: I'd say we should assume that a string path uses the host semantics, and you pass the correct `PurePath` if your hosts differ

coretl: Although now I've seen the code you wrote to do just that, then let's stick with what you've got...

coretl (approved):

Issue #709, opened 2024-12-17
 Should Devices have an ophyd sync compatible repr?


coretl:
At the moment `bps.scan` does this for the args:
https://github.com/bluesky/bluesky/blob/f655a730887e813c1905f89281cb564dd88d4e88/src/bluesky/plans.py#L1251

@burkeds reports:

> It looks like ophyd devices are returning string representations of themselves differently than ophyd devices.
> For example, in `scan`, the list of motors is populated by `list(map(repr, [motor1]))` which returns something that looks like:
> `["SynAxis(prefix='', name='motor1', read_attrs=['readback', 'setpoint'], configuration_attrs=['velocity', 'acceleration'])"]`
> An ophyd-async device returns the simple string representation of the object:
> `['<ophyd_async.sim.demo._sim_motor.SimMotor object at 0x7f2be5339510>']`
> Should ophyd-async devices have a similar return? This difference is breaking our analysis pipelines when moving between ophyd and ophyd-async devices

@tacaswell @danielballan thoughts?

tacaswell:
I am in favor of having nice reprs, but am a bit worried that analysis pipelines are depending on the exact details of the repr.  Everything you could want to know about the scan that are in that list should be available in a more machine-friendly form somewhere else in the documents.

coretl:
In which case I propose we add a repr with just the `name` for ophyd async: `SimMotor(name='motor1')`

@burkeds what exactly are the analysis pipelines reading from the repr? Could they get the same information elsewhere in the doc as @tacaswell suggests?

tacaswell:
To take a slightly harder stance, in general I do not think that parsibility of repr/str output is part of the "stable API" and can be broken at any time as the target consumer of these is mostly humans (our devices are for the most part too complex to be good candidates for round-trippable reprs, but even if they were the target is Python not ad-hoc parsing).

burkeds:
I'll also confirm that the people here working on analysis of Bluesky documents have interpreted the presence of this "standard" set of metadata in the plans to mean it is the "official" Bluesky metadata scheme that should be complied with and that schemes which do not comply with it are not "real" Bluesky schema.

coretl:
[...]

PR #1415, opened 2026-09-07
Add user friendly `__str__` and `__repr__` to `Device`

### LINKED ISSUE #1414: Add user_friendly `__str__` and `__repr__` for `Device`
When an error message occurs and it involves the device, it will display something like this:

`<ophyd_async.sim._motor.SimMotor object at 0x7ff5770c6e50>`

This isn't very useful when trying to display error messages back to a user involving a device. This came up when trying to refactor some of our plans in [dodal](https://github.com/DiamondLightSource/dodal/pull/2050) and trying to provide useful and user friendly error messages back to the user.

It would be good when logging, printing or raising an error involving a device it gave the device class and the name so it is much quicker to debug

Something like this instead

`ophyd_async.sim._motor.SimMotor[name="x_axis"]`

This a nice quality of life change that would make debugging and knowing exactly what device is involved much easier.

## Acceptance Criteria
- A better `__str__` and `__repr__` method added

oliwenmandiamond:
Fixes https://github.com/bluesky/ophyd-async/issues/1414

### DIFF
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -349,6 +349,12 @@ class Device(HasName):
             # Wait for it to complete
             await connect_task
 
+    def __repr__(self) -> str:
+        return f'{type(self).__name__}[name="{self.name}"]'
+
+    def __str__(self) -> str:
+        return repr(self)
+
 
 _not_device_attrs = {
     "_name",
--- a/src/ophyd_async/core/_device.py
+++ b/src/ophyd_async/core/_device.py
@@ -350,6 +350,8 @@ class Device(HasName):
             await connect_task
 
     def __repr__(self) -> str:
+        if self.name == "":
+            return super().__repr__()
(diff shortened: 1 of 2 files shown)


jwlodek (approved):

jwlodek:
Also resolves #709

jacopoabramo:
@oliwenmandiamond could you also add a test involving a device with signals? ~Or with children devices?~ What's the expected output?

EDIT: children devices are tested, I haven't looked at the changes appropriately. Sorry!

coretl:
@jwlodek this isn't *quite* what #709 suggested. That said it wanted `ClassName(name="my_name")`, and this PR makes `ClassName[name="my_name"]`. Do we want to go with #709 and ophyd compatibility? Or leave it like it is?

jwlodek:
I think parentheses are more natural here as it mimics an instantiation call, but I don't have a strong opinion on the format - I think clients relying on parsing repr strings probably should be doing what they need some other way.

PR #1418, opened 2026-09-11
Pin pyqt6 to <6.10, since 6.10+ needs a version of glibc that is newer than the one shipped with RHEL8.

jwlodek:
On older linux hosts (circa ~glibc 2.28) `uv sync` fails due to the pinned `PyQt6` wheels not being available. Since this is only needed for demos, move them out of the dev candidate.

1. `uv sync` on an older linux host before change fails
2. After change it succeeds

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,6 @@ dev = [
     "ophyd_async[ca]",
     "ophyd_async[pva]",
     "ophyd_async[tango]",
-    "ophyd_async[demo]",
     "inflection",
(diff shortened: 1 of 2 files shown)


coretl:
I'd prefer to cap the version of PyQt6 so it still works on older machines, and keep demo in the dev group:

```
"pyqt6<6.10",  # 6.10+ needs glibc 2.34; keep dev installable on RHEL8-era hosts
```

coretl (approved):

PR #1419, opened 2026-09-11
Make driver optional for ADHDFDataLogic

jwlodek:
We have a detector that has AD plugins, and so can use the HDF5 writer, but is not itself an ADDriver. Since the only thing the reference to the driver is used for is the attributes, we can omit it safely.

### DIFF
--- a/src/ophyd_async/epics/adcore/_data_logic.py
+++ b/src/ophyd_async/epics/adcore/_data_logic.py
@@ -167,16 +167,16 @@ class ADHDFDataLogic(DetectorDataLogic):
 
     :param array_description: Signals describing the NDArray shape and data type.
     :param path_provider: Callable that provides path information for file writing.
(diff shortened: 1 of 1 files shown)


sligara7 (approved):
Reviewed with you at beamline.  Looks good.

coretl:
What kind of a detector is this that has plugins but no driver? I didn't know that was possible in areaDetector...

jwlodek:
You can do this by subclassing `asynNDArrayDriver` instead of `ADDriver`. Then you can have a port driver that can produce NDArrays consumable by plugins, but without the common AD API. In this case this is for an in-house developed detector that originally had it's IOC as just a raw asynPortDriver, but we wanted to have access to file plugins etc. Porting it to an asynNDArrayDriver required minimal changes compared to reworking it into an ADDriver.

PR #1421, opened 2026-09-16
Fix the filling of optional TriggerableCommand or Device children

coretl:
`DeviceFiller._scan_for_annotations` only recognised `Optional[X]` via `typing.Union`, which `X | None` only produces when `X` is a subscripted generic alias. For a plain class (a bare `Device` or `Command` subclass), `X | None` produces `types.UnionType`
instead, so the annotation was silently dropped.

Fixed by also accepting `types.UnionType` (which is `typing.Union` on 3.14, so this stays correct across 3.11-3.14).

Also fixed a second bug: `check_filled` only resolved an unfilled-and-optional Signal/Device to `None`; an unfilled optional Command was left as a live, unconnected object, which then crashed p4p's native client (SIGABRT) instead of raising cleanly when connected. Fixed by including `_unfilled_command_backends` in `check_filled`'s check.


gkalua (approved):
Looks good to me.

PR #1423, opened 2026-09-17
Update SubsetEnum so its behaviour matches its intention

tomkane-dls:
A raw success from the PMAC comes back as a plain string and stays that way
```
if isinstance(value, str) and not isinstance(value, cls):
    return value
```

This means that a returned success will fail and return false with **is** so use **==** instead to ensure "Success" returns true
```
PmacStatus("Success")               -> 'Success'
'Success' is PmacStatus.SUCCESS     -> False
'Success' == PmacStatus.SUCCESS     -> True
```


coretl:
I'm not convinced this is the right fix, the fix should be to `SubsetEnum`, so instead of:
https://github.com/bluesky/ophyd-async/blob/4e27b36c28fe0050c4182bcfc71c99432dc70ad8/src/ophyd_async/core/_utils.py#L60-L61

Then maybe we do:
```python
try:
    return super().__call__(value, *args, **kwargs)
except ValueError:
    if isinstance(value, str):
        # This a string not listed in the subset, so allow it
        return value
    else:
        raise
```

or some kind of equivalent. We also need a test for `is` on a `SubsetEnum` to make sure we keep this behaviour

coretl (approved):

